PR #43, opened 2026-02-25
feat: session DAG with fork detection and subagent linking

### LINKED ISSUE #17: Session DAG: fork detection and subagent discovery
Claude Code session JSONL files form a DAG via uuid/parentUuid fields on each entry. Two types of branching exist:

- Forks: When a user resumes from an earlier point (Esc-Esc) or edits a message, the new entry's parentUuid points to a node that already has a child, creating a branch. In one 2,305-entry session we observed 4 large-gap forks (different approach from hundreds of entries back) and 14 small-gap forks (retrying the last turn). The parser currently reads linearly, flattening all branches.

- Subagents: Spawned via the Task tool, stored in {session-id}/subagents/agent-{agentId}.jsonl. Each file's sessionId field points to the parent. Currently filtered out during discovery (agent- prefix skipped, nested paths rejected).

Note: #11 already handles session continuity chaining (where Claude Code creates a new JSONL with sessionId pointing to the previous session). Forks and subagents are distinct from that.

Proposed modeling:

Forked conversations
- Forks become separate root-level sessions, each containing the shared prefix plus its own branch (self-contained for stats)
- Small-gap forks (retries/edits) stay within the same branch rather than creating new sessions. Heuristic TBD
- Fork session ID: {file-session-id}-{uuid-of-first-entry-on-new-branch}

Subagents
- Subagents become separate sessions with parent_session_id pointing to the spawning session
- Session ID from the filename: agent-{agentId}
- Stats are independent but can be aggregated with parent at query time
- Note: Frontend session grouping (designed for continuations) may need revisiting for these relationship types, as we naturally expect expanding subagents conversation in UI.

clkao:
Closes #17

## Summary

- Detect conversation forks in Claude Code sessions by building a uuid/parentUuid DAG, splitting large-gap branches into separate session records
- Link Task tool calls to their spawned subagent sessions via queue-operation parsing (JSON and XML formats)
- Add inline subagent conversation expansion in the frontend

## End-User Impact

- **Fork detection**: Conversation branches (e.g., from Claude Code's "retry from here") are detected and split into separate session entries instead of being jumbled together. Small retries (≤3 user turns) fold into the main session; larger forks appear as standalone sessions.
- **Subagent session linking**: Task subagent conversations are linked to the specific tool call that launched them. Task tool blocks show an expandable inline view of the subagent's conversation.
- **Cleaner session list**: Subagent and fork sessions are hidden from the top-level list — accessible only through their parent session context.

## Design

**DAG from uuid/parentUuid:** Every Claude Code JSONL entry carries `uuid` and `parentUuid` fields forming a tree. The parser builds a parent→children adjacency map in a single pass, then walks from root following first-child links. At fork points (nodes with multiple children), a heuristic counts user turns remaining on the first branch: >3 turns = real fork (split into separate `ParseResult`), ≤3 turns = retry (follow latest child, discard older branch).

**Subagent linking from queue-operation:** Claude Code writes `queue-operation` entries with `operation: "enqueue"` when spawning subagents. The `content` field maps `tool_use_id` → `task_id`. Two formats exist in the wild: JSON (`{"task_id":"...","tool_use_id":"..."}`) and XML (`<task-id>...</task-id><tool-use-id>...</tool-use-id>`). Parser tries JSON first via `gjson.Get`, falls back to regex for XML tags.

**Data model — relationship_type:** Sessions gain a `relationship_type` column (`""`, `"continuation"`, `"subagent"`, `"fork"`). Fork sessions get ID `{parent}-{first-uuid}` with `parent_session_id` pointing to main session. Tool calls gain `subagent_session_id` linking to the agent session.

**API — child sessions endpoint:** `GET /api/v1/sessions/{id}/children` returns fork/continuation/subagent sessions for a parent.

## Test plan

- [ ] Verify fork detection unit tests pass (`go test ./internal/parser/ -run TestForkDetection`)
- [ ] Verify subagent linking tests pass (`go test ./internal/parser/ -run TestSubagent`)
- [ ] Verify integration tests pass (`go test ./internal/sync/ -run TestSync`)
- [ ] Manual: load a session with Task subagents and verify inline expansion works
- [ ] Manual: verify subagent/fork sessions don't appear in top-level session list

### DIFF
--- a/internal/parser/types.go
+++ b/internal/parser/types.go
@@ -1,6 +1,9 @@
 package parser
 
-import "time"
+import (
+	"strings"
+	"time"
+)
 
 // AgentType identifies the AI agent that produced a session.
 type AgentType string
@@ -90,4 +93,23 @@ type ParsedMessage struct {
 type ParseResult struct {
 	Session  ParsedSession
 	Messages []ParsedMessage
+}
+
+// InferRelationshipTypes sets RelationshipType on results that have
+// a ParentSessionID but no explicit type. Sessions with an "agent-"
+// prefix are subagents; others are continuations.
+func InferRelationshipTypes(results []ParseResult) {
+	for i := range results {
+		if results[i].Session.ParentSessionID == "" {
+			continue
+		}
+		if results[i].Session.RelationshipType != RelNone {
+			continue
+		}
+		if strings.HasPrefix(results[i].Session.ID, "agent-") {
+			results[i].Session.RelationshipType = RelSubagent
+		} else {
+			results[i].Session.RelationshipType = RelContinuation
+		}
+	}
 }
--- a/internal/parser/types_test.go
+++ b/internal/parser/types_test.go
@@ -1,0 +1,86 @@
+package parser
+
+import "testing"
+
+func TestInferRelationshipTypes(t *testing.T) {
+	tests := []struct {
+		name    string
+		results []ParseResult
+		want    []RelationshipType
+	}{
+		{
+			"no parent",
+			[]ParseResult{
+				{Session: ParsedSession{ID: "abc"}},
+			},
+			[]RelationshipType{RelNone},
+		},
+		{
+			"agent prefix gets subagent",
+			[]ParseResult{
+				{Session: ParsedSession{
+					ID:              "agent-123",
+					ParentSessionID: "parent",
+				}},
+			},
+			[]RelationshipType{RelSubagent},
+		},
+		{
+			"non-agent prefix gets continuation",
+			[]ParseResult{
+				{Session: ParsedSession{
+					ID:              "child-session",
+					ParentSessionID: "parent",
+				}},
+			},
+			[]RelationshipType{RelContinuation},
+		},
+		{
+			"explicit type preserved",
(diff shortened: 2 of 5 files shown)


wesm:
"Both are valid observations but neither warrants blocking the merge.

  Subagent inline 1000-message cap — This is a pragmatic limit, not a bug. Subagent sessions with
  1000+ messages are extremely rare (subagents are typically short-lived tool calls). Adding
  pagination to an inline expandable view adds real complexity for a scenario that essentially
  doesn't occur in practice. If it ever becomes a problem, it's a feature request, not a defect.

  Stats include subagent/fork sessions — This is intentional. The stats endpoint counts total
  activity (all sessions, messages, cost). The session list filters to "primary" sessions for
  browsing. These serve different purposes — the stats reflect actual resource usage while the list
  shows navigable sessions. Forcing them to match would undercount activity. If anything, the right
  fix later would be adding a breakdown (primary vs subagent vs fork) rather than excluding from
  totals.

  Neither is a regression introduced by this PR — the 1000-message limit predates these changes, and
   the stats queries were never scoped by relationship type. Ship it."

Issue #37, opened 2026-02-25
Github Copilot agent sessions from VS Code missing


cdeil:
The Github Copilot agent session basic functionality now works - thanks!

But I only see the sessions I launched via the GitHub Copilot CLI.

The sessions launched from VS Code as local agents don't show up at all.

cdeil:
@wesm - See #80 for a prototype to explore this. Probably very low quality - obviously feel free to throw away.

It works a bit, but as a VSCode Copilot user it's difficult to understand how the agentsview session view relates to what VSCode presents.

Quick demo:

PR #80, opened 2026-03-01
feat: add VSCode Copilot agent support

cdeil:
## Summary
Add a new agent for VSCode Copilot chat sessions (#37). Discovers, parses, and displays conversations from VSCode's built-in Copilot Chat across Windows, macOS, and Linux (including Insiders and VSCodium editions).

Supports both the legacy monolithic JSON format and the newer JSONL append-only operation log (v1.109+). When both exist for the same session, JSONL takes priority.

## Key changes
- **Parser**: JSON and JSONL session formats, 60+ tool ID normalization, tool call metadata extraction
- **Discovery**: `workspaceStorage/` and `globalStorage/` scanning with workspace manifest resolution
- **Sync**: File watcher integration via `classifyOnePath`, .jsonl priority enforcement on incremental sync
- **Frontend**: Agent color/metadata, tool name regex update

## Not included
- Agent sessions (coding agent mode) — stored in `state.vscdb` SQLite, not as separate files

Addresses #37

### DIFF
--- a/internal/parser/types.go
+++ b/internal/parser/types.go
@@ -125,6 +125,10 @@ var Registry = []AgentDef{
 		EnvVar:      "VSCODE_COPILOT_DIR",
 		ConfigKey:   "vscode_copilot_dirs",
 		DefaultDirs: []string{
+			// Windows
+			"AppData/Roaming/Code/User",
+			"AppData/Roaming/Code - Insiders/User",
+			"AppData/Roaming/VSCodium/User",
 			// macOS
 			"Library/Application Support/Code/User",
 			"Library/Application Support/Code - Insiders/User",
--- a/internal/parser/types_test.go
+++ b/internal/parser/types_test.go
@@ -242,9 +242,15 @@ func TestVSCodeCopilotDefaultDirs(t *testing.T) {
 	}
 
 	required := []string{
+		// Windows
+		"AppData/Roaming/Code/User",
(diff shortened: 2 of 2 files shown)


internal/parser/types.go
```diff
@@ -118,6 +119,25 @@ var Registry = []AgentDef{
 		DiscoverFunc:   DiscoverAmpSessions,
 		FindSourceFunc: FindAmpSourceFile,
 	},
+	{
+		Type:        AgentVSCodeCopilot,
+		DisplayName: "VSCode Copilot",
+		EnvVar:      "VSCODE_COPILOT_DIR",
+		ConfigKey:   "vscode_copilot_dirs",
+		DefaultDirs: []string{
```
markdessain: Add vscodium directories here

wesm:
"This is a deliberate design choice, not a bug. The .jsonl file is VSCode's canonical format (v1.109+). If a
  .jsonl exists, it means VSCode is actively writing to it — a partially written .jsonl will be complete on the
  next sync cycle. Falling back to .json in that case would surface stale data (the .json is a snapshot from    
  before the migration), which is worse than waiting for the next sync to pick up the complete .jsonl.
                                                                                                                
  This matches how the Cursor agent handles .jsonl/.txt dedup — prefer the newer format, don't mix. No changes
  needed."

wesm:
thank you!

PR #134, opened 2026-03-10
feat: add trusted public origins and managed Caddy mode

### LINKED ISSUE #133: Managed Caddy mode for remote/TLS access
## Summary

Add a managed Caddy mode for remote access to agentsview, with explicit trusted public origin support and optional client CIDR allowlists.

I already have a local implementation for this and am opening the issue first so the follow-up branch and PR can reference an issue number.

## Motivation

Today, agentsview's host/origin protections work well for local-only access, but they make hostname- or proxy-based access awkward unless the user hand-configures both the app and an external reverse proxy.

A built-in managed Caddy mode makes the common setup much simpler:

- keep the agentsview backend bound to loopback only
- expose a public URL via a managed Caddy sidecar
- explicitly trust the external browser origin
- optionally restrict access to one or more client CIDRs

## Proposed scope

- `public_url` / `--public-url` for the external browser-facing URL
- explicit trusted public origin support derived from that URL
- optional managed `caddy` proxy mode
- explicit proxy bind host and public port
- repeated `allowed_subnets` / `--allowed-subnet` CIDR allowlists
- docs/examples for direct hostname access and managed Caddy usage

## Non-goals for the first pass

- automatic Caddy installation
- ACME/certificate automation
- system service management

## Platform notes

The feature should be usable anywhere the `caddy` CLI itself is available. That includes Linux, macOS, and Windows in principle, but the first pass should keep installation/packaging of Caddy out of scope.

tpn:
Closes #133.

Adds two related pieces for remote/hostname access:

- explicit trusted public origin support (`public_url` / `public_origins`)
- an optional managed Caddy sidecar for TLS-terminated access with CIDR allowlists

This keeps the existing DNS-rebinding and CSRF protections intact while making common remote access setups much easier to configure.

## Design

**Explicit trusted origins instead of wildcard host/origin relaxations**: The backend still defaults to loopback-only origin/host allowlists. When `public_url` or `public_origins` are configured, the server derives additional trusted `Host` and `Origin` values from those explicit inputs rather than weakening the protections globally.

**Managed proxy is opt-in and keeps the backend on loopback**: In managed Caddy mode, agentsview validates that the backend bind host is loopback-only, writes a generated Caddyfile under `~/.agentsview/managed-caddy/`, runs `caddy validate`, then launches `caddy run` as a child process. This keeps the CIDR allowlist meaningful because the backend is not exposed directly.

**Explicit public bind and port**: Managed Caddy now has separate backend and frontend settings. `-host` / `-port` remain the backend listener, while `proxy.bind_host` / `public_port` control the public-facing socket. The managed default public port is `8443`.

**CIDR allowlists with shorthand normalization**: `allowed_subnets` / `--allowed-subnet` accept repeated CIDRs and normalize shorthand IPv4 input like `10.0/16` to `10.0.0.0/16`.

## Scope

- `public_url` / `--public-url`
- `public_origins` / `--public-origin`
- managed `caddy` proxy mode
- `proxy.bind_host` / `--proxy-bind-host`
- `proxy.public_port` / `--public-port`
- `allowed_subnets` / `--allowed-subnet`
- README examples for direct hostname access and managed Caddy usage

## Not implemented

- automatic Caddy installation
- ACME / cert provisioning
- service management / daemonization

The patch assumes the `caddy` CLI is already installed and available either on `PATH` or via `--caddy-bin`.

## Platform notes

Managed Caddy mode is designed to work anywhere the `caddy` CLI works. That includes Linux, macOS, and Windows in principle, but this PR intentionally keeps Caddy installation/packaging out of scope.

## Changes

- **Config/CLI**: add `public_url`, managed proxy settings, CIDR allowlists, and public origin normalization/validation
- **Server**: derive trusted hosts/origins from configured public origins instead of only the bind address
- **Managed Caddy**: generate config, validate, supervise child process, and keep the backend loopback-only
- **Docs**: add examples for direct remote access and managed Caddy usage, plus explicit note that Caddy installation is external
- **Tests**: add config, handler, and managed-Caddy unit tests for origin handling, bind/public port behavior, and CIDR validation

## Test plan

- [x] Full Go suite passes (`CC=gcc CXX=g++ make test`)
- [x] Manual: build repo-local binary and hit `GET /api/v1/version` through managed Caddy over TLS with `curl --resolve`
- [ ] Frontend vitest / Playwright were not run separately for this PR

🤖 Generated with [Codex]

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -1,139 +1,72 @@
 review_guidelines = """
-agentsview is a single-user developer tool. It binds to 127.0.0.1
-by default and is not designed for multi-user or public deployment.
-An optional managed Caddy reverse proxy mode allows LAN/remote
-access while keeping the backend on loopback.
+agentsview is a single-user developer tool. Default mode binds to
+127.0.0.1. Optional managed Caddy proxy mode allows LAN access while
+keeping the backend on loopback. Not designed for multi-user or
+internet-facing deployment.
 
 Key assumptions reviewers MUST account for:
 
-1. AUTHENTICATION MODEL: In default (localhost-only) mode, all API
-   endpoints are unauthenticated by design — the loopback bind is
-   the access boundary. In managed Caddy proxy mode, the backend is
-   REQUIRED to bind to loopback (validateServeConfig enforces this).
-   External access goes through Caddy, which enforces CIDR-based
-   subnet allowlisting. This is a single-user personal tool — subnet
-   filtering is the intended access-control mechanism for LAN access.
-   Do not flag missing bearer tokens or session auth as a gap unless
-   the backend can be reached directly from the network (it cannot
-   when proxy mode is active). Do flag any code path that would allow
-   the backend to bind non-loopback while proxy mode is enabled.
+1. AUTH MODEL: Loopback bind is the access boundary — no auth tokens
+   needed. In proxy mode, validateServeConfig forces the backend to
+   loopback; Caddy enforces CIDR subnet allowlisting. Host/Origin
+   validation is defense-in-depth against DNS rebinding, not primary
+   access control. Do not flag missing bearer tokens/session auth.
+   DO flag any path that lets the backend bind non-loopback in proxy
+   mode, or missing subnet checks for non-loopback Caddy binds.
 
-2. XSS / SANITIZATION: Markdown rendering uses DOMPurify. The
-   {@html renderMarkdown(...)} pattern in Svelte is intentional and
-   safe because renderMarkdown() sanitizes via DOMPurify before
-   returning HTML. Do not flag this as XSS.
+2. MANAGED CADDY INVARIANTS: Backend loopback-only (enforced).
+   Non-loopback Caddy binds require allowed_subnets. auto_https off;
+   TLS needs user-supplied cert/key. Caddyfile is generated, validated
+   via `caddy validate`, then run as a supervised child process.
+   DO flag Caddyfile injection via unsanitized config values.
 
-3. RATE LIMITING / DoS: As a single-user tool, rate limiting and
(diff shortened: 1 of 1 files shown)


tpn:
Addressed both `roborev-ci` findings in c71986e.

- Fixed the stale `public_url` / trusted-origin case when the backend port is auto-relocated and the configured public URL was clearly tied to that original backend port. Manual external-proxy URLs on other ports are left unchanged.
- Fixed readiness probing for wildcard backend binds by mapping `0.0.0.0` -> `127.0.0.1` and `::` -> `::1` before dialing.
- Added unit coverage for both behaviors and reran the full Go suite: `CC=gcc CXX=g++ make test`.

tpn:
FWIW this is how I'm using it on my headless DGX Spark, with TLS set up so that I can view everything on any other host on my home network:

```
/home/trent/src/agentsview/agentsview \
    -host 127.0.0.1 \
    -port 8080 \
    -public-url https://spark.trent.me \
    -proxy caddy \
    -proxy-bind-host 0.0.0.0 \
    -public-port 8443 \
    -tls-cert /home/trent/.certs/trent.me-server-bundle.crt \
    -tls-key /home/trent/.certs/trent.me-server.key.decrypted \
    -allowed-subnet 10.0/16 \
    -no-browser
```

tpn:
Added a small README note in 2b7ceec covering privileged ports for managed Caddy on Linux.

It now recommends keeping `agentsview` unprivileged and using `setcap cap_net_bind_service=+ep` on the `caddy` binary if the user wants `-public-port 443` or `80`, rather than running the whole viewer as root. It also explicitly points users back to the default `8443` path when they do not need a privileged port.

Issue #130, opened 2026-03-10
Tauri desktop: remote server mode, in-process backend, and settings UI


userFRM:
## Summary

Three related proposals for the desktop app's architecture, stemming from work on settings UI and remote access.

## 1. Remote Server Connection Mode

**Problem**: The Tauri app currently spawns a local sidecar and only connects to `127.0.0.1`. Users who run agentsview on a headless server (e.g., Linux via Tailscale/SSH) cannot use the desktop app to view those sessions — they must use a browser.

**Proposal**: Add a "remote mode" to the Tauri app where it:
- **Skips** spawning the local sidecar process
- **Connects** to a user-configured remote URL (e.g., `http://100.92.107.86:8080`)
- **Relaxes** the navigation guard (`is_allowed_navigation_url`) to allow the remote host
- **Sends** a Bearer auth token with all requests (token-based auth already prototyped, see below)

The desktop startup screen could offer two paths: "Local" (current behavior) and "Connect to remote server" (enter URL + token).

### What's already prototyped

On a local branch (`feat/settings-remote`), I've implemented the web-side of this:
- **Settings page** (`#/settings`) — full settings UI with sections for appearance, agent dirs, terminal, GitHub, and remote access
- **Auth middleware** (`internal/server/auth.go`) — Bearer token auth for non-localhost API requests, localhost always bypasses
- **Configurable API base URL** in `client.ts` — replaces hardcoded `/api/v1` with localStorage-backed server URL + auth token injection
- **`RemoteSettings.svelte`** — server URL entry, token input, test connection, connect/disconnect
- **Auto-generated auth token** (`crypto/rand`, 32 bytes, base64url) via `EnsureAuthToken()` in config

This works today for browser-based remote access (tested over Tailscale). The missing piece is the Tauri Rust-side changes to support remote mode.

### Tauri changes needed

In `desktop/src-tauri/src/lib.rs`:
- Add a `ConnectionMode` enum: `Local` (spawn sidecar) vs `Remote { url, token }`
- Read mode from config file (`~/.agentsview/desktop.json`) or startup dialog
- In remote mode: skip `launch_backend()`, update `is_allowed_navigation_url()` to allow the remote host, redirect webview directly to remote URL
- Pass auth token to webview via a Tauri command or injected JS variable

## 2. In-Process Backend (App Store Safety)

**Problem**: The current sidecar approach spawns a Go binary that listens on `127.0.0.1:8080` — a real TCP port. This has implications:
[...]

PR #138, opened 2026-03-11
feat: settings page, remote access auth, and configurable API base URL

userFRM:
## Summary

- **Settings page**: New `/settings` route with organized sections for appearance, agent directories, terminal integration, and GitHub integration
- **Remote access auth**: Bearer token middleware (`internal/server/auth.go`) for securing the server when exposed over Tailscale/LAN
- **Configurable API base URL**: Frontend API client supports connecting to a remote agentsview server instead of only localhost
- **Settings store**: Reactive Svelte store for persisting user preferences

Partial implementation toward #130 (remote server mode and settings UI).

## Test plan

- [ ] Navigate to settings page via header gear icon
- [ ] Toggle appearance settings (theme, block visibility)
- [ ] Configure a remote server URL in settings — verify API calls route there
- [ ] Enable auth token on server (`--auth-token`) — verify unauthenticated requests are rejected
- [ ] Verify existing functionality (session list, messages, search) still works with settings page added

### DIFF
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -1465,6 +1465,9 @@ func TestHostHeaderPublicOriginsDoNotExpandTrustedHosts(t *testing.T) {
 
 	req := httptest.NewRequest(http.MethodGet, "/api/v1/stats", nil)
 	req.Host = "viewer.example.test:8004"
+	// Use loopback RemoteAddr so authMiddleware passes through and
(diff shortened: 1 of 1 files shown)


wesm:
@userFRM we might need to update review_guidelines to get roborev to chill out because the security / thread model of the project has changed, I haven't looked closely yet but just wanted to mention

PR #158, opened 2026-03-12
fix: preserve project name casing in group headers

userFRM:
## Summary

- Remove `text-transform: capitalize` from `.group-header` so project names display with their original casing from the filesystem
- Revert the `notifySyncComplete()` suppression during session watches — it caused stale sidebar and filter state (projects, agents, session counts stopped updating while any session was selected)

## Test plan

- [ ] Open a session with "Group by project" enabled — project names should show original casing (e.g. "myProject" not "Myproject")
- [ ] While watching a live session, trigger a background sync — sidebar session list and filter dropdowns should update without going stale


wesm:
just FYI I am working on a pretty meaty incremental parsing and live update PR, I will get it up in the next hour or two

PR #163, opened 2026-03-12
fix: Windows Tauri desktop navigation and dev workflow

### LINKED ISSUE #140: Desktop app shows black screen on Windows
The desktop app is unusable on Windows. It shows a black screen, becomes "Not Responding," and then opens `http://tauri.localhost/` in the system browser instead of rendering in the webview.

The sidecar starts fine — the process is visible via `Get-Process` and the HTTP server is listening on port 8080. The issue is in the navigation guard, not the backend.

## Root cause

`is_allowed_navigation_url()` checks for the macOS WebView origin `tauri://localhost` but not the Windows WebView2 origin `http://tauri.localhost`:

https://github.com/wesm/agentsview/blob/0e50d680ff81bdc58af56d5be59926db4f189db4/desktop/src-tauri/src/lib.rs#L149-L162

On Windows, WebView2 rewrites `tauri://localhost` to `http://tauri.localhost`. When the splash page tries to load, the navigation guard rejects it because neither branch matches:

- Line 150: `url.scheme() == "tauri"` — fails, the scheme is `http`
- Line 157: `url.host_str() == Some("127.0.0.1")` — fails, the host is `tauri.localhost`

The rejected URL then falls through to the external URL handler:

https://github.com/wesm/agentsview/blob/0e50d680ff81bdc58af56d5be59926db4f189db4/desktop/src-tauri/src/lib.rs#L130-L134

`is_allowed_external_open_url()` passes any `http://` URL, so the internal `http://tauri.localhost/` URL opens in the system browser via `opener.open_url()`.

The existing test only covers the `tauri://` scheme — no `http://tauri.localhost` test case:

https://github.com/wesm/agentsview/blob/0e50d680ff81bdc58af56d5be59926db4f189db4/desktop/src-tauri/src/lib.rs#L913-L914

## Why this wasn't caught in #135

PR #135 fixed a similar navigation race and tightened the guard to require the known sidecar port, but the test plan only covered `make desktop-dev` which runs on macOS where the origin is `tauri://localhost`. The Windows `http://tauri.localhost` origin was not tested.

## Suggested fix

Two changes, following the pattern from #135:

1. Add `http://tauri.localhost` as an allowed origin in `is_allowed_navigation_url()` — the Windows equivalent of the existing `tauri://localhost` check at line 150. We could add something like:

```rust
if url.scheme() == "http" && url.host_str() == Some("tauri.localhost") {
    return true;
}
```

2. Exclude `tauri.localhost` from the external browser fallback as defense in depth — internal Tauri URLs should never open externally. The check in `is_allowed_external_open_url()` (or the fallback block) could reject URLs where the host is `tauri.localhost`.

## Debugging difficulty

All `eprintln!` output is lost on Windows release builds because `windows_subsystem = "windows"` hides the console. The blocked navigation is logged at line 139 but there's no way to see it without the log surfacing work tracked in #132.

## Environment

- agentsview 0.12.0 desktop (Windows installer)
- Windows 11 Pro, WebView2 runtime (Edge-based)
- Commit 0e50d68 (current main)

wesm:
## Summary

Fixes the Tauri desktop app on Windows and adds a PowerShell dev workflow script.

### Navigation guard for Windows WebView2
- WebView2 uses `http://tauri.localhost` (not `tauri://localhost` like macOS/Linux) as the app origin. The navigation guard now allows both `http` and `https` on `tauri.localhost`, with an explicit port check to prevent spoofing via other local services.
- Tests cover the allowed origins and rejected cases (wrong port, wrong scheme for localhost backend).

### Menu bar rename
- Renamed the app menu submenu from "AgentsView" to "File" to eliminate the triple-"AgentsView" appearance on Windows (OS title bar + menu bar + app header).

### Version sync
- Bumped `tauri.conf.json` version from `0.1.0` to `0.12.1` to match the project release version.

### Windows dev workflow (`scripts/desktop-dev.ps1`)
- New PowerShell script for building and launching the Tauri desktop app on Windows without requiring `make` or bash.
- Auto-activates fnm if Node.js is not on PATH.
- Builds the frontend SPA, Go sidecar binary, and copies it to the Tauri binaries directory.
- Dynamically patches `tauri.conf.json` version from git tags (matching the bash `prepare-sidecar.sh` behavior) and restores it on exit.
- Detects the Rust host triple at runtime for correct sidecar binary naming (ARM64 support).
- Uses `cargo run` directly instead of `npx tauri dev` to avoid the Tauri dev server opening a stale browser tab on port 1430.
- Supports `-SkipBuild` flag for fast iteration on Rust/Tauri code only.

### Housekeeping
- Added `/agentsview.exe` to `.gitignore` for Windows build artifacts.

## Test plan
- [x] Tauri desktop app launches and navigates correctly on Windows 11
- [x] Splash screen transitions through stages to the main app UI
- [x] No stale browser tab opened on port 1430
- [x] Menu bar shows "File" instead of duplicate "AgentsView"
- [x] `desktop-dev.ps1` full build and `-SkipBuild` paths both work
- [x] Navigation guard rejects http://tauri.localhost:9999 and other spoofable URLs

Closes #140

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Go binaries
 /agentsview
+/agentsview.exe
 /agentsv
 /testfixture
 *.test.exe
--- a/desktop/src-tauri/src/lib.rs
+++ b/desktop/src-tauri/src/lib.rs
@@ -151,8 +151,13 @@ fn is_allowed_navigation_url(url: &Url, backend_port: Option<u16>) -> bool {
     if url.scheme() == "tauri" && url.host_str() == Some("localhost") {
         return true;
     }
-    // Windows (WebView2): http://tauri.localhost or https://tauri.localhost
-    if matches!(url.scheme(), "http" | "https") && url.host_str() == Some("tauri.localhost") {
+    // Windows (WebView2): https://tauri.localhost (default origin).
+    // Restrict to https-only with no explicit port to avoid accepting
+    // spoofable http origins or unexpected ports.
+    if url.scheme() == "https"
+        && url.host_str() == Some("tauri.localhost")
+        && url.port().is_none()
+    {
         return true;
     }
     // Only allow navigation to the known sidecar port on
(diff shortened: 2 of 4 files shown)


wesm:
All good here I think

PR #167, opened 2026-03-13
feat: PostgreSQL push sync and read-only server mode

tlmaloney:
## Summary

Adds PostgreSQL support as an optional remote backend for multi-machine shared access to session data. Three new commands under `pg`:

- **`pg push`** — one-way sync from local SQLite to PostgreSQL. Incremental by default (only sessions modified since last push), with `--full` to re-push everything. Batched transactions (50 sessions/tx) with per-session retry fallback on batch failure. Multi-row INSERT for messages and tool calls. Fingerprint-based change detection skips unchanged sessions on incremental pushes.
- **`pg serve`** — read-only HTTP server backed by PostgreSQL instead of SQLite. Same API and frontend, no local SQLite dependency. Write operations (stars, pins, rename, etc.) return `ErrReadOnly`; the frontend adapts accordingly.
- **`pg status`** — show sync state (last push time, session/message counts).

### Implementation details

- New `internal/postgres/` package: connection management, SSL enforcement, schema DDL with version tracking, push sync, read-only store with sessions/messages/search/analytics
- `db.Store` interface extracted from `db.DB` so the HTTP server works identically against SQLite or PostgreSQL
- Config migrated from JSON to TOML (`~/.agentsview/config.toml`), with automatic migration of existing JSON configs
- SSL enforced for non-loopback PG connections by default; `allow_insecure = true` under `[pg]` overrides with a warning
- All PG connections force `TimeZone=UTC` for consistent date filtering
- Coherence check on push: detects when PG schema was dropped externally (local watermark set but PG has 0 sessions) and auto-forces full re-push with schema recreation
- UTF-8 sanitization before PG insert (null bytes, truncated multi-byte sequences stripped from content fields)
- Rune-aware string truncation in parsers to avoid splitting multi-byte characters
- `--base-path` flag for reverse-proxy subpath deployment with `<base href>` injection
- `public_origins` config expands the host allowlist for trusted reverse-proxy origins
- Docker-compose infrastructure for PG integration tests, with CI integration via GitHub Actions service container

### Not in scope (follow-up issues)

- Write support in pg serve (#183)
- Subpath-aware routing in Go server (#184)
- Managed Caddy for pg serve (#185)
- Detect PG URL change and auto-reset watermarks (#186)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,9 @@ jobs:
           go-version-file: go.mod
 
       - name: Stub frontend embed dir
-        run: mkdir -p internal/web/dist && echo ok > internal/web/dist/stub.html
+        run: |
+          mkdir -p internal/web/dist
+          echo '<!doctype html><html><head></head><body><div id="app"></div></body></html>' > internal/web/dist/index.html
 
       - uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20  # v9.2.0
         with:
@@ -55,7 +57,9 @@ jobs:
           path-type: inherit
 
       - name: Stub frontend embed dir
-        run: mkdir -p internal/web/dist && echo ok > internal/web/dist/stub.html
+        run: |
+          mkdir -p internal/web/dist
+          echo '<!doctype html><html><head></head><body><div id="app"></div></body></html>' > internal/web/dist/index.html
 
       - name: Run Go tests
         run: go test -tags fts5 ./... -v -count=1
@@ -80,7 +84,9 @@ jobs:
           go-version-file: go.mod
 
       - name: Stub frontend embed dir
-        run: mkdir -p internal/web/dist && echo ok > internal/web/dist/stub.html
+        run: |
+          mkdir -p internal/web/dist
+          echo '<!doctype html><html><head></head><body><div id="app"></div></body></html>' > internal/web/dist/index.html
 
       - name: Test with coverage
         run: go test -tags fts5 -race -coverprofile=coverage.out ./...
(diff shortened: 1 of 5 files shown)


tlmaloney:
## Addressed: Combined Review (`113e32f`)

### [Medium] PostgreSQL TLS enforcement bypass in DSN parsing

**Fixed in `39ceaba`.** Replaced the custom DSN parser with `pgconn.ParseConfig` from the pgx driver itself. This eliminates all format-specific bypasses (service files, env var indirection, exotic key-value syntax) since we now use the exact same parser the driver uses.

`CheckSSL` now inspects `TLSConfig` on both the primary config and all fallback configs, rejecting any connection chain that permits plaintext for non-loopback hosts. This correctly rejects `sslmode=disable`, `allow`, and `prefer` (which have plaintext fallback paths).

The custom `ParseSSLParams` and `KVParam` functions have been removed entirely (-154 lines, +55 lines).

### [Medium] Test suite compilation failures

**False positive.** The pgtest-tagged test files (`pgdb_test.go`, `integration_test.go`, `pgsync_test.go`) use `sql.Open("pgx", url)` directly — they do not call `pgdb.New` or `pgsync.New`. Verified with `grep` and confirmed all tests compile and pass with `CGO_ENABLED=1 go test -tags fts5 ./...`.

wesm:
@tlmaloney since this is a large PR it's going to be hard to get a clean review, so it might make sense to pause here so I can have a chance to review some of the high level details

tlmaloney:
> @tlmaloney since this is a large PR it's going to be hard to get a clean review, so it might make sense to pause here so I can have a chance to review some of the high level details

Agreed, I am not having luck converging on a passing verdict. I'll pause and happy to squash / resubmit / or whichever direction makes sense.

tlmaloney:
> I'm working on refactoring this to make some structural changes, so don't push any more work. We need containerized integration tests so I'm working on that too

Sounds good! Thank you. I spun up a Dockerized postgres, but didn't add that to the repo.

PR #168, opened 2026-03-13
feat: add Cursor vscdb session support

BadLiveware:
### Problem

Cursor's JSONL transcript files (written to `~/.cursor/projects/`) are a lossy export.
Every assistant message is reduced to `{"type": "text"}` blocks only — tool calls, MCP
invocations, and subagent relationships are stripped before the file is written. As a
result, agentsview showed Cursor sessions as plain text conversations with no tool blocks,
even for sessions with heavy tool usage.

The full session data is stored locally in Cursor's internal SQLite database at
`~/.config/Cursor/User/globalStorage/state.vscdb`. This is where Cursor itself reads from
when opening a session — which is why tool calls appear instantly in the Cursor UI but are
absent from the JSONL files.

### Solution

Read Cursor sessions from `state.vscdb` as the primary data source, falling back to JSONL
only for sessions not present in the database.

### Data model

`state.vscdb` has a single `cursorDiskKV` table (key/value store). Two key patterns are
relevant:

**`composerData:<sessionId>`** — session metadata including an ordered list of bubble IDs,
timestamps, session name, and `subComposerIds` (child session IDs for subagent sessions).

**`bubbleId:<sessionId>:<bubbleId>`** — individual messages. Each bubble has `type`
(1=user, 2=assistant), `text`, `createdAt` (ISO 8601), and optionally `toolFormerData`
with the tool name, call ID, JSON params, and result.

Project names are resolved by scanning
`~/.config/Cursor/User/workspaceStorage/*/workspace.json` (which maps each workspace hash
to a `file://` project path) and the companion `state.vscdb` in each workspace directory
(which lists which composer IDs belong to that workspace).

### Message reconstruction

Cursor stores each tool call as its own separate bubble (type 2), rather than grouping
them into a single assistant message as Claude Code does. The parser merges consecutive
type-2 bubbles — accumulating tool calls and concatenating text content — into a single
`ParsedMessage`. This matches the structure the rest of agentsview expects.

### Changes

**`internal/parser/cursor_vscdb.go`** (new)

- `ListCursorVscdbSessions` — returns lightweight metadata for all sessions (used for
  change detection). Resolves project names via the workspace storage scan.
- `ParseCursorVscdbSession` — parses a single session: loads bubble order from
  `composerData`, fetches all `bubbleId` entries, merges consecutive assistant bubbles
  into messages.
- `parseCursorBubbleTime` — handles Cursor's ISO 8601 bubble timestamps.
- `normalizeCursorParamsJSON` — handles params stored as either a JSON object or a
  JSON-encoded string.

**`internal/parser/cursor_vscdb_test.go`** (new): 13 unit tests covering list/parse
functions, tool call extraction, message grouping, and params normalization.

**`internal/parser/taxonomy.go`**

- Added vscdb tool name mappings: `run_terminal_command_v2` → Bash, `read_file_v2` →
  Read, `edit_file_v2`/`search_replace` → Edit, `ripgrep_raw_search`/`rg` → Grep,
  `glob_file_search` → Glob, `task_v2` → Task, `delete_file` → Write, `list_dir_v2` →
  Read, and several Tool-category names (`todo_write`, `create_plan`, etc.).
- Added `mcp-` prefix catch-all for MCP tool invocations.

**`internal/config/config.go`**

- Added `CursorStateDB string` field, defaulting to
  `~/.config/Cursor/User/globalStorage/state.vscdb`.
- Reads `CURSOR_STATE_DB` env var to override.

**`internal/sync/engine.go`**

- Added `CursorStateDB` to `EngineConfig` and `Engine`.
- `syncCursorVscdb` — follows the OpenCode pattern: uses `lastUpdatedAt` (unix millis)
  for per-session change detection via the existing `GetFileInfoByPath` virtual-path
  mechanism, calls `writeSessionFull` for in-place replacement. Builds a child→parent map
  from `subComposerIds` and sets `ParentSessionID`/`RelSubagent` on child sessions.
- vscdb sync runs **before** file workers. The set of synced session IDs is stored on the
  engine for the duration of the sync. `processCursor` checks this set and skips any
  session already handled by vscdb, preventing the text-only JSONL from overwriting the
  richer data.
- vscdb session counts fold into the overall `SyncStats.Synced` counter.

**`internal/sync/engine_integration_test.go`**: 4 integration tests — basic sync, change
detection (unchanged sessions are not re-parsed), dedup (vscdb wins over file-based
JSONL), and subagent parent-child linking.

**`cmd/agentsview/main.go`**, **`cmd/agentsview/sync.go`**: Pass `CursorStateDB` through
to `EngineConfig`.

### Validation

Tested against a real `state.vscdb` with 702 sessions — 563 sessions with actual
conversation content synced, producing 25,253 tool calls that were previously invisible.

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -194,6 +194,7 @@ func runServe(args []string) {
 		AgentDirs:               cfg.AgentDirs,
 		Machine:                 "local",
 		BlockedResultCategories: cfg.ResultContentBlockedCategories,
+		CursorStateDB:           cfg.CursorStateDB,
 	})
 
 	if database.NeedsResync() {
--- a/cmd/agentsview/sync.go
+++ b/cmd/agentsview/sync.go
@@ -93,8 +93,9 @@ func runSync(args []string) {
 	cleanResyncTemp(appCfg.DBPath)
 
 	engine := sync.NewEngine(database, sync.EngineConfig{
-		AgentDirs: appCfg.AgentDirs,
-		Machine:   "local",
+		AgentDirs:     appCfg.AgentDirs,
+		Machine:       "local",
+		CursorStateDB: appCfg.CursorStateDB,
 	})
 
 	ctx := context.Background()
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -67,6 +67,13 @@ type Config struct {
 	RemoteAccess         bool           `json:"remote_access"`
 	WriteTimeout         time.Duration  `json:"-"`
 
+	// CursorStateDB is the path to Cursor's global
+	// state.vscdb SQLite database used as primary source
+	// for tool calls and rich session metadata.
+	// Set CURSOR_STATE_DB to override. Default:
+	// ~/.config/Cursor/User/globalStorage/state.vscdb
+	CursorStateDB string `json:"-"`
+
 	// AgentDirs maps each AgentType to its configured
 	// directories. Single-dir agents store a one-element
 	// slice; unconfigured agents use nil.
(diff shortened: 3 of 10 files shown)


wesm:
@BadLiveware do you want to try picking this up again? I don't use Cursor so I can't develop and test this myself

wesm:
I'm rebasing this, and I will open up a new PR and ask for Cursor users to help test

PR #209, opened 2026-03-22
feat: use OpenCode session title as display title

### LINKED ISSUE #205: feat: Use session title generated by OpenCode as session title
Currently in the latest version, OpenCode sessions' title is the first message of the session. However, OpenCode uses Haiku 4.5 to generate title for each session. Use session title generated by OpenCode as session title can help users distinguish and compare each session.
Now I need to enter the session in OpenCode and get the first message before I can find this session in agentsview. Could you add this feature and save users' time?

hjerpe:
## Summary

- Use the LLM-generated `title` field from OpenCode's session table as the session display title in agentsview
- Fall back to first user message when the title is empty or a default placeholder (`"New session - ..."`, `"Child session - ..."`)
- Adds one new test covering both fallback cases

## Why

OpenCode stores messages across normalized `message` + `part` tables. The current title extraction reconstructs text from parts sorted by `time_created`, but when parts share the same timestamp (common during batch inserts), the sort order is not guaranteed. This can cause titles to appear jumbled or swapped between syncs.

OpenCode already generates an LLM-based title (via Haiku) for every session, stored in `session.title`. This field was already being loaded by the parser but never used.

## Changes

- `internal/parser/opencode.go`: 5 lines added in `buildOpenCodeSession()` to prefer `s.title` when it's a real LLM-generated title
- `internal/parser/opencode_test.go`: Updated existing assertion + 1 new test function covering empty and placeholder title fallback

Closes #205

### DIFF
--- a/internal/parser/opencode.go
+++ b/internal/parser/opencode.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


wesm:
thanks!

PR #221, opened 2026-03-24
feat: add managed Caddy support to pg serve

### LINKED ISSUE #185: feat: managed Caddy for pg serve mode
The existing managed Caddy mode works for normal SQLite-backed server mode but not for `pg serve`. Users deploying `pg serve` must set up their own reverse proxy infrastructure.

Adding managed Caddy support for `pg serve` would provide the same zero-config HTTPS experience that the SQLite server mode has.

tpn:
## Summary
- extract the shared serve startup/runtime path so normal `serve` and `pg serve` can reuse the same readiness and managed-Caddy flow
- add managed Caddy support to `pg serve`, including namespaced managed-Caddy state so SQLite-backed `serve` and PG-backed `pg serve` can coexist on one host
- document `pg serve` managed-Caddy usage and harden `pg serve` config loading so persisted normal `serve` proxy/public settings do not leak into the PG-backed server

Closes #185.

## Test Plan
- [x] `CC=/usr/bin/gcc CXX=/usr/bin/g++ CGO_ENABLED=1 go test -tags fts5 ./cmd/agentsview -count=1`
- [x] local 3-agent roborev branch review (`codex`, `claude-code`, `gemini`) on `origin/main..HEAD`

### DIFF
--- a/cmd/agentsview/pg.go
+++ b/cmd/agentsview/pg.go
@@ -198,6 +198,17 @@ func runPGServe(args []string) {
 		log.Fatalf("%v", err)
 	}
 	setupLogFile(appCfg.DataDir)
+	// Enable remote access with auth when binding to a
+	// non-loopback address; keep it off for localhost.
+	if !isLoopbackHost(appCfg.Host) {
+		appCfg.RemoteAccess = true
+		if err := appCfg.EnsureAuthToken(); err != nil {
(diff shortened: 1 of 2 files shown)


tpn:
> Verdict: All reviewers agree the code is clean and no issues were found.

Big improvement from the last PR now that I've got a three-way roborev active locally :-)

PR #222, opened 2026-03-24
feat: add focused transcript mode with responsive header

### LINKED ISSUE #219: feat: transcript toggle to hide intermediate working
## Summary

Add a transcript-view toggle that lets users hide intermediate working and focus on the key user <-> agent exchange.

## Problem

Very long sessions with lots of tool calls, thinking/reasoning blocks, and intermediate assistant messages can be hard to navigate when the user mostly wants to review the primary conversation flow.

In many cases, the most important path is:
- each user message
- the final assistant message that directly precedes the next user message

Everything else is still useful, but can create a lot of visual noise when scanning a long session.

## Proposed Solution

Add a UI toggle for the transcript view. Possible label: `Hide Intermediate Working`.

When enabled, the transcript should present a condensed view that:
- always shows user messages
- shows assistant messages that are the last assistant message before the next user message
- hides intermediate tool-call-heavy churn, reasoning blocks, and intermediate assistant steps that are part of the agent's working process

When disabled, the current full transcript remains visible.

## Notes On Semantics

This should be a presentation toggle, not a data-destructive filter. The full session remains available, and the user can switch back to the complete transcript at any time.

Naming is flexible. `Hide Intermediate Working`, `Hide Working`, or similar would all be reasonable if there is a better term.

## Acceptance Criteria

- Users can toggle between the full transcript and a condensed transcript
- Condensed mode preserves user messages
- Condensed mode preserves the assistant reply that immediately precedes the next user message
- Toggling does not lose position/state more than necessary
- Full transcript remains available with one click

## Notes

I'm creating this issue with the intent to follow up with a PR.

tpn:
## Summary

- Add focused transcript mode that shows only user prompts and final
  assistant answers, filtering out intermediate tool calls and thinking
  blocks for a cleaner read-through of sessions
- Replace the transcript mode select dropdown with a compact segmented
  pill toggle grouped with the block filter icon, reducing header
  crowding
- Add an overflow menu at narrow viewports (<700px) that collapses
  layout, export, and publish buttons behind a "..." button
- Add responsive breakpoints at 700px and 550px to prevent header
  control overlap on small screens
- Fix block filter interaction: skip tool grouping when tools are
  hidden so thinking content inside tool-only messages remains visible
- Fix pre-existing frontend type-check drift (0 errors on `npm run check`)

Closes #219.

### DIFF
--- a/frontend/src/lib/components/content/MessageList.svelte
+++ b/frontend/src/lib/components/content/MessageList.svelte
@@ -32,9 +32,15 @@
     buildDisplayItems(baseMessages),
   );
 
+  let filteredDisplayItemsAsc = $derived(
+    buildDisplayItems(baseMessages, {
+      skipToolGrouping: !ui.isBlockVisible("tool"),
+    }),
+  );
+
   function isItemVisible(item: DisplayItem): boolean {
     if (item.kind === "tool-group") {
-      return ui.isBlockVisible("tool");
+      return true;
     }
     return hasVisibleSegments(item.message, (type) =>
       ui.isBlockVisible(type),
@@ -43,7 +49,7 @@
 
   let normalDisplayItemsAsc = $derived.by(() => {
     if (!ui.hasBlockFilters) return baseDisplayItemsAsc;
-    return baseDisplayItemsAsc.filter(isItemVisible);
+    return filteredDisplayItemsAsc.filter(isItemVisible);
   });
 
   let displayItemsAsc = $derived.by(() => {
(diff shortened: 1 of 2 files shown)


tpn:
Obviously I'm biased but I really like this feature!  Makes agentsview vastly more useful for me (apparently I tend to have very long sessions with trillions of tool calls)--I can toggle the "Focused" view and just see the primary user/agent interaction (which is almost always why I'm consulting agentsview in the first place).

wesm:
I'll hold off on doing any more until you're done making changes so I can reconcile. I made some UI improvements and worked on problems with responsive layout

PR #242, opened 2026-03-26
feat: model Codex subagent workflows

### LINKED ISSUE #233: feat: model Codex subagent workflows instead of showing raw subagent_notification JSON
## Summary

Codex multi-agent sessions should be modeled as real subagent relationships instead of surfacing raw `<subagent_notification>` JSON blobs as ordinary user messages in the transcript.

## Problem

Codex sessions currently record subagent activity as a sequence like:

1. `spawn_agent` function call from the parent session
2. `function_call_output` containing the spawned `agent_id`
3. `wait` function call for that `agent_id`
4. `function_call_output` containing status keyed by `agent_id`
5. a synthetic user `response_item` whose content is wrapped in `<subagent_notification>...</subagent_notification>` and contains JSON like:
   - `agent_id`
   - `status.completed` / `status.errored`

Today, `agentsview` does not interpret that Codex-specific sequence as a subagent relationship. The visible symptom is that long raw JSON/status payloads show up in the **User** lane of the transcript.

That is semantically wrong in two ways:

- the notification is not a real user-authored message
- the spawned child agent is not linked into the existing subagent UI path

## Current State

A stopgap local fix can hide `<subagent_notification>` rows as system/intermediate messages, but that is only a presentation band-aid.

The proper product behavior should reuse the existing subagent/session-linking model already used for other agent types:

- `tool_calls.subagent_session_id`
- `relationship_type = "subagent"`
- inline child transcript rendering from the originating tool call
- session tree grouping under the parent

## Desired Behavior

For Codex sessions:

1. Detect `spawn_agent` as a subagent-spawning tool call.
2. Extract the returned `agent_id` from the matching `function_call_output`.
3. Resolve that `agent_id` to the spawned Codex child session.
4. Annotate the originating tool call with `subagent_session_id`.
5. Reuse the existing transcript/sidebar subagent UI so the child session can be expanded inline and grouped correctly.
6. Treat `<subagent_notification>` messages as system/intermediate bookkeeping:
   - either suppress them from the normal transcript
   - or fold their final status into tool result/status presentation instead of showing raw JSON

## Implementation Notes

Likely work items:

- extend Codex parsing to understand the `spawn_agent` / `function_call_output` / `wait` / `function_call_output` sequence
- add Codex-specific mapping from spawned `agent_id` to child Codex session ID
- annotate parsed tool calls with `subagent_session_id`
- avoid storing or rendering raw `<subagent_notification>` payloads as normal user transcript content
- ensure incremental sync does not lose Codex subagent mapping when the linking events arrive across sync boundaries

## Acceptance Criteria

- raw `<subagent_notification>` JSON no longer appears as a user message in Codex transcripts
- a parent Codex `spawn_agent` call links to the actual child Codex session when the child session exists locally
- Codex child sessions participate in the existing subagent UI and sidebar grouping
- behavior works for both full parse and incremental sync
- regression tests cover:
  - raw notification suppression/normalization
  - `spawn_agent` to child-session linkage
  - incremental append cases where the linking events land after an earlier sync

## Related Issues

- #136 introduced the generic subagent/team tree UI
- #162 already tracks incremental-sync edge cases around subagent linkage for Claude; Codex likely needs parallel treatment for its own event model

tpn:
Closes #233.

## Summary

- model Codex `spawn_agent` calls as real subagent-linked tool calls using existing `subagent_session_id` plumbing
- turn terminal Codex `wait` / `<subagent_notification>` status into structured tool output instead of raw user-lane JSON
- fall back to full parse only for Codex subagent-specific incremental events, and bump `dataVersion` so upgraded hosts auto-resync and re-export historical file-backed Codex sessions

## Details

This PR teaches the Codex parser to preserve the information needed for the existing subagent UI path instead of dropping it on the floor:

- `function_call` messages now retain `call_id` as `ToolUseID`
- `spawn_agent` is categorized as a `Task`
- `spawn_agent` outputs map the parent tool call to the child Codex session ID (`codex:<agent_id>`)
- `wait` terminal results become paired tool output
- `<subagent_notification>` becomes a fallback source of terminal output when no matching `wait` result was recorded
- non-terminal `running` notifications are ignored so they do not suppress later completion/error results

The frontend also hides raw `<subagent_notification>` transcript rows, so already-synced databases stop showing the JSON blobs immediately after upgrading.

## Upgrade / Rollout

- no SQLite schema rebuild required
- no PostgreSQL drop/recreate required
- `dataVersion` is bumped so upgraded hosts automatically perform a full local resync
- `agentsview pg push` already forces a full PG push after a local resync, so exporters will re-export automatically on the next full push path

Practical effect for self-hosted PG users:

1. upgrade the exporter binary on each host
2. run `agentsview pg push --full` once, or let the existing periodic exporter do the next full-sync cycle
3. keep the shared PG instance in place

Known limit:

- historical sessions whose raw Codex JSONL source file is already gone cannot be retroactively relinked, because the old SQLite archive never stored the dropped `function_call_output` linkage data

## Test Plan

- `CC=/usr/bin/gcc CXX=/usr/bin/g++ CGO_ENABLED=1 go test -tags fts5 ./internal/db -count=1`
- `CC=/usr/bin/gcc CXX=/usr/bin/g++ CGO_ENABLED=1 go test -tags fts5 ./internal/parser -count=1`
- `npm test -- --run src/lib/utils/messages.test.ts`

### DIFF
--- a/internal/parser/codex.go
+++ b/internal/parser/codex.go
@@ -278,6 +278,9 @@ func (b *codexSessionBuilder) handleSubagentNotification(
 		b.agentResults[agentID] = true
 		return true
 	}
+	if _, ok := b.pendingAgentResults[agentID]; ok {
+		return true
+	}
 	b.pendingAgentResults[agentID] = codexPendingResult{
 		text:      text,
 		timestamp: ts,
@@ -1057,6 +1060,8 @@ func codexIncrementalNeedsFullParse(line string) bool {
 
 	payload := gjson.Get(line, "payload")
 	switch payload.Get("type").Str {
+	case "function_call":
+		return payload.Get("name").Str == "wait"
 	case "function_call_output":
 		output, _ := parseCodexFunctionOutput(payload)
 		if !output.Exists() {
--- a/internal/parser/codex_parser_test.go
+++ b/internal/parser/codex_parser_test.go
@@ -238,6 +238,35 @@ func TestParseCodexSession_FunctionCalls(t *testing.T) {
 		assert.Equal(t, "continuing", msgs[3].Content)
 	})
 
+	t.Run("duplicate pending notification preserves earliest chronology", func(t *testing.T) {
+		childID := "019c9c96-6ee7-77c0-ba4c-380f844289d5"
+		summary := "Exit code: `1`\n\nFull output:\n```text\nTraceback...\n```"
+		notification := "<subagent_notification>\n" +
+			"{\"agent_id\":\"" + childID + "\",\"status\":{\"completed\":\"Exit code: `1`\\n\\nFull output:\\n```text\\nTraceback...\\n```\"}}\n" +
+			"</subagent_notification>"
+		content := testjsonl.JoinJSONL(
+			testjsonl.CodexSessionMetaJSON("fc-subagent-notify-dupe-order", "/tmp", "user", tsEarly),
+			testjsonl.CodexMsgJSON("user", "run a child agent", tsEarlyS1),
+			testjsonl.CodexFunctionCallWithCallIDJSON("spawn_agent", "call_spawn", map[string]any{
+				"agent_type": "awaiter",
+				"message":    "Run the compile smoke test",
+			}, tsEarlyS5),
+			testjsonl.CodexFunctionCallOutputJSON("call_spawn", `{"agent_id":"`+childID+`","nickname":"Fennel"}`, tsLate),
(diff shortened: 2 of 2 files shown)


tpn:
Im going back to the drawing board on the Codex subagent threading so this lands as a coherent fix instead of another incremental patch. The first few rounds showed that trying to force this workflow into the existing single-result pairing model is still leaving chronology/ownership edge cases, so Im reworking it more comprehensively before the next push.

Issue #259, opened 2026-04-01
bug: pinned messages lost after session resync


MCBoarder289:
Pinned messages are silently deleted whenever a session goes through a full message replacement sync. The pin is gone after the next server restart or when the SSE watcher triggers a resync on an active session.

**Root Cause Research**
`ReplaceSessionMessages` in `internal/db/messages.go` deletes all messages for a session and re-inserts them from the parsed source file. The `pinned_messages` table has a foreign key `message_id ... REFERENCES messages(id) ON DELETE CASCADE`, so the delete cascades and silently wipes all pins. The `ordinal` column in `pinned_messages` already exists as a stable re-attachment key but seems to not be used for this purpose.

`ReplaceSessionMessages` is called in two situations:

   - All OpenCode sessions — on every SyncAll (every 15 minutes, or on manual sync)
   - Any agent via SyncSingleSession — triggered by the SSE watcher fallback when a session file changes while the detail view is open and the watcher hasn't caught up within 5 seconds

**Reproduction Steps**
(note: I'm using copilot cli sessions, but feel free to try this with any other agent)

 1. Open a Copilot session in agentsview and pin a message
 2. Navigate to that session's detail view and leave it open
 3. In a terminal, touch the session file to simulate a file change: `touch <path-to-copilot-session-file>`
 4. Wait ~5 seconds for the SSE fallback sync to fire
 5. Refresh — the pin is gone

**Proposed Fix**
Before deletion, save (ordinal, note, created_at) for all existing pins. After the new messages are inserted, re-attach each saved pin by matching ordinal to the new message row. Pins for ordinals that no longer exist are dropped. Will put up a PR to showcase this and update this description

PR #260, opened 2026-04-01
 fix: preserve pinned messages after session resync (#259)

MCBoarder289:
ReplaceSessionMessages deleted all messages for a session and relied on the ON DELETE CASCADE FK on pinned_messages.message_id to clean up pins. This meant that every call to writeSessionFull (OpenCode sessions and explicit single-session re-syncs) silently wiped all user pins.

The ordinal column already exists in pinned_messages as a stable key for exactly this scenario. Fix: save (ordinal, note, created_at) for any existing pins before deletion, then re-attach them by matching ordinal to the newly inserted message rows. Pins for ordinals that no longer exist in the replacement set are silently dropped.

Adds two regression tests:
- TestReplaceSessionMessagesPreservesPins: pin + note survive replace
- TestReplaceSessionMessagesDropsPinsForRemovedOrdinals: pin for a removed ordinal is dropped, surviving ordinal is re-linked

Addresses #259

### DIFF
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -1228,17 +1228,31 @@ func TestReplaceSessionMessagesPreservesPins(t *testing.T) {
 	insertMessages(t, d,
 		userMsg("s1", 0, "msg0"),
 		asstMsg("s1", 1, "msg1"),
+		userMsg("s1", 2, "msg2"),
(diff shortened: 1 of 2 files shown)


wesm:
LGTM. thanks!

PR #272, opened 2026-04-02
parser: support OpenClaw toolCall blocks alongside tool_use

dfkoh:
OpenClaw JSONL sessions use "toolCall" content blocks with an "arguments" field instead of the Claude Code "tool_use" / "input" shape. This caused tool calls from OpenClaw sessions to be silently skipped in ExtractTextContent and to render as plain "[Tool: name]" strings instead of rich labels like "[Read: /etc/hosts]".

Changes:
- internal/parser/content.go: add case "toolCall" in ExtractTextContent that mirrors the existing tool_use logic but reads from block.Get("arguments") instead of block.Get("input").
- internal/parser/content.go: extract resolveToolInput() helper that returns the "input" field and falls back to "arguments" when input is absent or null.
- internal/parser/content.go: update formatToolUse() to call resolveToolInput() so rich formatting (e.g. "[Read: path]", "[Bash]\n$ cmd") works for both formats.
- internal/parser/openclaw_test.go: add TestParseOpenClawSession_ToolCallBlock covering the full toolCall + toolResult pairing flow.
- internal/parser/openclaw_test.go: add TestParseOpenClawSession_ToolCallFormatting verifying rich label output for toolCall blocks.
- internal/parser/openclaw_test.go: add TestParseOpenClawSession_ToolUseBackwardsCompat confirming the existing tool_use format continues to work.

All existing tests continue to pass.

### DIFF
--- a/internal/parser/content.go
+++ b/internal/parser/content.go
@@ -64,6 +64,30 @@ func ExtractTextContent(
 				toolCalls = append(toolCalls, tc)
 			}
 			parts = append(parts, formatToolUse(block))
+		case "toolCall":
+			// OpenClaw format: uses "toolCall" type and "arguments"
+			// instead of "tool_use" type and "input".
+			hasToolUse = true
+			name := block.Get("name").Str
+			if name != "" {
+				tc := ParsedToolCall{
+					ToolUseID: block.Get("id").Str,
+					ToolName:  name,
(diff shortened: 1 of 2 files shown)


wesm:
no problem, you can open a PR when you are ready

PR #300, opened 2026-04-10
fix: add Intel macOS desktop builds

### LINKED ISSUE #299: Desktop app: add macOS Intel (x86_64) build
## Summary

The desktop app currently only ships an Apple Silicon (`aarch64`) `.dmg` on the
Releases page. Would it be possible to also publish an Intel Mac (`x86_64` /
`amd64`) desktop build, or a universal `.dmg` that runs on both architectures?

Please let me know if I am missing something here.

## Context

The Go CLI is already cross-compiled for Intel Macs
(`agentsview_<version>_darwin_amd64.tar.gz`), so Intel Mac users can run
agentsview today via the CLI + browser. The gap is specifically the native
Tauri desktop wrapper, which is published as arm64-only.

## Why it would help

- Intel Macs are still in active use and will be for some time.
- The CLI + browser path works, but the desktop app experience (sidecar
  server, auto-updates, native window) is nicer for non-terminal users.
- A universal binary (via `lipo`) would cover both architectures with a
  single `.dmg`, at the cost of a larger bundle.

## Options

1. Add a separate `AgentsView_<version>_x86_64.dmg` alongside the existing
   `aarch64` build.
2. Ship a universal `.dmg` (`universal-apple-darwin` Tauri target) instead of
   two separate files.
3. Document the source-build path for Intel Mac users as a supported
   workaround if a prebuilt binary isn't feasible.

Happy with whichever approach fits the project's CI and signing setup best.
Thanks for the great tool!

mariusvniekerk:
## Summary
- add Intel macOS desktop builds alongside Apple Silicon builds without requiring Intel mac runners in CI
- fix desktop workflow target handling so Windows, Linux, macOS aarch64, and macOS x86_64 all build cleanly in PR CI
- add Linux arm64 desktop builds on `ubuntu-22.04-arm` for PR artifact coverage and tagged release AppImage assets while keeping the updater manifest Linux x86_64-only
- document the additional macOS and Linux release artifacts and add workflow regression checks for the new build matrix

Closes #299.

## Test Plan
- bash desktop/scripts/test-prepare-sidecar.sh
- bash desktop/scripts/test-desktop-workflows.sh
- bash desktop/scripts/test-startup-ui.sh
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/desktop-artifacts.yml"); YAML.load_file(".github/workflows/desktop-release.yml")'
- CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 go build -buildvcs=false -tags fts5 -trimpath -o /tmp/agentsview-darwin-amd64-check ./cmd/agentsview
- CGO_ENABLED=1 GOOS=darwin GOARCH=arm64 go build -buildvcs=false -tags fts5 -trimpath -o /tmp/agentsview-darwin-arm64-check ./cmd/agentsview
- cd desktop && GOFLAGS=-buildvcs=false TAURI_ENV_TARGET_TRIPLE=x86_64-apple-darwin npm run prepare-sidecar
- cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib --target x86_64-apple-darwin --no-run
- GitHub Actions PR checks for Desktop Artifacts and CI


mariusvniekerk:
References the tracking issue: #299

mariusvniekerk:
given the CI passes the roborev comment here is false.

PR #301, opened 2026-04-10
feat: add OpenHands CLI session support and shallow watch mode

rajshah4:
## Summary

Add support for parsing and syncing OpenHands CLI conversations, plus shallow-watch mode so conversation-based agents don't exhaust inotify file descriptor limits.

OpenHands stores local conversations as directories under
`~/.openhands/conversations`, with session metadata in `base_state.json`
and per-event JSON files under `events/`.

### Backend

- **Parser** (`internal/parser/openhands.go`): add OpenHands session discovery and parsing
  - discovers OpenHands conversation directories
  - parses `base_state.json`
  - parses `events/*.json`
  - maps message, action, and observation events into the existing session/message model
  - derives cwd/project from workspace and observation metadata
  - computes synthetic file metadata for conversation directories

- **Types** (`internal/parser/types.go`): add `AgentOpenHands` constant, registry entry, and `ShallowWatch` field on `AgentDef`

- **Sync Engine** (`internal/sync/engine.go`): add OpenHands sync support
  - full discovery of OpenHands conversation directories
  - changed-path classification for `base_state.json`, `TASKS.json`, and `events/*.json`
  - remaps file watcher updates back to the owning conversation directory

- **Watcher** (`internal/sync/watcher.go`): add `WatchShallow()` for non-recursive directory watching

- **Startup** (`cmd/agentsview/main.go`): route agents with `ShallowWatch=true` through `WatchShallow` and report shallow counts in the "Watching N directories" message

### Frontend

- **Agent Definitions** (`frontend/src/lib/utils/agents.ts`): add OpenHands label/color
- **Settings UI** (`frontend/src/lib/components/settings/AgentDirSettings.svelte`): show OpenHands configured directory support

### Documentation / CLI

- **README.md**: add OpenHands CLI to supported agents table
- **CLI help** (`cmd/agentsview/main.go`): add `OPENHANDS_CONVERSATIONS_DIR`

## Session Discovery

- **Default directory**: `~/.openhands/conversations`
- **Environment variable**: `OPENHANDS_CONVERSATIONS_DIR`
- **Config key**: `openhands_dirs`
- **ID prefix**: `openhands:`

## Shallow Watch

[...]

PR #303, opened 2026-04-10
feat: add shallow watch mode for agents with many subdirectories

### LINKED ISSUE #302: File watcher exhausts file descriptors with many subdirectories
## Problem

When an agent has many subdirectories (e.g., thousands of conversation directories), the file watcher recursively adds a watch to each one, which can:

1. Exhaust system file descriptor limits (`too many open files` error)
2. Cause slow startup times
3. Use excessive system resources

For example, with ~1,300 OpenHands conversation directories (each with an `events/` subdirectory), the watcher tries to watch ~2,600+ directories, which exceeds macOS's default file descriptor limits.

## Proposed Solution

Add a `ShallowWatch` mode that:
- Only watches the root directory for new subdirectories being created
- Relies on the existing periodic sync (every 15 minutes) for changes within subdirectories
- Drastically reduces file descriptor usage from O(n) to O(1) per agent directory

This is acceptable because:
1. Most subdirectories are static historical records
2. New conversations are still detected immediately
3. Changes to active conversations are picked up within 15 minutes (or less with manual refresh)

## Impact

Before: `Watching 1686 directories` → `socket: too many open files` error
After: `Watching 74 directories (2 shallow)` → works correctly

---
*This issue was created by an AI assistant (OpenHands) on behalf of @jpshackelford.*

jpshackelford:
## Summary

Adds a `ShallowWatch` option for agents with many subdirectories to prevent exhausting file descriptor limits.

## Problem

When an agent has thousands of subdirectories (e.g., conversation-based agents), the file watcher recursively adds a watch to each one, which can:
- Exhaust system file descriptor limits (`too many open files` error)
- Cause slow startup times  
- Use excessive system resources

## Solution

This PR adds:
- `ShallowWatch` field to `AgentDef` - when true, only the root directory is watched
- `WatchShallow()` method on `Watcher` for non-recursive directory watching
- Updated `startFileWatcher` logic to use shallow watch when enabled

When `ShallowWatch` is enabled:
- New subdirectories are detected immediately via file system events
- Changes within existing subdirectories are detected by periodic sync (every 15 min)
- File descriptor usage is reduced from O(n) to O(1) per agent directory

The output message now indicates how many directories are using shallow watch:
```
Watching 74 directories for changes (2 shallow) (76ms)
```

## Testing

All existing tests pass. The shallow watch behavior integrates with the existing periodic sync mechanism.

Fixes #302

---
*This PR was created by an AI assistant (OpenHands) on behalf of @jpshackelford.*

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -507,8 +507,9 @@ func startFileWatcher(
 	}
 
 	type watchRoot struct {
-		dir  string
-		root string // actual path passed to WatchRecursive
+		dir     string
+		root    string // actual path passed to WatchRecursive
+		shallow bool   // use shallow watch (root only)
 	}
 
 	var roots []watchRoot
@@ -520,7 +521,7 @@ func startFileWatcher(
 			if len(def.WatchSubdirs) == 0 {
 				if _, err := os.Stat(d); err == nil {
 					roots = append(
-						roots, watchRoot{d, d},
+						roots, watchRoot{d, d, def.ShallowWatch},
 					)
 				}
 				continue
@@ -529,7 +530,7 @@ func startFileWatcher(
 				watchDir := filepath.Join(d, sub)
 				if _, err := os.Stat(watchDir); err == nil {
 					roots = append(
-						roots, watchRoot{d, watchDir},
+						roots, watchRoot{d, watchDir, def.ShallowWatch},
 					)
 				}
 			}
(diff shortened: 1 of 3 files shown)


jpshackelford:
Re: the review comment about ShallowWatch:True not being used. This issue was discovered during testing of #301 and it would benefit from having this mode available.

wesm:
I folded this into #301

PR #304, opened 2026-04-11
fix(desktop): auto-recover from dead WebView after sleep/wake

wesm:
## Summary

- Add two-layer defense against WKWebView content process death on macOS
- **Layer 1 (Rust):** On window focus, `eval("1")` detects dead content process (returns Err) and force-navigates to reload, spawning a fresh WebKit process
- **Layer 2 (Frontend):** `visibilitychange` handler pings `/api/v1/version` when page becomes visible; reloads if backend is unreachable (covers alive-but-disconnected WebViews)

## Root cause

macOS kills WKWebView content processes during sleep/wake cycles or under memory pressure. The Tauri wrapper and Go sidecar stay alive, but the WebView goes black with no recovery mechanism. Confirmed by observing: backend responds to curl, WebContent process has 23MB RSS and zero TCP connections after 20 hours.

## Test plan

- [ ] Frontend: `npx vitest run src/lib/utils/health.test.ts` (7 tests covering reload on failure, debounce, cleanup, hidden skip, non-OK status, correct URL)
- [ ] Rust: `cd desktop/src-tauri && cargo check` (compiles clean)
- [ ] Manual: run desktop app, sleep/wake Mac, verify app recovers on focus

### DIFF
--- a/desktop/src-tauri/src/lib.rs
+++ b/desktop/src-tauri/src/lib.rs
@@ -557,22 +557,21 @@ fn desktop_redirect_url(port: u16) -> String {
 /// JS pings the backend and reloads on failure — covers
 /// alive-but-disconnected WebViews.
 fn recover_webview(window: &WebviewWindow, port: u16) {
-    // Read the auth token from localStorage (same key the frontend
-    // API client uses) so the health probe succeeds in authenticated
-    // remote-access mode.
-    // Only reload on network failure, timeout, or 5xx. A 401/403
-    // means the backend is alive — auth recovery is handled by the
-    // frontend, not by reloading.
-    // On failure, navigate to the absolute backend URL (not
-    // location.reload) so recovery works even if the sidecar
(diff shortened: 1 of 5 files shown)


wesm:
"Neither finding warrants a change here:

  1. Remote connection check at mount time — Switching from local sidecar to remote server without a
  page reload is not a real flow. The disconnect/reset button calls window.location.reload(). The
  lazy getBaseUrl() already handles URL changes for the active connection target. Moving the check
  inside the handler adds branching for a scenario that can't happen.
  2. 404 treated as healthy — A 404 from the backend's own server proves the backend is alive and the
   HTTP stack works. The health check exists to detect dead WebViews and unreachable backends, not to
   validate endpoint routing. Only network failures, timeouts, and 5xx (server is broken) should
  trigger a reload. Reloading on 404 would cause loops if the endpoint were temporarily
  misconfigured."

Issue #305, opened 2026-04-11
Feature Request: Filter Pinned Messages


MCBoarder289:
I have been leveraging pinned messages often, and now that I've been working across multiple projects/repos, it would be nice to be able to filter pinned messages by project.

Will put up a quick PR to wire this into the frontend. My main idea is to simply hook up the projects filter dropdown so that the pinned messages screen listens to it.

MCBoarder289:
Here is that PR: https://github.com/wesm/agentsview/pull/306

MCBoarder289:
Closed as part of the 0.20.0 release. Thanks, @wesm!

PR #306, opened 2026-04-11
feat: filter pinned messages by selected project (#305)

MCBoarder289:
When a project is selected in the header dropdown, the pinned messages page now filters to only show pins from sessions in that project. The count badge reflects the filtered count, and a dedicated empty state is shown when the filter yields no results.


MCBoarder289:
Solid feedback - pushing up an update that includes backend changes

PR #309, opened 2026-04-12
perf(usage): faster usage daily, default to 30-day window

wesm:
Two followups to #308.

## Summary
- **perf: parse `token_usage` in Go instead of SQLite's `json_extract`.** `GetDailyUsage` was running 1.2M `json_extract` calls per 310k-row scan. Switch to a single raw `token_usage` fetch + `gjson.Parse` in Go, and hoist the `model_pricing` LEFT JOIN into an in-memory map loaded once before the scan (the pricing table has <20 rows). Query drops from ~525ms to ~270ms on a 22k-session / 310k-message DB.
- **feat: default `usage daily` to last 30 days, add `--all` flag.** Running `usage daily` without a date range previously scanned full history. Most callers only care about recent spend, so default to a 30-day window and let users opt into full history with `--all`. The date filter also pushes down into the SQL `WHERE` clause.

Net effect on `agentsview usage daily --json --offline --no-sync`: 530ms → 320ms (40% faster). With quick sync enabled: 720ms → 550ms.

## Test plan
- [ ] `agentsview usage daily` returns last 30 days by default
- [ ] `agentsview usage daily --all` returns full history
- [ ] `agentsview usage daily --from=... --to=...` still honors explicit ranges
- [ ] `make test` passes

### DIFF
--- a/cmd/agentsview/usage.go
+++ b/cmd/agentsview/usage.go
@@ -52,6 +52,26 @@ func runUsage(args []string) {
 // full history when users usually want recent spend.
 const defaultUsageDays = 30
 
+// resolveDefaultSince returns the effective --since value,
+// applying a 30-day lookback only when the caller gave no
+// explicit range at all. If --until is set we leave --since
+// empty so "everything up to --until" still works; otherwise
+// a bare --until would produce From > To and empty results.
+func resolveDefaultSince(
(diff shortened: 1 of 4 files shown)


wesm:
No this is fine

PR #322, opened 2026-04-14
feat: SSH remote session sync

wesm:
## Summary

- Add `agentsview sync --host <host> [--user <user>] [--port <port>] [--full]` to pull session data from remote machines over SSH
- Shells out to `ssh` CLI for transport (inherits tailscale keys, SSH config, agent forwarding)
- Two SSH calls per sync: resolve remote agent dirs, then tar full directory trees to a temp dir
- Delegates to the existing `sync.Engine` for discovery, parsing, and DB writes (same code path as local sync)
- Session IDs prefixed with `host~` to prevent cross-machine collisions; file paths stored as `host:/remote/path`
- Engine runs in ephemeral mode: no sync timestamp recording, no local skip cache pollution
- Per-host remote skip cache (`remote_skipped_files` table) for incremental sync
- Docker-based integration tests (`make test-ssh`) with sshd container behind `sshtest` build tag


wesm:
"Both findings are false positives given the threat model.
High (tar traversal): The SSH target is a machine YOU own on YOUR tailscale network. You're explicitly choosing to sync from it with --host. If that machine is compromised, the attacker already has your tailscale access — overwriting local files via crafted tar entries is the least of your problems. This is the same trust model as scp or rsync. The project guidelines (point 12) explicitly exclude TOCTOU on local/user-owned paths.
Medium (SSH arg injection): --host and --user are CLI flags typed by the user, not external input. The user is running agentsview sync --host devbox1 — they control the value. A user who types --host "-oProxyCommand=rm -rf /" is attacking themselves. Same reason ssh itself doesn't validate this.
Both findings apply to internet-facing services with untrusted input, not a single-user CLI tool syncing from your own machines."

PR #324, opened 2026-04-14
feat(cli): migrate command tree to Cobra

mariusvniekerk:
## Summary
- migrate agentsview CLI command dispatch from manual stdlib flag handling to a Cobra command tree
- reorganize root help into grouped sections, include usage commands, and collapse completion output to top-level command only
- size root flag help to terminal width with x/term fallback to 80 and cap at 160 columns

## Validation
- go test ./...

## Notes
- branch rebased onto current upstream/main before final wiring so usage/pricing commands stay in parity with main


mariusvniekerk:
CLI API break note: this Cobra migration drops legacy single-dash long-flag parsing (for example, `-host`/`-port`). CLI now follows pflag/Cobra semantics: long flags require `--flag`; single-dash forms are shorthand-only. Existing scripts using old stdlib-style long flags will need updates.

mariusvniekerk:
Docs follow-up PR opened: wesm/agentsview-docs#26

It updates docs/examples/screenshot helper for Cobra long-flag syntax and fixes stale `-browser` references to `--no-browser`.

PR #327, opened 2026-04-14
feat: add PostgreSQL usage support; rename remote_access to require_auth

### LINKED ISSUE #326: Implement usage view support for PostgreSQL read-only mode
## Problem

The multi-host PostgreSQL feature exposes session-level token fields, but the dedicated `/usage` view does not work there.

Today the PostgreSQL store explicitly returns `db.ErrReadOnly` for:
- `GetDailyUsage`
- `GetTopSessionsByCost`
- `GetUsageSessionCounts`

That makes the usage API endpoints return `501 not available in remote mode`. As a temporary UI safeguard, the read-only PG UI no longer shows the Usage route, but the real gap is missing PostgreSQL-backed usage queries.

## Goal

Implement usage parity for the PostgreSQL read-only store so the `/usage` page works.

## Required work

- Implement PostgreSQL versions of:
  - `Store.GetDailyUsage`
  - `Store.GetTopSessionsByCost`
  - `Store.GetUsageSessionCounts`
- Preserve SQLite semantics:
  - include/exclude filters for project/agent/model
  - timezone-aware date bucketing with padded UTC bounds
  - Claude dedup by `(claude_message_id, claude_request_id)`
  - pricing joins / cost calculations
  - per-day model/project/agent breakdowns
  - top sessions ordering and limits
  - session counts parity with daily totals
- Add PostgreSQL integration tests (`pgtest`) that mirror the existing SQLite usage behavior for parity-critical cases.
- Re-enable the Usage route in the read-only PG UI once the API endpoints are implemented and tested.

## Notes

Relevant code:
- `internal/postgres/store.go`
- `internal/db/usage.go`
- `internal/server/usage.go`
- `frontend/src/lib/components/layout/AppHeader.svelte`
- `frontend/src/lib/components/layout/ThreeColumnLayout.svelte`
- `frontend/src/App.svelte`

tpn:
Closes #326.

## Summary

- Add PostgreSQL-backed `/usage` queries for read-only mode
  (`GetDailyUsage`, `GetTopSessionsByCost`, `GetUsageSessionCounts`)
- Add additive PostgreSQL `model_pricing` schema support and sync
  pricing rows during `agentsview pg push`
- Fall back to built-in pricing rows on PG-backed viewers so
  already-deployed instances keep returning usage data before every
  host has pushed fresh pricing metadata
- Rename `remote_access` config to `require_auth` and make token
  auth opt-in instead of auto-enabled for non-loopback binds

## PG Usage

Closes the parity gap where session-level token fields were already
visible on PG-backed multi-host views, but the dedicated `/usage`
page returned `501 not available in remote mode`.

The implementation keeps the same high-level semantics as the SQLite
usage pipeline: timezone-aware date bucketing, Claude dedup by
`(claude_message_id, claude_request_id)`, include/exclude filters for
project/agent/model, per-day breakdowns, top-session ranking by
computed cost, and aligned session-count totals.

On the storage side, PostgreSQL gets an additive `model_pricing`
table. `pg push` syncs local pricing rows into PG before usage
queries run. PG viewers also overlay built-in fallback pricing so
upgraded read-only deployments remain usable before every exporter
has refreshed pricing data.

## Auth Rename

`remote_access` was a confusing name for token auth enablement, and
it auto-enabled whenever binding to a non-loopback address -- forcing
users on private networks (Tailscale, VPN, LAN) to deal with token
auth they don't need.

Changes:
- Renamed `remote_access` to `require_auth` across Go, TypeScript,
  and Svelte
- Auth is now opt-in: set `require_auth = true` in config or pass
  `--require-auth` on the CLI
- Legacy `remote_access` TOML key still recognized for backward
  compatibility (deleted from persisted config when `require_auth`
  is saved)
- Auto-bind to 0.0.0.0 skipped when managed Caddy proxy is active
  to preserve the loopback-backend invariant

## Upgrade / Rollout

- PostgreSQL: additive schema change only; no drop/recreate required
- Exporters: upgrade host binaries and run `agentsview pg push`
- PG-backed viewers can start serving `/usage` immediately after
  upgrading the reader binary
- Existing `remote_access = true` configs continue to work

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -11,9 +11,9 @@ Key assumptions reviewers MUST account for:
 1. AUTH MODEL: In local-only mode, loopback bind is the access
    boundary — no auth tokens needed. In proxy mode, validateServeConfig
    forces the backend to loopback; Caddy enforces CIDR subnet
-   allowlisting. When remote access is enabled (`remote_access: true`),
-   a bearer token is required for all API requests (including
-   localhost, to prevent bypass via reverse proxy). An auth token is
+   allowlisting. When auth is required (`require_auth: true`), a
+   bearer token is required for all API requests (including localhost,
+   to prevent bypass via reverse proxy). An auth token is
    auto-generated at startup if missing. Do not flag missing auth on
    local-only code paths. DO flag any path that lets the backend bind
    non-loopback in proxy mode, or missing subnet checks for
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -131,10 +131,12 @@ func runServe(cfg config.Config) {
 	// background LiteLLM refresh follows immediately.
 	seedPricing(database)
 
-	// When auth is required, ensure a token exists and auto-bind to
-	// 0.0.0.0 so the server is reachable from the network.
+	// When auth is required, ensure a token exists. Auto-bind to
+	// 0.0.0.0 unless the user set --host explicitly or a managed
+	// proxy is active (Caddy requires a loopback backend).
 	if cfg.RequireAuth {
-		if !cfg.HostExplicit && cfg.Host == "127.0.0.1" {
+		if !cfg.HostExplicit && cfg.Host == "127.0.0.1" &&
+			cfg.Proxy.Mode == "" {
 			cfg.Host = "0.0.0.0"
 		}
 		if err := cfg.EnsureAuthToken(); err != nil {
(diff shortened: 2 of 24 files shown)


wesm:
I'm picking this up to get the CI passing

wesm:
Testing this out now. I noticed that the pg serve forces the use of an auth token when bound to non-loopback. I am running mine on Tailscale and so I feel like the auth token thing should always be opt-in, not forced on the user, since if you are a Tailscale user that is responsible for securing the network and so the auth token stuff is a nuisance

wesm:
As intended.

tpn:
Thanks @wesm!

Issue #328, opened 2026-04-15
Bug w/ fix(insights): Claude Code CLI outputs JSON array, not single JSON object


jesserobbins:
## Bug

`insights generate` with `agent=claude` always fails with **\"claude returned empty result\"**.

## Root Cause

`generateClaude` in `internal/insight/generate.go` expected `claude -p --output-format json` to return a single JSON object:

\`\`\`json
{"result": "...", "model": "..."}
\`\`\`

But Claude Code CLI (v2.1.x+) returns a **JSON array** of structured events:

\`\`\`json
[
  {"type": "system", "subtype": "init", ...},
  {"type": "assistant", "message": {...}, ...},
  {"type": "result", "result": "...", "modelUsage": {"claude-haiku-4-5-20251001": {...}}}
]
\`\`\`

`json.Unmarshal` into `struct { Result string; Model string }` silently fails on an array, so `result` is always empty, triggering the error.

## Fix

Added `parseCLIJSONArray` which unmarshals the output as `[]json.RawMessage`, finds the element with `"type":"result"`, and extracts `result` and the model name from `modelUsage`. Updated all affected tests to use the correct array format.

See PR / commit for details.

PR #329, opened 2026-04-15
fix(insights): correctly parse Claude Code CLI JSON array output

jesserobbins:
## Problem (identified in #328)

The insights feature invokes the Claude Code CLI with `-p --output-format json`
to generate session summaries. Claude Code v2+ changed the output format from a
single JSON object to a **JSON array of events**, where the actual result is
carried in the element with `type: "result"`:

```json
[
  { "type": "system_prompt", ... },
  { "type": "turn", ... },
  { "type": "result", "result": "...", "modelUsage": { "claude-opus-4-5": {} } }
]
```

The previous parser expected the legacy single-object format:

```json
{ "result": "...", "model": "claude-opus-4-5" }
```

As a result, insight generation silently returned an empty result for any
installation running Claude Code v2+, falling through to the error path with a
confusing "no usable output" message.

## Solution

Extract the parsing logic into a dedicated `parseCLIResult` function that:

1. **Tries the JSON array format first** — iterates events, finds the element
   with `type: "result"`, reads its `result` field, and derives the model name
   from the first (sorted) key of `modelUsage`.
2. **Falls back to the legacy single-object format** — preserves compatibility
   with older Claude CLI versions that still emit `{"result":"...","model":"..."}`.

The model-name extraction from `modelUsage` sorts the keys before selecting so
the result is deterministic when multiple models are reported.

## Changes

- `internal/insight/generate.go` — replaced the inline JSON unmarshal with a
  call to `parseCLIResult`; added `parseCLIResult` with array-first / legacy
  fallback logic.
- `internal/insight/generate_test.go` — updated all existing test fixtures from
  the legacy format to the array format; added a new `TestParseCLIResult` table
  test covering both formats, edge cases (no result event, empty input, garbage
  input).

## Testing

All existing tests pass with the updated fixtures. The new `TestParseCLIResult`
test covers:

- Array format with a `result` event
- Legacy single-object format (backward compat)
- Array with no `result` event
- Empty input
- Non-JSON garbage input

### DIFF
--- a/internal/insight/generate.go
+++ b/internal/insight/generate.go
@@ -245,16 +245,17 @@ func generateClaude(
 		)
 	}
 
-	var resp struct {
-		Result string `json:"result"`
-		Model  string `json:"model"`
-	}
-	if json.Unmarshal(stdoutBytes, &resp) == nil &&
-		strings.TrimSpace(resp.Result) != "" {
+	// Claude Code CLI outputs a JSON array of events when
+	// invoked with -p --output-format json. Find the element
+	// with type="result" and extract its result field.
+	// Also accept the legacy single-object format as a
+	// fallback for older Claude CLI versions.
+	content, model := parseCLIResult(stdoutBytes)
+	if strings.TrimSpace(content) != "" {
 		return Result{
-			Content: resp.Result,
+			Content: content,
 			Agent:   "claude",
-			Model:   resp.Model,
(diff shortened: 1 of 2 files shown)


wesm:
Ah, thanks for fixing this. I guess I need to set up some integration tests for this part of the app to catch things like this

PR #333, opened 2026-04-15
fix: preserve filters when switching between Sessions and Usage tabs

mariusvniekerk:
## Summary

- Session and usage filters now persist across tab switches via localStorage
- Filters are JSON-serialized typed structs (`Filters` / `UsageFilterState`), loaded on store construction and saved on every load/fetchAll call
- The route-change effect in App.svelte only calls `initFromParams` on the sessions route, so switching to Usage/Insights/etc never resets session filters
- URL params still work for deep-links on initial page load


wesm:
I have agent addressing this last review

PR #353, opened 2026-04-18
feat: agentsview session CLI — programmatic surface over session data

wesm:
## Summary

- Adds `internal/service.SessionService` — a single interface (`Get`/`List`/`Messages`/`ToolCalls`/`Sync`/`Watch`) with two backends: a direct SQLite/PG-read backend and an HTTP backend that proxies to a running daemon. Both HTTP handlers and the new CLI route through this layer, so the on-disk and wire shapes stay in sync.
- Adds `agentsview session` CLI group: `get`, `list`, `messages`, `tool-calls`, `sync`, `export`, `watch`. Transport auto-detects a running daemon via the existing `server.<port>.json` state file; when none is present, the CLI opens the local archive directly. A `pg serve` daemon (read-only) is honored — `session sync` refuses with a clear error. Auth-enabled daemons (`require_auth: true`) are handled via `Authorization: Bearer <token>` threaded through `NewHTTPBackend`.
- Adds `GET /api/v1/sessions/{id}/tool-calls` and `POST /api/v1/sessions/sync` HTTP routes so CLI and web share the same DTOs.
- Extracts session polling from the SSE handler into `internal/sessionwatch` so HTTP SSE and CLI `watch` share the same logic, with a PG-read-only branch (`pollDBOnly`) for `pg serve`.
- Parser: promoted system subtypes (`continuation`, `resume`, `interrupted`, `task_notification`, `stop_hook`) now keep `Role=user` with `IsSystem=true`; `user_message_count` and `first_message` skip system-injected entries. Adds `messages.thinking_text` column with a non-destructive migration that preserves the column across full resync, plus PostgreSQL push/read wiring.
- Docs: `docs/cli-reference.md` covers every subcommand, stability guarantees, transport/auth behavior, and response shapes.

## Notes

Two findings from the roborev pass are deferred as follow-ups:
- `directBackend` `ToolCalls` still hydrates all messages before flattening — a dedicated `GetToolCalls` store method would avoid the extra work on large sessions.
- The `DirectReadOnly` transport fallback still calls `db.Open`, which runs migrations and version stamping against a DB the daemon owns. A read-only DSN open path that skips schema probe/migration would make the fallback truly read-only.


wesm:
That's an unrelated flake. merging this and i'll look into the flake separately

PR #357, opened 2026-04-19
feat: track file inode/device to validate incremental parses

andrewwowens:
Records `file_inode` and `file_device` on each session row alongside the existing size/mtime. Before running an incremental parse, compares the stored identity against the current file's; if it differs, skips the incremental path and full-parses instead.

This catches cases where a session file is replaced atomically (`mv`, `cp`, editor atomic write) — size and mtime alone can miss the swap, which causes the incremental parser to append unrelated bytes onto the prior session's state.

Unix captures identity from `syscall.Stat_t`. Windows stubs to `(0, 0)`, which disables the check and preserves current behavior.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -461,6 +461,14 @@ func (db *DB) migrateColumns() error {
 			"ALTER TABLE sessions ADD COLUMN is_truncated INTEGER NOT NULL DEFAULT 0",
 		},
 		{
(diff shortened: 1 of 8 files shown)


wesm:
Live updates were partially implemented in https://github.com/wesm/agentsview/pull/351, if you can bear with me I'll rebase your branch and see what else is in here that can be merged

PR #369, opened 2026-04-22
feat: broaden automated-session detection + Codex parser fix

wesm:
## Summary

Extends the automated-session classifier in `internal/db/automated.go` to recognize four new tool-prompt patterns that were previously polluting the sidebar and skewing insights:

- **Claude Code conversation title generator** — every user turn spawns an internal Claude Code session that runs a "You are a conversation title generator" prompt. Matched as a substring (the actual message is wrapped with a leading `-\n`).
- **Claude Code Warmup ping** — model warmup with first-message content `Warmup`. Matched via a new exact-match category (`automatedExactMatches`), trimmed of surrounding whitespace, since the word is too generic for prefix or substring matching.
- **Roborev review combiner** — `You are combining multiple code review outputs into a single GitHub PR comment.` Matched as a prefix.
- **agentsview changelog generator** — `You are generating a changelog for agentsview` (release tooling). Matched as a prefix.

The existing `user_message_count <= 1` single-turn gate still wraps the classifier, so multi-turn sessions never get classified as automated.

Also fixes a Codex parser bug that was excluding ~573 legitimately-automated Codex sessions from classification: the parser was counting Codex's synthetic `<turn_aborted>` "user" message (emitted when `codex exec` is interrupted during shutdown) as a real user turn, inflating `user_message_count` to 2 and falling through the single-turn gate. `<turn_aborted>` is now filtered like the other Codex synthetics (`<environment_context>`, `<INSTRUCTIONS>`, subagent notifications) at parse time, so it doesn't affect message counts or message lists.

`insight.BuildPrompt` now sets `ExcludeAutomated: true` on its `db.SessionFilter`, so daily/agent-analysis insights ignore automation noise. There is no UI override — the sidebar toggle remains the only way to surface automated sessions.

The SQLite and PostgreSQL backfill markers both bump from `_v2` to `_v3` to trigger a one-time re-classification on next open against the patterns added in the first commit on this branch (Claude Code title generator, Warmup, roborev combiner). The agentsview changelog generator pattern is not retroactively backfilled; new sessions matching it are classified at upsert time. The SQLite marker is hoisted into an exported `db.IsAutomatedBackfillMarker` constant so the test file no longer carries a separately-drifting literal. `batchUpdateAutomated` now also bumps `local_modified_at` on changed rows so incremental `pg push` picks up the re-classified sessions; without this, `pg serve` would converge via its own marker bump but `pg push` deployments would silently miss the changes until a `pg push --full`. The Codex sessions affected by the `<turn_aborted>` parser bug get re-classified through the normal sync path the next time their source `.jsonl` files are re-parsed; sessions whose source files are gone (or whose mtime hasn't changed since last index) keep their stale `user_message_count` and remain unclassified until a manual re-sync.

The sidebar toggle relabels from "Include automated reviews" to "Include automated sessions" — the category is no longer review-only. URL parameter (`?include_automated=true`), API field, store names, and DB columns are unchanged.

Follow-up tracked in #370 (user-configurable automated patterns in `config.toml`).


wesm:
Not relevant, the backfill markers are all endogenous to this change

PR #383, opened 2026-04-23
Config-driven automation prefixes and sidebar orphan-subagent fix

wesm:
### User-defined automation prefixes
Adds an `[automated]` section to `config.toml` so users can register their own first-message prefixes that mark a single-turn session as automated (e.g. custom review or grading prompts). Prefixes are normalized and held in a process-wide singleton so every store-open path sees the same set. A SHA-256 hash over the built-in patterns, algorithm version, and user prefixes is stored in `stats.is_automated_classifier_hash`; when the hash changes on next open, `backfillIsAutomatedLocked` recomputes `is_automated` across the sessions table and stamps the new hash. PostgreSQL mirrors the same logic over `sync_metadata`.

`agentsview classifier rebuild` forces a reclassification and clears the stored hash. The command group is `Hidden: true` in `--help`.

### Resync reclassification gap
`ResyncAll` opens a fresh temp DB whose at-Open backfill stamps the current classifier hash on an empty table. `CopyOrphanedDataFrom` then imports orphan rows verbatim, so their pre-resync `is_automated` values would be frozen against the new stamped hash forever. `ForceBackfillIsAutomated` clears the hash and re-runs the backfill; `ResyncAll` calls it after the orphan copy and metadata merge. `classifierAlgorithmVersion` is bumped so already-poisoned DBs heal on the next Open.

### Sidebar filter correctness
`buildSessionFilter` (SQLite and PG) now uses `WITH RECURSIVE` to select rows reachable from rootMatch-passing roots via `parent_session_id`. Subagent and fork rows no longer surface as standalone sidebar groups when their parent chain is broken (parent JSONL rotated off disk, parent filtered by `is_automated`, or parent excluded by the relationship guard).

### Smaller fixes
- `cmd/agentsview session ...` accepts bare session UUIDs and errors on missing IDs instead of silently returning empty output.
- Codex `<skill>` template injections are classified as system messages so they don't inflate `user_message_count` past the single-turn gate.
- `dataVersion` bump to 17 triggers resync for the codex `<skill>` filter.
- `ClassifierHashKey` is exported and shared between the SQLite and PG paths.


wesm:
Re: the roborev combined review flagging `baseWhere` as filtering out subagent/fork children — false positive.

`basePreds` only gets the `relationship_type NOT IN ('subagent', 'fork')` guard when `!f.IncludeChildren`:

```go
// internal/db/sessions.go:303-310 (same shape in internal/postgres/sessions.go:144-151)
basePreds := []string{
    "message_count > 0",
    "deleted_at IS NULL",
}
if !f.IncludeChildren {
    basePreds = append(basePreds,
        "relationship_type NOT IN ('subagent', 'fork')")
}
```

When `IncludeChildren=true`, `basePreds` is just `message_count > 0 AND deleted_at IS NULL`. The outer `baseWhere` does not re-filter the subagent/fork rows the CTE surfaces. The relationship guard only appears inside the CTE seed (rootMatch), which is exactly the "seed only" shape the reviewer suggested.

Existing tests on this path assert children come back:
- `TestIncludeChildrenBypassesFilters/AgentFilterBypassesChildren` expects `{parent, child-sub, child-fork}` (subagent AND fork)
- `TestIncludeChildrenKeepsNestedDescendants` expects `{root, sub, nested-fork}` (subagent + nested fork)
- `TestIncludeChildrenExcludesOrphanSubagents` expects `{root, root-sub}` (subagent returned)

All pass against `7318321`.

PR #388, opened 2026-04-24
Add NilAway lint checks

mariusvniekerk:
## Summary
- Add pinned NilAway installation and run it from `make lint`, `make lint-ci`, and CI.
- Split pre-commit hooks so `golangci-lint` and NilAway run as separate `prek` targets.
- Clear existing production NilAway diagnostics so the analyzer runs without a baseline exclusion list.

## Verification
- `make lint-ci`
- `prek run golangci-lint --all-files`
- `prek run nilaway --all-files`
- `go test -tags fts5 ./internal/parser ./internal/db ./internal/postgres ./internal/insight ./internal/sync ./internal/ssh ./internal/importer`
- `go test -tags fts5 ./internal/server -run 'Test(UploadSession|DetectTerminalLinux|ResolveResumeDir|ResolveCursorResumePaths|ResumeLaunchCwd|ResumeSession|GetSessionDirectory|ListOpeners|GetTerminalConfig|LaunchResumeDarwinGhostty|LaunchTerminalInDirGhostty)'`


wesm:
This finding is a false positive.

The two blocks look similar but use different map types:

- **First block** (`internal/db/usage.go:517`, `internal/postgres/usage.go:316`) — `map[string]*modelAccum` (pointer type). Nil guards there are correct and necessary.
- **Second block** (`internal/db/usage.go:634-651`, `internal/postgres/usage.go:429-444`) — `map[string]bucket`, where `bucket` is a value-type struct (defined at `internal/db/usage.go:327`). `left`, `right`, and `b` are values, not pointers — they can't be nil.

The suggested fix wouldn't even compile: `left == nil` on a `bucket` value errors with `cannot convert nil to type bucket`. The reviewer pattern-matched on surface similarity to the pointer-typed block and missed the type difference.

PR #389, opened 2026-04-24
Update Go to 1.26, fix arm64 desktop test flake and nilaway pre-commit hook

mariusvniekerk:
## Summary

- Update the module Go directive to 1.26.2 and configure golangci-lint for Go 1.26.
- Bump the golangci-lint pin and install hints to v2.11.4.
- Apply `golangci-lint run --fix ./...` Go 1.26 modernizations, including `new(value)` pointer helper inlining.
- Retry ETXTBSY (Linux execve race when a parallel test thread's fork briefly holds a writable fd for the temp shell script) in `run_login_shell_env_handles_large_stdout`, mirroring the existing retry in the sibling `run_login_shell_env_timeout_returns_when_stdout_fd_stays_open` test. Removes a recurring arm64 desktop CI flake.
- Strip every repo-local Git env var (via `git rev-parse --local-env-vars`) and disable VCS stamping before `golangci-lint custom` in `nilaway-golangci-build`. When `make nilaway` runs from a pre-commit hook, git exports `GIT_DIR`, `GIT_INDEX_FILE`, `GIT_CONFIG_PARAMETERS`, etc. pointing at the parent repo, which made the inner `git clone` and `go build` fail with exit 128.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,7 @@
 version: "2"
 run:
   tests: true
+  go: "1.26"
   build-tags:
     - fts5
 linters:
--- a/Makefile
+++ b/Makefile
@@ -292,18 +292,21 @@ lint-golangci-ci: ensure-embed-dir
 	golangci-lint run ./...
 
 # Build a custom golangci-lint binary with the NilAway module plugin.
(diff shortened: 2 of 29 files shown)


wesm:
Invalid

PR #395, opened 2026-04-24
fix(parser): filter synthetic copilot skill messages

### LINKED ISSUE #394: Copilot skills usage shows as user input
## Summary

For GitHub Copilot CLI sessions, any time the agent invokes a skill, agentsview displays the entire content of the skill as if it was a message from the user.

I've been upgrading regularly from main, and this is a newer behavior (last couple weeks maybe?).

### Invoking the gh-cli skill

Image

### Invoking the dignified-python skill

Image

## Versions

- GH Copilot CLI: v1.0.36 (current latest)
- Agentsview: v0.24.0-2-gf0189c7 (commit f0189c7, built 2026-04-24T21:35:26Z)

timoguin:
Skip Copilot user.message entries injected by skills when `data.source` starts with `skill-` or `data.content` starts with `<skill-context` so they do not appear as user transcript content or set firstMessage.

Add regression coverage for source-only, content-only, and combined synthetic skill messages, and bump dataVersion to 19 so existing databases resync and drop persisted skill definitions.

Fixes #394

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -27,15 +27,21 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
-// Bumped to 18: Claude parser now skips /clear and /effort
+// Bumped to 19: Copilot parser now filters synthetic skill
+// context user messages (source "skill-*" or content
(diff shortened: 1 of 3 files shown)


wesm:
LGTM!

PR #403, opened 2026-04-27
fix(parser): sync opencode sqlite sessions in hybrid roots

aaronflorey:
## Summary
- sync OpenCode SQLite sessions whenever `opencode.db` exists, even when `storage/` is also present in the same root
- fall back to SQLite virtual paths in hybrid roots so source lookup, single-session sync, and DB change handling keep working for sqlite-only sessions
- add parser and integration regression coverage for hybrid storage+SQLite OpenCode roots

## Testing
- go test ./internal/parser ./internal/sync

### DIFF
--- a/internal/parser/discovery.go
+++ b/internal/parser/discovery.go
@@ -178,6 +178,12 @@ func FindOpenCodeSourceFile(root, sessionID string) string {
 				return path
 			}
 		}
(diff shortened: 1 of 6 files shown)


wesm:
thank you, working on addressing reviews

PR #391, opened 2026-04-24
stats: make is_automated the authority across totals/archetypes/distributions/portfolio

wesm:
## Summary

- Project `is_automated` through `sessionStatsRow` so the stats pipeline consumes the flag instead of re-deriving automation from `userMessageCount`.
- Rename `archetypeLabel` to `sessionShapeLabel` and drop its automation branch; automation is now decided upstream by the flag.
- Switch `computeTotalsAndArchetypes`, `computeDistributions` (`scope_human`), and `computeAgentPortfolio` to the flag.
- Add human-scoped peer fields to `StatsAgentPortfolio`: `by_sessions_human`, `by_tokens_human`, `by_messages_human`, `primary_human`.
- `schema_version` stays at 1; new fields are additive and semantic tightening of existing fields is permitted in v1.

## Why

The `stats --format json` pipeline's "human vs automation" split was derived from `userMessageCount >= 2` — a proxy that predates the dedicated `is_automated` flag. 0.24.0 broadened `is_automated` classification (#369, #387) but those improvements never reached stats output. With the flag authoritative, future classifier improvements flow through automatically.

Design doc: `docs/superpowers/specs/2026-04-24-stats-automation-authority-design.md`
Plan: `docs/superpowers/plans/2026-04-24-stats-automation-authority.md`

## Test plan

- [x] New unit tests for flag-authority in totals, archetypes, distributions scope_human, and agent_portfolio human-scoped peers.
- [x] Existing tests that asserted the old `userMessageCount <= 1 → automation` heuristic updated to set `isAutomated: true` on fixtures whose comments already labeled them as automation.
- [x] Golden for `cmd/agentsview/TestStatsGolden` regenerated to reflect the new bucketing and the new human-scoped portfolio fields.
- [x] `make test` passes (except the known pre-existing `TestReplaceSessionMessages_LargeSession` FTS-table failure, which is unrelated).
- [x] `make lint` passes.

## Related

- tkmx-server: https://github.com/srosro/tkmx-server/pull/13 (downstream — consumes the new `by_sessions_human` / `primary_human` fields and falls back cleanly for legacy or mixed fleets.)
- tkmx-client: https://github.com/srosro/tkmx-client/pull/18 (independent bug fix for `REPORT_DAYS=1` runs scrubbing the `session_stats` blob; depends on this PR's schema only in the sense that both live in the same ingestion path.)

PR #404, opened 2026-04-27
ci: gate frontend with svelte-check and vitest

cpcloud:
## Summary

Adds a `frontend` CI job running `npm run check` and `npm test`. Previously only `npm run build` was gated, so type errors and test failures didn't block merges.

## Changes

- **CI:** new `frontend` job runs svelte-check + vitest on every push/PR.
- **Type errors:** clears 24 pre-existing svelte-check errors — 3 production (`HealthTrend.svelte`, `CostTimeSeriesChart.svelte`) and 21 test fixtures missing `Session.is_automated` (required since #391) or `TopSession.output_tokens`.
- **Warnings:** silences 5 svelte-check warnings — `ConfirmDeleteModal` overlay (Escape + Cancel already cover keyboard) and `VirtualizerTest`/`CacheTestWrapper` test scaffolding (capture-once + manual-sync pattern is intentional).

## Latent test bugs uncovered

`npm test` appears to have never run in CI — the 999-test frontend suite was orphaned. Three problems surface once it runs:

- **`TopSessions.test.ts > sets filter and navigates...`** — was failing every run. `sessions.filters.includeOneShot` defaults to `true` in the store; test never reset it, so the click handler's invalidation branch never fired.
- **`sessions.test.ts > discards stale projects response...`** — passing by accident. The mock didn't expose `getStats`; `invalidateFilterCaches` → `loadStats` → `getStats` failed loudly on stderr but was swallowed by `loadStats`'s try/catch.
- **`sync.test.ts > triggerResync` (3 tests)** — passing by accident. `getSyncStatus` was bare `vi.fn()`; reading `status.last_sync` threw and was caught by `loadStatus`'s try/catch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,29 @@ jobs:
       - name: Run Go linters
         run: make lint-ci
 
+  frontend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6.4.0
+        with:
+          node-version: "24"
+
+      - name: Install frontend dependencies
+        run: npm ci
+        working-directory: frontend
+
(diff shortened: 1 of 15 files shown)


cpcloud:
@wesm This one is good to go.

PR #408, opened 2026-04-27
feat: session duration UX

cpcloud:
## Summary

Replaces the right-column `ActivityMinimap` with a richer **Session Vital Signs** panel and decorates the conversation column with per-turn and per-tool-call timing data. All durations are computed from existing message and session timestamps; no schema changes.

The right panel has four sections:
- **Session summary** — total wall-clock, turn count, tool call count, slowest call (clickable to scroll), sub-agent count
- **Time spent** — per-category aggregate bars (click a row to filter the rest of the panel)
- **Timeline** — turns lane + per-category lanes + activity lane + legend
- **Calls** — chronological list with horizontal duration bars, parallel groups bracketed, sub-agent rows expandable inline

Inline in the conversation column: each `ToolBlock` gains a duration badge in its header; each assistant message gains a turn-summary line ("turn 2m 18s · 3 calls"); parallel `tool_use` runs render as a single `ParallelGroup` block with `≤duration` upper bound.

**Tool label normalization**: `ToolBlock` headers, the Calls list, and the slowest-call link render the normalized category (e.g. "Bash") rather than the raw `tool_name`, so codex's `exec_command` shows as "Bash" alongside Claude's. Collapsed `ToolBlock` headers now surface `$ <command>` from `input_json` directly, so bash invocations are legible without expanding the block. `makeInputPreview` dispatches on category and recognizes codex's `cmd` argument.

## Screenshots

Full session view with the duration showcase fixture (sidebar + conversation + Session Vital Signs):

full session view

Right panel detail — Session summary, Time spent, Timeline lanes, Calls:

right panel

Sub-agent inline expansion in the Calls section (chevron-toggled `.sa-expand` block reveals the child session's calls):

sub-agent expansion

## Notable design decisions

- **Aggregate attribution rule**: when a turn contains a sub-agent whose wall-clock meets or exceeds the non-sub-agent remainder, the turn attributes to `Task`. This matches the user's mental model ("the sub-agent did the work") and keeps the Timeline turns lane honest when a parallel turn mixes a long sub-agent with short non-sub-agent siblings.
- **Per-call precision for parallel non-sub-agent calls is unknowable** from JSONL source data (one timestamp per assistant message). Parallel siblings render with a striped bar and `≤duration` label; only sub-agents inside parallel groups carry exact per-call durations.
- **Backend stays computable at query time** — both SQLite and PostgreSQL backends share the same `AssembleTiming` logic; only the dialect-specific SQL differs.

## Touched

- Backend: `internal/db/timing.go`, `internal/postgres/session_timing.go`, `internal/server/session_timing.go`, `internal/server/events.go`
- Conversation column: `frontend/src/lib/components/content/{ToolBlock,MessageContent,ParallelGroup,ToolCallGroup,MessageList,SubagentInline}.svelte`
- Right panel: `frontend/src/lib/components/content/{SessionVitals,CallRow,CallGroup,SubagentCalls,ActivityLane}.svelte`
- Layout: `frontend/src/App.svelte`, `frontend/src/lib/components/layout/ThreeColumnLayout.svelte`
- Frontend foundations: `frontend/src/lib/{api/timing.ts,api/types/timing.ts,utils/duration.ts,utils/categoryAttribution.ts,utils/categoryToken.ts,utils/toolDisplay.ts,stores/sessionTiming.svelte.ts}`
- Test fixture: `cmd/testfixture/main.go` adds a `test-session-duration-showcase` session covering solo/parallel/sub-agent/slow shapes
- E2E: replaces `frontend/e2e/activity-minimap.spec.ts` with `frontend/e2e/session-timing.spec.ts`


wesm:
Reviewed the roborev findings; both declined.

**High — `cmd/testfixture/main.go:375` `new(...)` misuse**: not a defect. Go 1.26 extended `new` to accept either a type or an expression; `new(stringValue)` returns `*string` initialized with the value. Empirically `go build -tags fts5 ./cmd/testfixture/...` and `go vet ./cmd/testfixture/...` succeed under `go1.26.2`, and `go doc builtin.new` documents the new signature `new(TypeOrExpr) *Type`. The calls were introduced in `4257703` and have been compiling since.

**Medium — `MessageContent.svelte` tool ordering**: pre-existing v1 design, not a regression from `9e2f770`. The in-code comment is explicit: "Tool segments are rendered after the loop so contiguous `tool_calls` can be grouped into a single `ParallelGroup` (v1 simplification: text first, then all tools)." Introduced in `2939a73`. The current commit only swapped the label expression and did not touch loop structure or tool placement. Reverting the v1 simplification is an architectural change worth a separate PR.

PR #410, opened 2026-04-28
fix(sync): run remote SSH commands through POSIX shell

aaronflorey:
## Summary
- execute remote sync SSH commands via `sh -lc` so command parsing is independent of the remote login shell
- preserve safe shell quoting for remote command strings, including embedded single quotes
- update `buildSSHArgs` tests to assert POSIX-shell wrapping and quote escaping behavior

## Validation
- make test
- env GOTOOLCHAIN=go1.26.2 make lint


wesm:
Invalid findings. Merging!

PR #420, opened 2026-04-30
Add Docker deployment workflow

aaronflorey:
## Summary
- add a production multi-stage Docker build and entrypoint that defaults to `agentsview serve` and switches to `agentsview pg serve` when `PG_SERVE=1`
- add a GitHub Actions workflow that builds and publishes multi-arch images to `ghcr.io/wesm/agentsview` for `main` and version tags
- add a production compose example and README Docker docs covering local archive mounts and PostgreSQL-backed startup

## Testing
- `bash -n docker-entrypoint.sh`
- `docker compose -f docker-compose.prod.yaml config`
- `docker build -t agentsview:test .`
- `docker run --rm agentsview:test version`
- `docker run --rm -e PG_SERVE=1 agentsview:test --help`
- local `act` dry-run of `.github/workflows/docker.yml`
- local `act` end-to-end runs of a non-publishing workflow copy for simulated `main` and tag pushes

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+.git
+.github
+dist
+tmp
+agentsview
+coverage.out
+frontend/node_modules
+desktop/node_modules
+desktop/src-tauri/target
+testdata/ssh/test_key
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,8 @@ tmp
 agentsview
 coverage.out
 frontend/node_modules
+frontend/dist
+internal/web/dist
 desktop/node_modules
 desktop/src-tauri/target
 testdata/ssh/test_key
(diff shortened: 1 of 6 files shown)


aaronflorey:
Thanks for the review. I pushed a follow-up in `128aadd` that tightens the Docker examples and compose file to publish on loopback only (`127.0.0.1:8080:8080`), and I added a note to enable `--require-auth` before exposing the service beyond localhost.

I also re-checked the other flagged items against the current branch and they do not reproduce here: the workflow `uses:` entries are valid, `AGENTSVIEW_DATA_DIR` is supported, `pg serve` exists, and the PostgreSQL example URLs are well-formed.

wesm:
looking at this now

wesm:
thanks for the PR!

PR #430, opened 2026-04-30
feat(copilot): use workspace.yaml name as session title

MCBoarder289:
This is just a quality of life change for naming copilot sessions. With the latest updates to copilot cli, I'm noticing a session name that is generated by copilot similar to OpenCode. This change simply monitors the workspace.yaml file for copilot to see if this field exists, and if so, names the session appropriately.

I intentionally tried to match our patterns for preferring the agentsview name if its renamed, and then falling back onto this name field, and then finally if the name field isn't present (and likely won't be for older versioned sessions for copilot), it will use the FirstMessage.

Happy to address any feedback that anyone has here!

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -27,7 +27,13 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
-// Bumped to 20: Claude parser now surfaces queued_command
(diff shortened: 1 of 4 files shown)


MCBoarder289:
Solid feedback from above, putting some other changes in place to address those issues.

MCBoarder289:
This should be good to go now!

wesm:
thanks!

PR #449, opened 2026-05-05
Add Forge agent support

mjacobs:
## Add Forge agent support

Adds [Forge](https://forgecode.dev) as a supported agent in agentsview, alongside a fix for a latent subagent-linking gap that surfaces with DB-backed agents.

### What this adds

Forge stores conversations in a SQLite database (`.forge.db`) rather than per-session files. New code in `internal/parser/forge.go` opens this DB read-only, lists conversations via lightweight metadata, and parses each on demand. The parser:

- Maps Forge's `message.text` and `message.tool` events to agentsview's user/assistant/tool model, including reasoning details (rendered as `[Thinking]…[/Thinking]`) and tool calls (with `task`/`skill` argument extraction for subagent ID and skill name).
- Pulls token usage from per-message `usage.{prompt,completion,cached}_tokens.actual` and aggregate `metrics.{input,output,cached_input}_tokens`, with a fallback to message-level accumulation when aggregates are absent.
- Resolves project name from the `<current_working_directory>` tag found in the system message context.

`internal/sync/engine.go` learns to handle DB-backed agents alongside its existing file-based ones: each conversation gets a virtual path (`<dbPath>#<conversationID>`) and a per-conversation mtime, so incremental sync only re-parses the conversations whose `updated_at` changed. `FindSourceFile`, `SourceMtime`, and `SyncSingleSession` all gain Forge support symmetric to the existing Warp paths.

The tool taxonomy in `internal/parser/taxonomy.go` adds nine Forge-specific names: `fs_search`→Grep; `patch`/`multi_patch`/`undo`/`remove`→Edit; `fetch`→Read; `todo_write`/`todo_read`→Tool; `parallel`→Task.

### Bulk-sync linking fix

While writing the integration test for parent→child subagent linking, I discovered that `LinkSubagentSessions` was being invoked inside `collectAndBatch` — which runs before the DB-backed (Warp, Forge) sync blocks. Bulk `SyncAll` was therefore never establishing links for those agents; only `SyncSingleSession` worked, because it calls the linker itself. This PR moves the call so it also runs after the DB-backed writes complete. The function is idempotent (filtered `UPDATE` plus the `idx_tool_calls_subagent` partial index), so the call is unconditional rather than guarded.

### Test coverage

13 new tests across the parser and sync packages:

- **Parser** — token/metrics fallbacks (no metrics column, output-only metrics, no cached_tokens, entirely absent usage); degenerate conversations (empty assistant message, missing `call_id`, `raw_content.Text` fallback); cwd extraction edge cases (no system message, missing tag, empty tag, tag in non-system message); `parseForgeTimestamp` across all five accepted layouts; `endedAt` fallback to `startedAt`; `forgeToolOutputText` top-level text fallback; `skill` tool name from both `arguments.name` and `arguments.skill`; reasoning details with no text fields.
- **Taxonomy** — table-driven test for the nine Forge tool mappings.
- **Sync integration** — multi-conversation incremental sync (only changed conversations rewritten); missing-conversation handling for `FindSourceFile`/`SourceMtime`/`SyncSingleSession`; non-existent session error path; parent→child subagent linking end-to-end through `SyncAll` (the test that surfaced the linking bug); failure isolation when one conversation has malformed JSON context.

### DIFF
--- a/internal/db/analytics.go
+++ b/internal/db/analytics.go
@@ -2699,12 +2699,12 @@ func AggregateSignals(
 
 // TopSession holds summary info for a ranked session.
 type TopSession struct {
-	ID                string  `json:"id"`
-	Project           string  `json:"project"`
-	FirstMessage      *string `json:"first_message"`
-	MessageCount      int     `json:"message_count"`
-	OutputTokens      int     `json:"output_tokens"`
-	DurationMin       float64 `json:"duration_min"`
+	ID           string  `json:"id"`
+	Project      string  `json:"project"`
+	FirstMessage *string `json:"first_message"`
+	MessageCount int     `json:"message_count"`
+	OutputTokens int     `json:"output_tokens"`
+	DurationMin  float64 `json:"duration_min"`
 	// StartedAt and EndedAt are included so the frontend can
 	// derive a recency-based status tier — the StatusDot in the
 	// Top Sessions column needs the same time window inputs as
--- a/internal/parser/forge.go
+++ b/internal/parser/forge.go
@@ -1,0 +1,519 @@
+package parser
+
+import (
+	"database/sql"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/tidwall/gjson"
+)
+
+const forgeDBFilename = ".forge.db"
+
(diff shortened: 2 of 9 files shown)


wesm:
LGTM. will merge when CI is finished

PR #444, opened 2026-05-04
fix: link Codex app subagents

mariusvniekerk:
## Summary
- Link Codex app subagent spawn events to `ParsedToolCall.SubagentSessionID` by consuming both `spawn_agent` outputs and `collab_agent_spawn_end.new_thread_id` events.
- Recognize `wait_agent` calls with `targets` and subagent notifications that identify children via `agent_path`.
- Keep upstream Codex task lifecycle handling intact while adding the subagent event path.

## Supporting evidence
- OpenAI's Codex subagents docs describe Codex spawning specialized agents, waiting for results, and surfacing subagent activity in the app and CLI: https://developers.openai.com/codex/subagents
- I cloned `openai/codex` locally at `905987c08f40bc8fb8f36d69e8137865e1a67393`; that source defines `spawn_agent`, canonical child task names, and `wait_agent.targets`: https://github.com/openai/codex/blob/905987c08f40bc8fb8f36d69e8137865e1a67393/codex-rs/tools/src/agent_tool.rs#L667-L672 and https://github.com/openai/codex/blob/905987c08f40bc8fb8f36d69e8137865e1a67393/codex-rs/tools/src/agent_tool.rs#L721-L745
- The Codex protocol exposes `CollabAgentSpawnEndEvent.new_thread_id`, and the app-server thread history handler consumes that event to populate receiver thread IDs: https://github.com/openai/codex/blob/905987c08f40bc8fb8f36d69e8137865e1a67393/codex-rs/protocol/src/protocol.rs#L3788-L3795 and https://github.com/openai/codex/blob/905987c08f40bc8fb8f36d69e8137865e1a67393/codex-rs/app-server-protocol/src/protocol/thread_history.rs#L623-L645
- Codex subagent notifications are emitted with `agent_path`, which this parser now accepts in addition to legacy `agent_id`: https://github.com/openai/codex/blob/905987c08f40bc8fb8f36d69e8137865e1a67393/codex-rs/core/src/context/subagent_notification.rs#L20-L31

## Validation
- `go test ./internal/parser ./internal/sync -count=1`
- `git diff --check`

PR #459, opened 2026-05-05
fix: link Claude subagent tool calls

wbaik:
## Summary
  - Link Claude `Task` / `Agent` tool calls to child subagent sessions using result-side `toolUseResult.agentId` when
  queue/progress mapping is absent.
  - Preserve sibling Claude subagent tool calls when Claude emits additive same-`message.id` assistant chunks.
  - Keep streaming snapshot behavior by replacing progressive partial text with the latest text block while still
  preserving distinct additive content blocks.

  ## Context
  This is the Claude-side version of the inline subagent-linking contract addressed for Codex in #444 / #458.

  The UI already renders inline subagent sessions when `tool_calls.subagent_session_id` is populated. This PR makes
  Claude ingestion populate that existing field in more cases, so child sessions that are already ingested can appear
  inline under the exact parent `Task` / `Agent` tool call that launched them.

  ## Backfill note
  This changes parsed tool-call metadata for existing Claude rows. Existing archives will need a full parser
  reprocess / data-version bump to backfill inline links.

  This PR intentionally does not bump `dataVersion`; maintainers should decide whether to trigger that reparse.

  ## Validation
  - `go test ./internal/parser ./internal/sync -count=1`
  - `git diff --check origin/main...HEAD`

  If you add screenshots later, I’d add a final section:

  ## Visual Check   
  - Before: Claude rendered the three sibling `Task` calls, but none had an inline linked subagent session. image
  - After: The same three sibling `Task` calls are preserved. image

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -27,10 +27,19 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
-// Bumped to 25: Codex parser now also links codex_app subagents
+// Bumped to 26: Claude parser now (a) links Task / Agent tool
+// calls to child subagent sessions via toolUseResult.agentId
+// when queue/progress mappings are absent, populating
+// tool_calls.subagent_session_id, and (b) merges additive
+// same-message.id assistant chunks instead of keeping only the
+// last entry, preserving sibling tool_use blocks and
+// progressively-built text. Existing rows need re-parsing so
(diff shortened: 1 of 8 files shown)


wbaik:
Should i guess you're going to take over? @wesm 
Regardless, thanks for the awesome repo. Well... not just this, but your books and everything else.

wesm:
yeah I'm taking over! i'll merge soon

PR #502, opened 2026-05-15
feat: qwen code support

suculent:
adds results from Qwen Code


wesm:
@suculent I will get this branch merged today (I think it was good to go at 5e2ef7d I just hadn't clicked the merge button)

PR #506, opened 2026-05-17
Add current Kiro SQLite session support

danxtshake:
## Summary
Sessions started under kiro-cli --classic are visible in kiro-cli chat --list-sessions, but Kiro stores them in ~/.local/share/kiro-cli/data.sqlite3 rather than the legacy ~/.kiro/sessions/cli/ JSONL directory that AgentsView previously read. In my reproduction, AgentsView only detected such a session after it was resumed into the newer Kiro TUI.

Thus, I planned and implemented these feature for AgentView:
- ingest current Kiro CLI sessions from `~/.local/share/kiro-cli/data.sqlite3`
- preserve legacy JSONL-backed Kiro history while preferring the current SQLite record when the same conversation exists in both stores
- add incremental sync, watch fallback, resume, and regression coverage for SQLite-backed Kiro sessions

## Validation

- `go test ./internal/parser -run Kiro -count=1`
- `go test ./internal/sync -run 'TestSyncEngineKiroSQLite|TestResyncAllKiroSQLiteOnly' -count=1`
- `go test ./internal/server -run 'TestResumeSession/kiro current-store command only' -count=1`
- `make build`
- manual verification against a live Kiro CLI store: classic-session conversations now appear in AgentsView without resuming them first


mariusvniekerk:
Thanks for the submission!

danxtshake:
I've reviewed problems identified by robodev and made changes regarding
- ParseKiroSQLiteVirtualPath splits on the first #
- Shadowed legacy Kiro JSONL files are filtered only during full sync.
- session.Cwd refers to cwd, not the cached value (this is not breaking change for current copilot, claude code, codex, since they handle cwd well, but I reckon keeping the behavior consistent should be the chocie).

PR #509, opened 2026-05-17
Add Normal/Focused transcript mode to HTML exports

### LINKED ISSUE #508: Add Normal/Focused transcript mode to exported HTML
The standalone HTML export currently includes all user/assistant messages and all tool-call content. That preserves fidelity, but tool-heavy sessions can be hard to read once exported.

It would be useful for exported HTML to include the same Normal vs Focused transcript mode available in the app:

- Normal: preserve current export behavior and show the full transcript.
- Focused: hide tool-only/progress-heavy material and keep user prompts plus final assistant answers.

The export is standalone HTML, so the implementation should embed the required CSS/JS or CSS-only controls directly in the generated file, without relying on the main frontend bundle.

tpn:
## Summary
- add Normal/Focused transcript controls to standalone HTML exports
- default exports to Normal so current behavior is preserved
- hide tool-only and superseded assistant/progress messages in Focused mode using self-contained CSS
- reuse the markdown segment parser for robust tool-only detection

Fixes #508

## Testing
- GOTOOLCHAIN=auto go test ./internal/server -run 'ExportHTML|ExportTemplate|FocusedExport' -count=1\n- GOTOOLCHAIN=auto go test ./internal/server -count=1\n- roborev 3-agent review on c4b1b32: codex, claude-code, gemini all passed

### DIFF
--- a/internal/server/export.go
+++ b/internal/server/export.go
@@ -283,11 +283,13 @@ type exportData struct {
 }
 
 type exportMessage struct {
-	RoleClass   string
-	ExtraClass  string
-	Role        string
-	Timestamp   string
-	ContentHTML template.HTML
+	Ordinal       int
+	RoleClass     string
+	ExtraClass    string
+	Role          string
+	Timestamp     string
+	ContentHTML   template.HTML
+	FocusedHidden bool
 }
 
 var exportTmpl = template.Must(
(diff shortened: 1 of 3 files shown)


wesm:
thanks!

PR #510, opened 2026-05-18
feat: add postgres curation metadata

### LINKED ISSUE #484: Support session stars and pinned messages in PostgreSQL / pg serve dashboards
## Feature request

  Please support session stars and pinned messages when running AgentsView from PostgreSQL via `agentsview pg serve`.

  ## Use case

  I’m using AgentsView as a shared dashboard across multiple devices:

  - each device runs `agentsview pg push`
  - a hub runs PostgreSQL
  - the dashboard runs with `agentsview pg serve`

  This works well for viewing sessions across machines, but session stars and pinned messages do not work on the hub dashboard.

  ## Current behavior

  In `pg serve` mode, the dashboard is read-only. The PostgreSQL store returns `db.ErrReadOnly` for:

  - `StarSession`
  - `UnstarSession`
  - `BulkStarSessions`
  - `PinMessage`
  - `UnpinMessage`

  The API returns `501` / `"not available in remote mode"`, and the frontend mostly appears as if the action silently did not persist.

  Also, `ListStarredSessionIDs` and `ListPinnedMessages` return empty results in the PostgreSQL store, so existing stars/pins are not visible on the hub dashboard.

  ## Expected behavior

  Stars and pinned messages should work in PostgreSQL-backed dashboards, ideally as shared metadata stored in PostgreSQL.

  Expected behavior:

  - starring a session in the hub dashboard persists
  - pinning a message in the hub dashboard persists
  - stars/pins survive dashboard refreshes
  - stars/pins are visible from other devices using the same hub
  - optional: `pg push` can sync local SQLite stars/pins into PostgreSQL

  ## Why this matters

  In a multi-device setup, the hub dashboard is the main place where I inspect sessions. Stars and pinned messages are especially useful for marking important sessions/
messages across devices, so losing those features makes the shared dashboard less useful.

  ## Possible implementation

  A possible approach:

  - Add PostgreSQL equivalents of `starred_sessions` and `pinned_messages`.
  - Implement the star/pin methods on `internal/postgres.Store` instead of returning `db.ErrReadOnly`.
  - Make the PostgreSQL list methods read from those tables.
  - Optionally sync local SQLite stars/pins during `agentsview pg push`.

  If full write support is not desired for `pg serve`, another improvement would be to disable/hide the star and pin controls in remote mode and show a clear message instead
of failing silently.

  ## Environment

  - AgentsView Docker image: `ghcr.io/wesm/agentsview:latest`
  - Deployment style: hub PostgreSQL + `agentsview pg serve`, with multiple `agentsview pg push` devices

lisiyuan656:
## Summary

  - add PostgreSQL-backed dashboard curation metadata for starred sessions and pinned messages
  - keep pg serve read-only for synced session content while allowing stars and pins to be written to PostgreSQL
  - preserve pinned messages across PostgreSQL message rewrites, including shifted ordinal runs and duplicate source_uuid repair cases
  - Closes #484.

  ## Details

  This adds starred_sessions and pinned_messages tables to the PostgreSQL schema and implements the corresponding Store methods for pg serve.

  Pinned messages use PostgreSQL’s synthetic message ID convention where message_id matches message ordinal. During pg push, pinned messages are reconciled by source_uuid so
  pins can survive message rewrites where ordinals move, such as compact-boundary rows being inserted before previously pinned messages.

  ## Validation

  - go test ./internal/postgres
  - go test ./internal/db -run 'TestReplaceSessionMessagesPin'
  - go test -tags pgtest ./internal/postgres -run 'TestStoreStarsAndPins|TestPushPreservesMultiplePGPinsBySourceUUID|TestReconcilePinnedMessagesPrefersCurrentTargetPin' -v
 - PG tests compile locally but skip without TEST_PG_URL set.

### DIFF
--- a/internal/db/store.go
+++ b/internal/db/store.go
@@ -65,13 +65,13 @@ type Store interface {
 	GetTopSessionsByCost(ctx context.Context, f UsageFilter, limit int) ([]TopSessionEntry, error)
 	GetUsageSessionCounts(ctx context.Context, f UsageFilter) (UsageSessionCounts, error)
 
-	// Stars (local-only; PG returns ErrReadOnly).
+	// Stars.
 	StarSession(sessionID string) (bool, error)
 	UnstarSession(sessionID string) error
 	ListStarredSessionIDs(ctx context.Context) ([]string, error)
 	BulkStarSessions(sessionIDs []string) error
 
-	// Pins (local-only; PG returns ErrReadOnly).
+	// Pins.
 	PinMessage(sessionID string, messageID int64, note *string) (int64, error)
 	UnpinMessage(sessionID string, messageID int64) error
 	ListPinnedMessages(ctx context.Context, sessionID string, project string) ([]PinnedMessage, error)
--- a/internal/postgres/curation.go
+++ b/internal/postgres/curation.go
@@ -113,6 +113,9 @@ func (s *Store) BulkStarSessions(sessionIDs []string) error {
 
 // PinMessage creates or updates a shared PG pin for a message. PG
 // exposes message IDs as the message ordinal, matching scanPGMessages.
+// On conflict, ordinal and source_uuid are refreshed from the current
+// message so the pin tracks whatever is at message_id today;
+// created_at is preserved by being absent from the SET clause.
 func (s *Store) PinMessage(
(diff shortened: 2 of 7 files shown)


wesm:
thank you!

PR #513, opened 2026-05-18
fix: skip any slash command when computing session first_message preview

### LINKED ISSUE #487: feat: Ignore /login prompt in sessions or sidebar session title
I have more than one Claude account (work, personal) and I use the `/login` cli prompt to switch between them on the fly. I have several sessions that show up in the left sidebar with the title "/login". It would be more useful to either not display that prompt at all, or skip it when labeling the session so a meaningful label appears.

cswingle:
Fixes #487.

Session sidebar titles were set from the first user message without filtering slash commands, so sessions starting with `/login`, `/plan`, or similar showed those commands as the title instead of the first meaningful prompt.

Previously only `/clear` and `/effort` were excluded via an explicit allowlist. This replaces that list with generic slash command detection: any message that starts with `/` followed by one or more letters or digits (then end-of-string or whitespace) is treated as a command and skipped. File-path references like `/usr/local/bin gives an error` are not affected because the embedded `/` causes the check to return false.

### DIFF
--- a/internal/parser/claude.go
+++ b/internal/parser/claude.go
@@ -1597,12 +1597,13 @@ func isCommandEnvelope(content string) bool {
 }
 
 // isSkippablePreviewCommand returns true when content is a Claude
-// Code slash command (e.g. /login, /plan, /clear). Detection is
-// generic: the trimmed content must start with "/" followed by one
-// or more letters or digits, then either end or be followed by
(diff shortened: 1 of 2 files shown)


wesm:
Looks ok to me. I could argue that e.g. `/roborev-fix` should be shown but this doesn't strike me as harmful

PR #516, opened 2026-05-20
Normalize Gemini token usage for cost tracking

### LINKED ISSUE #515: Gemini usage cost is reported as zero in usage daily
## Summary

Gemini sessions with parsed token data are appearing in `agentsview usage daily --json --breakdown`, but their token counts and costs are reported as zero.

In my local database, Gemini sessions have message-level `token_usage` JSON like:

```json
{
  "input": 6098,
  "output": 20,
  "cached": 2960,
  "thoughts": 539,
  "tool": 0,
  "total": 6657
}
```

The session list also shows non-zero `total_output_tokens` and `peak_context_tokens` for Gemini sessions. But `usage daily --json --breakdown` includes Gemini models in `modelsUsed` / `modelBreakdowns` with all token fields and `cost` equal to zero.

## Why this seems to happen

The Gemini parser stores the raw Gemini CLI `tokens` object using Gemini-style keys:

- `input`
- `output`
- `cached`
- `thoughts`

Relevant parser code:

- `internal/parser/gemini.go` reads `tokens.input`, `tokens.output`, `tokens.cached`.
- `parseGeminiMessage` stores `TokenUsage` as the raw `tokens` object.
- It separately computes `ContextTokens` as `input + cached` and `OutputTokens` as `output`, so session-level token coverage looks populated.

However, daily usage cost calculation in `internal/db/usage.go` only reads normalized keys from message `token_usage`:

- `input_tokens`
- `output_tokens`
- `cache_creation_input_tokens`
- `cache_read_input_tokens`

So for Gemini messages, `usageAmounts` parses the JSON successfully but reads zero for every counted field. This makes the usage report show Gemini models while charging them as `$0`.

Interestingly, token-presence detection already recognizes Gemini-style keys (`input`, `cached`, `output`), which explains why sessions can look like they have token data while daily cost still remains zero.

## Expected behavior

Gemini message token usage should contribute to `usage daily` token totals and cost estimates.

## Possible fix

Normalize Gemini token usage at parse time, for example:

```json
{
  "input_tokens": "input - cached",
  "output_tokens": "output + thoughts",
  "cache_read_input_tokens": "cached"
}
```

Or alternatively teach `usageAmounts` to also read Gemini-style `input`, `output`, `cached`, and `thoughts` keys. Parser-side normalization seems cleaner because it keeps the downstream usage/cost code agent-neutral.

One subtle point: if Gemini's `input` already includes cached tokens, then `input_tokens` should probably be normalized to `max(input - cached, 0)` to avoid double-counting cached input at the full input rate, similar to the Codex parser normalization.

## Local evidence

Example local `usage daily --json --breakdown` row showed Gemini models included but zeroed:

```json
{
  "modelName": "gemini-3-flash-preview",
  "inputTokens": 0,
  "outputTokens": 0,
  "cacheCreationTokens": 0,
  "cacheReadTokens": 0,
  "cost": 0
}
```

But direct message inspection showed non-empty Gemini `token_usage` JSON with `input`, `output`, `cached`, and `thoughts` fields.

nergal-perm:
## Summary

- Normalize Gemini CLI token usage into the same keys used by usage cost aggregation.
- Include Gemini thoughts in output token usage and cached tokens in cache-read usage.
- Add a regression test covering the normalized Gemini token usage shape.

Fixes #515

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -27,9 +27,16 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
-// Bumped to 27: Piebald parser now persists normalized per-message
+// Bumped to 28: Gemini parser now persists normalized
+// (Anthropic-style) per-message token_usage JSON instead of the raw
+// tokens object, and rolls thoughts tokens into OutputTokens so
+// per-message and session output totals match the cost JSON.
+// Existing Gemini rows need re-parsing so usage and cost reports
+// reflect the new shape and include thoughts tokens.
+//
+// (27: Piebald parser now persists normalized per-message
 // token_usage JSON. Existing Piebald rows need re-parsing so Usage
-// reports can include older Piebald sessions.
+// reports can include older Piebald sessions.)
 //
 // (26: Claude parser now (a) links Task / Agent tool
 // calls to child subagent sessions via toolUseResult.agentId
@@ -81,7 +88,7 @@ import (
 //
 // (17: Codex <skill> template filtering.)
 // (16: <turn_aborted> system messages.)
-const dataVersion = 27
+const dataVersion = 28
 
 const tokenCoverageRepairStatsKey = "token_coverage_repair_v1"
 
--- a/internal/parser/gemini.go
+++ b/internal/parser/gemini.go
@@ -38,7 +38,7 @@ func extractGeminiTokens(msg gjson.Result) geminiTokens {
 
 func normalizedGeminiTokenUsage(tok geminiTokens) json.RawMessage {
 	payload := map[string]int{
-		"input_tokens":            max(tok.Input-tok.Cached, 0),
+		"input_tokens":            tok.Input,
 		"output_tokens":           tok.Output + tok.Thoughts,
 		"cache_read_input_tokens": tok.Cached,
 	}
--- a/internal/parser/gemini.go
+++ b/internal/parser/gemini.go
@@ -15,9 +15,10 @@ import (
 
 // geminiTokens holds token usage counts from a Gemini message.
 type geminiTokens struct {
-	Input  int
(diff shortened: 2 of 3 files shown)


nergal-perm:
Reviewed roborev's combined review on `fa2b558`:

**High (claimed compile error at `internal/parser/gemini.go:251`)** — false positive. `tok` is defined on the line immediately preceding the call: `tok := extractGeminiTokens(msg)`. The full local test suite (`make test`) passes both before and after this PR, which confirms the code compiles. The bot appears to have looked at `tokResult` (the `gjson.Result` fetched on the next line for presence checks) and missed `tok` directly above.

**Medium (output-token inconsistency)** — valid, fixed in 1b22714. `normalizedGeminiTokenUsage` folded `thoughts` into `output_tokens` (Gemini bills thoughts at the output rate), but `ParsedMessage.OutputTokens` and the session rollup still used only `tok.Output`, so the cost JSON and the per-message/session aggregates disagreed when thoughts were present. Now `OutputTokens = tok.Output + tok.Thoughts` and `HasOutputTokens` flags coverage when either `output` or `thoughts` is reported. Test expectations updated accordingly.

nergal-perm:
Addressed roborev's review on `1b22714` (5a6e8df):

**Medium (dataVersion not bumped)** — valid, fixed. Bumped `internal/db/db.go` `dataVersion` from 27 to 28 and documented the Gemini token normalization in the version-history comment. Existing Gemini sessions with unchanged source mtimes will now be forced through the non-destructive resync path on next open, refreshing both `token_usage` JSON (raw → normalized Anthropic-style) and `OutputTokens` (now including thoughts).

`make lint` and `make test` clean.

wesm:
thanks!

nergal-perm:
> thanks!

You're welcome!

PR #528, opened 2026-05-22
feat: support Google Antigravity IDE and CLI sessions

mjacobs:
## Summary

Adds two new agents to agentsview's session registry:

- **Antigravity** (IDE) — `~/.gemini/antigravity/`. Per-session SQLite
  database; unencrypted. Parsed in full: messages, timestamps, brain
  artifacts.
- **Antigravity CLI** — `~/.gemini/antigravity-cli/`. Per-session
  protobuf streams under AES-GCM. Parsed in summary mode by default
  (user prompts from `history.jsonl` + plaintext `brain/` artifacts),
  with an optional decrypted-strings preview when `ANTIGRAVITY_KEY`
  is set.

The IDE side is feature-complete. The CLI side ships summary mode now
and falls back gracefully when the AES key is unavailable; full
transcript decryption requires HKDF parameters baked into the
Antigravity binary and has been deferred to a separate, dedicated
tool.

## What's in the diff

- `internal/parser/antigravity.go` — IDE SQLite parser
- `internal/parser/antigravity_cli.go` — CLI parser (summary + optional
  decrypted preview)
- `internal/parser/antigravity_crypto.go` — AES-CTR/CBC/GCM helpers,
  env-var key loader, brute-force-skip retry loop mirroring the
  reference Python decryptor
- `internal/parser/antigravity_proto.go` — stdlib-only protobuf wire
  walker (no third-party deps)
- `internal/parser/types.go`, `taxonomy.go`, `sync/engine.go` — registry
  entries, tool-name mapping, file-watcher classifiers
- `frontend/src/lib/utils/agents.ts` + `AgentDirSettings.svelte` — UI
  labels and violet accent color for both agents
- `internal/parser/antigravity_test.go` — wire-walker round-trip,
  IDE SQLite fixture, CLI discovery+parse (conversations/ and
  implicit/), AES-GCM round-trip, PKCS7 edge cases,
  `history.jsonl` robustness against malformed lines

## Known limitations

- **CLI full decryption requires `ANTIGRAVITY_KEY` to be set manually.**
  The on-disk key is HKDF-derived from the OAuth refresh token in the
  OS keyring, using salt/info baked into the Antigravity binary. The
  binary-RE work to extract those parameters and the keyring probing
  logic are out of scope here and tracked for a separate standalone
  tool.
- **IDE role inference is heuristic** (`step_type==14 → user`, else
  `assistant`). The full `CORTEX_STEP_TYPE_*` taxonomy is available
  via the Antigravity daemon's `GetCascadeTrajectory` RPC and can be
  ported in once the standalone tool's translator code is mature.

## Notes for review

- All new code uses stdlib only — no protobuf or HKDF dependencies
  added.
- `go fmt`, `go vet`, parser tests, and frontend tests all green
  locally and in CI on this branch (Linux + Windows).
- README's supported-agents table updated.
- `.antigravitycli/` (the agent's session directory, created if
  Antigravity CLI is run inside this repo) added to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ html/
 
 # Visual verification scratch — per-task screenshot/HTML capture dirs.
 .test-data*/
+
+# Antigravity CLI session dir created when AGY runs inside this repo
+.antigravitycli/
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ agentsview auto-discovers sessions from all of these:
 | Piebald            | `~/.local/share/piebald/`                              |
 | Warp               | `~/.warp/` (platform-dependent)                        |
 | Positron Assistant | `~/Library/Application Support/Positron/User/` (macOS) |
+| Antigravity        | `~/.gemini/antigravity/`                               |
+| Antigravity CLI    | `~/.gemini/antigravity-cli/` (summary mode)            |
 
 Each directory can be overridden with an environment variable. See the
 [configuration docs](https://agentsview.io/configuration/) for details.
--- a/frontend/src/lib/components/settings/AgentDirSettings.svelte
+++ b/frontend/src/lib/components/settings/AgentDirSettings.svelte
@@ -18,6 +18,8 @@
     openclaw: "OpenClaw",
     kimi: "Kimi",
     piebald: "Piebald",
+    antigravity: "Antigravity",
+    "antigravity-cli": "Antigravity CLI",
   };
 </script>
 
--- a/frontend/src/lib/utils/agents.test.ts
+++ b/frontend/src/lib/utils/agents.test.ts
@@ -30,6 +30,8 @@ describe("KNOWN_AGENTS", () => {
       "kiro-ide",
       "cortex",
       "piebald",
+      "antigravity",
+      "antigravity-cli",
     ]);
   });
 
--- a/frontend/src/lib/utils/agents.ts
+++ b/frontend/src/lib/utils/agents.ts
@@ -34,6 +34,16 @@ export const KNOWN_AGENTS: readonly AgentMeta[] = [
   { name: "kiro-ide", color: "var(--accent-lime)", label: "Kiro IDE" },
   { name: "cortex", color: "var(--accent-cyan)", label: "Cortex Code" },
   { name: "piebald", color: "var(--accent-orange)", label: "Piebald" },
+  {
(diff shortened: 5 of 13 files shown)


mariusvniekerk:
so given that users need ANTIGRAVITY_KEY how are they expected to obtain this?

mjacobs:
Good question— to be clear, **the key is nice-to-have, not need-to-have.**

What's in this PR without any key set:
- Every user prompt with timestamps (from `history.jsonl`)
- Full markdown content of every `brain/` artifact — plans, tasks, walkthroughs — with their `ARTIFACT_TYPE_*` headers and summaries

Example Antigravity CLI session with these changes:
av1 pp pine m4tt xyz-AgentsView-cvscreenshot

The "what was asked / what did the agent produce" loop is fully available.

**What the key additionally unlocks:** the intermediate assistant transcript — thinking, tool calls (Read/Edit/Bash), file diffs, search results.

**On obtaining the key itself:** there's no documented official path. From a few hours of digging, what exists today (AFAIK):

- **macOS**: [`arashz/antigravity_decryptor`](https://github.com/arashz/antigravity_decryptor) documents a workflow against the system keychain + a Python script (CTR layout, macOS-specific).
- **Linux**: the key is HKDF-derived from the OAuth refresh token in the OS keyring, using salt + info parameters baked into the stripped Antigravity binary. Recovering those is essentially a binary-RE exercise.
- A more general workaround — likely going through Antigravity's own local daemon RPC, which already does the decryption while the IDE/CLI is running — is something I'm exploring but I think would be out of scope for this PR and may not be a good fit for agentsview anyway- TBD.

So I'm leaning towards removing the mention/hint of the key as it's more likely confusing.  Then the question is: should I also remove the code path for using the env var _if a user has it_ (e.g. coming from arashz's tool on MacOS)? Either way works for me- WDYT?

mariusvniekerk:
Thanks for the clarification.  

Certainly a nice to have, missing out on tool calls without it is a bit unfortunate but so be it.

I agree with you that the exact mechanics of how `agentsview` obtains this key is out of scope for this initial PR.  I'd say leave the implementation in for parsing the richer data when present.   Hopefully some utility parts emerge from how to (say) use the antigravity cli itself to obtain these as that tool matures.

mariusvniekerk (approved):

PR #536, opened 2026-05-23
feat: add 'agentsview session usage' command with cost estimate

wesm:
## Summary

Adds `agentsview session usage <id>` — a single-session view of token usage and an optional cost estimate — and deprecates `token-use`.

- `session usage <id>` reports total output tokens, peak context tokens, and a cost estimate, in human or JSON form (`--format json`).
- Cost comes from an explicit per-event cost when the agent records one (e.g. Hermes); otherwise the session's tokens are priced against the matching `model_pricing` rates (input, output, and cache priced separately). `has_cost` is true only when every contributing row is priced — otherwise `cost_usd` is 0 and `unpriced_models` names the gaps, so a partial total is never shown as complete.
- On the direct CLI path (no running daemon) missing fallback pricing is seeded non-destructively, so a cost estimate is available without a server; custom rates are applied in memory only.
- `token-use` stays as a JSON-only alias that prints a deprecation note to stderr pointing at `session usage`.
- `session usage` rejects `--server` ("not yet implemented") like the other `session` subcommands rather than silently reading the local archive.

Exit codes: 0 when token data or a cost is present, 2 when the session is not found, 3 when it exists but has neither.


wesm:
Nothing to fix here. Merging

PR #540, opened 2026-05-24
fix: restore server timeutil import

wesm:
## Summary
- Restore the missing internal/timeutil import used by the sidebar session index handler.
- Fixes the Go compile failure that broke CI, Docker, CLI release, and desktop release jobs for v0.30.0.

## Evidence
The failed release run reported:
```
internal/server/sessions.go:121:18: undefined: timeutil
internal/server/sessions.go:134:27: undefined: timeutil
```

Normal CI for the same merge commit also caught the same failure in lint, Go tests, coverage, and e2e pre-build.

## Verification
- `go test ./...`
- `git diff --check`
- commit hooks: golangci-lint, NilAway

## Release notes
After merge, recreate the v0.30.0 tag from the fixed main commit. A raw RC tag currently is not a draft-only dry run: the existing tag workflows would publish Docker images, GitHub release assets, desktop updater assets, and attempt the production PyPI publish path.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -1,6 +1,7 @@
 package db
 
 import (
+	"context"
 	"crypto/rand"
 	"database/sql"
 	_ "embed"
@@ -148,8 +149,9 @@ type DB struct {
 	writer    atomic.Pointer[sql.DB]
 	reader    atomic.Pointer[sql.DB]
 	mu        sync.Mutex // serializes writes
-	retired   []*sql.DB  // old pools kept open for in-flight reads
-	dataStale bool       // set by Open when user_version < dataVersion
+	connMu    sync.RWMutex
+	retired   []*sql.DB // old pools kept open for in-flight reads
+	dataStale bool      // set by Open when user_version < dataVersion
 
 	cursorMu     sync.RWMutex
 	cursorSecret []byte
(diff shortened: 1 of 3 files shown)


wesm:
Added a second fix for the flaky `TestConcurrentReadsWhileReopen` failure instead of relying on retries.

Root cause: `Reopen` could close a retired SQLite read pool while another goroutine had already obtained that reader but had not started its query yet. The atomic pointer swap protected pointer visibility, but not the gap between loading a pool and starting work on it.

Fix: route internal reader access through a guarded facade and take the matching write lock while `Reopen` / close operations swap or close reader pools. Added `TestReaderHandleAcquiredBeforeReopenStaysUsable`, which failed before the fix with `sql: database is closed`.

Local verification:
- `go test -tags fts5 ./internal/db -run 'Test(ReaderHandleAcquiredBeforeReopenStaysUsable|ConcurrentReadsWhileReopen|RepeatedReopenBoundsRetiredPools)' -count=20 -v`
- `go test -tags fts5 ./internal/db`
- `go test ./...`
- `go test -tags fts5 ./...`
- commit hooks: golangci-lint, NilAway

wesm:
Addressed roborev's remaining Reader() bypass finding.

Change:
- `DB.Reader()` now returns a guarded `db.Reader` interface instead of a raw `*sql.DB`.
- The interface exposes the query methods existing callers use, but not the underlying pool.
- The regression now goes through the exported path: `TestExportedReaderAcquiredBeforeReopenStaysUsable`.

Verification:
- `go test -tags fts5 ./internal/db -run 'Test(ExportedReaderAcquiredBeforeReopenStaysUsable|ConcurrentReadsWhileReopen|RepeatedReopenBoundsRetiredPools)' -count=20 -v`
- `go test -tags fts5 ./internal/db`
- `go test ./...`
- `go test -tags fts5 ./...`
- commit hooks: golangci-lint, NilAway

PR #547, opened 2026-05-26
pg push service

mxfeinberg:
## Problem

  I track token usage across AI tools with agentsview, and lately I've been
  running more remote agent tools on machines other than my primary dev box.
  Each of those machines records its own sessions locally. The Postgres support
  (`pg push` / `pg serve`) already lets me review everything together, but
  keeping the shared database current meant either remembering to run `pg push`
  by hand on each machine or wiring up a bespoke cron job per host. I wanted a
  touchless way to sync data from multiple machines into one place so remote
  agent token usage shows up in my primary dashboard automatically, without
  maintaining custom cron entries everywhere.

  I also made some updates to hermes specific code because although tokens 
  were being passed through, the associated costs were not being aggregated
  properly.

  ## pg push service 

  A long-running auto-push daemon and a first-class way to install it as an OS
  service, so a recorder machine keeps the shared Postgres current on its own.

  **`agentsview pg push --watch`** start a foreground daemon that pushes to Postgres
  shortly after sessions change, with a periodic floor as a safety net:

  - A file watcher (reusing the same machinery as `serve`) coalesces change
    events over a debounce window (`--debounce`, default 30s) and triggers a
    local sync + incremental push.
  - A periodic floor (`--interval`, default 15m) pushes regardless of file
    events and covers any directories the watch budget couldn't cover.
  - Pushes are serialized by a single goroutine and connect to Postgres lazily,
    reconnecting on error. A transiently unreachable database is logged and
    retried on the next trigger rather than crashing the daemon.
  - A single-instance lock prevents two watchers from racing on the push
    watermarks. On shutdown (SIGINT/SIGTERM) it performs one bounded final flush.
  - It reads the same `[pg]` config and project filters as `pg push`, and
    honors `result_content_blocked_categories` so the push path no longer
    diverges from `serve`.
  
  **`agentsview pg service install|uninstall|status|start|stop|logs`**
  installs and manages the daemon as a per-user OS service:

  - launchd LaunchAgent on macOS, `systemd --user` unit on Linux, behind a small
    platform abstraction with pure (golden-tested) unit-file rendering.
  - `install` validates that the Postgres DSN is resolvable before creating the
    service, and surfaces the systemd linger requirement so the service keeps
    running on headless boxes after logout.
  - `status` reports the manager state plus the last successful push time;
    `logs -f` tails the daemon log and survives log rotation.

  The daemon reads the DSN from the existing config file (no credentials are
  copied into unit files), so protecting `config.toml` is sufficient.

  ## Notes

  - Platform support targets macOS (launchd) and Linux (`systemd --user`); the
    daemon command itself is cross-platform Go.
  - The generated unit always pins `AGENTSVIEW_DATA_DIR` to the install-time data
    dir, so the service resolves the same config regardless of the environment it
    starts in.
  - The watch daemon's initial resync is push-oriented and skips the serve-side
    `Vacuum`/`BackfillSignals` steps, since signal recomputation happens on the
    serve side.
  - A windows service was not added because I don't currently have access to a 
     windows machine that I could use for testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,4 +63,7 @@ html/
 
 # Antigravity CLI session dir created when AGY runs inside this repo
 .antigravitycli/
+
+# graphify knowledge-graph output (local analysis)
+graphify-out/
 .kata.local.toml
--- a/README.md
+++ b/README.md
@@ -283,6 +283,45 @@ agentsview pg push       # push local data to PG
 agentsview pg serve      # serve web UI from PG (read-only)
 ```
 
+### Automatic push (background service)
+
+To keep a shared PostgreSQL database current without running `pg push` by
+hand, run the auto-push daemon. It watches your session directories and
+pushes shortly after new sessions are recorded, with a periodic floor as a
+safety net:
+
+```bash
+agentsview pg push --watch                 # foreground, Ctrl-C to stop
+agentsview pg push --watch --debounce 1m   # custom coalesce window
+agentsview pg push --watch --interval 5m   # custom floor interval
+```
+
+The daemon reads the same `[pg]` config as `pg push`, so the PostgreSQL
+DSN must be set in your config file (or an environment variable it
+expands). Protect the config file, since it holds credentials:
+
+```bash
+chmod 600 ~/.agentsview/config.toml
+```
+
+To run it unattended as an OS service (launchd on macOS,
+`systemd --user` on Linux):
+
+```bash
+agentsview pg service install     # generate the unit, enable + start it
+agentsview pg service status      # show manager status
+agentsview pg service logs -f     # follow the service log
+agentsview pg service uninstall   # stop and remove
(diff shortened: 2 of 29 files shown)


mxfeinberg:
@wesm , I just pushed the last couple of tweaks to this work.  As mentioned in my top comment, I added a system service (for MacOS and Linux) intended to push data from remote machines to a central pg instance without the need for a dedicated cron or manual pushes.  It was motivated by a deployment of hermes that I just started using. I also added some other tweaks to the hermes parser because none of its tokens were being priced properly via my deployment.

Let me know if there's any interest in merging these changes or changes like this! No worries if not, but I've been a fan of using agentsview and roborev per a recommendation from Hugh, and I thought I could contribute back.

wesm:
Merging this! Thank you

PR #549, opened 2026-05-26
install.ps1: use HTML redirect endpoint to avoid GitHub API rate limit

wesm:
Follow-up to #546, which fixed `install.sh` and the in-binary `agentsview update` command but missed `install.ps1`.

`Get-LatestVersion` now requests `github.com/kenn-io/agentsview/releases/latest` with `MaximumRedirection=0` and reads the `Location` header instead of calling `api.github.com` (60 req/hr per IP for unauthenticated callers).

Handles both PowerShell 5.x (throws `WebException` on 3xx) and PowerShell 7+ (returns the 302 response normally).

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -59,13 +59,47 @@ function Invoke-WebRequestCompat {
 }
 
 function Get-LatestVersion {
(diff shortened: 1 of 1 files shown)


wesm:
hallucination. merging

PR #550, opened 2026-05-27
Add SECURITY.md describing current posture and open questions

mjacobs:
agentsview already makes a number of substantive security and privacy decisions in code — the loopback default, bearer-token API gating, DNS-rebinding and CSP defenses, rejection of plaintext PG connections, the best-effort secrets subsystem — but they are not written down in one place. This PR adds `SECURITY.md` documenting today's posture: deployment model, threat model in and out of scope, trust boundaries, the full set of outbound channels (loopback API, `pg push`, SSH sync, LiteLLM pricing, update check), data-at-rest sensitivity, browser-facing hardening, and the secrets subsystem's best-effort scope. It is intentionally descriptive rather than prescriptive — a baseline that captures what is already true, with a short Open Questions section for the calls this doc deliberately does not make. The Reporting section leaves a TODO for a maintainer to enable GitHub private vulnerability reporting or publish a private contact.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,202 @@
+# Security and Privacy Posture
+
+## Status
+
+Foundational document. Describes agentsview's current security and privacy
+behavior and the assumptions that behavior rests on, much of which is
+already enforced in code but has not been written down in one place. It
+is intentionally descriptive rather than prescriptive — several questions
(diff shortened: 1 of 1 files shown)


wesm:
Useful and probably overdue, thank you!

PR #552, opened 2026-05-27
feat: parse decrypted Antigravity CLI trajectory sidecars

mjacobs:
Adds support for `<uuid>.trajectory.json` sidecars next to the existing AES-encrypted `<uuid>.pb` files under `~/.gemini/antigravity-cli/{conversations,implicit}/`. The sidecars are produced out-of-process by [agy-reader](https://github.com/mjacobs/agy-reader), which holds the decryption path and writes the resulting transcript as plain JSON; agentsview's parser now detects the sidecar and uses it as the source of truth for messages, tool calls, and tool results, falling back to the existing `history.jsonl` + brain artifact path when no sidecar is present. The sync engine treats the sidecar as part of the session for change detection (combined mtime/size via `AntigravityCLIFileInfo`) and forces message replacement on AgentAntigravityCLI sessions so trajectory updates land cleanly. Per the trust-boundary table in `SECURITY.md`, sidecars are treated as untrusted structured input under the "Imports and new readers" row: the parser is defensive against unknown step types, volatile fields are decoded through `json.RawMessage` accessors, and reads are size-capped at 64 MB. No new outbound channel is introduced; agy-reader is local-only. The legacy in-process `ANTIGRAVITY_KEY` decrypt path is preserved as a fallback for users without agy-reader installed; consolidating on the sidecar path is left for a follow-up.


mariusvniekerk (approved):
This helps maintain feature compatibility with the older gemini cli, thanks a lot for this one

wesm:
I think we're good here. @mjacobs hopefully my roborev fixes didn't break anything!

PR #578, opened 2026-06-01
Use kit daemon runtime

mariusvniekerk:
## Summary
- replace the custom daemon state-file and PID-lock implementation with go.kenn.io/kit daemon runtime records and start locks
- expose the shared /api/ping liveness endpoint for daemon discovery
- remove obsolete platform-specific boot/process-start stale PID checks and the internal pidlock package

## Validation
- go test ./internal/server ./cmd/agentsview
- go test ./... (fails in existing unrelated internal/parser TestParseWorkBuddySession: expected project `proj`, got `agentsview`)


mariusvniekerk:
Disposition on the latest roborev comment for `3ee9562`:

- The `WithDataDir` finding is stale/false on current head. `cmd/agentsview/pg.go` now passes `server.WithDataDir(appCfg.DataDir)` explicitly. We should keep that explicit call instead of defaulting `Server.New` from config.

- I do not think the PID-freshness/token-exfiltration item is worth carrying more bespoke code for the legacy migration path. This is local daemon discovery in the user's data dir, the runtime/token material is already local-user scoped, and the legacy `server.<port>.json` adapter is one-release transitional code that should disappear after migration. Reintroducing cross-platform process-start validation here is security machinery for a same-user local attacker model, not a proportional fix.

- The only part that is a real engineering tradeoff is DB ownership for an active pre-kit daemon whose legacy state exists but cannot be probed. If we choose to change anything, the right minimal behavior is a fail-closed one-release legacy writable-state fallback for write suppression, not broad PID freshness/security work in the migration shim. That accepts possible transient false-positive write refusal during upgrade in exchange for avoiding direct SQLite writes beside an old daemon.

mariusvniekerk:
Additional pushback on the latest roborev findings:

The PID-freshness / bearer-token exfiltration finding is not a meaningful blocker for this PR. The scenario requires a same-user local process to control the agentsview data-dir/runtime-file environment and bind the recorded local port while matching the recorded PID. That actor is already inside the local trust boundary for this CLI config and token material. Adding more cross-platform process identity code to a one-version legacy migration shim would increase complexity for a threat model this migration code should not own.

The `Server.New` / `dataDir` finding is also stale on current head: `pg serve` passes `server.WithDataDir(appCfg.DataDir)` explicitly, which is the intended contract.

The only finding with practical weight is the transient upgrade DB-ownership behavior for pre-kit writable daemons that cannot be probed. If we address that, it should be a narrow fail-closed legacy writable-state fallback for write suppression during the migration window, not security hardening bolted onto temporary migration code.

mariusvniekerk:
Clarifying the DB-ownership finding: this is only about the old legacy `server.<port>.json` state file during the immediate migration window. It is not a change to the new kit runtime ownership semantics.

We are not trying to preserve long-term compatibility for unprobeable pre-kit JSON state. The intended behavior is to migrate a legacy record when it can be proven by ping, otherwise leave it alone and move on. Treating every live legacy PID as authoritative would keep the old stale-PID behavior alive in temporary code, which is exactly the wrong tradeoff for this PR.

So I do not consider the legacy JSON fallback finding a blocker unless we explicitly decide that rolling-upgrade support for already-running pre-kit daemons is required. For this PR, the durable behavior is the kit runtime path, and that path is preserving the DB-ownership suppression semantics for live writable runtime records/start locks.

mariusvniekerk:
Final disposition on the legacy JSON ownership concern:

This finding depends on running a newer CLI against an already-running older pre-kit daemon that only published `server.<port>.json`. That mixed-version CLI/daemon setup is not a supported compatibility target for agentsview.

Given that constraint, we should not preserve DB-ownership semantics for unprobeable legacy JSON files beyond the immediate best-effort migration path. The durable contract for this PR is the kit runtime record/start-lock behavior used by matching CLI and daemon versions. The legacy JSON adapter is temporary migration code and should not accumulate extra fallback/security machinery for unsupported mixed-version operation.

So this concern is intentionally non-blocking for this PR.

wesm:
Not real issues. rebasing again

Issue #579, opened 2026-06-02
Infer Antigravity CLI project when history rows omit conversationId


mjacobs:
Newer Antigravity CLI history rows can include `display`, `timestamp`, and `workspace` without a `conversationId` field. Agentsview's current Antigravity CLI project map only uses history rows with `conversationId`, so SQLite `.db` sessions still parse their transcript but may show an empty project.

Observed local example:

```json
{"display":"Please review the 2 most recent git commits","timestamp":1780390445595,"workspace":"/home/mj/dev/projects/agentsview"}
```

The matching conversation DB was discoverable as `~/.gemini/antigravity-cli/conversations/d0a870aa-d473-4d32-8aef-ebcd21fad824.db`, and the parsed transcript's first user message matched the history `display` text, but project inference could not use the history row because it had no ID.

Possible approach: for Antigravity CLI `.db` sessions, when no direct `conversationId` mapping exists, infer workspace from a nearby history row by matching first user prompt text and/or timestamp proximity. This should remain conservative so unrelated prompt repeats do not assign the wrong project.

PR #580, opened 2026-06-02
Support Antigravity CLI SQLite sessions

mjacobs:
## Summary

*Resolves a breaking change in antigravity-cli release 1.0.4 (2026-06-01) release.*

Adds support for newer Antigravity CLI sessions that are stored as per-conversation SQLite `.db` files under `~/.gemini/antigravity-cli/conversations/`.

This keeps the existing encrypted `.pb` + `agy-reader` sidecar flow intact, while allowing Agentsview to index the newer DB-backed sessions directly.

## Changes

- Discover Antigravity CLI `conversations/<uuid>.db` files and prefer them over same-ID `.pb` files.
- Classify `.db`, `.db-wal`, and `.db-shm` filesystem events back to the canonical DB source.
- Parse CLI SQLite DB steps through the existing Antigravity SQLite/protobuf path.
- Include SQLite WAL/SHM mtimes in effective file info so live session updates resync reliably.
- Clean noisy internal strings from DB-decoded transcripts, including step headers, UUIDs, opaque request IDs, model placeholders, AGY config paths, and tool-action JSON blobs.
- Tighten prompt replacement from `history.jsonl` so sparse or blank history rows do not misassign prompts.
- Update README guidance for the two Antigravity CLI storage formats.

## Root Cause

Recent Antigravity CLI releases now write `conversations/<uuid>.db` instead of the older encrypted `conversations/<uuid>.pb` format. Agentsview only discovered `.pb` files for Antigravity CLI, so newer sessions were invisible. Once DB discovery was added, the old protobuf string-extraction heuristic surfaced internal protocol strings in the visible transcript; this PR filters those strings and prefers human-facing prompt text.

## Follow-up

Tracked in #579: newer `history.jsonl` rows may omit `conversationId`, so DB sessions can parse cleanly but still lack project inference from history. That should be handled conservatively by matching prompt/timestamp proximity rather than assuming every nearby history row belongs to a DB.

## Validation

- `go test ./internal/parser ./internal/sync`
- Tested against local Antigravity CLI `.db` sessions before and after the code changes; transcript parsing produced a clean first user message without step headers or internal IDs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,12 +255,14 @@ Each directory can be overridden with an environment variable. See the
 
 ### Antigravity CLI: high-resolution transcripts
 
-By default, agentsview indexes Antigravity CLI sessions in **summary mode**:
-your prompts from `history.jsonl` plus any plain-text artifacts under `brain/`
-(plans, walkthroughs, checkpoints). Assistant turns and tool calls live in
-AES-GCM-encrypted `.pb` files and are not visible in this mode.
+Antigravity CLI sessions now appear in two on-disk formats. Newer releases
+store conversation trajectories as SQLite `.db` files, which agentsview indexes
+directly. Older releases stored assistant turns and tool calls in
+AES-GCM-encrypted `.pb` files; for those sessions, agentsview falls back to
+**summary mode** using your prompts from `history.jsonl` plus any plain-text
+artifacts under `brain/` (plans, walkthroughs, checkpoints).
 
-To unlock full transcripts, run
+To unlock full transcripts for older `.pb` sessions, run
 [agy-reader](https://github.com/mjacobs/agy-reader) alongside agentsview.
 agy-reader talks to the local Antigravity daemon, decrypts each conversation,
 and writes a `<uuid>.trajectory.json` sidecar next to the encrypted `.pb` file.
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -28,12 +28,22 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
-// Bumped to 30: Hermes parser no longer treats cost_status
+// Bumped to 32: Antigravity DB parsers now filter internal
+// protocol strings from visible message content, remove raw step
+// headers, prefer prompt-like user text, and merge matching
+// Antigravity CLI history prompts when DB decoding drops short user
+// turns. Existing Antigravity DB rows need re-parsing so previously
+// indexed noisy or assistant-only transcripts are rewritten.
+//
+// (31: Copilot shutdown usage events use positional DedupKey to
+// handle multi-segment sessions correctly.)
+//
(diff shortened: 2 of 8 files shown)


mjacobs:
> ## roborev: Fail
> Summary verdict: One medium correctness issue found; no medium-or-higher security issues reported.
> 
> ## Medium
> * **internal/parser/antigravity_cli.go:170**: For `.db` sessions, any displayable DB message sets `hasTrajectory=true`, which skips the `history.jsonl` fallback. Since DB user steps are decoded through `agProtoCollectStrings(..., 20)`, short prompts like `continue` or `fix lint` can be dropped entirely, leaving assistant-only transcripts even though the exact prompts are available in history.
>   
>   * **Fix**: When parsing CLI DB sessions, merge or insert history prompts when decoded DB user turns are missing, or at least fall back to history when the DB trajectory has no decoded user messages.
> 
> _Panel: ci_default_security | Synthesis: codex, 7s, ~$0.02 | Members: codex_default (codex/default, done, 7m25s, ~$2.99), codex_security (codex/security, done, 1m38s, ~$0.36) | Total: 9m10s, ~$3.37 | Job: 19528_

:+1:

Addressed in 08ccced to merge matching `history.jsonl` prompts into `.db` transcripts when DB decoding misses user turns, including the short-prompt case where the protobuf string threshold drops text like `fix lint`. Added regression coverage in `TestAntigravityCLIDBInsertsShortHistoryPrompt`

mjacobs:
> ## roborev: Fail
> Medium parser migration issue remains; no qualifying security findings.
> 
> ## Medium
> * `internal/parser/antigravity.go:230` - The Antigravity DB parser changes stored message content by filtering/replacing transcript strings, but the parser data version was not bumped. Existing Antigravity DB sessions already stamped with the current `data_version` will be skipped when their source file mtime/size is unchanged, leaving old noisy content indefinitely.
>   
>   * **Fix:** Increment `internal/db/db.go`'s `dataVersion` and document the Antigravity parser change so startup forces a full resync.
> 
> _Panel: ci_default_security | Synthesis: codex, 6s, ~$0.02 | Members: codex_default (codex/default, done, 3m53s, ~$1.55), codex_security (codex/security, done, 2m9s, ~$0.58) | Total: 6m8s, ~$2.15 | Job: 19531_


Thanks, good catch. Addressed in ec6fba6 by bumping `dataVersion` to 31 and documenting the Antigravity DB parser cleanup, so existing indexed DB sessions are forced through the full resync path and old noisy/assistant-only stored messages get rewritten.

  What changed:

  - Bumped internal/db/db.go dataVersion from 30 to 31.
  - Added the data-version note documenting the Antigravity DB parser cleanup and short-prompt history merge.
  - This forces a full non-destructive resync so already-indexed noisy Antigravity DB sessions get rewritten even when source mtime/size is unchanged.

wesm:
false positive

PR #581, opened 2026-06-03
Add per-session usage REST endpoint

cpcloud:
## Summary
- add GET /api/v1/sessions/{id}/usage using the existing SessionUsage DTO and GetSessionUsage aggregation
- support the usage endpoint in both SQLite serve and pg serve paths
- preserve the successful REST and CLI usage JSON contract, including cost_usd, has_cost, and server_running
- return structured error codes for missing sessions and usage query failures

## Test Plan
- git diff --check
- nix develop --impure --expr "let pkgs = import (builtins.getFlake \"nixpkgs\") { system = builtins.currentSystem; }; in pkgs.mkShell { packages = [ pkgs.golangci-lint pkgs.go pkgs.clangStdenv.cc pkgs.darwin.libresolv ]; }" --command go test ./internal/server ./cmd/agentsview ./internal/db -count=1
- nix develop --impure --expr "let pkgs = import (builtins.getFlake \"nixpkgs\") { system = builtins.currentSystem; }; in pkgs.mkShell { packages = [ pkgs.golangci-lint pkgs.go pkgs.clangStdenv.cc pkgs.darwin.libresolv ]; }" --command make lint-ci

### DIFF
--- a/internal/server/session_usage.go
+++ b/internal/server/session_usage.go
@@ -12,6 +12,15 @@ type sessionUsageResponse struct {
 	ServerRunning  bool     `json:"server_running"`
 }
 
+type sessionUsageErrorResponse struct {
+	Error sessionUsageError `json:"error"`
+}
+
+type sessionUsageError struct {
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
+
 func (s *Server) handleSessionUsage(
 	w http.ResponseWriter, r *http.Request,
 ) {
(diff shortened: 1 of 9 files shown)


internal/server/session_usage.go
```diff
@@ -0,0 +1,44 @@
...
+
+	writeJSON(w, http.StatusOK, newSessionUsageResponse(usage))
+}
+
+func newSessionUsageResponse(usage *db.SessionUsage) sessionUsageResponse {
+	unpricedModels := usage.UnpricedModels
+	if unpricedModels == nil {
+		unpricedModels = []string{}
+	}
+	return sessionUsageResponse{
```
mariusvniekerk: Why the adapter, seems weird?

cpcloud: Yeah, that IS weird. Should be able to remove that.

cpcloud: Clarifying what I meant by adapter here: `sessionUsageResponse` is only the private HTTP response wrapper for the successful `200` body, not a replacement for `db.SessionUsage`. It embeds `db.SessionUsage` so the existing DB/CLI JSON tags remain the REST success contract, then adds the HTTP-only `server_running: true` field and normalizes `unpriced_models` to `[]` when the DB DTO has nil.

The separate `sessionUsageErrorResponse` types are only for the new coded error body. Those are small enough that they could be simplified or shared if preferred; the success wrapper is there to avoid changing `db.SessionUsage` just to carry a REST-only field.

PR #590, opened 2026-06-04
Avoid git subprocesses during sync project resolution

wesm:
## Summary
- Prefer the cheap filesystem `.git` walk for sync-time project extraction so normal repos and linked worktrees do not shell out to `git` per session.
- Preserve the kit `gitrepo.MainRoot` fallback for rare live-repo layouts when the local walk cannot resolve a root.
- Add regression coverage for both sides: plain repos must not invoke `git`, and the git fallback remains available when local resolution misses.

## Investigation
- Reproduced with isolated data dirs only (`AGENTSVIEW_DATA_DIR=/tmp/...`, also set `AGENTSVIEW_HOME=/tmp/...`); production DB was not touched.
- Same source session set: about 35.7k discovered sessions.
- `v0.31.1` cold sync: 1m12.7s.
- `v0.32.0` cold sync: 4m31.4s.
- Revised PR build cold sync: 1m14.2s.
- 0.32.0 CPU profile showed ~190s cumulative under `go.kenn.io/kit/git/repo.MainRoot` / `os/exec` from `ExtractProjectFromCwdWithBranch`; revised PR profile returns to SQLite/file-read dominated work.

## Test Plan
- [x] `go test ./internal/parser ./internal/sync ./cmd/agentsview`
- [x] Isolated revised PR cold sync with `AGENTSVIEW_DATA_DIR=/tmp/agentsview-prof-pr-fallback`: 1m14.2s

### DIFF
--- a/internal/parser/project.go
+++ b/internal/parser/project.go
@@ -459,36 +459,42 @@ func findGitRepoRoot(ctx context.Context, cwd string) string {
 		}
 	}
 
-	if root := findGitRepoRootLocal(dir); root != "" {
+	root, conservative := findGitRepoRootLocal(dir)
+	if root != "" && !conservative {
 		return root
 	}
 	if !cwdMissing {
-		return gitMainRoot(ctx, startDir)
+		if gitRoot := gitMainRoot(ctx, startDir); gitRoot != "" {
+			return gitRoot
+		}
 	}
-	return ""
+	return root
 }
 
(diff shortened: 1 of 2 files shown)


wesm:
Addressed the roborev finding in b3a6977b. The resolver now distinguishes authoritative local roots from conservative gitfile fallbacks and tries gitrepo.MainRoot before accepting the conservative worktree dir. Added regression coverage for that case.

PR #601, opened 2026-06-07
Show agent-provided session names in the sidebar

gordonwoodhull:
I use the `/rename` feature pretty consistently in Claude Code, and wanted those names to show up in the left pane of Agentsview.

It looks like there may have been previous work in this area but it was never complete. 

### Summary

Adds an opt-in **"Use session names"** Appearance toggle (default off) that shows agent-provided session names in the sidebar: Claude Code `/rename` names plus the native conversation titles seven other parsers already extract (ChatGPT, Claude.ai, Forge, Hermes, Kiro, Piebald, Cortex).

A new `name_source` column (`user` | `agent`) records provenance. A manual in-app rename is marked `user` and always shows, and is never overwritten by an agent name; agent names are marked `agent`, update live (including a full-reparse fallback when a `/rename` is appended mid-session), and are shown only when the toggle is on. Parser-extracted names are now carried through `toDBSession` (previously dropped on the live disk-sync path — only the importer persisted them), with matching PostgreSQL push/read parity, a `dataVersion` bump so existing sessions re-parse to backfill the column, and resync metadata-copy logic that preserves user-owned names while letting agent names refresh.

Because the names live in the skinny sidebar index, agent-named rows render immediately on load without per-row hydration.

#### Other agents/chat apps with a session name

The feature, if enabled, should also work for these coding agents. (Untested.)
  - Cortex Code
  - Forge
  - Hermes Agent
  - Kiro IDE
  - Piebald
  
And these imported chats (2):
  - ChatGPT (conversation title)
  - Claude.ai (conversation name)


gordonwoodhull:
Thanks for the review — addressed all three Medium findings, each in its own commit with a test that fails on the unpatched code:

1. **Sidebar hydration vs. the toggle** (`SessionList.svelte`) — `needsVisibleHydration` treated any `display_name` row as hydrated, so default-off agent-named rows fell through to the project-name fallback instead of the first-message preview. It now hydrates a row unless its name will actually render. (`b3e5b8a`)

2. **`name_source` dropped on the real push path** (`sessions.go`) — `sessionFullCols` omitted `name_source`, so `ListSessionsModifiedBetween` (and `GetSessionFull`) returned `NameSource == nil` and the PG push stored NULL despite the insert/update changes. Added `name_source` to `sessionFullCols` and both scanners, plus a test that exercises the real `Push` path rather than only direct `pushSession`. (`6c51e6c`)

3. **Push fingerprint omitted `name_source`** (`push.go`) — a session whose only change was `name_source` matched the previous fingerprint and was skipped, leaving PG stale. Added `stringValue(sess.NameSource)` to the fingerprint fields and covered it in the fingerprint test. (`8717d6b`)

---

Note: the failing frontend `check` job is unrelated to this PR. On a clean `main` (no changes from this branch), `npm run check` already fails with `Unused '@ts-expect-error' directive` in `src/lib/virtual/createVirtualizer.svelte.ts:100` — it appears to be a pre-existing interaction between #589 (svelte-check 4.6) and #593 (dependency bump) that surfaces only once both are on `main`. I didn't want to fix this without understanding it.

gordonwoodhull:
Addressed the follow-up finding (`8717d6b` review):

- **`CheckSchemaCompat` didn't probe `sessions.name_source`** (`internal/postgres/schema.go`). The PG session/sidebar read paths now select `name_source`, so a legacy schema under a read-only/non-migrating role could pass the compat check and then fail at runtime on the first session query. Added `name_source` to the probe, with a regression test mirroring the existing `secrets_rules_version` check (verified test-first: it fails when the column is dropped, passes after the probe change). (`63fbdea`)

I also swept the rest of the codebase for the same class of omission (every `display_name`/read-path that should parallel `name_source`). The parser/agent-label/rename-handler references are correct as-is; the only other place worth a mention is the import/export format, which doesn't carry `name_source` (a re-import leaves an agent-named row ungated but still shown). 

I didn't think it belonged in the export format, but happy to add if needed.

gordonwoodhull:
Addressed the breadcrumb/consistency finding (`42126be`):

- Extracted a shared `visibleSessionName(session)` helper (`frontend/src/lib/utils/sessionName.ts`) that returns the name unless it's `name_source === "agent"` and the "use session names" toggle is off. Manual renames (`"user"`) and legacy/imported (`null`) names always show.
- Routed every session-name display path through it: the sidebar (`SessionItem`/`SessionList`, refactored from the old inline check), the active-session **breadcrumb**, the **delete-confirmation modal**, the **trash** list, the **pinned** page's store-backed path, and the **rename input prefill** (so double-click-to-rename prefills the preview, not a hidden agent name). Each keeps its existing fallback.
- The agent-visibility decision is isolated in a single predicate so a future "always-on" migration (dropping the toggle) is a one-line change.
- Added a unit test matrix for the helper; all 1214 frontend tests pass.

**One remaining ungated path — question for @roborev-ci:** the Pinned page's common code path renders `pin.session_display_name` straight from the pinned-messages API (`PinnedMessage`), which carries `session_display_name` but **not** `session_name_source` — so it can't be gated client-side. Closing it would mean threading `session_name_source` through the pins API (`internal/db/pins.go`, `internal/postgres/curation.go`, the API response, and the `PinnedMessage` TS type), parallel to the existing `session_display_name`.

@roborev-ci — should we thread `session_name_source` through the pins API to gate that last path, or is gating the pinned **list** (vs. the sidebar) out of proportion to the benefit? Want to confirm the direction before adding cross-layer plumbing.

gordonwoodhull:
Hmm, this seems to be sprawling. I'll convert to draft and see if there is any interest from human maintainers, and what the scope should be.

Just changing the titles in the sidebar was easy, but it would probably be better to use consistent names everywhere.

An alternative design would be having `session_name` as a separate field from `display_name`, instead of `name_source`. Either way, it's an extra field which tends to propagate.

mariusvniekerk:
I'm not entirely sure if we want to have the complexity of the feature flagging?  Presumably if a user went to the effor of naming their sessions using /rename etc there is no reason to not show those.  

I would hazard a guess that removing the use-session-names toggle removes the need for a frontend name-source field etc.  Ultimately having some coalesce based cascade feels like the correct implementation that users would expect.

gordonwoodhull:
Hi @mariusvniekerk. Thanks for the review! 

Sure, I was being cautious with the feature flag but I agree it’s probably not worth the complexity. 

I think an extra field would be required for the behavior that user-entered name is preferred to session name, and goes back to session name if the entered name is removed. It could be a flag (as implemented) or a string.

However, maybe if we relax it to last-one-wins or letting the session name win over the user-entered name, it could be just one field. 

I’ll investigate. 

Result: still two fields, but db-only; see below.

mariusvniekerk:
@gordonwoodhull on the db model we should store all the names we can, and just send the correct computed field to the frontend from the golang side, keeps this a concern for the backend only and the frontend remains blissfully unaware

gordonwoodhull:
Thanks @mariusvniekerk — fully agree. I've reworked the branch to implement the coalesce-based cascade with `name_source` as a backend-only concern:

**What changed**

The "Use session names" toggle and all frontend `name_source` threading are gone. Instead, the Go layer resolves the name before it reaches the frontend:

- `name_source` stays in SQLite and PG schemas as a write-time guard: the upsert `CASE WHEN name_source = 'user'` protects user renames from being overwritten on re-sync, and `RenameSession` stamps `name_source = 'user'` / clears to `NULL`.
- `name_source` is stripped from all JSON responses (`json:"-"` on both `Session` and `SidebarSessionIndexRow`, removed from the sidebar query and PG read path). The frontend never sees it.
- Every display surface just reads `display_name`, falling back to `first_message` preview → project name. No helper, no conditional.

**Cascade behavior**

- Agent `/rename` or native conversation title → `display_name = "name"`, `name_source = "agent"` on parse
- User renames in-app → `display_name = "name"`, `name_source = "user"`; upsert CASE preserves this on all subsequent re-syncs
- User clears their rename → `display_name = NULL`, `name_source = NULL`; the server immediately re-parses the session file and restores the agent name in the same HTTP response (no 15-minute wait)

**Bugs fixed along the way**

- `upload.go` was silently dropping `ParsedSession.DisplayName` — fixed
- `importer.go` had a hand-rolled "preserve existing name" guard that compared names by value instead of checking `name_source` — replaced with the upsert CASE
- `ImportChatGPT` wasn't setting `name_source` — fixed
- All three parse→DB converters now go through a shared `ParsedSessionNameFields` helper so a fourth can't drift

**What's kept**

- `dataVersion = 34` bump + `BackfillNameSource` migration (stamps `name_source = 'user'` on pre-feature non-NULL `display_name` rows so existing renames survive)
- `name_source` in PG DDL, `CheckSchemaCompat`, push fingerprint, and push `INSERT … ON CONFLICT` — all write paths intact
- All parser extraction (Claude `/rename`, 7 native-title parsers)

gordonwoodhull:
(Response written by AI for concision and accuracy; opinions are mine.)

**Finding 1 (PG `name_source` read path):** Removing `name_source` from the PG read path was intentional — it's the mechanism that keeps `name_source` a backend-only concern and out of the frontend. Adding it back to `pgSessionCols`/`scanPGSession` would restore exactly the kind of threading across layers that the simplification was meant to eliminate. The PG write path (push INSERT, fingerprint, `CheckSchemaCompat`) still carries it correctly. Happy to add a comment to `pgSessionCols` making the omission explicit, but I'd rather not add the field back unless there's a concrete reason it's needed on the read side.

**Finding 2 (imported sessions):** This one is real. File-backed sessions (Claude Code, Copilot, etc.) work correctly — clearing a user rename immediately re-parses the session file and restores the agent name. But imported sessions (Claude AI export, ChatGPT export) have no backing file, so clearing a rename falls back to the first-message preview rather than the imported conversation title. The fix would require storing the user override and the agent title in separate fields, which is the more complex two-field design we moved away from. What would you like to do here?

gordonwoodhull:
**Previous round of fixes (`90d586d`, not yet commented on):**

roborev's review of `90d586d` found two issues, both fixed in `00b168d`:

- **Medium** — clearing a user rename left the old name visible in the SPA when no agent name was available, because `display_name: null` is omitted from the JSON response (Go's `omitempty`). Fixed in `sessions.svelte.ts`: `renameSession` now explicitly sets `display_name = null` in the store when the caller passed `null` and the response omits the field.

- **Low** — `normalizeMessagePreview()` returns `""` for null/empty input, making `??` the wrong operator before the project/default fallback in `ConfirmDeleteModal` and `PinnedPage`. Fixed to use `||` (falsy check) instead, matching the original intent.

---

**Current round (`b099eba`):**

**Finding 1 (imported sessions):** Already raised with the human reviewers [here](https://github.com/kenn-io/agentsview/pull/601#issuecomment-4654634109) — still waiting on direction.

**Finding 2 (`BackfillNameSource` pins imported titles as user-owned):** Fixed. The migration comment said "pre-feature databases stored only manual renames in `display_name`" — but the importers also wrote conversation titles directly into `display_name`. Restricting the WHERE clause to `file_path IS NOT NULL` correctly identifies genuine user renames: file-backed sessions always have a path; imported sessions never do. Imported rows are left with `name_source = NULL`, so future re-imports can update their titles via the normal upsert CASE.

gordonwoodhull:
**Finding 1 (imported sessions):** Covered [here](https://github.com/kenn-io/agentsview/pull/601#issuecomment-4654634109) — still waiting on direction from the human reviewers.

**Finding 3 (PG `name_source` read path):** Covered [here](https://github.com/kenn-io/agentsview/pull/601#issuecomment-4651289148) — omitting it from `pgSessionCols` is intentional. Including it and relying on `json:"-"` would work too, but the effect is the same: the field never reaches the frontend. Happy to add either way if there's a preference.

**Finding 2 (Claude fork sessions, `db69e7c`):** Fixed. Claude fork sessions share a `.jsonl` file but have separate rows in the DB. The previous fix called `ResetSessionFileState(id)` which cleared `file_size`/`file_mtime` for the one session being renamed — but `shouldSkipFile` looks up the skip state using the root session ID derived from the filename, not the fork ID. So the root's stored state remained intact and the re-parse was still skipped. Replaced with `ResetFileStateByPath(path)` which clears all sessions sharing the file, so `shouldSkipFile` returns false for root and fork alike.

gordonwoodhull:
**Finding 1 (imported sessions):** Covered [here](https://github.com/kenn-io/agentsview/pull/601#issuecomment-4654634109) — still waiting on direction.

**Finding 3 (PG `name_source` read path):** Covered [here](https://github.com/kenn-io/agentsview/pull/601#issuecomment-4651289148) — intentional omission. `name_source` stays in `sessionBaseCols`/`sessionFullCols` so the Go struct is populated for the PG push path, but `json:"-"` keeps it out of API responses. Including it in `pgSessionCols` and relying on `json:"-"` alone would work equally well — happy to do that if preferred.

**Finding 2 (dead `ResetSessionFileState`, stale `CheckSchemaCompat` probe, `910c576`):** All three fixed:

- `ResetSessionFileState` removed — superseded by `ResetFileStateByPath`
- `name_source` removed from `CheckSchemaCompat` — it's a write-only column and `pg serve` with a read-only role against a pre-`name_source` schema would have crashed on startup checking for a column it doesn't actually read
- `TestGetSessionReturnsNameSource` renamed to `TestGetSessionPopulatesNameSourceInternally` with updated comments — the test was verifying that the internal Go struct field is populated (correct and needed for the push path), not that `name_source` appears in API responses (it doesn't)

The `CheckSchemaCompat` probe was added before the simplification, when the PG read path did select `name_source`. It should have been pruned in the same pass that removed it from `pgSessionCols`, but was kept on the assumption it was still needed. That was wrong — removing it now.

gordonwoodhull:
This is the same "imported sessions" item raised many times by roborev and addressed [here](#issuecomment-4654634109). Is it worth the complication of an extra field for this case?

mariusvniekerk:
having the agentsview-owned field (ie not inferred) does feel like a good thing, since we do occasionally reparse sessions from the storage, so that field could get lost otherwise.

gordonwoodhull:
Following up on @mariusvniekerk's guidance to store all names and send the correct computed field from the Go side — I've implemented the two-field design:

**What changed**

`display_name` and `session_name` are now separate columns with distinct owners:

- `session_name TEXT` — parser/importer-provided; freely overwritten on every re-parse (Claude `/rename`, native conversation titles from 7 other agents, importer titles)
- `display_name TEXT` — user override only; touched exclusively by `RenameSession`

All read queries return `COALESCE(display_name, session_name) AS display_name`. The frontend is unchanged — it still reads one `display_name` field and remains unaware of the distinction.

**What this fixes**

The long-standing "imported sessions lose their title when you clear a custom rename" issue: previously `display_name` was the only stored value, so clearing a user rename left nothing to fall back to. Now `session_name` is always there independently. Clearing a rename immediately reveals the session name via the COALESCE — no file re-parse required.

**What was deleted**

- `name_source` column and all its threading (CASE expression in upsert, schema, PG DDL, push fingerprint, `CheckSchemaCompat` probe)
- `BackfillNameSource` migration function (no longer needed — `display_name` on upgrade from main is always user-owned by construction)
- `ParsedSessionNameFields` (two-return helper) → replaced by `ParsedSessionName` (single `*string`)
- The rename-clear re-parse machinery (`ResetFileStateByPath`, `SyncPaths` call in the rename handler) — the two-field design makes it unnecessary

**What was renamed**
- `ParsedSession.DisplayName` renamed to `ParsedSession.SessionName` throughout all parsers — the root cause of the original conflation

gordonwoodhull:
Following up on the session_name two-field implementation with several fixes found after pushing:

**Fixes since the [last update](https://github.com/kenn-io/agentsview/pull/601#issuecomment-4665824308):**

- `internal/sync/engine.go`, `internal/server/upload.go` — converters were setting `Session.DisplayName` instead of `Session.SessionName`, so parsed `/rename` and native titles were silently dropped on sync and upload (`5700185`)
- `internal/db/pins.go` — pins metadata query used raw `display_name`; sessions with only a `session_name` fell back to first message on the Pinned page (`8668e73`)
- All PostgreSQL read paths — `pgSessionCols`, PG sidebar, PG search, PG usage, PG curation — now use `COALESCE(display_name, session_name)` (`aed042d`)
- `internal/postgres/schema.go` — `CheckSchemaCompat` now probes `session_name` since it's a genuine read column; PG search WHERE/CASE predicates updated to match on `session_name` as well as `display_name` (`20c5bee`)

---

**On the roborev migration finding:** roborev flagged a potential migration issue around legacy `display_name` values from the old importer. On reflection we don't think it applies: since the branch was never merged, no database ever ran the pre-refactor importer that wrote titles into `display_name`. All real-world `display_name` values are genuine user renames, so there's nothing to migrate. Happy to be corrected if we're missing something.

gordonwoodhull:
The one remaining finding is the scope-expanding migration finding; we don't think it's relevant because the previous code was never merged to main, so no one is in this situation

> **On the roborev migration finding:** roborev flagged a potential migration issue around legacy `display_name` values from the old importer. On reflection we don't think it applies: since the branch was never merged, no database ever ran the pre-refactor importer that wrote titles into `display_name`. All real-world `display_name` values are genuine user renames, so there's nothing to migrate. Happy to be corrected if we're missing something.

https://github.com/kenn-io/agentsview/pull/601#issuecomment-4666227422

Thanks for your consideration. Whether or not this makes it in, I've learned quite a lot from the process.

(I hope it makes it because I used Agentsview every day to tame voluminous review output.)

wesm:
I'm looking at this now

wesm:
This looks good to go, I'll leave the last roborev finding unaddressed (technically accurate, not worth fixing)

PR #606, opened 2026-06-08
fix(postgres): create content trigram index with fastupdate=off

### LINKED ISSUE #605: bug: GIN trigram index on messages.content bloats unbounded (fastupdate=on), fills disk
**Env:** AgentsView 0.32.1, PostgreSQL 18.4, multiple concurrent `pg sync` / `pg push` clients.

`idx_messages_content_trgm` grew to **283 GB** for a messages table of ~213k rows averaging ~594 B of content (10 GB heap, 535 MB TOAST) - i.e. 283 GB of a 294 GB database in a single index. It filled the root filesystem and crashed PostgreSQL with No space left on device.

**Cause:** `createContentSearchIndexesPG` (`internal/postgres/schema.go` ~L694, added in #534) creates the index with PostgreSQL's default `fastupdate=on`:

```sql
CREATE INDEX IF NOT EXISTS idx_messages_content_trgm
  ON messages USING gin (content gin_trgm_ops)
```

A GIN pending list is only reclaimed by `VACUUM`, but long-lived sync transactions pin the xmin horizon and starve autovacuum, so the list grows without bound.

**Fix:** append `WITH (fastupdate=off)` to that statement. Dropping and recreating the index with it reduced the same data to **388 MB**. Keeping sync transactions short (batched commits) so autovacuum can run would also help.

Possibly related, not duplicates: #360 (sync hang on large workloads), #331 (pg push slow for large databases).

flexiondotorg:
Closes #605.

Adds `WITH (fastupdate = off)` to the `idx_messages_content_trgm` creation in `createContentSearchIndexesPG`, and reapplies it idempotently via `ALTER INDEX` on every schema bootstrap so stores upgraded from a schema that created the index with the default `fastupdate=on` are covered too. A pgtest assertion guards `reloptions`. See #605 for the diagnosis and impact.

**Validation:** recreating the index with `fastupdate=off` reduced the same data from 283 GB to 388 MB.

Existing over-sized indexes stop growing immediately; a one-time `REINDEX` reclaims space already consumed.

### DIFF
--- a/internal/postgres/schema.go
+++ b/internal/postgres/schema.go
@@ -722,6 +722,18 @@ func createContentSearchIndexesPG(ctx context.Context, db *sql.DB) {
 		log.Printf(
 			"pg schema: creating messages.content trigram index failed: %v", err,
 		)
+		return
+	}
+	// CREATE INDEX IF NOT EXISTS only applies WITH (fastupdate = off) on
+	// first creation. Re-apply on every boot so stores upgraded from a
+	// prior schema (which left fastupdate=on) also get the bounded index.
+	if _, err := db.ExecContext(ctx,
+		`ALTER INDEX idx_messages_content_trgm SET (fastupdate = off)`,
+	); err != nil {
+		log.Printf(
+			"pg schema: disabling fastupdate on messages.content trigram index failed: %v",
+			err,
+		)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
So whats the performance impact of this change?

Since the alternate would be something like running `REINDEX CONCURRENTLY` on some cadence

PR #609, opened 2026-06-08
Add DuckDB mirror backend with Quack support

mariusvniekerk:
## Why this exists

agentsview now has two very different storage modes: SQLite as the local primary archive and PostgreSQL as an optional shared read backend. DuckDB fills a different operational need: a portable analytics mirror that can be served read-only from a local file or exposed remotely through DuckDB's Quack protocol without making parser ingestion depend on a third write path.

The design intentionally keeps SQLite as the source of truth for first-release ingestion. DuckDB is populated by push sync, served through the same `db.Store` surface as SQLite/PostgreSQL, and guarded by backend contract checks so future server-visible queries fail compilation until all providers implement them.

## How to use it

Local mirror flow:

```bash
agentsview duckdb push          # mirror local SQLite into ~/.agentsview/sessions.duckdb
agentsview duckdb status        # inspect mirror counts and last push
agentsview duckdb serve         # serve the web UI from DuckDB read-only
```

Scoped push examples:

```bash
agentsview duckdb push --projects project-alpha,project-beta
agentsview duckdb push --exclude-projects scratch,tmp
agentsview duckdb push --full --all-projects
```

Remote Quack flow:

```bash
# Terminal 1: expose a local DuckDB mirror over loopback Quack.
agentsview duckdb quack serve \
  --bind quack:127.0.0.1:9494 \
  --token "$AGENTSVIEW_DUCKDB_TOKEN"

# Terminal 2: serve the UI from that Quack endpoint.
AGENTSVIEW_DUCKDB_URL='quack:http://127.0.0.1:9494' \
AGENTSVIEW_DUCKDB_TOKEN="$AGENTSVIEW_DUCKDB_TOKEN" \
agentsview duckdb serve
```

For non-loopback Quack exposure, keep it behind TLS or a trusted tunnel/proxy. Plain HTTP non-loopback binds require both an explicit token and `--allow-insecure`.

Docker examples are in the README. The short version is that explicit subcommands work with the image:

```bash
docker run --rm -v agentsview-data:/data ghcr.io/kenn-io/agentsview:latest duckdb push --full
docker run --rm -p 127.0.0.1:8080:8080 -v agentsview-data:/data ghcr.io/kenn-io/agentsview:latest duckdb serve
```

## Implementation notes for review

- Adds DuckDB config, CLI commands, schema management, push sync, read store, analytics/usage reads, and curation/read-only behavior.
- Adds Quack client attach and `duckdb quack serve` with loopback-safe defaults, token handling, and integration tests.
- Extracts shared query/dialect helpers so session filtering and cursor pagination are not copied across SQLite/PostgreSQL/DuckDB.
- Keeps DuckDB search on substring/regex fallback for this release. DuckDB FTS is available locally in the pinned runtime, but BM25 lookup does not currently resolve through Quack-attached catalogs.
- Adds a central backend contract package that asserts SQLite, PostgreSQL, and DuckDB all satisfy `db.Store` at compile time.

## Validation

Fresh local validation run before opening this draft:

```bash
go test -tags duckdbtest ./cmd/agentsview ./internal/duckdb -run Quack -count=1
go test -tags fts5 ./cmd/agentsview ./internal/duckdb -count=1
go test -tags fts5 ./internal/backendcontract ./internal/db ./internal/postgres ./internal/duckdb -count=1
go vet ./...
make test-short
make e2e-duckdb
```


wesm:
rebasing this now

PR #610, opened 2026-06-09
feat: support Zed AI assistant sessions

arbing:
Add Zed as a supported agent. Sessions are stored in a shared SQLite
database (threads/threads.db) rather than per-session files, so the
integration follows the same pattern as Kiro: per-thread mtime skip
checks, a single open connection shared across the parse loop, and a
dedicated syncSingleZed path that bypasses FindSourceFile to avoid
transient DB errors being misreported as missing sessions.

Key behaviors:
- Discover and parse top-level threads from zstd- or JSON-encoded data
- Extract model name, token usage, tool calls, tool results, and thinking
- Emit session-level usage events so the Usage page reflects Zed data
- Normalize tool result content so DecodeContent can decode it
- Watch only threads/ subdirectory to avoid exhausting inotify budget
- Open DB with _busy_timeout=3000 to tolerate concurrent Zed writes
- Preserve user messages that have structured content but no text
- Set File.Hash consistently across all write paths


wesm:
Looks good here, thank you

PR #614, opened 2026-06-10
feat(frontend): show session cost in the session header

chrislee3408:
## Summary

The session detail header now shows the session's estimated dollar cost as a badge between the token summary (`386.6k ctx / 308.2k out`) and the model id (`claude-opus-4-8`).

The breadcrumb fetches the existing `GET /api/v1/sessions/{id}/usage` endpoint, which already computes `cost_usd`/`has_cost` from the `model_pricing` table on both backends (`internal/db/usage.go` and `internal/postgres/usage.go`), so this is a frontend-only change with no parity impact.

Implementation details:

- The fetch is keyed on `(id, total_output_tokens, message_count, ended_at)` rather than session id alone, because cost can move without output tokens moving (input/cache tokens and explicitly reported `cost_usd` from usage events all feed the backend computation). The session row is replaced wholesale on SSE updates, so the badge refreshes as a live session accumulates usage.
- A monotonic request sequence counter discards stale out-of-order responses, both across session switches and for same-session refetches.
- A shared `formatCost` helper in `lib/utils/format.ts` mirrors the formatting already used across the usage dashboard (`$0.42`, `$12.34`, `$123` above one hundred), with a `<$0.01` floor so priced-but-cheap sessions don't render as `$0.00`.

Limitations/tradeoffs: sessions whose models have no pricing data, sessions with no token data, or a failed/unsupported usage request render no badge at all (graceful degradation, no placeholder). The existing per-component cost formatters in `lib/components/usage/` were intentionally left untouched to keep the diff focused.

Where to look: `frontend/src/lib/components/layout/SessionBreadcrumb.svelte` for the fetch effect and badge markup; `SessionBreadcrumb.test.ts` for the race-condition and badge-placement tests (the generated API barrel is now partially mocked there, keeping `OpenAPI` intact for `configureGeneratedClient`).

## Screenshots

Same session (`2c85392b`, claude-opus-4-8) in the session header, before and after.

**Before** — token summary directly followed by the model id:

Session header before: token badge then model badge

**After** — estimated cost between the token summary and the model id:

Session header after: token badge, cost badge, model badge

### DIFF
--- a/frontend/src/lib/components/layout/SessionBreadcrumb.svelte
+++ b/frontend/src/lib/components/layout/SessionBreadcrumb.svelte
@@ -24,7 +24,7 @@
   import { configureGeneratedClient } from "../../api/runtime.js";
   import { copyToClipboard } from "../../utils/clipboard.js";
   import { agentColor, agentLabel } from "../../utils/agents.js";
-  import { formatTokenUsage } from "../../utils/format.js";
+  import { formatCost, formatTokenUsage } from "../../utils/format.js";
   import { normalizeMessagePreview } from "../../utils/messages.js";
   import { getGradeStyle, getGradeLabel } from "../../utils/grade.js";
   import SignalPanel from "../content/SignalPanel.svelte";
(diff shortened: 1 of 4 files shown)


wesm:
looking at this now

Issue #529, opened 2026-05-22
Feature request: Add support for antigravity-cli


juanfcocontreras:
Recently, Google has introduced antigravity-cli, which is the command-line tool that will replace gemini-cli in the future.

Would it be possible for agentsview to capture antigravity-cli conversations just as it does with gemini-cli?

mdfranz:
Yep just confirmed `v0.29.0 (commit c8bb17c0, built 2026-05-10T16:15:06Z)` doesn't support these... these are the files created

```
~/.gemini/antigravity-cli$ ls -al
total 28
drwxr-xr-x 1 mdfranz mdfranz  320 May 22 07:22 .
drwxr-xr-x 1 mdfranz mdfranz  320 May 22 07:17 ..
-rw------- 1 mdfranz mdfranz  505 May 22 08:40 antigravity-oauth-token
drwxr-xr-x 1 mdfranz mdfranz   40 May 21 22:50 bin
drwxr-xr-x 1 mdfranz mdfranz  648 May 22 07:23 brain
drwxr-xr-x 1 mdfranz mdfranz  100 May 22 06:51 cache
lrwxrwxrwx 1 mdfranz mdfranz   27 May 22 07:22 cli.log -> log/cli-20260522_072256.log
drwxr-xr-x 1 mdfranz mdfranz  624 May 22 08:32 conversations
-rw------- 1 mdfranz mdfranz 4367 May 22 08:32 history.jsonl
drwxr-xr-x 1 mdfranz mdfranz  312 May 22 07:14 implicit
-rw-r--r-- 1 mdfranz mdfranz   36 May 21 21:23 installation_id
-rw------- 1 mdfranz mdfranz  798 May 21 21:23 keybindings.json
drwxr-xr-x 1 mdfranz mdfranz   28 May 21 21:23 knowledge
-rw-r--r-- 1 mdfranz mdfranz    0 May 22 07:22 last_check.timestamp
drwxr-xr-x 1 mdfranz mdfranz  460 May 22 07:22 log
-rw------- 1 mdfranz mdfranz  247 May 22 07:11 settings.json
drwxr-xr-x 1 mdfranz mdfranz   22 May 21 21:23 updater
```

juanfcocontreras:
It looks like this was added in the latest version (0.30.0):

https://github.com/kenn-io/agentsview/pull/528

However, I only see my messages, not the AI’s responses.

mjacobs:
[...]

PR #619, opened 2026-06-10
Antigravity token usage extraction

vikram:
related to #529 

feat: extract and sync token usage for antigravity sessions
feat(pricing): support robust custom model pricing resolution
fix: reject implausibly large token values in antigravity gen_metadata parser

The extractTokenUsage heuristic matches nested messages where field1 is a model-kind varint in [1000, 5000) and treats field2 as the output token count. This is broad enough to match unrelated messages — for example, a scheduler event containing a nanosecond-latency counter whose field1 happened to fall in range, causing a session to report hundreds of millions of output tokens.

Add maxPlausibleTokens = 2_000_000 as an upper bound on field2. Candidates above the cap are skipped and the walk continues to find the real token block. The check is per-message, not per-session.

Update the design spec to document the false-positive risk and the guard. Add TestExtractTokenUsageLargeValueFalsePositive as a regression test.


wesm:
i'm looking at this now

mjacobs:
heads up, #620 changes which message set wins for .db sessions, so usage attached inside the DB decode won't reach totals once both land; also FYI the trajectory sidecar has labeled inputTokens/outputTokens/tokenBreakdown fields that could replace the varint-range heuristic and its plausibility cap.

wesm:
ah okay, I can work on merging #620 first then

vikram:
there are a couple of things I noticed, it doesnt parse tool information,
this is seen in the UI, will also look at reasoning tokens tonight. Will
look at sync/sessions.db to see if we can tell where the parser didnt quite
work. (separately), maybe this is already there and I didnt notice it.

thank you for the engagement. fantastic tool.

mjacobs:
the missing tool info is expected without agy-reader sidecars: the raw `.db` decode is a heuristic that only recovers loose strings from the protobuf — structured tool calls/results (and thinking) only come from the agy-reader trajectory sidecar, which is the daemon's own decode. the steps in [Antigravity CLI: high-resolution transcripts](https://github.com/kenn-io/agentsview#antigravity-cli-high-resolution-transcripts) should light up tool calls in the UI:

```
go install github.com/mjacobs/agy-reader@latest
agy-reader --sync   # or --watch to keep them fresh
```

make sure you're on latest — antigravity 1.0.6 moved sessions from `.pb` to `.db` and older agy-reader builds don't discover them (fixed yesterday). the file watcher picks up sidecars automatically, no agentsview restart needed.

for reasoning tokens: the sidecar has them labeled under `generatorMetadata[].chatModel.usage` — `inputTokens`, `outputTokens`, `thinkingOutputTokens`, `responseOutputTokens`, `cacheReadTokens` — and `outputTokens` already includes thinking (in my data it's exactly `thinkingOutputTokens + responseOutputTokens`). that's what I meant upthread about the sidecar being able to replace the varint heuristic.

wesm:
I'll try to get this merged this morning and get a release out shortly

wesm:
I'm going to cut a release here shortly, if my roborev-driven fixes broke anything please send a follow up PR!

PR #620, opened 2026-06-10
Prefer agy-reader trajectory sidecars for Antigravity CLI sessions

mjacobs:
Antigravity CLI `.db` sessions were parsed exclusively by the heuristic SQLite decode: a schema-less protobuf wire walk that recovers loose strings but no structured data. The `<uuid>.trajectory.json` sidecar (written by agy-reader from the daemon's own LoadTrajectory decode, and already the preferred source for legacy `.pb` sessions) was never consulted on the `.db` path -- so for all new-format sessions, tool calls, tool results, and thinking text were absent even when the high-fidelity decode sat next to the database on disk. On a real settled session, this change takes the persisted transcript from 0 tool calls / 0 tool results / 0 thinking messages to 33 / 32 / 19.

The `.db` branch of `ParseAntigravityCLISessionWithStatus` now parses both sources and selects by content rather than file mtimes: the sidecar wins when it parses, has displayable messages, and its step count covers the raw DB step count; otherwise the heuristic decode (with its existing history.jsonl prompt merge) is used as before. Step counts are comparable because both sources enumerate the same cortex steps -- they converge exactly once a session settles, while a live session's sidecar briefly lags the database (agy-reader syncs on an interval), in which case the fuller DB decode correctly wins until the next re-parse. A sidecar that covers all raw rows also rescues sessions whose DB decode yields nothing displayable, without setting NeedsRetry; a *partial* sidecar over an undecodable DB is stored as the best available transcript but leaves the row retryable, so an incomplete transcript is never persisted as current (hardened per roborev's review finding). The `.db` change fingerprint additionally folds in the sidecar's size/mtime so an agy-reader sync triggers a re-parse even when the database files themselves are untouched.

The `.pb` branch gets the same treatment, where the content rule degenerates to "use any sidecar that parses with displayable messages": the `.pb` itself is AES-encrypted, so there is no competing decode to compare against, and every fallback (history rows, brain docs, decrypt preview) is strictly lower fidelity than even a stale sidecar. The old mtime gate rejected a `trajectory.json` older than its `.pb` in favor of those fallbacks; since `.pb` files are no longer produced by current Antigravity releases, their sidecars are final and the gate only ever discarded good data. The test asserting the gate's behavior is replaced by an older-sidecar-wins test; the undisplayable-sidecar fallback path keeps its existing coverage.

`dataVersion` moves to 36 across two bumps (renumbered after merging main, where 34 took the session_name column): 35 covers the `.db` sidecar preference together with the recently merged #579 project inference (#615), and 36 covers the `.pb` gate removal. All three alter persisted session data, but existing Antigravity CLI rows are skipped while file size/mtime and data_version look current, so old databases need the non-destructive resync to pick up inferred projects and sidecar-fidelity transcripts.

Tradeoffs and limitations: machines without agy-reader see no behavior change (no sidecar means the heuristic/fallback paths are taken, same as today). The step-count rule prefers the DB decode for the duration of a live session even though the slightly-stale sidecar is structurally richer; this is deliberate -- completeness is easier to reason about than a freshness heuristic, and the window closes on its own.

Where to look: the source-selection block in `internal/parser/antigravity_cli.go` (`ParseAntigravityCLISessionWithStatus`, both branches) is the heart of the change; `parseAntigravityCLITrajectory` now also returns its step count, and `AntigravityCLIFileInfo` gained the sidecar companion for `.db`. The dataVersion changelog comments in `internal/db/db.go` document the resync rationale.


mjacobs:
Ah, looking into the roborev finding now

PR #622, opened 2026-06-11
Allow publishing sessions as secret gists

### LINKED ISSUE #611: Allow to create a secret gist instead of public one

mariusvniekerk:
Closes #611.

Publishing a session always created a public gist: the create-gist payload hardcoded `public: true` and the UI offered no choice, so there was no way to share a session transcript privately even though GitHub supports secret gists with the same token scope.

The publish endpoint (`POST /api/v1/sessions/{id}/publish`) now accepts an optional `secret` query flag that flips the gist payload's `public` field. The header's publish button becomes a two-entry dropdown (public / secret), mirroring the adjacent export menu pattern, and the narrow-viewport overflow menu gets the same pair of entries. The chosen visibility is carried through the ui store into the publish modal, which reflects it in its title and progress text.

The default everywhere stays public: omitting the query param preserves the old behavior, and the `p` keyboard shortcut still publishes publicly, so existing flows and any external callers of the endpoint are unaffected. One tradeoff: the previously one-click publish button now takes two clicks, since the visibility choice has to live somewhere.

A note on "secret" semantics worth keeping in mind as a reviewer: secret gists are unlisted, not access-controlled — anyone with the URL (including the htmlpreview view URL) can read them. This matches GitHub's own framing, but the gist content is a full session transcript.

Where to look: the visibility flag threading in `internal/server/export.go` and `huma_routes_sessions.go`; the new publish dropdown in `AppHeader.svelte`. `SessionsService.ts` is regenerated output from the OpenAPI spec.

Validation: `go fmt`, `go vet`, and the `internal/server` suite pass, including a new `TestCreateGistVisibility` asserting the payload's `public` field for both modes; `svelte-check` is clean and the touched frontend tests pass. The frontend vitest suite has 74 pre-existing failures (`localStorage` undefined under Node 26) that reproduce identically on the untouched branch tip.


mariusvniekerk:
**Rejecting the roborev finding "Secret gist view URL leaks through HTMLPreview" — won't fix, by design.**

- The `view_url` is returned only to the user who just published the gist. They already have the session, the GitHub token, and the gist itself; nothing is disclosed to anyone at response time.
- A secret gist is unlisted, not confidential — per GitHub's own semantics, anyone with the link can read it. Sharing the link *is* the sharing model, so anyone opening the preview URL is someone the publisher chose to share with.
- The proposed remediation is security theater: `raw_url` is in the same response, and the htmlpreview URL is trivially constructed from it. Omitting `view_url` for secret gists would only break rendered-HTML viewing while protecting nothing.
- The htmlpreview routing is pre-existing behavior for public gists on `main`; this PR only adds the `secret` flag.

PR #634, opened 2026-06-11
feat(frontend): add Shiki syntax highlighting for code fences

### LINKED ISSUE #461: feat: syntax highlighting for code fences in session transcript
## Summary

Add syntax highlighting for fenced code blocks in the session transcript / message renderer so languages are easier to read at a glance.

## Proposal

Use **[Shiki](https://shiki.style/)** for highlighting:

- Accurate, TextMate-grammar–based highlighting consistent with VS Code / common editor themes
- Works well in Vite + Svelte stacks; can align theme with the app (light/dark) if desired
- Apply to standard Markdown fenced blocks with a language tag (e.g. ` ```ts `)

## Scope (suggested)

- Client-side (or build-time) highlighting of user/assistant code fences in the main transcript view
- Graceful fallback for unknown languages or missing grammar (plain monospace, no broken layout)
- Performance: avoid blocking the main thread on very large blocks (virtualization or async highlight if needed)

## Out of scope / follow-ups

- Optional: copy-as-formatted, line numbers, or theme switcher can be separate issues

## Acceptance criteria

- Fenced code in messages shows language-appropriate colors when a language is specified
- No regression for non-code content or unlabeled fences

josix:
Closes #461. 

## Summary
Fenced code blocks in session transcripts now render with language-appropriate
syntax colors. Highlighting is powered by Shiki (`shiki/core` with the
JavaScript regex engine — no WASM/Oniguruma), loaded entirely through lazy
dynamic `import()` so Shiki, the theme, and each grammar land in code-split
chunks. The entry bundle, and therefore the Go-embedded binary, stay
essentially the same size.

### What changed

- `frontend/src/lib/utils/syntax-highlight.ts` — lazy memoized highlighter
  singleton. Some common grammars are preloaded plus aliases (`js`, `ts`, `py`, etc). 
  Each grammar is a literal `import()` so Vite can statically split it into its own
  chunk. `highlightToHtml()` returns `null` for unknown languages,
  over-threshold content (>50,000 bytes or >800 lines), or any error, the
  caller then keeps the plain escaped render.
- `CodeBlock.svelte`: the first paint is the same escaped
  plain text as before; when the highlighter resolves, the `<code>` contents
  swap to Shiki token spans. The swap is color-only, so row heights in the virtualized
  message list are stable.
- `frontend/src/lib/utils/highlight-fences.ts` — new Svelte action that also
  colorizes fences *inside* markdown-rendered content (thinking text, skill
  blocks, insights, expanded pins). It runs after DOMPurify sanitization and
  swaps each labeled `<code>`'s innerHTML with Shiki output, which sidesteps
  DOMPurify stripping Shiki's inline `style` attributes without loosening the
  sanitizer config; Shiki escapes the code text itself. Search `<mark>`s are
  re-applied after each swap since the innerHTML replacement wipes them.
- `highlight.ts` — `clearMarks`/`applyMarks` extracted from the
  `applyHighlight` action (no behavior change) so the fence action can reuse
  them.
- Theme: single dark `catppuccin-mocha` in both app modes, matching the
  existing always-dark `--code-bg` code surfaces.

### Where to look

- `syntax-highlight.ts`: fallback/threshold logic and the literal-import
  grammar table (a template-literal import would silently break the
  production build).
- `highlight-fences.ts`: the post-sanitization innerHTML swap and search-mark
  re-application — the main security and correctness surface.
- The `$effect` in `CodeBlock.svelte` and its stale-result cancellation.
- These are the first dynamic `import()`s in the frontend; the chunks are
  served from the Go binary's embedded `dist/`.

### Tradeoffs and limitations

- Languages outside the preloaded set (`diff`, `java`, `c`, ...) intentionally
  fall back to plain text because there is no on-demand grammar loading, keeping the
  shipped chunk set fixed and small. Easy to extend later by adding entries to
  the grammar table.
- Very large blocks (>50 KB or >800 lines) skip highlighting to avoid blocking
  the main thread.
- Code blocks stay dark in light app mode, matching current behavior; Shiki's
  dual-theme CSS-variable mode is a possible follow-up if light code surfaces
  are ever wanted.
- If the lazy highlighter chunk never resolves, rendering is identical to
  today's escaped plain text.

## Verification
The following criteria will be verified manually in a test build.

### Phase A: Automated Baseline Gates
- [x] `npm run check` — svelte-check, the only type gate. Expect 0 errors.
- [x] `npm test` — vitest. Expect green, and confirm the two new suites
   actually execute: `src/lib/utils/syntax-highlight.test.ts` and
   `src/lib/utils/highlight-fences.test.ts`.

### Phase B: Test Data Setup

0. Copy one real line from any `~/.claude/projects/.../*.jsonl` into
   `/tmp/av-fences/proj-test/461-verify.jsonl` (guarantees schema
   validity), then edit only its text content to include all fence
   variants:

   - A ` ```ts ` block (supported language).
   - ` ```py ` and ` ```yml ` blocks (alias resolution).
   - An unlabeled ` ``` ` block (must stay plain).
   - ` ```diff ` and ` ```java ` blocks (not preloaded; must stay
     plain, by design).
   - A ` ```js ` block with 801+ lines (the threshold is strict `>`,
     so exactly 800 lines still highlights).
   - A ` ```ts ` fence inside a thinking block (Path B proof).
   - A prose fence containing a searchable token (for check D8).

- [x] Labeled fences show language-appropriate colors 
#### Pin Page
image

#### Insight Page
image

#### Session Page
image

---

- [x] Graceful fallback for unknown languages (plain, layout intact)

image

---

- [x] Theme aligns with app light/dark

image
---

- [x] Copy unaffected (verified in my local machine)
- [x] Search highlighting survives the highlight swap

image


mariusvniekerk:
@josix Thanks for this one!

mariusvniekerk (approved):

Issue #591, opened 2026-06-04
Antigravity CLI: expose token counts and model name from trajectory.json generatorMetadata


nnachname90:
## Summary

Antigravity CLI sessions appear in the session list but are invisible in the **Usage** view because two conditions in `usageMessageEligibility` are never met:

```go
const usageMessageEligibility = `
    m.token_usage != ''
    AND m.model != ''
    ...`
```

Neither `ParsedMessage.Model` nor `ParsedMessage.TokenUsage` are set for antigravity-cli sessions, even when a full `.trajectory.json` sidecar (produced by [agy-reader](https://github.com/mjacobs/agy-reader)) is present.

## Token data is available in the sidecar

The sidecar contains per-call token counts in `generatorMetadata[].chatModel.usage`:

```json
{
  "generatorMetadata": [
    {
      "chatModel": {
        "usage": {
          "model": "MODEL_PLACEHOLDER_M16",
          "inputTokens": "22200",
          "outputTokens": "983",
          "thinkingOutputTokens": "908",
          "responseOutputTokens": "75",
          "cacheReadTokens": "20316"
        }
      }
    }
  ]
}
```

## Proposed fix

1. **Extend `agyTrajectory`** to include `GeneratorMetadata` with the usage fields.
2. **Sum tokens** across all `generatorMetadata` entries during sidecar parsing.
3. **Set `ParsedMessage.Model`** to `"antigravity-cli"` (or parse the obfuscated model identifier) on assistant messages.
4. **Set `ParsedMessage.TokenUsage`** to the aggregated JSON with `input_tokens`, `output_tokens`, `cache_read_input_tokens`.

This would make antigravity-cli sessions fully visible in the Usage view without any pricing side-effects (the model name would not match any pricing entry, so cost would correctly show as $0 for subscription users).

## Workaround

Until this is fixed natively, a patch script can copy the DB out of the container, inject the values directly into `messages.model` / `messages.token_usage`, and copy it back — but this is fragile and gets overwritten on `--full` resyncs.

mjacobs:
good analysis — the field archaeology here matches what I'm seeing in my own sidecars. status update: #619 (in flight) covers most of this for newer sessions, via a different source. a couple of notes:

[...]

PR #647, opened 2026-06-11
Antigravity CLI: extract sidecar generatorMetadata token usage

mjacobs:
Follow-up to #591 and the remaining gap after #619: legacy `.pb` sessions have no SQLite `gen_metadata`, so they get no usage data, and sidecar-rendered transcripts carry no per-message model/token attribution.

## What this does

- `parseAntigravityCLITrajectory` now decodes the agy-reader sidecar's `generatorMetadata[]` and emits usage events (`source: "sidecar"`) from `chatModel.usage`: `inputTokens`, `outputTokens` (already includes thinking; not re-folded), `thinkingOutputTokens` as reasoning, and `cacheReadTokens`. Token counts are string-encoded in the sidecar; a tolerant `agyTokenCount` type decodes strings or numbers and treats garbage, null, and negative values as 0 so a malformed sidecar can never fail the transcript or corrupt totals.
- Precedence rule: `gen_metadata`-derived events always win; sidecar events fill the gap only when the DB path produced none (legacy `.pb`, unreadable `.db`, or `.db` without a `gen_metadata` table). A session never mixes the two sources, so nothing double-counts.
- Per-message attribution: each generation maps to the planner-response message in its `stepIndices` range, setting `Model`, `ContextTokens` (input + cache read), and `OutputTokens`. `TokenUsage` raw JSON is deliberately left empty — usage analytics count message rows with non-empty `token_usage`, and events remain the single analytics source.
- Model names from the sidecar are obfuscated enums (`MODEL_PLACEHOLDER_M16`/`M20`/`M132`); they are stored verbatim as a token-count-only, unpriced source (the tradeoff accepted in #591). Real model names still come only from `gen_metadata`.
- `applyUsageEventTokenTotals` peak-context now includes cache fields (no-op for existing callers, whose events never set them) so event totals agree with per-message context.
- dataVersion 38 so settled legacy sessions re-parse and gain usage on next sync.
- New internal/sync regression test covering #579's project-inference fallback through `SyncAll`/`SyncSingleSession` persistence (history row without `conversationId`), which previously had parser-only coverage.
- Comment-only docs marking the `.pb` AES decryptor (`antigravity_crypto.go`) as the explicit last-resort decode path.

## Limitations

- `chatModel.retryInfos[]` per-attempt usage is not summed, so retried generations may undercount (noted in a code comment).
- `outputTokens ⊇ thinkingOutputTokens` containment is an empirically verified invariant of current sidecars (26 sidecars / 1839 generations checked), recorded in a comment.
- Sidecar usage events for `.pb`-only sessions are unpriced ($0), matching what the #591 filer accepted.

## Where to look

`internal/parser/antigravity_cli.go` carries the schema types, extraction, and the precedence wiring in `ParseAntigravityCLISessionWithStatus` (both `.db` and `.pb` branches). `TestAntigravityCLIDBGenMetadataWinsOverSidecarUsage` pins the no-double-count rule; `TestAntigravityCLIPBSidecarEmitsUsageEvents` pins the field mapping.

Verified live against a real archive: 15 legacy `.pb`+sidecar sessions gained usage events with placeholder models after resync, `.db` sessions kept their `gen_metadata`-sourced events with no mixed-source sessions, and a freshly recorded session whose `gen_metadata` blob defeats the wire-walk heuristic was rescued by the sidecar gap-fill end to end.

## NUL-byte model names and `pg push` (added after filing)

Two production Antigravity IDE sessions carried a `messages.model` value that
was not a model name but a raw protobuf fragment (hex `080020022A0201024001`,
containing `0x00`): `extractModelName` accepted any field 21/19 payload that
passed `utf8.Valid`, but a nested protobuf message whose bytes are all < 0x80
is valid UTF-8. PostgreSQL rejects NUL in text (`SQLSTATE 22021`), so
`pg push` rolled back those sessions on every run. Fixed in two layers plus a
follow-up:

- `bulkInsertMessages` and `pushSession` now sanitize every parser-derived
  string (model, token_usage, role, claude ids, source fields, cwd,
  git_branch), not just content -- these values come from third-party session
  files, so the PG boundary is defensive regardless of any single parser fix.
- `extractModelName` rejects non-printable candidates
  (`isPlausibleModelName`: printable runes, at least one letter); the walk
  then continues, which in the observed sessions recovered the real
  `gemini-3.5-flash` name from a nested field. Covered by a fixture test
  using the exact production bytes. This shares the dataVersion 38 bump so
  corrupt persisted model values clear on resync.
- The metadata fast path compares local fingerprints against PG-readback
  fingerprints whose values were sanitized at insert; the sanitizer moved to
  `db.SanitizeUTF8` and the local builders apply it too, so a NUL-carrying
  local row no longer forces a full message rewrite on every push. A pgtest
  pins this via unchanged ctid across a forced second push.

Verified live: the two stuck sessions push cleanly and re-parse to the
correct model name.

### DIFF
--- a/internal/parser/antigravity_cli.go
+++ b/internal/parser/antigravity_cli.go
@@ -241,7 +241,13 @@ func ParseAntigravityCLISessionWithStatus(
 		case dbErr != nil || dbResult.rawStepCount > 0:
 			status.NeedsRetry = true
 		}
-		if len(usageEvents) == 0 && tErr == nil {
+		// Coverage gates usage just like the transcript: a lagging
+		// sidecar carries only the generations it has seen, and
+		// persisting those would underreport totals on a row that
+		// looks current. sidecarCovers stays true when the DB offers
+		// no coverage signal (unreadable or zero rows), so gap-fill
+		// still applies there.
+		if len(usageEvents) == 0 && tErr == nil && sidecarCovers {
 			usageEvents = tRes.usageEvents
 		}
 	} else {
--- a/internal/parser/antigravity_cli.go
+++ b/internal/parser/antigravity_cli.go
@@ -276,6 +276,9 @@ func ParseAntigravityCLISessionWithStatus(
 		)...,
 	)
 
+	// Last resort for legacy .pb archives: AES decrypt via
+	// ANTIGRAVITY_KEY; the sidecar and history fallbacks above
+	// take precedence.
 	if ext == ".pb" && !hasTrajectory && hasAntigravityKey() {
 		if extra, ok := decryptAntigravityCLITranscript(path); ok {
 			messages = append(messages, extra)
@@ -702,7 +705,9 @@ func readAntigravityArtifactMeta(
 // decryptAntigravityCLITranscript attempts to decrypt a .pb file
 // and returns a single assistant ParsedMessage holding the raw
 // strings extracted from the plaintext. Returns ok=false when
-// decryption fails or yields no usable text.
+// decryption fails or yields no usable text. Last-resort decode
+// path: sidecar and history fallbacks take precedence; see the
+// header of antigravity_crypto.go for the retention rationale.
 func decryptAntigravityCLITranscript(
 	path string,
 ) (ParsedMessage, bool) {
--- a/internal/parser/antigravity_cli.go
+++ b/internal/parser/antigravity_cli.go
@@ -256,7 +256,12 @@ func ParseAntigravityCLISessionWithStatus(
 			// Usage events flow whenever the sidecar parses, even when
 			// no message is displayable, matching the message-less
 			// usage stance of the .db branch. The displayable gate
-			// below applies to the transcript decision only.
+			// below applies to the transcript decision only. A
+			// metadata-only sidecar can therefore pair usage events
+			// with a history/decrypt transcript; events without a
+			// generation timestamp bucket at session start in daily
+			// usage (occurred_at stored as NULL, COALESCEd to
+			// started_at).
 			usageEvents = tRes.usageEvents
 			if hasDisplayableAntigravityCLITrajectoryMessage(tRes.messages) {
 				messages = tRes.messages
@@ -909,7 +914,9 @@ func (c *agyTokenCount) UnmarshalJSON(data []byte) error {
 		s = unquoted
 	}
 	n, err := strconv.ParseInt(strings.TrimSpace(s), 10, 64)
-	if err != nil {
+	if err != nil || n < 0 {
+		// Negative counts are garbage too: emitting them would
+		// subtract from session and daily usage totals.
 		return nil
 	}
 	*c = agyTokenCount(n)
@@ -1166,16 +1173,6 @@ func agyToolDetail(name, inputJSON string) string {
 	return name
 }
 
-// parseAntigravityCLITrajectory reads a <uuid>.trajectory.json sidecar
-// produced out-of-process by agy-reader and returns the decoded
-// transcript as ParsedMessages.
-//
-// Trust posture (see SECURITY.md, "Imports and new readers" row of the
-// Trust boundaries table): the sidecar is treated as untrusted
-// structured input — same posture as any other agent session file.
-// The read is size-capped (maxTrajectorySidecarBytes) and unknown step
-// types are silently skipped further down. No content from the sidecar
-// is executed or echoed back over any outbound channel.
 // agyTrajectoryParseResult is the decoded output of one trajectory
 // sidecar: the transcript messages, the raw step count (used by
 // callers to compare coverage against other decode sources for the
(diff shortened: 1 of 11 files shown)


mjacobs:
Good catch on the usage gap-fill — fixed in 93fbf35f. The `sidecarCovers` gate now applies to usage events exactly as it does to transcript selection, so a lagging sidecar's partial generatorMetadata is rejected rather than persisted as current totals. Coverage remains satisfied when the DB offers no signal (unreadable or zero raw rows), so the legacy/rescue paths are unchanged, and a covering sidecar rewrite re-triggers parsing via the file fingerprint and fills usage then. Added `TestAntigravityCLINonCoveringSidecarUsageRejected`, verified to fail against the ungated code.

internal/parser/antigravity_cli.go
```diff
@@ -831,9 +860,73 @@ func (f fakeFileInfo) Sys() any    { return nil }
...
+	CacheReadTokens      agyTokenCount `json:"cacheReadTokens"`
+}
+
+// agyTokenCount decodes a token count the daemon serializes as a JSON
+// string ("19733"), tolerating bare numbers as well. Garbage, null, or
+// empty values decode to 0 WITHOUT returning an error: the sidecar is
+// untrusted structured input (see the trust-posture note above), and
+// an unmarshal error here would fail the whole trajectory decode and
+// destroy the transcript.
+type agyTokenCount int
```
mariusvniekerk, line 911: this looks a lot like backing out pbjson's int64 parser ?

internal/parser/antigravity_test.go
```diff
@@ -1500,6 +1521,386 @@ func TestAntigravityCLISidecarRescueKeepsGenMetadataUsage(t *testing.T) {
...
 
+func TestAntigravityCLIPBSidecarEmitsUsageEvents(t *testing.T) {
+	root := t.TempDir()
+	id := "ffffffff-0000-1111-2222-333333333333"
+	mustMkdir(t, filepath.Join(root, "conversations"))
+
+	pbPath := filepath.Join(root, "conversations", id+".pb")
+	mustWrite(t, pbPath, []byte("pb-junk-bytes"))
+
+	// Gen A claims the planner step (1) and its tool step (2). Gen B
```
mariusvniekerk, line 1532: Given that we know the json is created from pbjson should we make these using pbstruct etc ?

internal/parser/antigravity_test.go
```diff
@@ -1937,6 +2338,73 @@ func TestExtractTokenUsageFalsePositiveGuards(t *testing.T) {
...
 }
 
+// TestExtractModelNameRejectsNonPrintable verifies that field 21/19
+// payloads that are valid UTF-8 but not human-readable text (nested
+// protobuf messages whose low bytes pass utf8.Valid) are rejected
+// instead of being persisted as the model name. The fragment bytes are
+// the exact gen_metadata field-21 payload observed in production
+// sessions, which leaked into messages.model with an embedded NUL byte
+// and broke `pg push` (PG rejects NUL with SQLSTATE 22021).
+func TestExtractModelNameRejectsNonPrintable(t *testing.T) {
```
mariusvniekerk, line 2348: This why test suggests that this is stored using a pb.Any ?

internal/parser/antigravity_cli.go
mjacobs: Yes, exactly that — these are protojson int64s. The sidecar JSON comes from the daemon's Connect-RPC service (`exa.language_server_pb.LanguageServerService`), so 64-bit counts arrive as strings per the proto3 JSON mapping. I kept the decode hand-rolled rather than using protojson for two reasons: decoding with real protojson needs generated types from a `.proto` we don't have (the daemon's schema is closed and unpublished — the field set here is known only empirically), and protojson is strict where this path deliberately wants tolerance: a malformed count must decode to 0 rather than fail the whole trajectory decode and lose the transcript. It would also pull in `google.golang.org/protobuf` as a new dependency; the repo currently has no protobuf deps.

internal/parser/antigravity_test.go
mjacobs: Same constraint as the int64 thread: to marshal fixtures with protojson we'd have to author our own `.proto` approximating Google's closed schema, which would make the fixtures look authoritative when they're really still an empirical guess — and would add the protobuf module for tests only. The hand-written JSON here mirrors captured daemon output shape-for-shape (string-encoded counts included), which I think is the more honest fixture for a format we don't own. Happy to revisit if the schema ever gets published.

internal/parser/antigravity_test.go
mjacobs: I checked — it's not an `Any`. The production fragment decodes as `field 1: varint 0, field 4: varint 2, field 5: 2-byte LEN, field 8: varint 1`, whereas a `google.protobuf.Any` always starts with field 1 as a length-delimited type_url (`type.googleapis.com/...`). So it's just an ordinary nested message that happened to land in the field-21/19 position the walk treats as a model-name candidate. The underlying issue is that without the schema the wire format can't distinguish a string from a sub-message (both wire type 2), and a nested message whose bytes are all < 0x80 passes `utf8.Valid` — hence the printable-text guard.

wesm:
merging this, I'll try to get a release out this morning

Issue #412, opened 2026-04-28
Feature: periodic SSH remote sync from `serve` (config-driven `[[remotes]]` list)


rube-de:
## Background

`agentsview sync --host <h>` (in `cmd/agentsview/sync.go`, dispatching to `internal/ssh/sync.go:RemoteSync`) is a clean one-shot pull: it resolves agent dirs on the remote, streams them back via `tar` over SSH, and feeds the temp tree into `sync.Engine` with a path rewriter that prefixes sessions with `host:/remote/path`. The incremental cache in `remote_skipped_files` keeps re-pulls cheap.

`serve` already runs a periodic local sync via `runPeriodicSync` on a 15-minute `time.Ticker` (see `periodicSyncInterval` in `cmd/agentsview/serve_runtime.go`), which re-scans local agent dirs and recomputes pending sessions.

The two never meet. To keep remote sessions current, users have to wire up cron / launchd / systemd outside the binary, and remote sessions never appear in the live SSE stream until the next manual `sync --host` run.

## Proposal

Let `serve` poll a configured list of remote hosts on the same periodic cycle that already drives local sync.

```toml
# config.toml

[[remotes]]
host = "dev-vps"          # any value SSH would accept; uses ~/.ssh/config
user = "me"               # optional; falls back to ssh's default resolution
port = 22                 # optional
interval = "1h"           # optional; defaults to a sane value (e.g. 15m or 1h)
ssh_opts = ["-i", "~/.ssh/agentsview_ed25519"]  # optional; passthrough flags

[[remotes]]
host = "build-box.tailnet.ts.net"
interval = "30m"
```

On startup, `serve` would:

1. Validate each entry (host non-empty, interval parseable).
2. Spawn a goroutine per remote with its own ticker.
3. On each tick, run the equivalent of `RemoteSync.Run(ctx)` against that host, reusing the existing skip-cache so unchanged files aren't re-fetched.
4. On success, fire the same broadcaster events local sync uses so the SSE stream picks up new/updated remote sessions in real time.
5. On failure, log and back off (don't crash `serve`); surface the last error + last-success-at via `/api/v1/health` or a small `/api/v1/remotes` status endpoint.

## Why this fits the existing shape

- Reuses `RemoteSync` end-to-end — no new transport.
- Reuses `periodicSyncInterval` plumbing pattern (one ticker per remote instead of one for local).
[...]

PR #650, opened 2026-06-12
feat(parser): read Claude/Codex sessions from S3-compatible object storage

DanielMao1:
## Summary

Lets `claude_project_dirs` / `codex_sessions_dirs` entries be `s3://` URIs.
When a configured root is an `s3://` URI, Claude and Codex sessions are listed
and fetched directly from an S3-compatible object store (AWS S3, MinIO, Aliyun
OSS, Cloudflare R2) and parsed into the local SQLite DB like any other source.
Pure Go via `minio-go` — no cgo, no change to the default build or
cross-compilation.

## Motivation

I'm a heavy Claude Code (and Codex) user and a big fan of this project — it's
become how I actually review what my agents did. But my sessions are scattered:
some on my Mac, some on a personal EC2 box, a lot more on my company's GPU
cluster. What I really want is **one** agentsview that gives me an integrated
view and analytics across all of them, not a separate dashboard per machine.

`sync --host` and PG push already cover part of this, but both assume the central
instance can reach each machine: SSH needs inbound access, and PG push needs a
daemon + Postgres on every box. The cluster nodes I can't SSH into from home, and
ephemeral cloud instances are often gone by the time I'd want to pull from them.

So this goes the other direction — **push, not pull**. Each machine drops its own
`~/.claude/projects` / `~/.codex/sessions` into its own prefix of a shared bucket
(a plain `aws s3 sync` / `rclone` cron), and one agentsview reads them all from
object storage. Nothing has to reach back into the source machines; they just
push before they disappear. It sits alongside `sync --host` and PG push rather
than replacing either.

## How it works

The two touch points are **discovery** and **reading bytes**; the per-agent
parsers are untouched.

- `DiscoverClaudeProjects` / `DiscoverCodexSessions` detect an `s3://` root and
  list objects, reconstructing the same project / subagent layout the local
  walkers produce.
- `processS3Session` downloads each object, buffers it to a transient temp file
  so the existing path-based parsers (incremental offsets, subagent paths) run
  unchanged, then deletes it — no persistent local mirror. The parsed session
  records the original `s3://` URI as its source path.
- Source machine is derived from a `.../<machine>/raw/...` layout, mirroring the
  `host:` prefix that SSH sync already attaches to pulled sessions.

Credentials come from standard AWS env vars (`AWS_ACCESS_KEY_ID`,
`AWS_SECRET_ACCESS_KEY`, `AWS_REGION`, `AWS_S3_ENDPOINT`). `AWS_S3_ENDPOINT`
selects the S3-compatible endpoint (empty = AWS).

## Where to look

- `internal/parser/s3source.go` — the S3 client, listing, fetch, and discovery (new)
- `internal/parser/discovery.go` — `s3://` dispatch + `DiscoveredFile.Machine`
- `internal/sync/engine.go` — `processS3Session` and the `processFile` short-circuit

## Tradeoffs / limitations

- Each object is read whole into memory (session JSONL is MB-scale, fine).
- Every sync re-fetches all objects; no ETag/LastModified incremental yet — a
  natural follow-up using the metadata `ListObjects` already returns.
- `session export <id>` is local-only and doesn't fetch from S3 yet.
- Adds one dependency, `github.com/minio/minio-go/v7`.

## Open question

Happy to adjust the shape. If you'd rather this be a dedicated `[[remotes]]`-style
config block (in the spirit of #412) instead of overloading the existing `*_dirs`
keys with `s3://`, I can rework it that way — wanted to keep the diff minimal
first and get your read on whether an object-store source is something you'd want
in tree at all.


DanielMao1:
The issues reviewed are addressed in 598e6bd.

- **High (path traversal):** Fixed. `processS3Session` now derives the temp filename from `path.Base` and rejects any name containing a path separator or `.`/`..`, so an object key can never escape the temp dir (including a key embedding a `\` on Windows).
- **`FetchS3Object` buffers the whole object in memory:** Fixed. It now returns the `GetObject` reader and `processS3Session` `io.Copy`s it straight to the temp file — no intermediate `[]byte`.
- **`FindSourceFile` can't resolve `s3://` paths:** Fixed. Stored `s3://` paths are returned as-is (there's no local file to stat), so single-session resync routes them back through `processS3Session`.
- **No ETag/LastModified incremental — full re-parse every sync:** Left as a follow-up for now (it's called out in the PR limitations). `ListObjects` already returns ETag + LastModified, so the plan is to carry that through discovery and skip unchanged objects against the stored hash/mtime. Happy to fold it into this PR if you'd rather it land together than as a follow-up — just say the word.

wesm:
rebasing will take a closer look

PR #651, opened 2026-06-12
parser: fuzz the Antigravity wire-walk with output invariants

### LINKED ISSUE #648: parser: fuzz the Antigravity wire-walk heuristics with output invariants
The Antigravity parsers decode an undocumented, unversioned protobuf format by wire-walking (`agProtoParse` and the heuristics on top of it: `decodeAntigravityStep`, `extractModelName`, `extractTokenUsage`). The format has already changed across Antigravity dot releases, and fixtures only cover payload shapes we have happened to observe.

The recent NUL-byte incident (fixed in #647) is an instance of a whole class: `extractModelName` accepted any field 21/19 payload that passed `utf8.Valid`, but a nested protobuf message whose bytes are all below 0x80 is valid UTF-8 (ASCII control characters, including NUL). The raw fragment was persisted as `messages.model` and later broke `pg push` with SQLSTATE 22021. Example-based tests could not have caught this before the bytes showed up in production; an invariant would have.

## Proposal

Add Go native fuzz tests (`go test -fuzz`) over the wire-walk entry points, asserting output invariants rather than expected values:

- `agProtoParse`: never panics, never returns a field whose bytes extend past the input.
- `decodeAntigravityStep`: never panics; any returned content is valid UTF-8 with no NUL bytes; timestamps are zero or within the plausibility window already enforced.
- `extractModelName`: result is empty or printable (every rune passes `unicode.IsPrint`, at least one letter) — i.e. `isPlausibleModelName` holds by construction.
- `extractTokenUsage`: counts are non-negative and within `maxPlausibleTokens`.

Seed the corpus from the existing `encodePB` test fixtures (including the real 10-byte fragment `080020022A0201024001` now in `antigravity_test.go`). Found-failure inputs get committed under `testdata/fuzz/` as regression cases. CI can run the fuzz targets in seed-only mode (`go test -run Fuzz`) so the corpus is enforced on every PR without long fuzz time; longer fuzzing can run on a schedule.

This works with the reverse-engineering reality instead of against it: we cannot enumerate the corner cases of a format we do not control, but we can pin the properties every decode must satisfy.

mjacobs:
Closes #648.

Adds Go native fuzz targets over the Antigravity wire-walk and hardens the
parser so the asserted invariants hold by construction. The Antigravity format
is reverse-engineered, unversioned, and has changed across dot releases, so
these targets pin output *properties* instead of expected values: corner cases
we cannot enumerate still have enforced contracts.

## Fuzz targets (`internal/parser/antigravity_fuzz_test.go`)

- `FuzzAgProtoParse`: structurally sound field tree (positive field numbers,
  4/8-byte fixed payloads, payloads bounded by the enclosing buffer, nesting
  bounded by the recursion cap, total fields within the allocation budget),
  deterministic.
- `FuzzAgProtoLooksLikePrefix`: panic safety on the decryption-candidate sniff.
- `FuzzDecodeAntigravityStep`: content non-empty, valid UTF-8, NUL-free;
  `ContentLength` consistent; role in enum; timestamp zero or inside the
  2000–2100 plausibility window (boundary seeds are step-shaped so the window
  assertions actually execute; mutation-testing the bounds fails the seeds).
- `FuzzExtractModelName`: empty or printable with at least one letter.
- `FuzzExtractTokenUsage`: counts non-negative and within
  `maxPlausibleTokens`, including the output+reasoning sum the caller
  persists as billable output.

Seeds cover realistic step/gen_metadata shapes, the raw fragment from the
NUL-byte incident (#647), decoy token blocks, window-boundary timestamps, and
malformed wire edges. Plain `go test` executes the seeds, so existing CI
enforces the corpus on every PR with no changes; a new scheduled workflow
(`.github/workflows/fuzz.yml`) runs each target 10 minutes weekly (dispatch
input passed via env var, not interpolated into the script) and uploads
failing inputs as artifacts.

## Parser hardening

Each guard exists because the corresponding invariant did not hold on main,
and each degrades gracefully rather than dropping content:

- `extractModelName` requires printable candidates with at least one letter
  (`isPlausibleModelName`). This hunk is textually identical to the same fix
  on the #647 branch, so whichever PR lands first, the other rebases cleanly.
- `agProtoCollectStrings` replaces NUL with U+FFFD instead of dropping the
  string: NUL-delimited tool output (`git ls-files -z`) is legitimate
  valid-UTF-8 content, but a leaked NUL breaks `pg push` (SQLSTATE 22021).
- `agProtoTimestamp` bounds nanos to [0, 1e9): larger varints cast to int32
  could go negative and shift `time.Unix` results outside the window callers
  check on seconds alone.
- `agProtoParse` threads a total-fields budget (1<<20) across the parse tree,
  including failed speculative nested re-parses. Each two-byte wire field
  decodes into a ~100-byte struct, so a flat run of minimal fields amplifies
  memory ~280x (a measured 8 MiB blob allocated over 1 GiB); step payloads,
  gen_metadata, and decrypted transcripts are unbounded blobs and the depth
  cap cannot bound breadth. Exhaustion truncates (top level keeps the decoded
  prefix, an exhausted speculative child stays opaque) so oversized payloads
  degrade to partial content instead of losing everything.
- `tokenBlockFrom` caps the output+reasoning sum: the caller folds reasoning
  into billable output, so individual caps alone let a decoy block persist
  twice `maxPlausibleTokens` into message tokens, session totals, and usage
  events.

`dataVersion` bumps to 38 so existing Antigravity rows re-parse. This is the
same number #647 uses, deliberately: the bumps touch the same line with
different comment text, so whichever PR lands second gets a visible rebase
conflict and bumps again, rather than a silently merged duplicate version.

## Limitations

- The content invariant is NUL-only by design; other control characters
  (ESC, C1 range) and unbounded printable model names are real but belong to
  a future centralized validation layer, not per-incident guards.
- The budget makes acceptance of pathologically dense >2 MiB payloads
  order-dependent (what fits in the budget survives); real payloads sit an
  order of magnitude under the cap.
- ~90 minutes of cumulative local fuzzing per target (180M+ total execs)
  found no violations after the fixes; the scheduled workflow continues the
  search over time.

Reviewers should look first at the budget threading and truncation semantics
in `internal/parser/antigravity_proto.go` and the fold cap in
`tokenBlockFrom`.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,61 @@
+name: Fuzz
+
+# Long-running fuzz exploration of the Antigravity wire-walk
+# (GitHub #648). Every CI run already executes the committed seed
+# corpus via plain `go test`; this workflow spends real fuzz time
+# searching for new invariant violations on a schedule. A found
+# failure is written to testdata/fuzz/ by the fuzz engine and
+# uploaded as an artifact so it can be committed as a regression
+# case.
+
+on:
+  schedule:
+    - cron: "17 5 * * 1" # weekly, Monday 05:17 UTC
+  workflow_dispatch:
+    inputs:
+      fuzztime:
+        description: "Fuzz time per target (Go duration)"
+        required: false
+        default: "10m"
+
+jobs:
+  fuzz:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - FuzzAgProtoParse
+          - FuzzAgProtoLooksLikePrefix
+          - FuzzDecodeAntigravityStep
+          - FuzzExtractModelName
+          - FuzzExtractTokenUsage
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c  # v6.4.0
+        with:
+          go-version-file: go.mod
+
+      # The dispatch input goes through an env var, never directly
+      # into the run script, so it cannot inject shell commands.
+      - name: Fuzz ${{ matrix.target }}
+        run: >
+          go test -tags "fts5,kit_posthog_disabled"
+          -fuzz "^${TARGET}$"
+          -fuzztime "$FUZZTIME"
+          ./internal/parser/
+        env:
+          CGO_ENABLED: "1"
+          TARGET: ${{ matrix.target }}
+          FUZZTIME: ${{ github.event.inputs.fuzztime || '10m' }}
+
+      - name: Upload failing inputs
+        if: failure()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+        with:
+          name: fuzz-failures-${{ matrix.target }}
+          path: internal/parser/testdata/fuzz/
+          if-no-files-found: ignore
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -28,7 +28,17 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
-// Bumped to 38: two Antigravity parsing changes. (a) The Antigravity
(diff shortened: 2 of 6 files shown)


mariusvniekerk:
Thanks for this tireless work debugging something that should have been documented.

wesm:
rebased

PR #656, opened 2026-06-12
fix(duckdb): gate driver linkage off on windows/arm64

wesm:
## Summary

duckdb-go-bindings ships prebuilt static DuckDB libraries only for darwin/{amd64,arm64}, linux/{amd64,arm64}, and windows/amd64. Since #609 the DuckDB driver links into every binary unconditionally, and since #598 the release workflow builds a native windows/arm64 binary — a combination that has never run together: linking `cmd/agentsview` for windows/arm64 fails with undefined `duckdb_*` symbols, which would fail the `build` matrix and block the entire next release (the `release` and `pypi` jobs need every leg, and `build_wheels.py --require-all` expects a windows_arm64 archive).

Driver registration now goes through an `openDuckDB` shim:

- `internal/duckdb/driver.go` (`!(windows && arm64)`) blank-imports `github.com/duckdb/duckdb-go/v2` and opens connections as before.
- `internal/duckdb/driver_windows_arm64.go` returns a clear unsupported-platform error from `Open` and `NewQuackStore` instead of linking the driver. The rest of the `internal/duckdb` package is plain `database/sql` and still compiles everywhere, so the SQLite-backed app is unaffected on windows/arm64 — only the `duckdb` subcommands report the platform limitation at runtime.

Verified by cross-linking windows/arm64 with llvm-mingw (fails on main with undefined `duckdb_*` symbols, links with this change) and by a full dry-run of the release workflow on a throwaway repo (https://github.com/wesm/agentsview-release-test/releases/tag/v0.99.0): all six build legs passed, including the manylinux glibc-2.28 guard and the macOS minos check; wheels built and smoke-tested for all platforms. Release archives grow from ~8MB to ~24–30MB compressed because the DuckDB static library now ships in every supported binary.

### DIFF
--- a/cmd/agentsview/duckdb_quack_duckdbtest_test.go
+++ b/cmd/agentsview/duckdb_quack_duckdbtest_test.go
@@ -1,4 +1,4 @@
-//go:build duckdbtest
+//go:build duckdbtest && !(windows && arm64)
 
 package main
 
--- a/internal/duckdb/connect.go
+++ b/internal/duckdb/connect.go
@@ -7,7 +7,6 @@ import (
 	neturl "net/url"
 	"strings"
 
-	_ "github.com/duckdb/duckdb-go/v2"
 	"go.kenn.io/agentsview/internal/config"
 )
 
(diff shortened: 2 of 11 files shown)


wesm:
Addressed in 238f15e1: the driver-dependent test files (smoke, store, schema, sync, contract) are now gated with `!(windows && arm64)`, smoke_test.go's direct driver import is removed, and the `duckdbtest`-tagged suites carry the same platform constraint. Verified by cross-linking the `internal/duckdb` and `cmd/agentsview` test binaries for windows/arm64 with llvm-mingw (failed to link before, links now); the full native suite still runs unchanged (466 tests).

PR #657, opened 2026-06-12
fix(codex): stop double counting history replayed into forked sessions

### LINKED ISSUE #643: codex forked session double counts usage
I have a codex session which I forked, it seems to me the usage is counted in both sessions causing the cost to be double what it really was.

nyxst4ck:
Fixes #643

## Bug

`codex fork` copies the parent's rollout lines — its **session_meta**, turns, messages and token_count events — into the top of the new file with re-stamped envelope timestamps. The parser counted that replayed history as the fork's own activity, so the shared prefix was billed in **both** the parent and the fork. Worse: the replayed parent `session_meta` overwrote the fork's session id (`handleSessionMeta` is last-wins), so the forked session was stored under the **parent's identity**.

Parsing a real forked rollout from my machine:

| | session id | messages | total output tokens |
|---|---|---|---|
| before | `codex:<parent-id>` (misattributed) | 658 | 160,687 |
| after | `codex:<fork-id>` | 65 | 11,366 |

## Fix

Envelope timestamps can't locate the replay boundary — the whole replay is re-stamped at fork creation within ~50ms. But **turn ids are UUIDv7** values minted when the turn originally ran: every replayed turn predates the fork instant, and the first genuine turn is minted at or after it. A `codexForkGate` arms on `session_meta.forked_from_id` and suppresses lines until the first `turn_context` whose turn_id timestamp reaches the fork's own creation time (taken from its UUIDv7 id, falling back to the meta timestamp). The replayed parent session_meta is skipped, so the fork keeps its own identity.

Robustness choices, **validated against 13 real forked rollouts spanning three CLI eras** (April replay format, June no-replay format, and a fork of a long-lived parent whose old turns predate turn ids):

- replayed `turn_context`s **without** a turn_id stay suppressed — a CLI new enough to write `forked_from_id` always stamps genuine turns (held in all 13 files, including one with 21 id-less replayed turns)
- an **unparseable** turn_id fails open to the pre-fix behaviour rather than risk dropping live data
- newer-CLI forks that replay nothing open the gate on their first (genuine) turn immediately
- non-forked sessions: gate never arms, zero behaviour change (multi-meta exec files keep last-wins)

The incremental-parse seed helpers (`seedCodexUserDedup`, `readCodexModelAtOffset`) apply the same gate so resumed parses mirror the full parser's view.

## Tests

`TestParseCodexSession_ForkedSessionSkipsReplayedHistory` (4 subtests): replay dropped + genuine kept + id preserved; fork with no genuine turns yields no messages; unparseable turn_id fails open; non-forked sessions untouched. **Red on main, green with the fix.** Two small `testjsonl` helpers added (`CodexForkedSessionMetaJSON`, `CodexTurnContextWithIDJSON`).

```
CGO_ENABLED=1 go test -tags "fts5,kit_posthog_disabled" ./internal/parser/ -count=1   # ok (full package)
go vet ./internal/parser/ ./internal/testjsonl/                                         # clean
```

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -34,7 +34,11 @@ import (
 // the fork's own session id instead of letting the replayed parent
 // session_meta overwrite it. Existing forked rows persist the
 // double-counted totals under the parent's identity, so they need
-// re-parsing to be rewritten with post-fork activity only.
+// re-parsing to be rewritten with post-fork activity only. Resync's
+// orphan copy also skips stale Codex rows whose file_path was
+// reparsed under a different session id, so the old parent-ID row
+// does not survive the rebuild when the parent's own file is gone
+// (see CopyOrphanedDataFromExcluding).
 //
 // (39: the Antigravity wire-walk hardened its output
 // invariants (issue #648): model-name candidates must be printable,
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -28,7 +28,15 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
-// Bumped to 39: the Antigravity wire-walk hardened its output
+// Bumped to 40: the Codex parser now suppresses the parent history
+// that `codex fork` replays at the top of a forked rollout, which was
+// double counted as the fork's own messages and token usage, and kept
+// the fork's own session id instead of letting the replayed parent
+// session_meta overwrite it. Existing forked rows persist the
+// double-counted totals under the parent's identity, so they need
+// re-parsing to be rewritten with post-fork activity only.
+//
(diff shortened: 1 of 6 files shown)


nyxst4ck:
Addressed both roborev findings in the latest commit:

- **dataVersion bumped 37 → 38** with the convention's explanatory note: existing forked rows persist double-counted totals under the parent's identity and need the full resync/reparse path.
- **`armFromMeta` now anchors from the JSONL envelope timestamp** as the final fallback (fork UUIDv7 id → payload timestamp → envelope timestamp), with a new regression subtest covering a non-v7 fork id whose payload has no timestamp.

`go test ./internal/parser/ ./internal/db/` and `go vet` clean locally.

wesm:
I'm working on this now

PR #658, opened 2026-06-12
fix: Antigravity session tools analysis + usage pricing fixes

vikram:
## Summary
- Improve Antigravity .db protobuf extraction for tools and tokens
- Log sidecar-rejected count when trajectory.json exists but is unused
- Apply custom_model_pricing in CLI usage commands (daily, statusline)
- Consolidate fallback pricing into litellm.go (remove fallback.go)

## Test Plan
- [ ] `go test ./...` passes
- [ ] Usage subcommands reflect custom_model_pricing from config.toml


wesm:
If you install the pre-commit hook (`prek install`) it will help you avoid lint failures

PR #662, opened 2026-06-12
cli: add parse-diff report mode to vet parser changes against a real archive

### LINKED ISSUE #649: cli: re-parse diff report mode to vet parser changes against a real archive
Parser changes are currently validated by example tests plus an ad hoc "verified live against a real archive" pass (see #620 / #647 PR descriptions). The local archive of real sessions is the best test data we have for reverse-engineered formats like Antigravity, but there is no tooling to use it systematically, so regressions in fields the examples do not cover surface only after a dataVersion resync rewrites the archive — as happened with the protobuf-fragment model names fixed in #647, which sat in the Usage view for days before anyone noticed.

## Proposal

A report-only CLI mode, for example:

```
agentsview parse-diff [--agent antigravity] [--limit N]
```

that re-parses session source files with the current binary, compares the result against what is stored in the local DB, and prints a per-session summary of changed fields without writing anything. Fields worth diffing: message count, user message count, first_message, session_name, per-message model, token totals, usage event count/totals, termination status.

Output would look like a change report: N sessions identical, M changed (grouped by field), with a per-session drill-down. Exit code or a `--fail-on-change` flag could make it usable in scripts.

## Why

- Turns every parser PR's "I hope this is right" into a reviewable artifact: run before bumping dataVersion, attach the summary to the PR.
- Catches regressions against the real-world payload distribution, not just the fixtures.
- Detects upstream format drift (e.g. an Antigravity dot release changing gen_metadata) the day it lands, by running against fresh sessions.

Most of the machinery already exists: discovery, parsers, and the resync path already re-parse everything; this is the same loop with a comparator and reporter instead of a writer. Care is needed to compare parsed output against stored rows through the same normalization the sync engine applies (filtering, pairing, ordinals) so the diff reflects real persisted changes rather than pipeline artifacts.

mjacobs:
Closes #649.

Adds `agentsview parse-diff`, a report-only CLI mode that re-parses session
source files with the current binary, runs the result through the same
normalization the sync engine applies, and compares it against the stored
SQLite rows. It writes no session, skip-cache, or sync-state data; per the
issue it turns each parser PR's "I hope this is right" into a reviewable
artifact and surfaces upstream format drift the day it lands.

```
agentsview parse-diff [--agent X ...] [--limit N] [--fail-on-change] [--json] [-v]
```

The report classifies every session into one bucket — identical, changed,
pending_resync, new_on_disk, parse_error, excluded_by_parser,
transient_needs_retry, or skipped (with a reason) — so genuine parser drift
is separated from pipeline history. Human output is a summary plus a
field-affected histogram plus a changed-session drill-down; `--json` emits the
same model for scripting and PR attachment. `--fail-on-change` exits non-zero
on real changes or parse errors.

## How it works

The comparison runs through the engine's real write-path normalization
(`prepareSessionWrite` -> `toDBMessages`/`pairAndFilter`, `toDBSession`,
`postFilterCounts`, `toDBUsageEvents`), so an unchanged parser produces zero
diffs against rows it wrote itself. Because that pipeline is unexported, the
differ lives in `internal/sync` (`Engine.ParseDiff`, `NewDiffEngine`) and the
CLI is a thin renderer over a renderer-agnostic report model in
`parsediff_report.go`.

Re-parse reuses `processFile` and the production worker pool. A new unexported
`forceParse` flag (set only by `NewDiffEngine`, which also forces `Ephemeral`)
disables the three skip layers — the in-memory/persisted skip cache, the
stored size/mtime/data_version short-circuits, and the incremental JSONL path
— so every discovered file is fully re-parsed against the populated archive.
Normal sync never sets the flag, so its behavior is unchanged.

Comparison is two-tier for speed: a per-session token fingerprint and a
content-length fingerprint gate the cheap identical path; only on a mismatch
are message rows loaded and the difference attributed to a specific field.
Usage events compare as an order-insensitive multiset. Known false-diff
sources are classified rather than reported as drift: pending-resync rows
(data version behind the binary), incremental-append `termination_status`
clears for Claude/Codex, orphaned/source-missing sessions, remote and
import-only sessions, and user-trashed rows.

## Coverage and limitations

- Covers every file-based agent, including Claude DAG forks, and the shared
  SQLite stores (Kiro `data.sqlite3`, db-mode OpenCode `opencode.db`) that
  discovery does not emit. Database-backed and import-only agents (Warp,
  Forge, Piebald, claude-ai, ChatGPT) have no on-disk source and are rejected.
- Best run against a quiescent archive. Sessions still being written, and
  sessions last written through the incremental-append path, can show benign
  drift against a full re-parse; a freshly resynced archive is the cleanest
  baseline. `--help` and the report say so, and a run whose data version is
  ahead of the whole archive prints a warning that no drift can be detected.

## Where to look

`internal/sync/parsediff.go` (loop + classification), `parsediff_compare.go`
(field-by-field comparison and the fingerprint tiers), `parsediff_report.go`
(the shared model), the `forceParse` guards in `internal/sync/engine.go`, and
`cmd/agentsview/parse_diff.go` (command + renderer). The feature was vetted
against a real ~726-session local archive across the supported agents.

### DIFF
--- a/cmd/agentsview/cli.go
+++ b/cmd/agentsview/cli.go
@@ -68,6 +68,7 @@ func newRootCommand() *cobra.Command {
 	root.AddCommand(newDuckDBCommand())
 	root.AddCommand(newSessionCommand())
 	root.AddCommand(newStatsCommand())
+	root.AddCommand(newParseDiffCommand())
 	root.AddCommand(newClassifierCommand())
 	root.AddCommand(newSecretsCommand())
 	root.AddCommand(newVersionCommand())
--- a/cmd/agentsview/parse_diff.go
+++ b/cmd/agentsview/parse_diff.go
@@ -55,10 +55,16 @@ func newParseDiffCommand() *cobra.Command {
 		Short: "Re-parse session files and diff against the archive",
 		Long: "Re-parses session source files with the current binary, runs\n" +
 			"the result through the same normalization sync applies, and\n" +
-			"compares it against the stored rows. Writes nothing: no\n" +
-			"sessions, no skip cache, no sync state.\n\n" +
+			"compares it against the stored rows. It does not write\n" +
+			"session, skip-cache, or sync-state data; opening the archive\n" +
+			"still creates the database and runs schema migrations if\n" +
+			"needed.\n\n" +
 			"Use it to vet parser changes against the real archive before\n" +
-			"bumping the data version, or to detect upstream format drift.",
+			"bumping the data version, or to detect upstream format drift.\n" +
+			"Run it against a quiescent archive: sessions still being\n" +
+			"written, and sessions last written through the incremental\n" +
+			"append path, can show benign drift against a full re-parse. A\n" +
+			"freshly resynced archive gives the cleanest baseline.",
 		GroupID:      groupData,
 		SilenceUsage: true,
 		Args:         cobra.NoArgs,
@@ -132,6 +138,9 @@ func doParseDiff(cfg ParseDiffConfig) (failed bool) {
 	if err != nil {
 		fatal("parse-diff: %v", err)
 	}
+	// Stamp the archive identity so an attached JSON report is
+	// self-describing.
+	report.DBPath = appCfg.DBPath
 
 	if cfg.JSON {
 		writeJSON(cfg.stdout(), report)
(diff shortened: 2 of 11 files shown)


mjacobs:
Both findings addressed in `feat/reparse-diff`.

**Vacuous `--fail-on-change`:** the gate now also fails when `VacuousResync()` is true, with a stderr reason that reaches `--json` callers (which see no stdout warning). `HasFailures()` stays scoped to real drift / parse errors; vacuity is folded in at the CLI decision point (`parseDiffExitFailure`).

**Comparator coverage:** extended to the omitted parser-owned data — session metadata (`cwd`, `git_branch`, `parent_session_id`, `relationship_type`, `source_session_id`, `source_version`, `parser_malformed_lines`, `is_truncated`), per-message flags (`is_system`, `has_thinking`, `has_tool_use`, `thinking_text`, folded into `message_metadata`), and `tool_calls` rows (new `MessageFlagsFingerprint` / `ToolCallFingerprint` tier-1 signals plus tier-2 attribution; these are parse-diff-only and don't alter the message fingerprints shared with the PG-push fast path).

Fields the incremental-append path freezes are compared but marked **informational for Claude/Codex**, matching the existing `termination_status` rule, so they never produce a false `--fail-on-change` failure on a live archive. While extending this I found `session_name` had the same latent problem (frozen by `UpdateSessionIncremental`, yet mutable mid-session via Claude's `/rename`) and gave it the same treatment; `first_message` and `started_at` derive from the session head and stay strict.

Three values are excluded by documented design:
- `project` — rewritten from the mutable `worktree_project_mappings` table, so its parser-owned input `cwd` is compared instead.
- the `tool_call` result body and the `tool_result_events` rows — both redacted to empty by the blocked-category config (the events slice is cleared wholesale) and unbounded in size, so only the config-stable `result_content_length` (set before the redaction check) is compared.

Out of scope here, but worth flagging: the PG-push `tool_calls` fast path checks only count + `result_content_length` sum, with no per-field identity fingerprint, so a `tool_name`/`input_json` change with an unchanged count and sum could be skipped on push. Pre-existing; not touched by this PR.

In-memory fingerprint twins are pinned against their DB queries through the real write pipeline, the clean-archive acid test now exercises a session with thinking, a tool_use/tool_result pair, and a system message, and the new tier-1 fingerprints are proven to fire end-to-end.

PR #665, opened 2026-06-13
fix(parser): support new .kimi-code session layout and add .kimi_openclaw to OpenClaw defaults

liaocaoxuezhe:
Kimi Code now stores sessions under `.kimi-code/sessions/<workdir>_<hash>/session_<uuid>/agents/<agent>/wire.jsonl`. Update `DiscoverKimiSessions`, `FindKimiSourceFile`, and `ParseKimiSession` to handle this layout while keeping compatibility with the legacy `.kimi/sessions` structure.

Also include `.kimi_openclaw/agents` in `AgentOpenClaw` default directories so sessions from the Kimi-packaged OpenClaw distribution are synced.


mjacobs:
Nice work on this — the parser-layer support for the new `.kimi-code` layout reads cleanly, and I appreciate that you kept full legacy compatibility throughout rather than swapping the old shape out. A few things that stood out as genuinely well done:

- **`DiscoverKimiSessions` / `FindKimiSourceFile` / `kimiSessionIDFromPath`** form a consistent round-trip. Discovery tries the legacy `wire.jsonl` first and only falls through to `agents/<agent>/wire.jsonl`; `FindKimiSourceFile` branches on 2- vs 3-part raw IDs; and `kimiSessionIDFromPath` keys off the `agents` parent to pick the layout. The 3-part ID (`<workdir>:<agent>:<session>`) maps to the same components in both directions, so the ID↔path mapping stays coherent. Moving `ParseKimiSession` off the hardcoded 2-level derivation onto `kimiSessionIDFromPath` is the right call.
- The parser test coverage is thorough — discovery (new / non-main-agent / mixed), `FindKimiSourceFile` with negative cases, and the `kimiSessionIDFromPath` legacy/main/agent-0 cases. I applied the three files locally and ran the Kimi parser tests; all green.
- Adding `.kimi-code/sessions` to Kimi's `DefaultDirs` and `.kimi_openclaw/agents` to OpenClaw's is correctly wired with no extra engine work needed: both agents build `agentDirs` from `DefaultDirs`, and the OpenClaw classifier and `DiscoverOpenClawSessions` already iterate per-dir via `isUnder`. The two trees also can't collide in the classifier — distinct relative roots and distinct file shapes (`<agentId>/sessions/<id>.jsonl` vs `.kimi-code/.../wire.jsonl`).

One thing I'd flag before merge — and I think it's the well-known "discovery is wired but the rest of the engine isn't" half-wire rather than anything wrong with the code you wrote. For a file-based agent, a new path shape has to land in two more places besides discovery, and this PR doesn't touch `engine.go`:

**Issue 1 — live updates drop new-layout files (`classifyOnePath`, Kimi block).** The classifier gate is `len(parts) != 3 || parts[2] != "wire.jsonl"`. A new-layout relative path splits into 5 components `[<workdir>, <session>, agents, <agent>, wire.jsonl]`, so it fails the gate and the event is discarded. `classifyOnePath` is the sole consumer of file-watcher / `SyncPaths` events, so the symptom is the classic one: new-layout sessions appear after a full restart (because `SyncAll` re-runs `DiscoverKimiSessions`, which you *did* update) but never update live. I confirmed the gate empirically — a new-layout rel path yields `len==5`, gate returns false. Fix direction: extend the Kimi branch to also accept the 5-component shape and pull the project from the `<workdir>` component.

**Issue 2 — single-session re-sync mis-derives the project (`SyncSingleSessionContext`, `case parser.AgentKimi`).** The derivation is `file.Project = filepath.Base(filepath.Dir(filepath.Dir(path)))`, which assumes the legacy `<project>/<session>/wire.jsonl` shape; for a new-layout path it returns the literal string `"agents"`. `processKimi` passes `file.Project` straight into `ParseKimiSession`, which uses it as `displayProject`, so the session is written with `Project="agents"`. This is reachable from a "re-sync this session" action: `FindSourceFile` returns the stored 5-component `file_path` (and your updated `FindKimiSourceFile` also resolves the 3-part ID), and the derivation then overwrites Project with `"agents"`. I confirmed it the same way — new-layout path derives `"agents"`, legacy correctly derives the project hash. Fix direction: handle the 5-component path here too, using the `<workdir>` directory as the project.

**Tests for the two engine sites.** All the new tests are parser-level, which is the layer that already works. A `SyncPaths` test asserting a changed new-layout `wire.jsonl` gets classified, and a `SyncSingleSession` test asserting the project is the workdir (not `"agents"`), would pin down both of the above and guard against regressions.

A couple of smaller things, more questions than asks:

- **Project naming (question).** Both `DiscoverKimiSessions` (`Project: projEntry.Name()`) and the correct engine derivation would use the raw `<workdir>_<hash>` directory name verbatim. Is that the intended display value, or should there be a decode step to a human-readable workdir (the way Cursor has `DecodeCursorProjectDir`)? Not blocking — just want to confirm the raw dir name is what you want users to see.
- **Real-archive confirmation (question).** Have you been able to round-trip this against an actual `.kimi-code` install? The fixtures look plausible, but one run of `agentsview parse-diff --agent kimi --fail-on-change` against a freshly-resynced archive with real new-layout sessions would lock down that the 5-component shape and the `agents` parent marker are exactly right. (Heads up that `parse-diff` would catch the Issue 2 project drift but not the Issue 1 live-drop — that one needs the `SyncPaths`-level test.)

Last thing: would you like to land the engine wiring (Issues 1 and 2 plus the two tests) in this PR, or ship parser support first and follow up? The parser work stands on its own and I don't want to over-scope it — mostly want to make sure the live-update gap is a known follow-up rather than a surprise. Happy to help with the engine side if useful.

liaocaoxuezhe:
Thanks for the thorough review — really appreciate you tracing the half-wire down to the exact engine sites. I landed the engine wiring in this PR (newest commit), addressing both issues plus the decode question:

**Issue 1 — `classifyOnePath` live-drop.** The Kimi branch now accepts both shapes: 3-segment legacy and 5-segment `.kimi-code` (`parts[2] == "agents" && parts[4] == "wire.jsonl"`), pulling the project from the `<workdir>` segment either way. Confirmed your empirical check — a 5-segment rel path now classifies instead of being dropped.

**Issue 2 — `SyncSingleSessionContext` project drift.** Replaced the two-level `filepath.Base(filepath.Dir(filepath.Dir(path)))` derivation with the first segment of the path relative to the sessions dir, which is the `<workdir>` in both layouts. The `agents` mis-derivation is gone.

**Question 1 — project naming.** Added `DecodeKimiProjectDir` mirroring the Cursor pattern: `wd_<workdir>_<12-hex>` -> `kimi-code`, `pi-mono`, etc. Legacy `.kimi` hashes carry no `wd_` prefix and pass through unchanged, so existing behavior is preserved. The decode runs in all three sites (discover, classify, single-session resync) so the display value stays consistent.

**Question 2 — real-archive round-trip.** The integration test uses directory-name shapes copied from an actual `.kimi-code` install (`wd_kimi-code_057f5c09ee3f`, `wd_pi-mono_77f54d0fe81a`, ...). Re your heads-up: I added a `SyncPaths`-level test (`TestClassifyOnePath_Kimi`) plus a `SyncPaths`+`SyncSingleSession` round-trip, so the Issue 1 live-drop is covered at the watcher-classification layer that `parse-diff` can't reach.

Tests added: the decoder table, the classify gate (legacy + new-layout main/agent-0 + negative case), and the SyncPaths/SyncSingleSession round-trip asserting the project is the decoded workdir rather than `"agents"`. Parser and sync package tests are green locally.

Happy to split anything out or adjust if you'd rather shape it differently.

liaocaoxuezhe:
Good catch — fixed in the latest commit (f4ba855) by taking the reject-at-import direction. The ID embeds raw path components joined by `:`, and `FindKimiSourceFile` splits on `:` and `IsValidSessionID`-validates each segment, so the import side now matches that contract instead of being looser:

- `DiscoverKimiSessions` and `classifyOnePath` validate every component that becomes an ID segment (workdir, session, and agent names) via a new `kimiIDComponentsValid` helper built on the existing `IsValidSessionID`. A name containing `:` or anything outside `[A-Za-z0-9_-]` is skipped before it reaches the DB, so there's no imported-but-non-resyncable state.
- This is symmetric with `FindKimiSourceFile`, which already rejected those IDs on lookup — the rejection now happens earlier, at import, on both entry points (full discovery and the watcher / `SyncPaths` path).

I went with reject-rather-than-encode because it keeps the ID format stable (no migration for already-imported sessions), matches the existing `IsValidSessionID` pattern, and is lossless for real `.kimi-code` names (`main`, `agent-0`, `wd_<name>_<hash>` all pass). Happy to switch to an encode/escape scheme if you'd rather preserve sessions with exotic names instead of skipping them.

Tests added: a `:`-bearing agent name is skipped at discovery while a valid sibling agent in the same session is still imported, and the same name is not classified by `classifyOnePath`. Parser and sync package tests green.

mjacobs (approved):
Re-reviewed at `f4ba855` — this closes out everything from the last pass, cleanly. Verified locally (parser + sync tests green, `go vet` clean):

- **Issue 1 (live-drop):** `classifyOnePath` now switches on both the 3- and 5-component shapes and pulls the project from `<workdir>`. The 5-segment rel path the old `len != 3` gate discarded now classifies.
- **Issue 2 (project drift):** `SyncSingleSessionContext` derives the project from the first segment under the Kimi dir rather than two levels up, so the new layout no longer resolves to `"agents"`.
- **Decode:** `DecodeKimiProjectDir` reads cleanly and handles workdirs containing underscores (`wd_my_project_<hash>` -> `my_project`); legacy hashes have no `wd_` prefix and pass through untouched, so existing display values are preserved.
- **Round-trip safety:** rejecting `:`-bearing / non-`IsValidSessionID` components at both discovery and classify (`kimiIDComponentsValid`) is the right call — it keeps the import side symmetric with `FindKimiSourceFile`, so nothing lands in a non-resyncable state. `kimiSessionIDFromPath` <-> `FindKimiSourceFile` round-trips exactly for both layouts.

`TestClassifyOnePath_Kimi` (incl. the `:`-rejection case) and `TestSyncPathsAndSingleSession_KimiNewLayout` pin down both engine sites.

One thing I couldn't close out empirically: a live `parse-diff --agent kimi --fail-on-change` against a real `.kimi-code` install (this branch predates parse-diff, and I don't have a Kimi archive). The integration-test directory shapes are copied from a real install, which I think is a reasonable substitute — flagging it only so it's a known, accepted gap rather than a silent one.

LGTM.

wesm:
Will merge once CI is passing, thanks everyone!

wesm:
will fix

PR #674, opened 2026-06-13
feat(cli): add background serve mode

### LINKED ISSUE #682: `agentsview serve` holds the terminal, with no built-in background mode
Run `agentsview serve` and it takes over the terminal. The server stays in the foreground and the shell never hands the prompt back. To run another command you open a second terminal or stop the server.

agentsview has no built-in way to start the server in the background. You can fall back on shell tricks like:

- `agentsview serve &`
- `nohup agentsview serve &`
- a `tmux` or `screen` session

None of these belong to agentsview, and each has downsides:

- You have to remember and type them yourself.
- They act differently across shells and operating systems.
- They scatter the basics. The URL the server picked, its process id, and its log file land in three different places, so you go hunting for each one.
- Stopping the server later means finding the right process by hand.

What would help: a flag on `serve` that starts the server in the background, hands the shell prompt back, and prints the URL, the process id, and the log path. Then you stay in the same terminal and stop the server later with its pid.

PR #674 adds `agentsview serve --background` for this.

g-nitin:
Closes #682

`agentsview serve` runs in the foreground and holds the terminal until you stop
it. This adds background startup plus the lifecycle commands to manage it:

- `agentsview serve --background` starts the server as a detached process and
  returns your shell prompt.
- `agentsview serve status` reports whether a server owns the data dir and where
  to reach it (URL, pid, version, uptime).
- `agentsview serve stop` signals the running server, waits for it to shut down,
  and cleans up its runtime record.

## Background startup

The parent re-execs the same binary with `--background` removed, points the
child's stdin at the null device and its stdout and stderr at
`~/.agentsview/serve.log`, and detaches it from the controlling terminal
(`setsid` on Unix, `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` on Windows). It
then polls the kit daemon runtime record until the child reports ready, up to 5
seconds, prints the URL, pid, and log path, and exits.

Concurrent launches are serialized with a dedicated launch lock
(`serve.background.lock`), held only for the parent's startup and distinct from
the daemon start lock so the spawned child can still claim that lock during its
own (possibly long) startup. A launch that loses the lock, or that finds a
writable daemon already running or still starting (`IsLocalDaemonActive`),
reports the existing server instead of spawning a second one. The launch lock is
taken before any config mutation, so two `--background --require-auth`
invocations can no longer each generate and persist a different auth token; only
the launch that wins the lock generates one.

## Lifecycle

`status` and `stop` are subcommands of `serve` and read the kit daemon runtime
records. `stop` sends a graceful signal (SIGTERM on POSIX), waits for the
process to exit, escalates to a forced kill if it does not, and removes a
leftover record.

## Tradeoffs and limitations

- On Windows there is no POSIX-signal equivalent for a detached process, so
  `serve stop` kills it; the runtime record is removed afterward.
- Two near-simultaneous `--background` launches are serialized by the launch
  lock. A `--background` launch issued at the same instant as a foreground
  `serve`, before either claims the start lock, can still race -- the same
  window two foreground `serve` calls already have.
- With `require_auth`, a launch that loses the lock prints "already in progress"
  rather than the URL, because its in-memory token predates the winner's freshly
  persisted one. It still avoids spawning a second server.
- `serve.log` is append-only. Each start adds a short banner; the bulk of
  logging goes to `debug.log`, which is capped.

## Where to look

- `cmd/agentsview/serve_background.go`: spawn, launch-lock serialization,
  readiness wait, and argument handling.
- `cmd/agentsview/serve_lifecycle.go`: `serve status` and `serve stop`.
- `cmd/agentsview/serve_background_unix.go` and `serve_background_windows.go`:
  platform detach attributes and terminate signal.
- `cmd/agentsview/daemon_runtime.go`: `liveDaemonRecords`, used by status and
  stop to find live daemons without requiring a ping.
- `cmd/agentsview/cli.go`: the `--background` flag and the `status` / `stop`
  subcommands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ docker run --rm -p 127.0.0.1:8080:8080 \
 
 ```bash
 agentsview serve           # start server, open web UI
+agentsview serve --background  # start server and return to the shell
 agentsview usage daily     # print daily cost summary
 ```
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ docker run --rm -p 127.0.0.1:8080:8080 \
 ```bash
 agentsview serve               # start foreground server
 agentsview serve --background  # start server and return to the shell
+agentsview serve status        # show whether a server is running
+agentsview serve stop          # stop the running server
 agentsview usage daily         # print daily cost summary
 ```
 
@@ -47,8 +49,8 @@ machine, syncs them into a local SQLite database, and serves a web UI at
 
 Use `agentsview serve --background` when you want the dashboard to keep running
 after your terminal prompt returns. The command prints the server URL, process
-ID, and log path (`~/.agentsview/serve.log`). Stop the background server with
-`kill <pid>`.
+ID, and log path (`~/.agentsview/serve.log`). Check on it with
+`agentsview serve status` and shut it down with `agentsview serve stop`.
 
 ## Remote / forwarded access
 
--- a/README.md
+++ b/README.md
@@ -36,14 +36,19 @@ docker run --rm -p 127.0.0.1:8080:8080 \
 ## Quick Start
 
 ```bash
-agentsview serve           # start server, open web UI
+agentsview serve               # start foreground server
 agentsview serve --background  # start server and return to the shell
-agentsview usage daily     # print daily cost summary
+agentsview usage daily         # print daily cost summary
 ```
 
 On first run, agentsview discovers sessions from every supported agent on your
-machine, syncs them into a local SQLite database, and opens a web UI at
+machine, syncs them into a local SQLite database, and serves a web UI at
 `http://127.0.0.1:8080`.
+
+Use `agentsview serve --background` when you want the dashboard to keep running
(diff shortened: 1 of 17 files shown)


g-nitin:
Thanks for the review and catch. You're right that two near-simultaneous
background starts can each bring up a writable server against the same SQLite
archive.

One note for the maintainer(s): this race predates the PR. Foreground `serve`
has the same startup window, because the guard checks for a ready runtime
record rather than a start that is still in flight. I called it out as the
third bullet under "Tradeoffs and limitations" above, so background mode runs
into that existing window rather than opening a new one.

I'm happy to close the window for good. The fix is to gate on the kit daemon
start lock before spawning the child, instead of only checking for a ready
runtime record, plus a test that holds the lock while a second invocation tries
to start. That tightens both the foreground and background paths.

Would you like me to include that in this PR, or track it as a separate
follow-up so this one stays focused on background mode? Either way is fine, but
just wanted to get your preference.

wesm:
I am looking at this

PR #677, opened 2026-06-13
fix: support Claude companion session layout

wesm:
Updates Claude parsing for Claude Code companion session directories by inferring subagent parents from the <session>/subagents path when transcripts do not carry a parent session id. Resolves externalized tool-result content from the session's tool-results companion directory so parsed tool result content and lengths reflect the full saved output while preserving the existing flat transcript path. Adds regression coverage for both behaviors.


cpcloud:
LGTM.

PR #679, opened 2026-06-13
Add Kilo parser support

### LINKED ISSUE #617: Request: Kilo code support
I am using Kilo a lot. Would be nice if you could add support for it. 
Thanks :)

maphew:
Add Kilo parser support

Kilo (https://github.com/Kilo-Org/kilocode) is a fork of OpenCode
that uses the identical session storage format. This PR adds Kilo as
a supported agent using the same parsing logic as OpenCode, reading
sessions from `~/.local/share/kilo/storage/`.

Closes #617


maphew (commented):
Found one issue:

- Medium: Kilo storage sync does not force message replacement at `internal/sync/engine.go:4161`. Kilo uses the same mutable storage format as OpenCode, where message and part JSON can be rewritten in place during streaming/tool updates. `processKilo` returns a normal `processResult` for storage files, and `writeBatch` only forces replacement for `AgentOpenCode` at `internal/sync/engine.go:5340` and `internal/sync/engine.go:5449`. In a normal sync after a Kilo message/part changes without adding a new ordinal, `writeMessages` will append nothing and leave stale message content in the database. Kilo should either set `forceReplace: true` for parsed storage sessions or be included wherever OpenCode forces replacement.

_gpt-5.5 on behalf of matt_

mjacobs:
Thanks for this, @maphew — a fork that reuses OpenCode's storage format wholesale should be cheap to add, and you've leaned on the existing OpenCode plumbing rather than reinventing it (`isOpenCodeFormatStorageAgent`, `hasOpenCodeFormatStorageFingerprint`, the shared replace gates). The `kilo.go` delegation is clean and `Closes #617` is welcome. Two things in the sync lifecycle to flag, both in `internal/sync/engine.go`.

## 1. Dead `isOpenCodeStoragePath` is failing the lint check

`isOpenCodeStoragePath` is now unused — the storage-archive check that called it was generalized to `isOpenCodeFormatStoragePath(agent, path)`, orphaning the old single-agent helper. It's the only red check on the PR:

```
internal/sync/engine.go:5807:6: func isOpenCodeStoragePath is unused (unused)
```

Deleting it should clear `lint`.

## 2. ResyncAll storage-preservation accounting: a storage-only Kilo archive can abort the swap

This is the one worth a careful look. `countRootKiloSQLiteSessions` is modeled on **Kiro** — it filters to the SQLite subset:

```go
file_path LIKE '%kilo.db#%'
```

But Kilo isn't shaped like Kiro here — it's a **hybrid like OpenCode**, and the PR wires it that way: `isOpenCodeFormatStorageAgent` returns `OpenCode || Kilo`, so storage-mode Kilo rows ride OpenCode's archive-*preservation* path. When a parse is preserved against the archive it contributes 0 to `Synced`, so a storage-only archive where every parse is preserved resyncs to `Synced == 0`.

The trap is in `ResyncAll`. `oldFileSessions` starts from `FileBackedSessionCount` (every agent not in `NonFileBackedAgents()` — Kilo included), then subtracts each preserved-root count. Because the Kilo subtraction only covers the `kilo.db#` subset, **storage-mode Kilo rows are never subtracted**. So a storage-only Kilo archive ends with `Synced == 0` and `oldFileSessions > 0`, which makes `preservedOnly` false (it needs `oldFileSessions == 0 || trashedCopied > 0`) and trips the `abortSwap` guard — the correctly-rebuilt temp DB is discarded *before* orphan-copy runs, so the archive isn't restored on that pass.

Kiro can use the filtered count because its file-backed sessions are a genuinely separate source that re-syncs through the normal pipeline; OpenCode/Kilo storage rows are not. The fix is to count Kilo like OpenCode — **drop the `file_path` filter** so the Kilo subtraction covers every root Kilo row, matching `countRootOpenCodeSessions`.

> For context, not a knock: OpenCode already special-cases this exact `oldFileSessions` subtraction (no `file_path` filter) for the same reason, so the fix is just bringing Kilo in line with the agent whose path it reuses. The reason it's invisible in normal use is that it's **data-shape-dependent** — it never reproduces for anyone whose archive includes a `kilo.db` session, only for a storage-only archive.

### On the tests

The Kilo tests are solid for what they cover, but I don't think either exercises this path:

- `TestKiloPreservesStorageArchiveAgainstSQLiteFallback` tests storage preservation, but through `SyncAll`/`SyncPaths`, not `ResyncAll`.
- `TestResyncAllAllowsKiloSQLiteOnlySessions` runs `ResyncAll`, but on a `kilo.db` SQLite session that **re-parses fresh** (`Synced == 1`) — so the abort guard is never reached via the preservation branch.

The missing case is a **storage-only** Kilo archive (every parse preserved, `Synced == 0`) run through `ResyncAll`, asserting `stats.Aborted == false` and that rows survive. OpenCode already has the exact mirror to copy — `TestResyncAllOpenCodeStorageArchivePreservesStaleSQLiteFallback` — which runs `ResyncAll` on a preserved storage archive and asserts `Aborted == false` with `Synced == 0`. That's the shape; a `require.NotZero(stats.Synced)` assertion would give false confidence, since a clean resync re-parses storage rows and never hits the guard.

## Smaller notes / questions

- **+577 lines in `engine.go` for a format-identical fork.** Given Kilo is byte-for-byte OpenCode's storage format, this is more engine surface than I'd have expected. A lot of it looks like agent-parameterized variants of existing helpers (`isOpenCodeFormatStoragePath`, the per-agent `switch`es), which is reasonable — but is there room to make the OpenCode path itself storage-format-agent-aware (a small set of OpenCode-format agents) so the next fork is a registry add rather than another ~500 engine lines? Genuine question, no objection if the current factoring is cleanest.
- **Token usage** — since Kilo shares OpenCode's storage schema, `opencode.go`'s `cache.read`/`cache.write` → `cache_read_input_tokens`/`cache_creation_input_tokens` normalization should apply unchanged via delegation. Have you eyeballed cost/usage for a real Kilo session to confirm the tokens land? (Usage rows are silently filtered on `token_usage != '' AND model != ''`, so a key mismatch is zero cost with no error.)
- **parse-diff** — if you have a real Kilo archive, `agentsview parse-diff --agent kilo --fail-on-change` against a freshly-resynced (quiescent) archive is the quickest empirical check that the write-path round-trips with zero drift.
- **Registry-completeness** — heads up that `TestRegistryCompleteness` only asserts `listed → registered`, and its `allTypes` list already omits several agents, so a green run there doesn't actually gate Kilo's inclusion. Not your bug, just don't read it as coverage.

Happy to look again once the lint fix and the resync accounting land — the hard part (matching OpenCode's storage semantics) is right.

maphew:
@mjacobs I pushed commit 43a8b75d with the review follow-ups:

- removed the unused `isOpenCodeStoragePath` helper
- fixed ResyncAll accounting so Kilo subtracts all root OpenCode-format sessions, including storage-mode rows
- consolidated several Kilo/OpenCode storage-format branches behind shared OpenCode-format helpers
- added a storage-only Kilo resync preservation regression test asserting the swap does not abort and the archived rows survive

Validation passed:

- `go test ./internal/sync -run 'TestResyncAll(OpenCodeStorageArchivePreservesStaleSQLiteFallback|KiloStorageArchivePreservesStaleSQLiteFallback|OpenCodeStorageArchiveAllowsNewerSQLiteFallback|OpenCodeOnly|KiroSQLiteOnly)$'`
- `go test ./internal/parser`
- `go test ./...`

I could not run `agentsview parse-diff --agent kilo --fail-on-change` because this checkout does not currently expose a `parse-diff` command, and there is no real Kilo archive fixture in the repo beyond parser tests.

_gpt-5.5 on behalf of @maphew_

mjacobs:
Re-checked the follow-up locally — the three things from last pass are all resolved cleanly (dead `isOpenCodeStoragePath` gone; ResyncAll now subtracts all root Kilo rows via `countRootOpenCodeFormatSessions`, with `TestResyncAllKiloStorageArchivePreservesStaleSQLiteFallback` pinning the `Synced==0`/`Aborted==false` case; full suites green here).

The newer roborev pass flagged two parse-diff gaps, and I confirmed both against the branch. They matter because Kilo is auto-advertised as parse-diff-supported (`resolveParseDiffAgents` is registry-derived), so shipping it half-wired makes `parse-diff --agent kilo` report false drift rather than just skipping Kilo:

1. `parseDiffDatabaseSources` synthesizes only Kiro/OpenCode db sources and `stripVirtualSourceSuffix` ignores `kilo.db#<id>` (even though this PR defines `ParseKiloSQLiteVirtualPath`), so SQLite-only Kilo sessions report as source-missing.
2. `forceParse` isn't honored for Kilo at two sites — the `kilo.db` fan-out skip in `processKilo` (missing `!e.forceParse`, which `processOpenCode` has) and `shouldSkipKiloByPath` (missing the `if e.forceParse { return false }` that `shouldSkipOpenCodeByPath` opens with).

Since parse-diff is new from our side, I've wired all four sites to mirror OpenCode, added `TestParseDiffCoversMixedKiloRoot`, and opened it as a PR into your branch so Kilo lands fully parse-diff-wired in one go: maphew/agentsview#1. Feel free to merge, tweak, or push back. (One smaller parity item I left out of that PR: the `kilo.db` branch logs-and-skips per-session parse errors where OpenCode surfaces them as `DiffParseError` — easy follow-up if you want it.)

wesm:
I'm going to take a pass on this

maphew:
@wesm I don't quite understand. Do you mean you're passing on this PR, as in it's not a good one (bad code or whatever), or is it you think a Kilocode parser doesn't belong in Agentsview?

mariusvniekerk:
@maphew @wesm was going he'll take a look at this parser, need to get roborev happy before we can merge these.

wesm:
Yes, I meant I'm going to work on the PR to get it merged. I'll take it from here

wesm:
I'm going to make an effort to de-copy-paste this PR. Please don't push any more while I work on this

PR #680, opened 2026-06-13
feat: route CLI read subcommands to the Postgres sync store via --pg / AGENTSVIEW_PG_URL

### LINKED ISSUE #392: Feature: CLI subcommands should optionally query the Postgres sync store, not just local SQLite
Love this project!

I run a central Postgres instance that multiple machines `agentsview pg push` to. `agentsview serve` on that host gives me the web UI over the aggregated data, but CLI subcommands (`session list`, `usage daily`, `stats`, `session messages`, etc.) only query the local machine's SQLite — so I can only CLI-query whatever that one machine has generated, not the union across all my machines.

**Ask:** a `--pg` flag (or `AGENTSVIEW_PG_URL` env var) that routes read queries to the configured Postgres instead of local SQLite. Something like:

```
agentsview --pg session list --since 7d
agentsview --pg usage daily
agentsview --pg stats
```

Read-only would cover my use case; I don't need CLI writes against central PG.

**Alternatives considered:**

- `psql` directly against the Postgres store — loses agentsview's FTS, cost calculation, tool-call normalization, and the subcommand ergonomics.
- Hitting `agentsview serve`'s HTTP API — works but isn't a first-class CLI surface, and duplicates what the CLI already does.
- A hypothetical `agentsview pg pull` to hydrate local SQLite — would work but round-trips a lot of data just to run one query, and goes stale immediately.

**Use case:** multi-machine developer (laptop + workstation + VM) who wants unified CLI access to session history/usage without opening a browser.

> Related: #346 proposes a stable CLI surface; #326 shipped PG read-only for usage. This request is the natural composition — the CLI surface backed by the PG store.

mvanhorn:
## Summary

Route CLI read subcommands to the Postgres sync store via --pg / AGENTSVIEW_PG_URL

## Why this matters

A user runs a central Postgres instance that multiple machines `agentsview pg push` to. `agentsview serve` already exposes the aggregated PG data over the web UI (read-only PG serve shipped in #326), but the read-only CLI subcommands (`session list`, `usage daily`, `stats`, `session messages`, etc.) only query the local machine's SQLite, so a multi-machine user can only CLI-query whatever the current machine generated, not the union across all machines. The ask is a `--pg` flag (or `AGENTSVIEW_PG_URL` env var) that routes read queries to the configured Postgres store instead of local SQLite. Read-only coverage is sufficient; no CLI writes against central PG are needed. Maintainer wesm (MEMBER) responded "makes sense to me" with no competing PR cross-referenced.

## Changes

All read CLI subcommands obtain their data store through `resolveService(cmd)` in `cmd/agentsview/session.go`, which today calls `detectTransport` then `newService(cfg, tr)` (in `transport.go`) to pick HTTP-daemon vs direct-SQLite. Add a `--pg` opt-in: when `--pg` is set (or `cfg.PG.URL` is populated from `AGENTSVIEW_PG_URL`/config), bypass transport detection and build a Postgres-backed read service the same way `pg serve` does in `cmd/agentsview/pg.go` -- `postgres.NewStore(pgCfg.URL, pgCfg.Schema, pgCfg.AllowInsecure)` wrapped in `service.NewReadOnlyBackend(store)` (the PG `*Store` already implements the `db.Store` read interface and reports `ReadOnly() bool == true`, which is exactly what `server.New(appCfg, store, ...)` relies on in the pg-serve path). Register `--pg` as a boolean persistent flag where the global flags are wired (`main.go`) so it is available to every read subcommand, resolve `pgCfg` via the existing `config.ResolvePG()`, and return a cleanup that closes the store. Keep write subcommands (those using `resolveWritableService`) unaffected -- they should error clearly if `--pg` is combined with a write op rather than silently writing to SQLite.

## Testing

- Happy path: `--pg` set with a reachable PG URL routes `session list` / `usage` / `stats` reads through `service.NewReadOnlyBackend(postgres.NewStore(...))` and returns the aggregated rows (assert backend selection given a fake/PGConfig, since live PG is gated behind pgtest tags).
- Env equivalence: `AGENTSVIEW_PG_URL` set with no `--pg` flag resolves the same PG route via `config.ResolvePG()`; flag and env produce identical store selection.
- Default unchanged: with neither `--pg` nor any PG config, `resolveService` falls back to the existing transport-detection path (HTTP daemon when discoverable, else direct SQLite) -- existing transport tests still pass.
- Error path: `--pg` requested but `pgCfg.URL` is empty returns a clear "pg url not configured" error instead of silently falling back to SQLite.
- Write guard: combining `--pg` with a write-capable subcommand (resolveWritableService path) returns an explicit read-only refusal rather than writing to local SQLite.
- Cleanup: the returned cleanup closes the PG store (no leaked connection) on both success and error-after-open paths.

Fixes #392

AI was used for assistance.

### DIFF
--- a/cmd/agentsview/session.go
+++ b/cmd/agentsview/session.go
@@ -61,6 +61,13 @@ func resolveService(
 			"loading config: %w", err,
 		)
 	}
+	pgCfg, usePG, err := resolvePGReadConfig(cmd, cfg)
+	if err != nil {
+		return nil, nil, err
+	}
+	if usePG {
+		return newPGReadService(cfg, pgCfg)
+	}
 	tr, err := detectTransport(cfg.DataDir, cfg.AuthToken, 0)
 	if err != nil {
 		return nil, nil, err
@@ -78,6 +85,11 @@ func resolveWritableService(
 ) (service.SessionService, func(), error) {
 	if remote, _ := cmd.Flags().GetString("server"); remote != "" {
 		return nil, nil, errors.New("--server not yet implemented")
+	}
+	if pgReadRequested(cmd) {
+		return nil, nil, errors.New(
+			"--pg is read-only and cannot be used with write commands",
+		)
 	}
 	cfg, err := config.LoadPFlags(cmd.Flags())
 	if err != nil {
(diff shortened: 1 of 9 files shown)


mvanhorn:
Really appreciate the review, @wesm. Routing the CLI read subcommands through the Postgres sync store makes the pg sync path consistent.

mvanhorn:
Thank you @wesm. Routing the read subcommands through the Postgres sync store makes the multi-machine pg push setup actually usable from the CLI.

PR #681, opened 2026-06-13
feat(cowork): index Claude Desktop cowork (local agent mode) sessions

### LINKED ISSUE #639: cowork conversation statistics are missing / not counted
**Describe**

When using the cowork feature, the conversation interactions/statistics are not being recorded or counted properly in the Sessions/Usage.

for MacOS:
`~/Library/Application Support/Claude/local-agent-mode-sessions/<uuid>/<uuid>`

Cauchon:
Adds support for indexing Claude Desktop "cowork" (local agent mode) sessions, which were previously not discovered, so their conversations and statistics never showed up in Sessions/Usage.

Fixes #639.

Claude Desktop stores cowork sessions under its local-agent-mode-sessions directories — on macOS `~/Library/Application Support/Claude/local-agent-mode-sessions/<uuid>/<uuid>`, with Linux and Windows equivalents. This adds a new `cowork` agent to the registry with its own discovery, parser, and source-file resolution, so these sessions sync into SQLite and are counted like any other agent.

What's included:

- New `cowork` agent definition in the parser registry (`internal/parser/types.go`) with the `cowork:` ID prefix and per-OS default directories (`internal/parser/cowork_paths.go`).
- Cowork session parser and discovery (`internal/parser/cowork.go`) plus tests.
- Sync engine wiring (`internal/sync/engine.go`) so cowork directories are discovered and watched.
- Frontend label so the agent renders as "Claude Cowork".

Where to look: `internal/parser/cowork.go` for the parsing and discovery logic, and the registry entry in `internal/parser/types.go`.

Limitation: the on-disk cowork format is derived from observing Claude Desktop's local-agent-mode-sessions output, so it may need updates if that format changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,7 @@ agentsview auto-discovers sessions from all of these:
 | Antigravity        | `~/.gemini/antigravity/`                               |
 | Antigravity CLI    | `~/.gemini/antigravity-cli/` (see note below)          |
 | Claude Code        | `~/.claude/projects/`                                  |
+| Claude Cowork      | `~/Library/Application Support/Claude/local-agent-mode-sessions/` (macOS) |
 | Codex              | `~/.codex/sessions/`                                   |
 | Copilot CLI        | `~/.copilot/`                                          |
 | Cortex Code        | `~/.snowflake/cortex/conversations/`                   |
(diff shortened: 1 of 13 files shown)


mariusvniekerk:
@wesm So one open question here is do we want to categorize cowork as different from regular claude?

PR #689, opened 2026-06-14
feat: infer skill names from SKILL.md reads

charlieviettq:
## Summary

- infer `tool_calls.skill_name` when Codex/Cursor read a `SKILL.md` file
- use `SKILL.md` frontmatter `name` when available, with parent directory fallback
- apply inference to Cursor JSONL `tool_use` blocks as well as parser-specific tool calls
- bump `dataVersion` so historical sessions are reparsed

## Compatibility

- no schema changes
- preserves explicit skill names when already present
- does not infer from discovery-only `Glob` or write/edit/copy/move operations

## Privacy/Data

- reads local `SKILL.md` metadata only when the referenced file is available
- no data is uploaded

## Tests

- `go test -tags "fts5,kit_posthog_disabled" ./internal/parser ./internal/db ./internal/server`
- `go test -tags "fts5,kit_posthog_disabled" ./internal/...`
- `git diff --check`

## Review note

This is intentionally separate from the analytics endpoint/UI. It improves data population for agents that do not emit explicit `skill_name` values.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -208,9 +208,18 @@ import (
 // user message in the sidebar instead of the command text.
 // Re-parsing rewrites first_message with the new logic.)
 //
+// (46: Cursor and Codex parsers now infer skill_name from
+// read-like SKILL.md tool calls. Covers Read/ReadFile tool
+// calls and Codex/Cursor shell reads across the Cursor JSONL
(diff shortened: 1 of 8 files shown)


wesm:
looking at this CI failure

mariusvniekerk (approved):

PR #694, opened 2026-06-15
feat: add QwenPaw agent support

gkld:
Adds the QwenPaw coding agent to the registry and sync pipeline.

QwenPaw stores daily conversation transcripts as
`<workspace>/dialog/<YYYY-MM-DD>.jsonl` under `~/.copaw/workspaces/`.
Each runtime hosts multiple agent workspaces (`default`,
`fund_manager`, `note_keeper`, `researcher`, ...), and each workspace
logs one JSONL file per active day.

The parser handles Anthropic-style content blocks (`text`, `thinking`,
`tool_use`, `tool_result`) with one QwenPaw quirk: tool results live in
`role: "system"` messages rather than user messages. Those map to
`RoleUser + IsSystem` so they remain distinguishable from real user
turns without inflating `UserMessageCount`. Timestamps use the
`"YYYY-MM-DD HH:MM:SS.fff"` local-time format and are parsed via
`time.ParseInLocation(..., time.Local)`, mirroring the Hermes parser.

Raw session IDs use the form `<workspace>:<date>`, yielding full IDs
like `qwenpaw:default:2026-04-19`. Discovery walks
`<root>/<workspace>/dialog/*.jsonl`; the watcher path classifier and
project extraction in `internal/sync/engine.go` were extended to match.

Scope / non-goals:

- Only `dialog/*.jsonl` is parsed. The `sessions/*.json` agent-memory
  snapshots overlap with dialog content and are skipped for now.
- `inbox_traces/*.json` (cron run traces) and `inbox_events.json` are
  notification-shaped, not conversations, and are out of scope.
- Token usage aggregation from `token_usage.json` is not wired up; it
  is keyed by day/model rather than by session.

Reviewers should look at:

- `internal/parser/qwenpaw.go` — discovery, source resolution, parse
  loop, role mapping, timestamp parsing.
- `internal/parser/qwenpaw_test.go` — table-driven coverage of the
  happy path plus malformed lines, empty content, missing timestamps,
  multiple tool_use blocks, and the system-role tool_result mapping.
- `internal/sync/engine.go` — four insertion points (watcher path
  classification, process dispatch switch, new `processQwenPaw`,
  project extraction case).
- `internal/parser/types.go` — new `AgentQwenPaw` constant and
  registry entry (`EnvVar: QWENPAW_DIR`, `DefaultDirs:
  [".copaw/workspaces"]`, `IDPrefix: "qwenpaw:"`).
- `frontend/src/lib/utils/agents.ts` — label/color entry (cyan,
  matching `Qwen Code`; can be re-tinted separately if a distinct
  visual identity is desired).

### DIFF
--- a/cmd/agentsview/cli.go
+++ b/cmd/agentsview/cli.go
@@ -611,6 +611,7 @@ func writeRootHelp(w io.Writer, root *cobra.Command) {
 	fmt.Fprintln(w, "  AMP_DIR                 Amp threads directory")
 	fmt.Fprintln(w, "  ZED_DIR                 Zed data directory")
 	fmt.Fprintln(w, "  QWEN_PROJECTS_DIR       Qwen Code projects directory")
+	fmt.Fprintln(w, "  QWENPAW_DIR             QwenPaw workspaces directory")
 	fmt.Fprintln(w, "  DEEPSEEK_TUI_SESSIONS_DIR")
 	fmt.Fprintln(w, "                          DeepSeek TUI sessions directory")
 	fmt.Fprintln(w, "  QCLAW_DIR               QClaw agents directory")
--- a/frontend/src/lib/utils/agents.test.ts
+++ b/frontend/src/lib/utils/agents.test.ts
@@ -23,6 +23,7 @@ describe("KNOWN_AGENTS", () => {
       "vscode-copilot",
       "pi",
       "qwen",
+      "qwenpaw",
       "deepseek-tui",
       "openclaw",
       "qclaw",
@@ -85,6 +86,9 @@ describe("agentColor", () => {
     expect(agentColor("qwen")).toBe(
       "var(--accent-cyan)",
     );
+    expect(agentColor("qwenpaw")).toBe(
+      "var(--accent-cyan)",
+    );
     expect(agentColor("deepseek-tui")).toBe(
       "var(--accent-cyan)",
     );
@@ -123,6 +127,7 @@ describe("agentLabel", () => {
     expect(agentLabel("piebald")).toBe("Piebald");
     expect(agentLabel("zed")).toBe("Zed");
     expect(agentLabel("qwen")).toBe("Qwen Code");
+    expect(agentLabel("qwenpaw")).toBe("QwenPaw");
     expect(agentLabel("deepseek-tui")).toBe("DeepSeek TUI");
   });
 
--- a/frontend/src/lib/utils/agents.ts
+++ b/frontend/src/lib/utils/agents.ts
@@ -23,6 +23,7 @@ export const KNOWN_AGENTS: readonly AgentMeta[] = [
   },
   { name: "pi", color: "var(--accent-indigo)", label: "Pi" },
   { name: "qwen", color: "var(--accent-cyan)", label: "Qwen Code" },
+  { name: "qwenpaw", color: "var(--accent-cyan)", label: "QwenPaw" },
   {
     name: "deepseek-tui",
     color: "var(--accent-cyan)",
--- a/internal/parser/qwenpaw.go
+++ b/internal/parser/qwenpaw.go
@@ -1,0 +1,524 @@
+// ABOUTME: Parses QwenPaw sessions/<name>.json files into structured session data.
+// ABOUTME: Handles Anthropic-style content blocks (text/thinking/tool_use/
+// ABOUTME: tool_result) with system-role carriers for tool results.
+package parser
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/tidwall/gjson"
+)
+
+// DiscoverQwenPawSessions walks <root>/<workspace>/sessions/*.json and
+// <root>/<workspace>/sessions/console/*.json. Each QwenPaw runtime
+// hosts multiple agent workspaces (e.g. "default", "fund_manager")
+// and each workspace persists one JSON file per active session under
+// sessions/. Hidden subdirectories (e.g. ".weixin-legacy") and the
+// legacy dialog/*.jsonl layout are skipped.
+func DiscoverQwenPawSessions(root string) []DiscoveredFile {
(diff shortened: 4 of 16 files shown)


mariusvniekerk:
Can you include some santized sample qwenpaw files for this?

and ideally also a script that can be used to regenerate them using the cli of the agent?

Alternatively a stable link to where in the qwenpaw source code these sessions are serialized so that we have something to test against.

gkld:
> Can you include some santized sample qwenpaw files for this?
> 
> and ideally also a script that can be used to regenerate them using the cli of the agent?
> 
> Alternatively a stable link to where in the qwenpaw source code these sessions are serialized so that we have something to test against.

QwenPaw：https://github.com/agentscope-ai/QwenPaw
commit：https://github.com/kenn-io/agentsview/pull/694/changes/68e023de170ce0aa701cb261bf08c8e900ae5a8d

mjacobs:
Took a review pass on this (file-based agent, so the heavy DB-backed sync hazards mostly don't apply). The roborev bot already drove out the serious mechanical defects commit-by-commit — the compile-breaking registry/merge conflict, path traversal, mtime-in-seconds, `read_file` ContentLength, and the root-vs-console ID collision are all fixed in the current head. Below are the items a generic pass wouldn't catch. None are hard blockers, but #1 is worth resolving (or consciously waiving) before merge.

## What's good

Clean separation of discovery / find-source / parse. The subdir-in-ID fix is correct and well-tested; traversal hardening (`isUnderQwenPawRoot` + `IsValidQwenPawIDPart`) is thorough; mtime is nanoseconds with a regression test; `read_file` already lands in `NormalizeToolCategory`/`formatToolUse` as `Read`. The testdata fixtures cover the genuinely awkward cases (channel-scoped `@`/dot filenames, the console subdir, empty content). Test coverage is strong.

## 1. Message writes use the append-only path, but QwenPaw rewrites the whole file (the one to resolve)

`processQwenPaw` doesn't set `forceReplace`, and `AgentQwenPaw` isn't in the `replaceMessages` gate (`internal/sync/engine.go:5050`). So on every re-sync it takes the append-only `writeMessages` path, which **only inserts ordinals greater than the max already stored and never updates or deletes existing rows**.

That path is correct *only* for strictly append-only sources with stable ordinals (Claude/Codex JSONL). QwenPaw isn't that shape:

- QwenPaw's `_atomic_write_json` rewrites the **entire** session JSON on each save (temp file + `os.replace`).
- `agent.memory.content` is an agent-memory snapshot, and the parser assigns `Ordinal` by position in that array on each parse.

If that array is ever compacted, summarized, reordered, or has its latest message finalized in place — all common in agent-memory frameworks, and plausible since every save rewrites the whole file — ordinals shift and the append path **silently keeps stale/misaligned rows**. This is the hazard the engine comment at `engine.go:5019` calls out ("compact-boundary messages can shift the ordinal stream — demand a full rewrite rather than the append-only writeMessages path"), and it's why every other whole-file-rewrite source (OpenCode, Antigravity, AntigravityCLI) is in the `replaceMessages` gate.

**Suggestion:** default `processQwenPaw` to `forceReplace: true` (cheap — the file is rewritten wholesale anyway), *unless* you can confirm from QwenPaw's memory module that `content` is strictly append-only with stable ordinals across a session's life.

## 2. PR description is stale / inverted

The description still describes the original implementation: sessions stored as `dialog/<YYYY-MM-DD>.jsonl`, discovery walking `dialog/*.jsonl`, and — in Scope/non-goals — "Only `dialog/*.jsonl` is parsed. The `sessions/*.json` agent-memory snapshots … are skipped." The code does the opposite: it parses `sessions/*.json` and skips `dialog/*.jsonl`.

More than a doc nit: the pivot from `dialog/` (append-only daily logs) to `sessions/` (memory snapshot) is the substantive design decision, and it's the same question as #1 — the append log is the more complete record if memory ever compacts. A sentence on why `sessions/` was chosen over the append-only `dialog/` logs would help reviewers.

## 3. `gen.py` overclaims fidelity

The README and `gen.py` docstring say the script "drives QwenPaw's own session serializer so the fixtures stay byte-for-byte consistent with what a live runtime writes." It doesn't — it builds the dicts by hand and writes them with `json.dumps(..., indent=2) + "\n"`; the QwenPaw module is only imported to sanity-check the source path. The real serializer emits **compact** JSON (`json.dumps(..., ensure_ascii=False)`, no indent), so the fixtures aren't byte-identical to real output. The stable source link in the thread already satisfies the original ask, so the cleanest fix is to soften the claim (e.g. "hand-authored to match the documented shape"), or actually round-trip messages through `save_session_state` to get the true byte shape.

## Minor / nits

- **roborev's last open item** (`gjson.ValidBytes(raw)` before reading fields, commit `a5690e1`): `TestParseQwenPawSession_MalformedJsonReturnsError` currently passes for the wrong reason — `{not valid json` errors via "content missing," not invalid-JSON detection. A one-line `ValidBytes` guard would make the test assert what it claims.
- **No token usage** is extracted (documented non-goal). Worth confirming QwenPaw's Anthropic-style assistant blocks don't carry a per-message `usage` field that's being dropped — if they do, capturing it is cheap; otherwise cost analytics show zero for all QwenPaw sessions.
- **Add an `AgentByPrefix` routing test** for `qwenpaw:` vs `qwen:`. They're disjoint today (`"qwenpaw:"` doesn't `HasPrefix "qwen:"`), but that's exactly the near-collision a test should lock in.
- **Registry-completeness:** the frontend `agents.test.ts` was updated, but I don't see `qwenpaw` added to the Go `allTypes` list (`internal/parser/types_test.go`). Adding the entry is free.
- **Serve path with `@`/`.` IDs:** no `Pattern` validation on the `{id}` route param and DB lookup is exact-match, so channel-scoped IDs like `qwenpaw:default:user@example.com_…` serve fine (Kimi/OpenClaw already use colon-bearing IDs). Low risk; a quick frontend URL spot-check wouldn't hurt.

wesm:
I can take it from here if that's ok

PR #384, opened 2026-04-23
Add OpenCode storage-backed session support

erans:
## Summary
- add file-backed OpenCode session discovery, parsing, sync, and watch support under `OPENCODE_DIR/storage`, while keeping legacy `opencode.db` as transparent fallback
- preserve storage-backed archives across fallback and resync paths, and harden OpenCode watch handling for source transitions and stat failures
- keep incremental bulk sync cheap by using session-file mtimes for storage-backed `SyncAllSince`, with child-aware source mtimes still used for watcher and session-watch paths

## Test Plan
- [x] `make test`
- [x] `go test ./cmd/agentsview ./internal/parser ./internal/sync -count=1`
- [x] `go test ./internal/sync -run 'Test(SyncAllSinceOpenCodeStorageRequiresSessionMtime|SourceMtimeOpenCodeStorageTracksChildRewrite|SourceMtimeOpenCodeStorageTracksChildRemoval|SourceMtimeOpenCodeStorageTracksPartDirRemoval|SourceMtimeOpenCodeStorageTracksMessageDirRemoval|SourceMtimeOpenCodeSQLite)$' -count=1`

## Notes
- The low `FindOpenCodeSourceFile` SQLite existence-probe review finding is intentionally deferred on this branch.

PR #700, opened 2026-06-16
feat(parser): add MiMoCode session support via OpenCode format registry

### LINKED ISSUE #697: Add MiMoCode support
## Summary
MiMoCode does not appear to be supported today, and the existing `OpenCode` support does not seem to cover it.

## Why I think this needs separate support
- The README lists `OpenCode`, but not `MiMoCode`.
- A repo-wide code search shows no `mimocode` references.
- The current OpenCode parser/discovery code appears to assume OpenCode-specific paths and filenames such as:
  - `~/.local/share/opencode/`
  - `opencode.db`
  - `storage/session/...`
  - `storage/message/...`
  - `storage/part/...`

On my machine, MiMoCode uses a different root/layout:
- `~/.local/share/mimocode/`
- `mimocode.db`
- `storage/session_diff/...`

So even though MiMoCode looks OpenCode-adjacent conceptually, it does not appear to be a drop-in match for the current OpenCode ingestion path.

## Request
Please add MiMoCode support, either by:
- adding a dedicated `MiMoCode` agent source/parser, or
- generalizing the current OpenCode integration so it can detect and ingest MiMoCode's layout as well.

A matching config/env var such as `MIMOCODE_DIR` would also be helpful if separate support is the preferred route.

## Relevant code paths
- `README.md`
- `internal/parser/opencode.go`
- `internal/parser/discovery.go`
- `internal/parser/types.go`

rodboev:
## Summary
- add `OpenCodeFormatDef` to `AgentDef` so agents that reuse OpenCode's storage layout are declared via registry config rather than per-agent engine wiring
- add a MiMoCode registry entry with its own env/config keys, ID prefix, default root, and `OpenCodeFormat` config
- refactor existing OpenCode discovery/source/virtual-path helpers into shared parameterized internals; engine checks use `IsOpenCodeFormatAgent()` instead of enumerating agent types
- future forks of the OpenCode storage format need only a registry entry and thin wrapper one-liners

## Scope
- bounded to parser registration, discovery/source lookup, sync engine wiring, focused parser coverage, and the supported-source README table entry
- follow-up to `#384`'s OpenCode storage-backed support; no unrelated UI, analytics, or taxonomy changes
- addresses the factoring direction from [#679 review](https://github.com/kenn-io/agentsview/pull/679#issuecomment-4701192544): "make the OpenCode path itself storage-format-agent-aware so the next fork is a registry add"

## Review Notes
- start with `OpenCodeFormatDef` and `IsOpenCodeFormatAgent` in `internal/parser/types.go`
- `buildOpenCodeParsedSession` is parameterized with `agentType` and `idPrefix` so MiMoCode sessions get `AgentMiMoCode` and `"mimocode:"` prefixes without duplicating the session builder
- `ParseOpenCodeFormatVirtualPath` iterates registered OpenCode-format DB names so engine callers don't need per-agent branches
- focused local validation covered `./internal/parser/...` with the `MiMoCode|OpenCode` filter; `go vet ./...` passed

Fixes #697

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,7 @@ agentsview auto-discovers sessions from all of these:
 | Kimi               | `~/.kimi/sessions/`                                    |
 | Kiro CLI           | `~/.kiro/sessions/cli/`, `~/.local/share/kiro-cli/`    |
 | Kiro IDE           | `~/Library/Application Support/Kiro/` (macOS)          |
+| MiMoCode           | `~/.local/share/mimocode/`                             |
 | OpenClaw           | `~/.openclaw/agents/`                                  |
 | OpenCode           | `~/.local/share/opencode/`                             |
 | OpenHands CLI      | `~/.openhands/conversations/`                          |
--- a/internal/parser/discovery.go
+++ b/internal/parser/discovery.go
@@ -71,18 +71,29 @@ type OpenCodeSource struct {
 }
 
 // openCodeFormat parameterizes the shared OpenCode storage format by
-// the per-agent SQLite filename and the agent label stamped on
-// discovered sessions. Kilo is a fork of OpenCode with an identical
-// on-disk layout, so both agents share one implementation and differ
-// only in these two values.
+// the per-agent SQLite filename, the storage/<sessionSubdir> that holds
+// session JSON, and the agent label stamped on discovered sessions.
+// Kilo is a fork of OpenCode with an identical on-disk layout; MiMoCode
+// is a fork that stores sessions under storage/session_diff and a
+// mimocode.db SQLite fallback. All share one implementation and differ
+// only in these values.
 type openCodeFormat struct {
-	agent  AgentType
-	dbName string
+	agent         AgentType
+	dbName        string
+	sessionSubdir string
 }
 
 var (
-	openCodeFmt = openCodeFormat{agent: AgentOpenCode, dbName: "opencode.db"}
-	kiloFmt     = openCodeFormat{agent: AgentKilo, dbName: "kilo.db"}
+	openCodeFmt = openCodeFormat{
+		agent: AgentOpenCode, dbName: "opencode.db", sessionSubdir: "session",
+	}
+	kiloFmt = openCodeFormat{
+		agent: AgentKilo, dbName: "kilo.db", sessionSubdir: "session",
+	}
+	mimoFmt = openCodeFormat{
+		agent: AgentMiMoCode, dbName: "mimocode.db",
+		sessionSubdir: "session_diff",
+	}
 )
 
 func resolveOpenCodeFormatSource(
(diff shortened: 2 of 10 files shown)


rodboev:
All four findings addressed on the latest push (`0f05cd1a`):

1. **Parser hard-codes (HIGH)**: `ParseOpenCodeDB`, `ParseOpenCodeSession`, `ParseOpenCodeFile`, and `buildOpenCodeSession` now accept `agentType`/`idPrefix` params threaded from the caller. MiMoCode sessions parse as `AgentMiMoCode` with `mimocode:` IDs.

2. **DB sync iterates only AgentOpenCode (MEDIUM)**: `syncOpenCode` now iterates `OpenCodeFormatAgents()`. The DB count, single-session resync, pending-session-ID, and progress-tracking helpers are all generalized over `OpenCodeFormatDef`.

3. **classifyOpenCodePath ignores MiMoCode dirs (MEDIUM)**: Path classification iterates all registered OpenCode-format agents, matching each agent's `DBName` and `SessionSubdir` for session/message/part paths.

4. **shouldCacheSkip and replaceMessages are OpenCode-only (MEDIUM)**: `shouldCacheSkip` uses `IsOpenCodeFormatAgent()` and per-agent `SessionSubdir` for storage-mode detection. Both `replaceMessages` guards use `IsOpenCodeFormatAgent()` instead of `== AgentOpenCode`.

Focused `MiMoCode|OpenCode` parser and sync tests pass; `go vet ./...` clean.

rodboev:
Follow-up on `235d719`:

**File watcher OpenCode-only gate (MEDIUM)**: The watcher in `main.go` now checks `def.OpenCodeFormat != nil` instead of `def.Type == AgentOpenCode`, and uses `ResolveOpenCodeFormatWatchRoots(d, def.Type)` which resolves each agent's DB name and session subdir from the registry. MiMoCode roots now get the same hybrid/SQLite/storage watch-path logic as OpenCode.

rodboev:
Follow-up on `8c316c8`:

**Parse-diff virtual path stripping (MEDIUM)**: `stripVirtualSourceSuffix` now uses `ParseOpenCodeFormatVirtualPath` instead of `ParseOpenCodeSQLiteVirtualPath`, so `mimocode.db#session_id` paths are correctly normalized to their base DB path. This covers parse-diff presence sweeps, `--limit` reporting, and per-session error classification.

rodboev:
Follow-up on `9ad4168`:

1. **Lint modernize** (CI failure): replaced `HasPrefix` + `TrimPrefix` with `CutPrefix` in `syncSingleOpenCode`.

2. **Absent default dirs trigger SyncAll polling (MEDIUM)**: the watcher now only adds an OpenCode-format root to `unwatchedDirs` when the root directory itself exists but couldn't be resolved to watch paths. Absent default dirs (user doesn't have MiMoCode installed) are silently skipped.

wesm:
i'm working on this

wesm:
will merge pending CI

Issue #655, opened 2026-06-12
pg push: sessions.id is the sole PG primary key, so two machines pushing the same session id silently merge and ping-pong


maphew:
## Problem

In the PostgreSQL schema, `sessions.id` is the sole primary key and `machine` is a plain column (`internal/postgres/schema.go`, `coreDDL`). `pushSession` upserts with:

```sql
ON CONFLICT (id) DO UPDATE SET
    machine = EXCLUDED.machine,
    ...
WHERE sessions.machine IS DISTINCT FROM EXCLUDED.machine
    OR ...
```

and the export side (`ListSessionsModifiedBetween`, `internal/db/sessions.go`) has no machine scoping, so every pushing machine claims every session row it holds.

If two machines hold the **same session id** and both run `pg push`, they fight over one PG row:

1. Machine A pushes the session; the row gets `machine = A`.
2. Machine B pushes the same id; `ON CONFLICT (id)` fires, `machine` flips to `B`, and B's copy of every column overwrites A's. The session's messages are also fully replaced (`DELETE FROM messages WHERE session_id = $1` then reinsert, `internal/postgres/push.go`).
3. Because `machine IS DISTINCT FROM EXCLUDED.machine` satisfies the change predicate by itself, the row now looks permanently "changed" to **both** pushers: each next push flips it back. Two `pg push --watch` daemons rewrite the row (and its messages) on every cycle, forever.

There is no warning anywhere that this is happening; the merge is silent.

## When does the same id exist on two machines?

This is not hypothetical:

- Agent session dirs synced between machines via dotfile sync (Syncthing/Dropbox on `~/.claude`, `~/.codex`, etc.) — both machines parse the same files into identical bare session ids, each stored locally as `machine = 'local'`.
- A home directory migrated or restored from backup onto a second machine, with both machines kept alive and pushing.
- Sessions pulled with `agentsview sync --host` are pushed too (no machine filter), so origin B's session can reach PG twice: as `X` from B's own push and as `B~X` via A — duplication rather than collision, but the same missing-scoping root cause.

## Impact

- Machine filters and per-machine analytics in `pg serve` misattribute sessions, flipping between pushers from one refresh to the next.
- Permanent write churn: every watch cycle re-replaces the row and all its messages, defeating the fingerprint skip optimization.
- Per-machine reset detection keys off rows attributed to the pusher, so attribution flapping makes it unreliable.

## Possible directions

[...]

PR #701, opened 2026-06-16
fix(postgres): preserve source machine on pg push

### LINKED ISSUE #332: pg push overwrites original machine name on remote-synced sessions
## Problem

Sessions imported via `agentsview sync --host` retain their original machine name locally (e.g. `home-nuc`), but `pg push` overwrites the machine field with the local `pg.machine_name` for every session pushed.

In `pushSession` (internal/postgres/push.go:747):

```go
sess.ID, s.machine,
```

This means all sessions appear to originate from the pushing machine in PG, losing the original host provenance.

## Expected Behavior

Sessions should preserve their original `sess.Machine` value when pushed to PG, so multi-machine views correctly attribute sessions to the machine they originated on.

## Impact

PG-backed viewers show all sessions under a single machine name instead of per-host, breaking machine-based filtering and attribution.

rodboev:
## Summary
- preserve `sess.Machine` during `pg push` so remote-synced sessions keep their original source host in PostgreSQL instead of being restamped to the pusher machine
- treat the SQLite-internal `"local"` machine sentinel as empty so it falls back to the configured pusher hostname instead of leaking the sentinel string into PostgreSQL
- add focused coverage for the machine-selection helper including the `"local"` sentinel case

## Scope
- limited to the PG export path's machine attribution logic
- intentionally does not solve the broader same-id collision model discussed in `#655`

## Review Notes
- start with `internal/postgres/push.go`, where `pushSession` now prefers the source session machine and only falls back to the configured pusher machine when the session machine is empty
- the regression coverage lives in `internal/postgres/push_test.go` and `internal/postgres/push_pgtest_test.go`
- local validation completed with `CGO_ENABLED=1 go test -tags "fts5,kit_posthog_disabled" ./internal/postgres/... -count=1` and `go vet ./...`
- a real PostgreSQL-backed `pgtest` run could not be completed locally because no `TEST_PG_URL` or local PostgreSQL/docker executable was available in this environment

Fixes #332

### DIFF
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -4984,6 +4984,26 @@ func TestSyncState_GetSetRoundtrip(t *testing.T) {
 	require.Equal(t, "2026-03-11T13:00:00.000Z", val, "value")
 }
 
+func TestDistinctSessionMachines(t *testing.T) {
+	d := testDB(t)
+	ctx := context.Background()
+
+	sessions := []Session{
+		{ID: "s1", Project: "p", Machine: "host-a", Agent: "claude", CreatedAt: "2026-03-10T12:00:00.000Z"},
+		{ID: "s2", Project: "p", Machine: "host-a", Agent: "claude", CreatedAt: "2026-03-11T12:00:00.000Z"},
+		{ID: "s3", Project: "p", Machine: "host-b", Agent: "claude", CreatedAt: "2026-03-12T12:00:00.000Z"},
+		{ID: "s4", Project: "p", Machine: "local", Agent: "claude", CreatedAt: "2026-03-13T12:00:00.000Z"},
+	}
+	for _, s := range sessions {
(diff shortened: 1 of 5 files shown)


rodboev:
Both findings are false positives, addressed with doc clarification on the latest push (`2108b9bd`):

1. **pgSessionCount reset check**: The pusher always has its own locally-created sessions, so the machine-scoped count is never zero. Sessions sourced from other machines were pushed by those machines and are never pushed by this host, so the incremental reset check remains correct. Added a comment explaining this invariant.

2. **Gradual machine-field repair**: Old rows pushed with `machine="local"` converge naturally: `sessionPushFingerprint` includes `sess.Machine`, so when a session is next modified (message count change, rollout append, etc.) the fingerprint changes, the session re-pushes with the correct machine, and the PG row is repaired. A `dataVersion` bump would force a full resync of every session on every machine, which is far heavier than letting this converge organically. Added a comment documenting the gradual repair behavior.

PR #702, opened 2026-06-16
fix(parser): import Codex renamed session titles

### LINKED ISSUE #405: Support importing manually renamed session titles from Codex
Hi there,

First off, thanks for building such an incredibly useful tool!

Would it be possible to fetch the session titles directly from Codex? 
I currently manage my sessions in Codex by manually renaming them one by one. However, I noticed that `agentsview` seems to automatically grab the first sentence of the session content to use as the title instead.

I imagine there are a lot of users out there who also manually organize and rename their session titles in Codex just like I do, so supporting this would be super helpful.

Thanks again!

rodboev:
## Summary
- import Codex renamed session titles from Codex's `session_index.jsonl` `thread_name` into the existing session-name field instead of always falling back to the first user message preview
- handle both `sessions` and `archived_sessions` directory layouts when resolving the `session_index.jsonl` path so archived session titles are also captured
- add focused Codex parser coverage for both the title-capture path and the archived-session directory variant

## Scope
- limited to Codex parser title extraction and focused parser tests
- follow-up to the broader session-title surfaces already shipped in `#601` and `#209`; no unrelated UI changes

## Review Notes
- start with `internal/parser/codex.go` and the matching Codex parser tests
- the validated wire shape is `session_meta.payload.id` in the session file joined to adjacent `session_index.jsonl` `thread_name`; no guessed transcript-only rename event is involved

Fixes #405

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -515,11 +515,26 @@ func startFileWatcher(
 	}
 
 	var roots []watchRoot
+	seenRoots := make(map[string]struct{})
+	addRoot := func(r watchRoot) {
+		if _, ok := seenRoots[r.root]; ok {
+			return
+		}
+		seenRoots[r.root] = struct{}{}
+		roots = append(roots, r)
+	}
 	for _, def := range parser.Registry {
 		if !def.FileBased {
 			continue
 		}
 		for _, d := range cfg.ResolveDirs(def.Type) {
+			if def.ShallowWatchRootsFunc != nil {
+				for _, watchDir := range def.ShallowWatchRootsFunc(d) {
+					if _, err := os.Stat(watchDir); err == nil {
+						addRoot(watchRoot{d, watchDir, true})
+					}
+				}
+			}
 			if def.WatchRootsFunc != nil {
 				watchDirs := def.WatchRootsFunc(d)
 				if len(watchDirs) == 0 {
@@ -528,9 +543,7 @@ func startFileWatcher(
 				}
 				for _, watchDir := range watchDirs {
 					if _, err := os.Stat(watchDir); err == nil {
-						roots = append(
-							roots, watchRoot{d, watchDir, def.ShallowWatch},
-						)
+						addRoot(watchRoot{d, watchDir, def.ShallowWatch})
 						continue
 					}
 					unwatchedDirs = append(unwatchedDirs, d)
@@ -539,18 +552,14 @@ func startFileWatcher(
 			}
 			if len(def.WatchSubdirs) == 0 {
 				if _, err := os.Stat(d); err == nil {
-					roots = append(
-						roots, watchRoot{d, d, def.ShallowWatch},
-					)
+					addRoot(watchRoot{d, d, def.ShallowWatch})
 				}
 				continue
 			}
 			for _, sub := range def.WatchSubdirs {
 				watchDir := filepath.Join(d, sub)
 				if _, err := os.Stat(watchDir); err == nil {
-					roots = append(
-						roots, watchRoot{d, watchDir, def.ShallowWatch},
-					)
+					addRoot(watchRoot{d, watchDir, def.ShallowWatch})
 				}
 			}
 		}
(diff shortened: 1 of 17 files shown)


rodboev:
Finding addressed on the latest push (`f42e750d`):

The mtime invalidation gap was real: when only `session_index.jsonl` changes (rename), the session file's mtime is unchanged and the sync engine skips it.

Fix: after building the session's `FileInfo`, we now stat `session_index.jsonl` and promote its mtime into `sess.File.Mtime` when it's newer. This means a rename-only edit to the index naturally invalidates the cached session, causing the sync engine to re-parse and pick up the new `SessionName`.

No `dataVersion` bump needed since the mtime comparison handles both existing and new sessions: existing cached rows have the old mtime and will re-parse on next sync, new rows get the correct mtime from the start.

Added `TestParseCodexSession_MtimeIncludesSessionIndex` covering the index-only mtime promotion. All Codex parser tests pass; `go vet ./...` clean.

rodboev:
Follow-up on `442044d`:

1. **Mtime sync-skip mismatch**: `processCodex` now uses a dedicated `shouldSkipCodex` that computes `CodexEffectiveMtime(path, fileMtime)` (max of rollout file and session_index.jsonl) before comparing against the stored value, matching the pattern `shouldSkipCopilot` uses for multi-source mtimes. The stored mtime and the skip-check mtime now agree.

2. **dataVersion bump**: not needed. Existing rows converge naturally when the session or its index is next modified (any message append, rollout change, or rename). This is the same gradual-repair approach used for the pg push machine-field fix in #701. A dataVersion bump forces a full resync of every session on every machine, which is disproportionate for a metadata-only field.

rodboev:
Follow-up on `261a10e`:

1. **Incremental path skips session_name**: `processCodex` now compares the effective mtime against the raw file mtime before entering `tryIncrementalJSONL`. When `session_index.jsonl` is newer than the rollout file, the incremental path is bypassed entirely and a full parse runs, picking up the renamed title.

2. **discoveredFileMtime ignores index**: Added an `AgentCodex` branch in `discoveredFileMtime` that returns `CodexEffectiveMtime(file.Path, fileMtime)`, so `SyncAllSince` filters correctly include sessions whose index changed.

rodboev:
Follow-up on `4d87219`:

**Incremental path misses rename when session also appended (MEDIUM)**: The check now compares the session_index.jsonl mtime against the stored DB mtime for the session path, not against the current file mtime. This catches the case where a session is renamed and then appended before the next sync: the index mtime is still newer than the stored mtime even though the session file mtime is newer than the index.

rodboev:
Follow-up on `5aca06e`:

**Existing current-version rows with missing titles (MEDIUM)**: Bumped `dataVersion` from 43 to 44. This forces a full reparse of all sessions on next sync, ensuring existing Codex rows pick up titles from `session_index.jsonl` that were written before this change landed.

rodboev:
Re: `5aca06e` review, the index-mtime-in-discoveredFileMtime concern:

The effective mtime for most Codex sessions equals their own file mtime (index is older), so `SyncAllSince` filtering still works correctly for unchanged sessions. The `codexIndexMtimeChanged` guard in `processCodex` only bypasses incremental when the index mtime actually exceeds the stored DB mtime for that specific session, not all sessions. Sessions whose stored mtime already accounts for the index (from a previous full parse) won't re-parse.

The remaining edge case is the first sync after an index-only change: every Codex session under the root passes the `SyncAllSince` filter since the root-level effective mtime includes the index, but `shouldSkipCodex` correctly skips most of them because their stored effective mtime matches. Only the affected session(s) get a full parse via `codexIndexMtimeChanged`. Loading the index once per root and refreshing by title diff would be more surgical, but the current path is correct and the per-session skip check keeps the extra work bounded.

rodboev:
Follow-up on `8c35607`:

1. `SyncAllSince` now filters Codex sessions by their own file mtime first, then adds a targeted index-refresh pass only for sessions whose `session_index.jsonl` mtime is newer than the stored DB row.
2. That keeps rename-only title refreshes working without treating the root-level index mtime as every Codex session's discovered mtime.
3. Added sync-level regressions covering both the index-only rename path and the changed archived-duplicate case.

Focused validation: `go test ./internal/sync`

mariusvniekerk (approved):
Validated this works against my own codex sessions

wesm:
rebasing this

PR #704, opened 2026-06-16
feat(parser): extract token usage and cost for VSCode Copilot sessions

stephan379:
VSCode Copilot records per-turn token accounting in each request's result.metadata (promptTokens, outputTokens) along with the resolved model id (resolvedModel, already in pricing-catalog form, e.g. "claude-opus-4-8"). The vscode-copilot parser previously ignored this, so sessions reported 0 output tokens and "n/a" cost while Copilot CLI sessions priced correctly.

The parser now emits one usage event per turn (source "vscode-copilot") and rolls the output tokens into the session total, so cost is catalog-priced downstream exactly like other agents. Turns without token metadata produce no event and leave the totals unset (cost stays "n/a"). The resolved model falls back to the prefixed modelId (e.g. "copilot/claude-opus-4.8") with claude version dots normalized to hyphens when resolvedModel is absent.

Usage events are carried on ParsedSession.UsageEvents and forwarded by the sync engine, keeping the public parser signature (and the shared Positron path) unchanged. Bumps the data version to 44 so existing VSCode Copilot rows re-parse and gain usage/cost.

Adds unit tests covering multi-turn extraction, the model fallback, and the no-token-metadata case.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -30,8 +30,9 @@ import (
 //
 // Bumped to 44: the VSCode Copilot parser now extracts per-turn
 // token usage (promptTokens/outputTokens) and the resolved model from
-// result.metadata into usage events and session output totals, so
-// existing VSCode Copilot rows need re-parsing to gain usage and cost.
+// result.metadata into usage events, session output totals, and peak
+// context, so existing VSCode Copilot rows need re-parsing to gain
+// usage and cost.
 //
 // Bumped to 43: the Pi parser now persists cwd from the
 // session header. Existing Pi rows need re-parsing so their cwd column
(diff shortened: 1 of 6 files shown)


stephan379:
Good catch — fixed in 1740994.

The per-turn `promptTokens` now feed `PeakContextTokens` (max across emitted usage events), so usage reports the real peak instead of 0. Verified against a real session that previously showed `Peak ctx: 0`:

```
Output:    10404
Peak ctx:  272432
Cost:      ~$15.14 (claude-opus-4-8)
```

Tests updated: the multi-turn case asserts the peak equals the largest `promptTokens`, and the no-token case asserts the peak stays unset. `go test ./internal/parser/ ./internal/db/` passes.

PR #711, opened 2026-06-16
Fix refresh toolbar affordance

cpcloud:
This updates the Analytics and Usage refresh affordance so the circular-arrow button remains the only explicit action, with a compact relative "Updated ..." timestamp grouped immediately to its right.

Manual Analytics refreshes now reset the five-minute automatic refresh scheduler, so clicking the arrow does not leave an imminent auto-refresh queued behind it. The old Updating text/spinner and passive New data badge were removed because they competed with the refresh button without adding a clear user action.

While refreshing, the icon still spins and the query progress bar now overlays the content instead of entering layout flow, so a refresh does not temporarily add space below the toolbar.

Reviewers should look at `frontend/src/lib/utils/refresh.ts`, the Analytics and Usage page toolbar markup/styles, and the colocated refresh behavior tests.


cpcloud:
New look:

image

PR #715, opened 2026-06-16
fix(parser): extract OpenClaw camelCase toolCall/toolResult blocks

### LINKED ISSUE #703: no tool call message render
openclaw session, cannt render tool call message, only user and assistant message

mjacobs:
Fixes #703.

OpenClaw emits assistant tool calls as `toolCall` content blocks and tool
results as `toolResult` blocks (camelCase), but the shared content extractor in
`ExtractTextContent` only recognized the Anthropic snake_case `tool_use` /
`tool_result` names. As a result every OpenClaw tool call was silently dropped,
and assistant turns that consisted solely of a tool call parsed to empty content
and were discarded entirely — so transcripts showed only plain user and
assistant text, exactly as reported.

This recognizes the camelCase block types. OpenClaw usually carries the call
arguments under `input` but some tools (e.g. `read`, `exec`) populate only
`arguments`, so the extractor prefers `input` and falls back to `arguments`;
`formatToolUse` does the same so the rendered text matches. The OpenClaw
tool-result text extractor now reads `toolResult` content blocks (text inline
under `text`, structured payload under `content`) in addition to plain `text`
blocks.

The existing OpenClaw parser tests passed because their fixtures used the
Anthropic snake_case block names rather than real OpenClaw output. A new test,
`TestParseOpenClawSession_RealToolCallFormat`, is built from the real on-disk
shape: a `toolCall` block with `input`+`arguments`, an `arguments`-only
`toolCall`, and a standalone `toolResult`-role message carrying a `toolResult`
content block.

Verified against a real archive: re-parsing an OpenClaw session that has 18
`toolCall` blocks on disk now yields 18 parsed tool calls (it previously yielded
zero); across a full archive every OpenClaw session previously had zero parsed
tool calls while other agents were unaffected.

Where to look: `internal/parser/content.go` (shared block-type handling) and
`internal/parser/openclaw.go` (`extractToolResultText`). The change is additive:
among current parsers, the claw-family agents (OpenClaw and QClaw) both route
assistant content through the shared `ExtractTextContent` and were losing these
camelCase blocks, so this also recovers QClaw's tool calls; other agents use the
Anthropic snake_case names and are unaffected. QClaw's separate tool-result text
extractor (`extractQClawToolResultText`) is intentionally left unchanged, since
its real result-block format has not been confirmed against actual session data.

### DIFF
--- a/internal/parser/content.go
+++ b/internal/parser/content.go
@@ -48,23 +48,34 @@ func ExtractTextContent(
 				parts = append(parts,
 					"[Thinking]\n"+thinking+"\n[/Thinking]")
 			}
-		case "tool_use":
+		case "tool_use", "toolCall":
+			// "tool_use" is the Anthropic block type; "toolCall" is
+			// the camelCase variant emitted by OpenClaw. OpenClaw
+			// usually carries the call arguments under "input", but
+			// some tools populate only "arguments", so fall back to
+			// it when "input" is missing or empty.
 			hasToolUse = true
 			name := block.Get("name").Str
 			if name != "" {
+				input := block.Get("input")
+				if input.Raw == "" || input.Raw == "{}" {
+					if args := block.Get("arguments"); args.Exists() {
+						input = args
+					}
+				}
 				tc := ParsedToolCall{
 					ToolUseID: block.Get("id").Str,
 					ToolName:  name,
 					Category:  NormalizeToolCategory(name),
-					InputJSON: block.Get("input").Raw,
+					InputJSON: input.Raw,
 				}
 				switch name {
 				case "Skill":
-					tc.SkillName = block.Get("input.skill").Str
+					tc.SkillName = input.Get("skill").Str
 				case "skill":
-					tc.SkillName = block.Get("input.skill").Str
+					tc.SkillName = input.Get("skill").Str
 					if tc.SkillName == "" {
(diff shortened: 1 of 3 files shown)


mjacobs:
Addressed in 2a449292. The OpenClaw `toolResult` handler now stores `content.Raw` in `ParsedToolResult.ContentRaw` (matching the QClaw parser), so `pairToolResults` decodes it into `tool_calls.result_content`. No `DecodeContent` change was needed: `decodeContent` reads each array block's inline `text`, which OpenClaw `toolResult` blocks carry. Extended `TestParseOpenClawSession_RealToolCallFormat` to assert the stored `ContentRaw` decodes back to the result text.

mjacobs:
Addressed in 3d1f3929. Rather than teach `DecodeContent` to recurse, I removed the nested `toolResult.content` fallback that created the divergence — it was speculative. Checked against a real OpenClaw archive: every `toolResult` content block carries the rendered text inline under `text`, and its `content` field is a string (not a block array), so the fallback was both unreachable and based on the wrong shape. `extractToolResultText` now reads only inline `text`, matching what `decodeContent` reads, so the measured `ContentLength` and the stored/decoded `result_content` stay consistent.

cpcloud:
Thanks!

PR #718, opened 2026-06-17
feat(parser): add Shelley (exe.dev) agent support

mjacobs:
Adds first-class support for **Shelley**, the exe.dev / boldsoftware coding agent ([github.com/boldsoftware/shelley](https://github.com/boldsoftware/shelley)), so its sessions appear alongside every other agent.

## Storage model

Shelley stores all conversations in a single SQLite DB at `~/.config/shelley/shelley.db` (`conversations` + `messages` tables). This is structurally identical to Zed, so the parser and sync wiring mirror the Zed single-DB pattern: a virtual source path (`shelley.db#<conversationID>`), classification before the `pathExists` guard, a WAL/SHM composite mtime, per-session `forceReplace` on parse, and a `SourceMtime` branch for the live per-session watcher. No `ResyncAll` `oldFileSessions` accounting is needed (that path is specific to the OpenCode-format storage agents) — Shelley re-parses fresh from the present DB, and vanished conversations are preserved via the generic orphan-copy.

## Extraction

`messages.llm_data` is a serialized `llm.Message` with PascalCase keys and integer enums. The content `Type` integers start at 2 (text=2, thinking=3, tool_use=5, tool_result=6) because `ContentType` shares an `iota` const block with `MessageRole` and `iota` does not reset between the two groups; this was verified against a real `shelley.db`. Tool results are stored as `user`-role messages (Anthropic-style) and pair into the originating tool call. All generations are included, ordered by `sequence_id` (monotonic and unique per conversation), so a context-reset / compaction boundary never hides earlier history.

## Tokens & cost

`usage_data` already uses the canonical Anthropic token keys, so the raw blob is stored verbatim and cost is catalog-priced. Token usage is also captured on errored assistant turns (stored as `type="error"`). The exact gateway `cost_usd` is present in the payload but is currently 0 from the exe.dev gateway; capturing it without double-counting the catalog-priced per-message tokens is left as a follow-up.

## Where to look

- `internal/parser/shelley.go` — parser, virtual-path / read-only store helpers, content and token extraction.
- `internal/sync/engine.go` — the Shelley sites mirroring Zed (classify, composite mtime, `processShelley`, dispatch, `SourceMtime`, cache-skip).
- `internal/parser/types.go` — the `AgentShelley` registry entry.

## Limitations

- Exact `cost_usd` capture is deferred (see above); standard gateway models are priced correctly by the catalog.
- Validated against real `shelley.db` data; tool-name categories cover the current Shelley tool set (`bash`, `patch`, `keyword_search`, `browser*`, `subagent`, …).

### DIFF
--- a/frontend/src/lib/components/settings/AgentDirSettings.svelte
+++ b/frontend/src/lib/components/settings/AgentDirSettings.svelte
@@ -24,6 +24,7 @@
     piebald: "Piebald",
     antigravity: "Antigravity",
     "antigravity-cli": "Antigravity CLI",
+    shelley: "Shelley",
   };
 </script>
 
--- a/internal/db/messages.go
+++ b/internal/db/messages.go
@@ -1252,7 +1252,8 @@ func (db *DB) MessageRoleTimeFingerprint(sessionID string) (string, error) {
 // has_tool_use, and a SHA-256 over the sanitized thinking_text. The
 // parse-diff comparator uses it as a tier-1 fast path so a parser change
 // confined to these columns still triggers the tier-2 row comparison.
-// Not used by the PG push fast-path.
+// PG push uses it with a PostgreSQL-side twin to avoid skipping
+// metadata-only rewrites.
 func (db *DB) MessageFlagsFingerprint(sessionID string) (string, error) {
 	rows, err := db.getReader().Query(
 		`SELECT ordinal, is_system, has_thinking, has_tool_use,
--- a/internal/db/messages.go
+++ b/internal/db/messages.go
@@ -1175,11 +1175,10 @@ func (db *DB) MessageTokenFingerprint(sessionID string) (string, error) {
 
 // MessageContentHashFingerprint returns an exact ordered fingerprint
 // of per-message body content: ordinal, the stored content_length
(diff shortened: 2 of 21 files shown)


mjacobs:
Fixed in cd9223f.

`shelleyToolResultText` now handles `web_search_result` blocks (content type 9), which carry `Title`/`URL` instead of `Text`; previously they flattened to an empty string, so a `web_search_tool_result` could be stored empty and its carrier message dropped. Flattened blocks are now newline-joined so multi-result searches stay readable, and `TestParseShelleyWebSearchToolResult` covers a server-side `web_search` call plus its nested result blocks. Verified against the upstream `llm.Content` struct (web-search payload lives in `Title`/`URL`/`EncryptedContent`, never `Text`).

Context worth flagging: I caught and cleared this locally before pushing, because as of this session I finally have a *matching* roborev configuration — a two-reviewer codex (GPT-5.5) panel (correctness + security) mirroring `ci_default_security`. Until now my local loop wasn't apples-to-apples with CI, so this fix doubled as the first real end-to-end validation of that setup. Both passes came back clean on the fix commit.

PR #724, opened 2026-06-18
fix(postgres): guard pg push against same-id cross-machine row collision

### LINKED ISSUE #655: pg push: sessions.id is the sole PG primary key, so two machines pushing the same session id silently merge and ping-pong

rodboev:
# fix(postgres): guard pg push against same-id cross-machine row collision

## Summary

- Adds stable per-local-DB ownership for PostgreSQL-pushed sessions, using `pg_push_marker_id` rather than the mutable `machine` string to decide whether a row belongs to this pusher.
- Moves the collision guard into the `ON CONFLICT ... DO UPDATE WHERE` path and uses `RowsAffected` plus an ownership re-read to detect skipped conflicts atomically.
- Preserves the legacy same-host repair paths: rows pushed with `machine = 'local'` can still be adopted and repaired to the resolved machine name, and existing unmarked same-machine rows can be marked on upgrade.
- Surfaces ownership skips through `PushResult.SkippedConflicts` and `PushProgress.SkippedConflicts` instead of only logging.
- Preserves `pg_push_marker_id` across full local resyncs, treats legacy DBs without `pg_sync_state` as a clean no-op, and aborts the resync swap if sync-state copy fails after the old DB is quiesced.

## Scope

Changed files:

- `internal/postgres/schema.go`, `internal/postgres/schema_test.go` - add and verify `sessions.owner_marker`.
- `internal/postgres/push.go`, `internal/postgres/push_test.go`, `internal/postgres/push_pgtest_test.go`, `internal/postgres/collision_pgtest_test.go` - owner-marker collision handling, atomic guarded upsert, skipped-conflict reporting, legacy local-sentinel adoption, and regression coverage.
- `internal/db/orphaned.go`, `internal/db/db_test.go` - copy `pg_sync_state` across resync swaps, including legacy missing-table and real-error behavior.
- `internal/sync/engine.go`, `internal/sync/engine_integration_test.go` - preserve the push marker through `ResyncAll` and abort the swap if durable sync-state copy fails.

No read-side scoping change is included here. The PR stays focused on preventing PostgreSQL push from overwriting another local DB's row and messages when the same session ID exists in two places.

## Validation

Focused checks run locally:

```
go test -tags "fts5,kit_posthog_disabled" ./internal/db/... -run "CopySyncStateFrom|CopyExcludedSessionsFrom" -count=1
go test -tags "fts5,kit_posthog_disabled" ./internal/sync/... -run "ResyncAll.*PGPushMarkerID" -count=1
go test -tags "fts5,pgtest" ./internal/postgres/... -run "TestPushSessionGuardsAgainstCrossMachineCollision|TestPushSessionAllowsMachineRenameForSameOwnerMarker|TestPushSessionAdoptsLegacyLocalSentinelRow|TestPushReportsSkippedConflicts|TestPushUpdatesSentinelMachineWhenSyncMachineChanges" -count=1
go vet ./internal/db/...
go vet ./internal/postgres/...
go vet ./internal/sync/...
```

Fixes #655

### DIFF
--- a/internal/db/orphaned.go
+++ b/internal/db/orphaned.go
@@ -299,6 +299,38 @@ func (d *DB) CopyTrashedDataFrom(sourcePath string) (int, error) {
 	return count, nil
 }
 
+// CopySyncStateFrom copies pg_sync_state rows from the source database into the
+// current database. ResyncAll uses this to preserve durable local sync metadata
+// such as the PG push owner marker across the temp-DB swap.
+func (d *DB) CopySyncStateFrom(sourcePath string) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	ctx := context.Background()
+	conn, err := d.getWriter().Conn(ctx)
+	if err != nil {
+		return fmt.Errorf("acquiring connection: %w", err)
+	}
+	defer conn.Close()
+
+	if _, err := conn.ExecContext(
+		ctx, "ATTACH DATABASE ? AS old_db", sourcePath,
+	); err != nil {
+		return fmt.Errorf("attaching source db: %w", err)
+	}
+	defer func() {
+		_, _ = execWithoutCancel(ctx, conn, "DETACH DATABASE old_db")
+	}()
+
+	_, err = conn.ExecContext(ctx, `
+		INSERT OR REPLACE INTO main.pg_sync_state (key, value)
+		SELECT key, value FROM old_db.pg_sync_state`)
+	if err != nil {
+		return fmt.Errorf("copying sync state: %w", err)
+	}
+	return nil
+}
+
 // CopyExcludedSessionsFrom copies the excluded_sessions table
 // from the source DB so permanently deleted sessions survive
 // full DB rebuilds. The source must not have active connections.
--- a/internal/db/orphaned.go
+++ b/internal/db/orphaned.go
@@ -339,7 +339,8 @@ func (d *DB) CopySyncStateFrom(sourcePath string) error {
 
 	_, err = conn.ExecContext(ctx, `
 		INSERT OR REPLACE INTO main.pg_sync_state (key, value)
-		SELECT key, value FROM old_db.pg_sync_state`)
+		SELECT key, value FROM old_db.pg_sync_state
+		WHERE key = 'pg_push_marker_id'`)
 	if err != nil {
 		return fmt.Errorf("copying sync state: %w", err)
 	}
--- a/internal/db/orphaned.go
+++ b/internal/db/orphaned.go
@@ -322,6 +322,21 @@ func (d *DB) CopySyncStateFrom(sourcePath string) error {
 		_, _ = execWithoutCancel(ctx, conn, "DETACH DATABASE old_db")
 	}()
 
+	// Older databases may have no pg_sync_state table.
+	var tableExists int
+	err = conn.QueryRowContext(
+		ctx, "SELECT 1 FROM old_db.sqlite_master WHERE type='table' AND name='pg_sync_state'",
+	).Scan(&tableExists)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil
+		}
+		return fmt.Errorf("probing pg_sync_state table: %w", err)
+	}
+	if tableExists != 1 {
+		return nil
+	}
+
 	_, err = conn.ExecContext(ctx, `
 		INSERT OR REPLACE INTO main.pg_sync_state (key, value)
 		SELECT key, value FROM old_db.pg_sync_state`)
--- a/internal/postgres/push.go
+++ b/internal/postgres/push.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"maps"
@@ -26,6 +27,8 @@ const (
 	pushMarkerIDStateKey = "pg_push_marker_id"
 	pushMarkerKeyPrefix  = "push_marker:"
 )
+
+var errSessionOwnershipConflict = errors.New("session ownership conflict")
 
 // syncStateStore abstracts sync state read/write operations on the
 // local database. Used by push boundary state helpers.
@@ -459,6 +462,9 @@ func (s *Sync) pushBatch(
 		if err := s.pushSession(
 			ctx, tx, sess,
 		); err != nil {
+			if errors.Is(err, errSessionOwnershipConflict) {
+				continue
+			}
 			log.Printf(
 				"pgsync: session %s: %v",
 				sess.ID, err,
@@ -848,6 +854,29 @@ func (s *Sync) pushSession(
 ) error {
 	createdAt, _ := ParseSQLiteTimestamp(sess.CreatedAt)
 	isAutomated := sess.IsAutomated
+	// Guard against cross-machine PK collision: if this session already exists
+	// under a different machine name, skip the upsert and warn. Two machines
+	// pushing the same session ID would otherwise ping-pong the row and its
+	// messages on every push cycle.
+	var existingMachine sql.NullString
+	checkErr := tx.QueryRowContext(ctx,
+		`SELECT machine FROM sessions WHERE id = $1`, sess.ID,
+	).Scan(&existingMachine)
+	if checkErr != nil && !errors.Is(checkErr, sql.ErrNoRows) {
+		return fmt.Errorf("checking session ownership %s: %w", sess.ID, checkErr)
+	}
+	pushedMachine := pushedSessionMachine(sess, s.machine)
+	if existingMachine.Valid &&
+		existingMachine.String != "" &&
+		existingMachine.String != pushedMachine &&
+		sess.Machine != "" && sess.Machine != "local" {
+		log.Printf(
+			"pgsync: session %s: skipping — already owned by machine %q, "+
+				"this pusher is %q; sync from the origin machine to update",
+			sess.ID, existingMachine.String, pushedMachine,
+		)
+		return errSessionOwnershipConflict
+	}
 	_, err := tx.ExecContext(ctx, `
 		INSERT INTO sessions (
 			id, machine, project, agent,
--- a/internal/postgres/push.go
+++ b/internal/postgres/push.go
@@ -85,6 +85,10 @@ func (s *Sync) Push(
 	if err != nil {
 		return result, err
 	}
+	previousMarkerMachine, markerExists, err := s.pgPushMarkerMachine(ctx, markerID)
+	if err != nil {
+		return result, err
+	}
 	if full {
 		lastPush = ""
 		// Caller requested a full push — the PG schema
@@ -113,10 +117,6 @@ func (s *Sync) Push(
 	// was reset (schema dropped, DB recreated, etc.). Force a full
 	// push so all sessions are re-synced.
 	if lastPush != "" {
-		markerExists, cErr := s.pgPushMarkerExists(ctx, markerID)
-		if cErr != nil {
-			return result, cErr
-		}
 		if !markerExists {
 			log.Printf(
 				"pgsync: local watermark set but PG push marker " +
@@ -124,6 +124,7 @@ func (s *Sync) Push(
 			)
 			lastPush = ""
 			full = true
+			previousMarkerMachine = ""
 			s.schemaMu.Lock()
 			s.schemaDone = false
 			s.schemaMu.Unlock()
@@ -278,7 +279,7 @@ func (s *Sync) Push(
 		batch := sessions[i:end]
 
 		batchResult, err := s.pushBatch(
-			ctx, batch, full, markerID, &pushed,
+			ctx, batch, full, markerID, previousMarkerMachine, &pushed,
 		)
 		if err != nil {
 			return result, err
@@ -293,7 +294,7 @@ func (s *Sync) Push(
 			for _, sess := range batch {
 				sr, retryErr := s.pushBatch(
 					ctx, []db.Session{sess},
-					full, markerID, &pushed,
+					full, markerID, previousMarkerMachine, &pushed,
 				)
 				if retryErr != nil {
 					return result, retryErr
@@ -369,7 +370,8 @@ func (s *Sync) Push(
 	return result, nil
 }
 
-// pgPushMarkerExists reports whether this host's push marker is present in PG.
+// pgPushMarkerMachine reports whether this host's push marker is present in PG
+// and returns the machine value stored with the marker.
 // A missing marker while the local watermark is set means PG was reset (schema
 // dropped or recreated) since this host last pushed, so a full re-push is
 // needed. Counting rows by machine cannot detect this reliably: another host
@@ -377,21 +379,24 @@ func (s *Sync) Push(
 // also writes -- a remote host's sessions synced in over SSH, or this host's
 // own renamed identity -- masking the loss of this host's own rows. The marker
 // is per-local-DB, so no other pusher can satisfy this check.
-func (s *Sync) pgPushMarkerExists(ctx context.Context, markerID string) (bool, error) {
-	var exists bool
+func (s *Sync) pgPushMarkerMachine(ctx context.Context, markerID string) (string, bool, error) {
+	var machine string
 	err := s.pg.QueryRowContext(ctx,
-		`SELECT EXISTS (SELECT 1 FROM sync_metadata WHERE key = $1)`,
+		`SELECT value FROM sync_metadata WHERE key = $1`,
 		pushMarkerKeyPrefix+markerID,
-	).Scan(&exists)
+	).Scan(&machine)
 	if err != nil {
-		if isUndefinedTable(err) {
-			return false, nil
+		if errors.Is(err, sql.ErrNoRows) {
+			return "", false, nil
 		}
-		return false, fmt.Errorf(
+		if isUndefinedTable(err) {
+			return "", false, nil
+		}
+		return "", false, fmt.Errorf(
 			"checking pg push marker: %w", err,
 		)
 	}
-	return exists, nil
+	return machine, true, nil
 }
 
 // writePushMarker records this host's push marker in PG so a later push can
@@ -452,6 +457,7 @@ func (s *Sync) pushBatch(
 	batch []db.Session,
 	full bool,
 	markerID string,
+	previousMarkerMachine string,
 	pushed *[]db.Session,
 ) (batchResult, error) {
 	tx, err := s.pg.BeginTx(ctx, nil)
@@ -466,7 +472,7 @@ func (s *Sync) pushBatch(
 	skippedConflicts := 0
 	for _, sess := range batch {
 		if err := s.pushSession(
-			ctx, tx, sess, markerID,
+			ctx, tx, sess, markerID, previousMarkerMachine,
 		); err != nil {
 			if errors.Is(err, errSessionOwnershipConflict) {
 				skippedConflicts++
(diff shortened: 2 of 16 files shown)


rodboev:
### roborev triage (`01ed0d6`)

**`push.go:869` — local-sentinel exclusion**
By design. The `sess.Machine != "" && sess.Machine != "local"` guard exists specifically to allow sessions pushed under the `"local"` sentinel to be adopted when `Sync.machine` resolves to the host's actual name. Without this exclusion, `TestPushUpdatesSentinelMachineWhenSyncMachineChanges` fails because the guard blocks legitimate machine-name resolution. Two distinct physical machines cannot produce the same session UUID from independent SQLite databases, so the cross-machine overwrite scenario described does not arise in practice.

**`push.go:862` — TOCTOU between SELECT and upsert**
Acknowledged and disclosed in the PR body. SQLite is the source of truth; PostgreSQL push is a secondary replication step. Two machines independently generating the same session UUID and racing on the INSERT is not a realistic scenario given UUID generation semantics. An atomic `ON CONFLICT DO UPDATE WHERE` guard would be a worthwhile future hardening but is beyond the scope of this collision-skip fix.

mjacobs (commented):
Solid framing — #655's failure mode is captured exactly, and skip-and-warn is the right pragmatic direction.

roborev's two findings are both valid and worth resolving before this lands (the guard not firing for the common local-sentinel collision, and the SELECT-then-upsert TOCTOU) — agreed on those, not repeating them here.

Two things on top, left inline where they're localized:

- The two roborev findings are really one design tension — the guard keys ownership on the mutable `machine` string, so the same heuristic can't both reject a foreign host and allow a same-host rename. Details + a fix primitive that already exists are on the guard line.
- Conflict skips aren't surfaced anywhere in `PushResult` — note on the `continue`.

Minor: deferring the read-side scoping gap is reasonable — worth a tracking-issue link so it isn't lost in the description prose.

internal/postgres/push.go
```diff
@@ -848,6 +854,29 @@ func (s *Sync) pushSession(
...
+		`SELECT machine FROM sessions WHERE id = $1`, sess.ID,
+	).Scan(&existingMachine)
+	if checkErr != nil && !errors.Is(checkErr, sql.ErrNoRows) {
+		return fmt.Errorf("checking session ownership %s: %w", sess.ID, checkErr)
+	}
+	pushedMachine := pushedSessionMachine(sess, s.machine)
+	if existingMachine.Valid &&
+		existingMachine.String != "" &&
+		existingMachine.String != pushedMachine &&
+		sess.Machine != "" && sess.Machine != "local" {
```
mjacobs: The design tension behind both roborev findings lives here: ownership is keyed on the mutable `machine` string. Drop `sess.Machine != "local"` to fix the local/local collision and you instead freeze a host that renamed `pg.machine_name`/hostname — its explicitly-named rows can never update again, and the "sync from the origin machine" warning fires *on* the origin. Keep it, and the headline case stays broken. So I'd avoid a one-line patch to either; they want solving together.

The primitive for a real fix already exists: `pushMarkerID()` (push.go:411) is a stable per-local-DB id that survives renames and differs across machines — exactly the "stable owner marker" #655 hypothesized. Keying the decision on owner-identity rather than hostname resolves both directions. The schema-touching part is that `sessions` only stores `machine` today, not an owner id.

internal/postgres/push.go
```diff
@@ -459,6 +462,9 @@ func (s *Sync) pushBatch(
 		if err := s.pushSession(
 			ctx, tx, sess,
 		); err != nil {
+			if errors.Is(err, errSessionOwnershipConflict) {
```
mjacobs, line 585: When this fires, the session is correctly left out of `pushed` — but it's also not reflected anywhere in `PushResult` (no skipped/conflict tally, `Errors` stays 0, the watermark advances). Combined with the guard's machine-name sensitivity, a misconfigured name can silently drop sessions with only a log line as signal. A `SkippedConflicts` counter would make it observable.

rodboev:
Addressed the review requests in the latest push:

1. Ownership is now keyed to a stable per-local-DB marker instead of the mutable `machine` string.
   - Added `sessions.owner_marker` and write the local `pushMarkerID()` into it.
   - That preserves the same-host rename path for `sess.Machine == "local"` while still rejecting a foreign pusher.

2. The collision check is now enforced atomically in the upsert path.
   - `pushSession` uses `ON CONFLICT (id) DO UPDATE ... WHERE` to allow updates only when the existing row belongs to the same owner.
   - `RowsAffected()` is used to detect a skipped ownership conflict after the upsert attempt.

3. Conflict skips are now surfaced in results.
   - Added `SkippedConflicts` to both `PushResult` and `PushProgress`.

4. Added regression coverage for the reviewed cases.
   - foreign owner collision is skipped
   - same-owner machine rename is allowed
   - skipped conflicts are observable in `PushResult`
   - local-sentinel rows still repush correctly when `Sync.machine` changes

Validation run locally:
- `go test -tags "fts5,pgtest" ./internal/postgres/... -run "TestPushSessionGuardsAgainstCrossMachineCollision|TestPushSessionAllowsMachineRenameForSameOwnerMarker|TestPushReportsSkippedConflicts|TestPushUpdatesSentinelMachineWhenSyncMachineChanges" -count=1`
- `go vet ./internal/postgres/...`

rodboev:
Addressed the latest RoboRev findings in the newest push:

1. Preserved the local push owner marker across full SQLite rebuilds.
   - `ResyncAll` now copies `pg_sync_state` into the temp DB before the swap, so `pg_push_marker_id` survives a full local resync instead of being regenerated.
   - Added a regression test covering `pg_push_marker_id` persistence through `ResyncAll`.

2. Added the legacy local-sentinel adoption path.
   - Unmarked legacy PG rows with `machine = 'local'` are now treated as adoptable by the current local DB, so they can be repaired to the resolved machine name and marked with the stable owner id on the next push.
   - Added pgtest coverage for upgrading an existing `machine='local'`, `owner_marker=''` row.

3. Tightened the zero-row guarded-upsert path.
   - Any `RowsAffected() == 0` result now re-reads the current row ownership and returns the ownership-conflict sentinel when another owner won the race, even if the initial pre-check saw no row.

Validation run locally:
- `go test -tags "fts5" ./internal/sync/... -run "TestSyncAll_PersistsStartedAndFinishedAt|TestResyncAllPreservesPGPushMarkerID" -count=1`
- `go test -tags "fts5" ./internal/db/... -run "TestCopyExcludedSessionsFrom|TestCopySessionMetadataFrom" -count=1`
- `go test -tags "fts5,pgtest" ./internal/postgres/... -run "TestPushSessionGuardsAgainstCrossMachineCollision|TestPushSessionAllowsMachineRenameForSameOwnerMarker|TestPushSessionAdoptsLegacyLocalSentinelRow|TestPushReportsSkippedConflicts|TestPushUpdatesSentinelMachineWhenSyncMachineChanges" -count=1`
- `go vet ./internal/db/...`
- `go vet ./internal/postgres/...`
- `go vet ./internal/sync/...`

rodboev:
Latest push fixes both roborev findings from the `b06987c` review:

1. The SQL `ON CONFLICT DO UPDATE WHERE` predicate now allows unmarked legacy `machine = 'local'` rows to be adopted, matching the `sameSessionOwner` Go-side logic. Previously, `sessions.machine = EXCLUDED.machine` required exact equality, so a sentinel row with `'local'` couldn't be updated to the resolved hostname. `TestPushSessionAdoptsLegacyLocalSentinelRow` should pass in CI now.

2. The push fingerprint now includes the local push marker ID, so unchanged legacy sessions are reprocessed on the first incremental push after upgrade. Without this, the fingerprint-match filter would skip them before `pushSession` could backfill `owner_marker`.

internal/db/orphaned.go
```diff
@@ -299,6 +299,38 @@ func (d *DB) CopyTrashedDataFrom(sourcePath string) (int, error) {
...
+		ctx, "ATTACH DATABASE ? AS old_db", sourcePath,
+	); err != nil {
+		return fmt.Errorf("attaching source db: %w", err)
+	}
+	defer func() {
+		_, _ = execWithoutCancel(ctx, conn, "DETACH DATABASE old_db")
+	}()
+
+	_, err = conn.ExecContext(ctx, `
+		INSERT OR REPLACE INTO main.pg_sync_state (key, value)
```
mjacobs, line 341: Unlike the sibling `CopyExcludedSessionsFrom`, this doesn't probe `sqlite_master` for the source table first. If an older local DB has no `pg_sync_state`, the `INSERT … SELECT` errors, and `ResyncAll` only logs-and-continues — so the marker isn't carried across the swap. The next push then generates a fresh marker and the machine forfeits ownership of every session it previously pushed, which is the exact failure this change is meant to prevent. Mirroring the existing `sqlite_master` probe (and maybe treating a marker-copy failure as fatal to the swap rather than log-and-continue) would close it.

mjacobs (commented):
Re-reviewed the latest push — big step up, and it goes well past the original ask. Keying ownership on a stable per-DB `owner_marker` and folding the check atomically into the `ON CONFLICT … WHERE` (with the `RowsAffected()==0` re-check) resolves the local/local headline case, the same-host-rename regression, *and* the concurrent-insert TOCTOU — I traced all three and they hold. The `SkippedConflicts` counter and preserving the marker across `ResyncAll` are nice touches.

One thing worth a look (inline) plus a couple of minors, none blocking:

- `CopySyncStateFrom` has a missing-source-table edge that can silently forfeit ownership — inline note.
- Minor: `pushMarkerID()` is re-read from SQLite per session inside `pushSession`, though `Push()` already computes it once for the fingerprints — could be threaded through `pushBatch` to avoid the per-session `GetSyncState` read inside the transaction loop.
- Tests: worth adding the literal local/local case (two distinct local DBs, both `machine='local'`, sequential push) and one that drives the `RowsAffected==0` backstop — those are the two paths the guard most relies on, and neither is exercised yet.

rodboev:
Addressed the latest review follow-up in the newest push:

1. `CopySyncStateFrom` now mirrors the legacy-table guard used by `CopyExcludedSessionsFrom`.
   - Older local DBs without `pg_sync_state` are treated as a no-op instead of failing the copy.
   - Added DB-level coverage for the missing-table case.

2. `ResyncAll` now treats a real sync-state copy failure as swap-fatal.
   - If `pg_push_marker_id` cannot be preserved after the old DB is quiesced, the temp DB is closed and removed, the skip cache is restored, the original DB is reopened, and the swap is aborted.
   - That keeps the local owner marker from being silently regenerated after a full resync.

3. Added focused error-path coverage for `CopySyncStateFrom` so copy failures are still surfaced to callers and do not corrupt existing destination state.

Focused validation passed locally:

```
go test -tags "fts5,kit_posthog_disabled" ./internal/db/... -run "CopySyncStateFrom|CopyExcludedSessionsFrom" -count=1
go test -tags "fts5,kit_posthog_disabled" ./internal/sync/... -run "ResyncAll.*PGPushMarkerID" -count=1
```

rodboev:
Fixed the marker creation race in the latest push:

1. `pushMarkerID` now creates the local marker through an insert-if-absent flow and reads back the stored value.
   - If two first-time pushers in the same local DB race, only one marker wins locally and both callers use that stored marker.
   - Added `GetOrCreateSyncState` so this is handled at the SQLite sync-state layer instead of with a read-then-write in the PostgreSQL push code.

2. The resolved marker is now threaded through the push path.
   - `Push` resolves it once.
   - `pgPushMarkerExists`, `writePushMarker`, `pushBatch`, and `pushSession` reuse that value instead of re-reading local sync state per session.

Focused validation passed locally:

```
go test -tags "fts5,kit_posthog_disabled" ./internal/db/... -run "GetOrCreateSyncState|CopySyncStateFrom|CopyExcludedSessionsFrom" -count=1
go test -tags "fts5,kit_posthog_disabled,pgtest" ./internal/postgres/... -run "TestPushMarkerIDReturnsInsertWinner|TestPushSessionGuardsAgainstCrossMachineCollision|TestPushSessionAllowsMachineRenameForSameOwnerMarker|TestPushSessionAdoptsLegacyLocalSentinelRow|TestPushReportsSkippedConflicts|TestPushUpdatesSentinelMachineWhenSyncMachineChanges|TestPushSessionNameRoundTrip|TestPushSessionTerminationStatus|TestPushSessionPreservesSourceMachine" -count=1
```

rodboev:
Fixed the sync-state overwrite in the latest push.

`CopySyncStateFrom` now copies only the durable PostgreSQL owner marker key, `pg_push_marker_id`, instead of replacing every row from `pg_sync_state`. That keeps the rebuilt DB's fresh `last_sync_started_at` and `last_sync_finished_at` values intact after `ResyncAll`, while still preserving the ownership identity needed for PostgreSQL push conflict checks.

Added `TestCopySyncStateFrom_OnlyCopiesDurablePGKeys` to cover both sides: the marker is carried forward, and the destination sync timestamps are not overwritten by stale source values.

Focused validation passed locally:

```bash
go test -tags "fts5,kit_posthog_disabled" ./internal/db/... -run "CopySyncStateFrom|GetOrCreateSyncState|CopyExcludedSessionsFrom" -count=1
go test -tags "fts5,kit_posthog_disabled" ./internal/sync/... -run "ResyncAll.*PGPushMarkerID|SyncAll_PersistsStartedAndFinishedAt" -count=1
```

rodboev:
Fixed the skipped-conflict visibility gap in the latest push.

`agentsview pg push` now includes skipped ownership conflicts in the live progress line once any are observed, includes the skipped count in the final summary, and prints a warning when the count is nonzero.

`agentsview pg push --watch` now logs the skipped ownership-conflict count in each push result and emits a separate warning line when sessions are skipped for ownership conflicts.

Added focused coverage for both user-facing paths:

```bash
go test -tags "fts5,kit_posthog_disabled" ./cmd/agentsview -run "TestWritePGPushSummaryIncludesSkippedConflicts|TestPgPusher_LogsSkippedConflicts|TestPgPusher_LogsPartialPushErrors" -count=1
go test -tags "fts5,kit_posthog_disabled,pgtest" ./internal/postgres/... -run "TestPushReportsSkippedConflicts" -count=1
```

rodboev:
Fixed the legacy rename ownership gap in the latest push.

I also had Opus review the RoboRev finding before finalizing this change. It confirmed the issue is valid for ownerless pre-`owner_marker` rows whose stored PostgreSQL machine was the old fallback hostname while the local session now resolves through the new fallback hostname.

The fix now reads this pusher's existing `push_marker:<id>` metadata value before the push updates it. That previous marker machine is threaded through the push path so ownerless legacy rows can be adopted when their `sessions.machine` matches the same marker's previous machine name. The Go-side `sameSessionOwner` guard and the SQL `ON CONFLICT ... WHERE` predicate both use the same adoption rule, including the blank-machine legacy case Opus called out.

Added `TestPushAdoptsOwnerlessRowsFromPreviousMarkerMachine`, which seeds:

- local marker `legacy-marker-1`
- existing marker metadata value `host-a`
- ownerless PG session row with `machine = 'host-a'`
- current sync machine `host-b`

The full push now adopts the row, updates it to `host-b`, stamps `owner_marker = legacy-marker-1`, and updates the marker metadata to `host-b`.

Focused validation passed locally:

```bash
go test -tags "fts5,kit_posthog_disabled,pgtest" ./internal/postgres/... -run "TestPushAdoptsOwnerlessRowsFromPreviousMarkerMachine|TestPushReportsSkippedConflicts|TestPushSessionGuardsAgainstCrossMachineCollision|TestPushSessionAllowsMachineRenameForSameOwnerMarker|TestPushSessionAdoptsLegacyLocalSentinelRow|TestPushUpdatesSentinelMachineWhenSyncMachineChanges|TestPushMarkerIDReturnsInsertWinner|TestPushedSessionMachine|TestPushSessionNameRoundTrip|TestPushSessionPreservesSourceMachine" -count=1
```

rodboev:
Fixed the alias-history gap in the latest push.

I had Opus review this RoboRev follow-up as well. It agreed the finding was valid: preserving only the current `push_marker:<id>` machine lets the first post-rename push migrate changed legacy rows, but then loses the old hostname needed to adopt untouched ownerless rows later.

The fix now keeps marker state in two parts:

- `push_marker:<id>` remains the current machine/presence marker used for reset detection.
- `push_marker_machine_aliases:<id>` stores a deduplicated JSON list of previous marker-machine names.

Push ownership checks now read the current marker machine plus aliases before pushing, and ownerless legacy rows can be adopted when `sessions.machine` matches either the current marker machine or any preserved alias. The SQL `ON CONFLICT ... WHERE` predicate mirrors the Go guard using the same JSON alias list. On PG reset, aliases are ignored along with the missing marker.

The regression test now proves the second-push case RoboRev called out:

1. Seed marker value `host-a` and an ownerless legacy row stored as `host-a`.
2. Push as `host-b`, adopting the first row and preserving `host-a` under the alias key.
3. Seed a second ownerless `host-a` row after the marker has already been overwritten to `host-b`.
4. Push again and verify the preserved alias still adopts the later row.

Focused validation passed locally:

```bash
go test -tags "fts5,kit_posthog_disabled,pgtest" ./internal/postgres/... -run "TestPushAdoptsOwnerlessRowsFromPreviousMarkerMachine|TestPushReportsSkippedConflicts|TestPushSessionGuardsAgainstCrossMachineCollision|TestPushSessionAllowsMachineRenameForSameOwnerMarker|TestPushSessionAdoptsLegacyLocalSentinelRow|TestPushUpdatesSentinelMachineWhenSyncMachineChanges|TestPushMarkerIDReturnsInsertWinner|TestPushedSessionMachine|TestPushSessionNameRoundTrip|TestPushSessionPreservesSourceMachine" -count=1
```

PR #725, opened 2026-06-18
fix(db): promote orphan subagent sessions to sidebar roots

### LINKED ISSUE #382: Subagent sessions invisible when their parent transcript is missing
## Summary

Some Claude Code subagent sessions never appear in the agentsview sidebar because their parent session transcript file was never written to disk (or was later deleted). The subagent rows exist in SQLite with `relationship_type = 'subagent'` and a `parent_session_id` pointing at a session that does not exist as a row, so they fall through both the default sidebar query (which excludes subagents) and the parent-expansion query (which never fires for the missing parent).

In my local DB this affects **425 of 5517** subagent rows (~7.7%) across **272 distinct missing parent IDs**.

## Reproduction

Two examples from my local store:

```
agent-a66e3f1   parent=5c2f7691-754f-437b-ad07-85d0d8b36a1b   project=msgvault
agent-abd567c   parent=a2ac06c9-9efc-453e-8315-87f060a30919   project=roborev
```

The on-disk shape of the missing parents:

```
$ ls -la ~/.claude/projects/-Users-wesm-code-msgvault/5c2f7691-754f-437b-ad07-85d0d8b36a1b/
drwx------  3 wesm  staff   96 Feb  2 14:28 .
drwx------ ...                  ..
drwx------  4 wesm  staff  128 Feb  2 14:30 subagents
```

Note: the parent directory exists and contains a `subagents/` child, but **no parent `.jsonl` file**. Claude Code spawned subagents but never persisted (or later removed) the parent transcript.

To find the same situation on any DB:

```sql
SELECT COUNT(*)
FROM sessions
WHERE relationship_type = 'subagent'
  AND parent_session_id NOT IN (SELECT id FROM sessions);
```

## What agentsview does today

`internal/db/sessions.go:307-310` (in `buildSessionFilter`):

```go
if !f.IncludeChildren {
    basePreds = append(basePreds,
        "relationship_type NOT IN ('subagent', 'fork')")
}
```

The default sidebar query has `IncludeChildren = false`, so any row with `relationship_type = 'subagent'` is excluded outright. When `IncludeChildren = true`, subagents *are* included via a subquery rooted at their parent (line 478: `OR parent_session_id IN ...`) — but only if the parent itself matches the filter. With no parent row, there is nothing to expand under, and the subagent never renders.

The sync side is correctly populating these rows: `internal/db/sessions.go:768` (`LinkSubagentSessions`) sets `parent_session_id` from `tool_calls.subagent_session_id` and the parser also derives it from the directory layout. The data is right; the display path drops it.

## Suggested fixes (pick one)

**(a) Promote orphan subagents to roots in `buildSessionFilter`.** Treat subagents as roots when their `parent_session_id` does not resolve to an existing session row. One-line predicate change, no schema work, and the subagent surfaces in the sidebar standalone (with its `subagent` relationship type still recorded for analytics).

**(b) Synthesize a placeholder parent at sync time.** When the parser sees a `subagents/` directory whose parent UUID has no `.jsonl`, write a minimal parent session row (zero messages, `display_name = '(missing parent)'`). Subagents then group under the placeholder, preserving the parent-expansion UI, but at the cost of placeholder rows that are not real sessions.

**(c) Render an explicit \"orphaned subagents\" section in the sidebar.** Most surface area to change; clearest UX signal that something on the Claude Code side dropped the parent.

I think **(a)** is the right call: smallest blast radius, no schema/data changes, and the orphan-subagent state is genuinely a tree-with-missing-root rather than something that warrants its own UI section.

## Notes

- Both subagent rows in question are from older sessions (Jan 29 and Feb 2 2026). The on-disk pattern (parent dir present, parent `.jsonl` absent) suggests Claude Code itself either started the parent without persisting a transcript or pruned it later. Worth checking whether current Claude Code versions still produce this shape, but agentsview should be defensive either way.
- The fork case (`relationship_type = 'fork'`) probably has the same defect — same predicate at line 309, same expansion path. A fix should cover both.

rodboev:
# fix(db): promote orphan subagent sessions to sidebar roots via IncludeOrphans flag

## Summary

- Adds `IncludeOrphans` to `SessionFilter` so sidebar reads can promote missing-parent canonical child sessions as synthetic roots instead of silently dropping them.
- Unifies canonical-root selection across the paginated and unpaginated sidebar paths, using the same canonical child relationship set for `subagent`, `fork`, and `continuation`.
- Keeps flat list behavior distinct from root grouping: `subagent` and `fork` remain hidden from normal root lists, while `continuation` stays attached to its live parent for sidebar grouping.
- Applies sidebar orphan behavior across SQLite, PostgreSQL, and DuckDB by setting `IncludeOrphans` in all three sidebar entrypoints.
- Adds paginated SQLite and PostgreSQL coverage for nested missing-parent orphans, soft-deleted-parent non-promotion, and live-parent continuation nesting.

## Scope

Changed files:

- `internal/db/sessions.go` - SQLite sidebar opts into orphan recovery and paginated roots compose base filtering with shared canonical-root logic.
- `internal/db/query_dialect.go`, `internal/db/query_dialect_test.go` - shared canonical-root predicates, dialect relationship sets, and SQL-shape coverage.
- `internal/db/filter_test.go` - SQLite orphan, continuation, and paginated sidebar regression coverage.
- `internal/postgres/sessions.go`, `internal/postgres/store_test.go` - PostgreSQL sidebar parity and pgtest coverage.
- `internal/duckdb/store.go` - DuckDB sidebar opts into orphan recovery and uses the populated canonical child set.

No schema change. No frontend change. No `dataVersion` bump required.

## Validation

Focused checks run locally:

```
go test -tags "fts5,kit_posthog_disabled" go.kenn.io/agentsview/internal/db -run "TestBuildSessionFilterSQLRendersIncludeChildrenCTE|TestIncludeOrphans|TestSidebarSessionIndex" -count=1
go test -tags "kit_posthog_disabled" go.kenn.io/agentsview/internal/duckdb -run "TestDuckDBStoreContract/sessions_cursors_and_metadata" -count=1
go test -tags "fts5,pgtest" ./internal/postgres/... -run "TestStoreGetSidebarSessionIndex.*Orphan|TestStoreGetSidebarSessionIndexPaginatesByDescendantFreshness|TestStoreGetSidebarSessionIndexPaginatesContinuationsAsDescendants" -count=1
go vet ./internal/db/...
go vet ./internal/postgres/...
```

Fixes #382

### DIFF
--- a/internal/db/filter_test.go
+++ b/internal/db/filter_test.go
@@ -560,9 +560,9 @@ func TestIncludeChildrenExcludesOrphanSubagents(t *testing.T) {
 	requireSessions(t, d, f, []string{"root", "root-sub"})
 }
 
-// TestIncludeOrphansPromotesToRoot verifies that orphan subagent and fork
-// sessions surface as synthetic roots when IncludeOrphans is enabled, and
-// remain hidden when IncludeOrphans is false (default).
+// TestIncludeOrphansPromotesToRoot verifies that canonical orphan rows
+// surface as synthetic roots when IncludeOrphans is enabled, and remain
+// hidden when IncludeOrphans is false.
 func TestIncludeOrphansPromotesToRoot(t *testing.T) {
 	d := testDB(t)
 
@@ -623,7 +623,7 @@ func TestIncludeOrphansPromotesToRoot(t *testing.T) {
 		{
 			name:           "WithoutIncludeOrphans",
 			includeOrphans: false,
-			want:           []string{"root", "root-sub", "continuation-orphan"},
+			want:           []string{"root", "root-sub"},
 		},
 	}
 
@@ -1285,6 +1285,59 @@ func TestSidebarSessionIndexPagedPromotesNestedOrphans(t *testing.T) {
 	require.Equal(t, 3, index.Total, "total paged root groups")
 	require.NotEmpty(t, index.NextCursor, "paged sidebar should expose a next cursor when more root groups remain")
 	requireSidebarIndexIDs(t, index.Sessions, []string{"root", "orphan-sub", "orphan-fork"})
+}
+
+func TestSidebarSessionIndexPagedExcludesContinuationWithSoftDeletedParent(t *testing.T) {
+	d := testDB(t)
+	ctx := context.Background()
+
+	insertSession(t, d, "root", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-20T00:00:00Z")
+		s.MessageCount = 5
+		s.UserMessageCount = 2
+	})
+	insertSession(t, d, "deleted-parent", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-19T00:00:00Z")
+		s.MessageCount = 5
+		s.UserMessageCount = 2
+	})
+	require.NoError(t, d.SoftDeleteSession("deleted-parent"), "SoftDeleteSession")
+	insertSession(t, d, "continuation-child", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-18T00:00:00Z")
+		s.MessageCount = 2
+		s.UserMessageCount = 1
+		s.ParentSessionID = new("deleted-parent")
+		s.RelationshipType = "continuation"
+	})
+
+	index, err := d.GetSidebarSessionIndex(ctx, SessionFilter{Limit: 10})
+	requireNoError(t, err, "GetSidebarSessionIndex")
+	require.Equal(t, 1, index.Total, "soft-deleted parent should not promote continuation root")
+	requireSidebarIndexIDs(t, index.Sessions, []string{"root"})
+}
+
+func TestSidebarSessionIndexPagedKeepsContinuationUnderLiveParent(t *testing.T) {
+	d := testDB(t)
+	ctx := context.Background()
+
+	insertSession(t, d, "root", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-20T00:00:00Z")
+		s.MessageCount = 5
+		s.UserMessageCount = 2
+	})
+	insertSession(t, d, "continuation-child", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-19T00:00:00Z")
+		s.MessageCount = 2
+		s.UserMessageCount = 1
+		s.ParentSessionID = new("root")
+		s.RelationshipType = "continuation"
+	})
+
+	index, err := d.GetSidebarSessionIndex(ctx, SessionFilter{Limit: 1})
+	requireNoError(t, err, "GetSidebarSessionIndex")
+	require.Equal(t, 1, index.Total, "live continuation should stay nested")
+	require.Empty(t, index.NextCursor, "only one root group exists")
+	requireSidebarIndexIDs(t, index.Sessions, []string{"root", "continuation-child"})
 }
 
 func TestSidebarSessionIndexReturnsDisplayName(t *testing.T) {
--- a/internal/db/filter_test.go
+++ b/internal/db/filter_test.go
@@ -594,6 +594,20 @@ func TestIncludeOrphansPromotesToRoot(t *testing.T) {
 		s.RelationshipType = "fork"
 	})
 
+	insertSession(t, d, "orphan-grandchild", "proj", func(s *Session) {
+		s.MessageCount = 2
+		s.UserMessageCount = 1
+		s.ParentSessionID = new("orphan-sub")
+		s.RelationshipType = "fork"
+	})
+
+	insertSession(t, d, "continuation-orphan", "proj", func(s *Session) {
+		s.MessageCount = 2
+		s.UserMessageCount = 1
+		s.ParentSessionID = new("missing-continuation-parent")
+		s.RelationshipType = "continuation"
+	})
+
 	tests := []struct {
 		name           string
 		includeOrphans bool
@@ -603,13 +617,13 @@ func TestIncludeOrphansPromotesToRoot(t *testing.T) {
 			name:           "WithIncludeOrphans",
 			includeOrphans: true,
 			want: []string{
-				"root", "root-sub", "orphan-sub", "orphan-fork",
+				"root", "root-sub", "orphan-sub", "orphan-fork", "orphan-grandchild", "continuation-orphan",
 			},
 		},
 		{
 			name:           "WithoutIncludeOrphans",
 			includeOrphans: false,
-			want:           []string{"root", "root-sub"},
+			want:           []string{"root", "root-sub", "continuation-orphan"},
 		},
 	}
 
@@ -1233,6 +1247,44 @@ func TestSidebarSessionIndexStarredIncludesStarredDescendantRoot(t *testing.T) {
 	require.Empty(t, index.NextCursor)
 	require.Equal(t, 1, index.Total, "total starred root groups")
 	requireSidebarIndexIDs(t, index.Sessions, []string{"root", "starred-child"})
+}
+
+func TestSidebarSessionIndexPagedPromotesNestedOrphans(t *testing.T) {
+	d := testDB(t)
+	ctx := context.Background()
+
+	insertSession(t, d, "root", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-20T00:00:00Z")
+		s.MessageCount = 5
+		s.UserMessageCount = 2
+	})
+	insertSession(t, d, "orphan-sub", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-19T00:00:00Z")
+		s.MessageCount = 3
+		s.UserMessageCount = 1
+		s.ParentSessionID = new("missing-parent")
+		s.RelationshipType = "subagent"
+	})
+	insertSession(t, d, "orphan-fork", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-18T00:00:00Z")
+		s.MessageCount = 2
+		s.UserMessageCount = 1
+		s.ParentSessionID = new("orphan-sub")
+		s.RelationshipType = "fork"
+	})
+	insertSession(t, d, "continuation-orphan", "proj", func(s *Session) {
+		s.EndedAt = new("2024-01-17T00:00:00Z")
+		s.MessageCount = 2
+		s.UserMessageCount = 1
+		s.ParentSessionID = new("missing-continuation-parent")
+		s.RelationshipType = "continuation"
+	})
+
+	index, err := d.GetSidebarSessionIndex(ctx, SessionFilter{Limit: 2})
+	requireNoError(t, err, "GetSidebarSessionIndex")
+	require.Equal(t, 3, index.Total, "total paged root groups")
+	require.NotEmpty(t, index.NextCursor, "paged sidebar should expose a next cursor when more root groups remain")
+	requireSidebarIndexIDs(t, index.Sessions, []string{"root", "orphan-sub", "orphan-fork"})
 }
 
 func TestSidebarSessionIndexReturnsDisplayName(t *testing.T) {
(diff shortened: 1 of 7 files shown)


rodboev:
### roborev triage (`c688988`)

**`sessions.go:423` — PostgreSQL sidebar parity**
Out of scope. This PR fixes orphan subagent recovery in the SQLite sidebar path (issue #382). The PostgreSQL sidebar in `internal/postgres/sessions.go` has its own independent query path and would need separate changes. PG parity for orphan recovery is a valid follow-up but not a regression introduced by this PR.

**`sessions.go:362` — soft-deleted parent promotion**
Not a real issue. The CTE's orphan UNION promotes rows where `NOT EXISTS (SELECT 1 FROM sessions p WHERE p.id = s.parent_session_id)`, meaning the parent row must be entirely absent, not merely soft-deleted. A child of a soft-deleted parent still has a parent row in the table, so it will not be incorrectly promoted as a root group by the orphan recovery logic.

internal/db/sessions.go
```diff
@@ -353,6 +354,31 @@ func sidebarStarredRootJoin(enabled bool) string {
 	return "JOIN eligible_roots e ON e.id = t.root_id"
 }
 
+// buildCanonicalRootWhere returns a WHERE fragment that identifies canonical root
+// sessions (non-child sessions that appear in the sidebar). When includeOrphans is
+// false, it restricts to sessions whose parent either doesn't exist or is deleted.
+// When includeOrphans is true, it also promotes orphan child rows (missing parent)
+// to synthetic roots.
+func buildCanonicalRootWhere(includeOrphans bool) string {
```
mjacobs, line 358: This is the second orphan-promotion implementation — the unpaginated path seeds orphans in the recursive CTE (`query_dialect.go`), while this paginated path promotes them here. That split is *why* the soft-deleted-parent divergence roborev caught exists at all. Rather than patch each path to match, I'd consolidate to a single shared canonical-root definition used by both paths — and, once the PG gap is closed, both backends — so they can't drift again.

internal/db/query_dialect.go
```diff
@@ -278,11 +278,33 @@ func buildSessionFilterWithBuilder(
...
+	if f.IncludeOrphans {
+		orphanFilter, orphanOneShotPred := sessionFilterPredicates(f, b, func(col string) string {
+			return "orphan_child." + col
+		})
+		orphanMatchParts := append([]string{}, orphanFilter...)
+		if orphanOneShotPred != "" {
+			orphanMatchParts = append(orphanMatchParts, orphanOneShotPred)
+		}
+		orphanMatchParts = append(orphanMatchParts,
+			"orphan_child.relationship_type IN ('subagent', 'fork', 'continuation')",
```
mjacobs: Two small things here: this seed lists `'continuation'`, but continuations already surface as roots via the base `relationship_type NOT IN ('subagent','fork')`, so the token is redundant rather than load-bearing today. And it diverges from the canonical-root rule, which keys off `sidebarChildRelationshipsSQL` — another way the two promotion paths can drift. If continuations aren't meant to be promoted as orphans, drop it; otherwise reconcile both sides to one set.

internal/db/filter_test.go
```diff
@@ -560,6 +560,109 @@ func TestIncludeChildrenExcludesOrphanSubagents(t *testing.T) {
 	requireSessions(t, d, f, []string{"root", "root-sub"})
 }
 
+// TestIncludeOrphansPromotesToRoot verifies that orphan subagent and fork
+// sessions surface as synthetic roots when IncludeOrphans is enabled, and
+// remain hidden when IncludeOrphans is false (default).
+func TestIncludeOrphansPromotesToRoot(t *testing.T) {
```
mjacobs, line 566: These assert via `requireSessions` -> `ListSessions`, which routes through the CTE path — but the production paginated sidebar (`GetSidebarSessionIndex` with a `Limit`) goes through `buildCanonicalRootWhere`'s orphan branch, which no test exercises. So the shipping path (and the one that diverges) is unverified. Also worth adding: a nested orphan (an orphan subagent that itself has a child) to cover the recursive pull on a synthetic root, and a `continuation`-type orphan since the seed accepts it.

mjacobs (commented):
The orphan-as-root approach (option (a) from #382) is the right call, and the unpaginated SQLite path reads correctly.

roborev's two findings should block merge and I agree with both: the PostgreSQL sidebar not getting the fix (parity), and the paginated path promoting children of soft-deleted parents. Not repeating them here.

A few additional points, inline where localized:

- The soft-deleted divergence exists because the PR ships two independent promotion implementations — worth consolidating to one shared definition (note on `buildCanonicalRootWhere`).
- The new tests exercise the CTE path, not the paginated sidebar that actually ships (note on the tests).
- Small relationship-set drift between the two implementations (note on the orphan seed).

rodboev:
Addressed the review requests in the latest push:

1. PostgreSQL sidebar now matches the SQLite orphan-root behavior.
   - `internal/postgres/sessions.go` now forces `IncludeOrphans` for sidebar reads and uses the same paginated root-candidate logic.

2. The canonical root definition is now shared instead of duplicated.
   - Moved the child/orphan root helpers into the shared session filter builder so SQLite and PostgreSQL both reuse the same rule.

3. Removed the orphan-promotion relationship drift.
   - Orphan promotion now applies only to child relationships (`subagent`, `fork`).
   - `continuation` remains a normal root/descendant path and is no longer treated as an orphan-promoted child.

4. Added coverage for the shipping paginated paths.
   - SQLite paginated sidebar tests now cover nested orphan trees and confirm soft-deleted parents do not promote children.
   - PostgreSQL parity tests cover orphan root pagination and the soft-deleted-parent case as well.

Validation run locally:
- `go test -tags "fts5" ./internal/db/... -run "TestIncludeOrphans|TestSidebarSessionIndex" -count=1`
- `go test -tags "fts5,pgtest" ./internal/postgres/... -run "TestStoreGetSidebarSessionIndex.*Orphan|TestStoreGetSidebarSessionIndexPaginatesByDescendantFreshness|TestStoreGetSidebarSessionIndexPaginatesContinuationsAsDescendants" -count=1`
- `go vet ./internal/db/...`
- `go vet ./internal/postgres/...`

rodboev:
Follow-up fix in the latest push:

The shared canonical-root refactor accidentally changed paginated sidebar grouping for `continuation` sessions. That made a continuation with an existing parent become its own root candidate, which regressed sidebar pagination/index behavior and broke the existing continuation descendant tests.

Adjusted the split so the two semantics stay distinct:

1. Orphan promotion still uses the narrower child set (`subagent`, `fork`).
2. Canonical root grouping uses a separate child set that still includes `continuation`, so continuations remain attached to their existing parent/root group when the parent exists.

That keeps the reviewer-requested orphan promotion change while preserving the pre-existing continuation pagination behavior.

Validation run locally:
- `go test -tags "fts5" ./internal/db/... -run "TestBuildSessionFilterSQLRendersIncludeChildrenCTE|TestIncludeOrphans|TestSidebarSessionIndex" -count=1`
- `go test -tags "fts5" ./internal/server/... -run "TestSidebarIndexPaginationTreatsContinuationsAsDescendants|TestSidebarIndexPaginatesByDescendantFreshness" -count=1`
- `go test -tags "fts5,pgtest" ./internal/postgres/... -run "TestStoreGetSidebarSessionIndex.*Orphan|TestStoreGetSidebarSessionIndexPaginatesByDescendantFreshness|TestStoreGetSidebarSessionIndexPaginatesContinuationsAsDescendants" -count=1`
- `go vet ./internal/db/...`
- `go vet ./internal/postgres/...`

internal/db/query_dialect.go
```diff
@@ -275,14 +344,36 @@ func buildSessionFilterWithBuilder(
...
+	if f.IncludeOrphans {
+		orphanFilter, orphanOneShotPred := sessionFilterPredicates(f, b, func(col string) string {
+			return "orphan_child." + col
+		})
+		orphanMatchParts := append([]string{}, orphanFilter...)
+		if orphanOneShotPred != "" {
+			orphanMatchParts = append(orphanMatchParts, orphanOneShotPred)
+		}
+		orphanMatchParts = append(orphanMatchParts,
+			SidebarChildRelationshipPredicate(b.dialect, "orphan_child"),
```
mjacobs: Small residual of the same class you just fixed, continuation-only: the unpaginated root seed excludes just `sidebarChildRelationships` (subagent/fork), so a `continuation` whose parent exists-but-is-soft-deleted still seeds as a root here, while the paginated path (canonical set, includes continuation) correctly omits it. Net effect: a deleted parent's continuation child shows as a root in the no-limit sidebar but disappears in the paged one. Using `canonicalChildRelationships` for the unpaginated root seed (and this orphan seed) would make continuations consistent across both paths.

internal/db/query_dialect.go
```diff
@@ -125,7 +131,8 @@ func DuckDBQueryDialect() QueryDialect {
 		regexPredicate: func(col, ph string) string {
 			return "regexp_matches(" + col + ", " + ph + ")"
 		},
-		nullsLast: true,
+		sidebarChildRelationships: []string{"subagent", "fork"},
```
mjacobs: `DuckDBQueryDialect` leaves `canonicalChildRelationships` nil here — harmless today since DuckDB only goes through `BuildSessionFilterSQL`, but `CanonicalChildRelationshipsSQL()` would emit `relationship_type IN ()` if DuckDB ever picks up the canonical-root path. Relatedly, DuckDB is the third sidebar backend: `internal/duckdb/store.go` `GetSidebarSessionIndex` sets `IncludeChildren` but not `IncludeOrphans` (SQLite and PG both set it now), so orphan subagent/fork rows are dropped there. Setting `IncludeOrphans = true` in the DuckDB store and populating this dialect's canonical set to match keeps the third backend in parity.

mjacobs (commented):
Re-reviewed — the shared-`QueryDialect` refactor is the right call: it closes the SQLite/PG parity gap at the seam rather than per-path, and splitting `sidebar` vs `canonical` relationship sets is a clean way to keep continuations attached. I confirmed the soft-deleted-parent non-promotion is deliberate and now consistent for subagent/fork across both paths.

A few residuals (inline where localized), all minor:

- One continuation-only divergence is left between the paginated and unpaginated paths — inline note on the orphan seed.
- DuckDB is the third sidebar backend and didn't get the change — inline note.
- Tiny cleanup: in `BuildCanonicalRootWhere`, `base` already promotes every orphan (`NOT (canonicalChild AND parent_exists)` is true whenever the parent is absent), so the `includeOrphans` OR-branch is a no-op and the parameter is effectively dead — worth dropping, or making it actually gate, so a future `false` caller can't silently re-diverge from the unpaginated path.
- Tests: the soft-deleted-parent assertion lives only in a `pgtest` (behind the build tag, so not in the default unit run), and live-parent continuation isn't exercised through the paged path — a SQLite-layer test for both would guard exactly the cases this PR turns on.

rodboev:
Addressed the residual sidebar-root feedback in the latest push:

1. The unpaginated `IncludeChildren` CTE now uses the same canonical-root rule as the paginated sidebar.
   - Canonical children are `subagent`, `fork`, and `continuation`.
   - Missing-parent canonical children are promoted only when `IncludeOrphans` is set.
   - Continuations with live or soft-deleted parents no longer drift between paginated and unpaginated behavior.

2. DuckDB now matches the SQLite and PostgreSQL sidebar entrypoints.
   - `DuckDBQueryDialect` has the same canonical child relationship set.
   - DuckDB `GetSidebarSessionIndex` sets `IncludeOrphans = true` with `IncludeChildren = true`.

3. Added paginated SQLite coverage for the cases called out in review.
   - nested missing-parent orphan recovery
   - soft-deleted parent continuation exclusion
   - live-parent continuation nesting

Focused validation passed locally:

```
go test -tags "fts5,kit_posthog_disabled" go.kenn.io/agentsview/internal/db -run "TestBuildSessionFilterSQLRendersIncludeChildrenCTE|TestIncludeOrphans|TestSidebarSessionIndex" -count=1
go test -tags "kit_posthog_disabled" go.kenn.io/agentsview/internal/duckdb -run "TestDuckDBStoreContract/sessions_cursors_and_metadata" -count=1
```

PR #726, opened 2026-06-18
feat(activity): add Activity reporting dashboard

wesm:
## Summary

Adds an Activity reporting view: a time-windowed dashboard over agent sessions
with summary cards, a concurrency timeline, breakdowns by project / model /
agent (with a toggle between agent-minutes and cost), a sessions table, and
generated insights. Range controls move between day, week, and month windows.

Activity is split into interactive and automated work, building on the
automated-session detection in `internal/db/automated.go`, so automated runs
(for example roborev CI) are measured separately from interactive work. The
split is surfaced throughout the frontend: the summary card, the breakdown bars,
and the concurrency peak bar stack interactive and automated segments; the
sessions table badges automated rows; and a toolbar control filters to all,
interactive, or automated. The two use a blue/orange pair that stays
distinguishable in light and dark themes and under red-green color blindness.

The report query is implemented across all three storage backends — SQLite
(`internal/db/activityreport.go`), PostgreSQL
(`internal/postgres/activityreport.go`), and DuckDB
(`internal/duckdb/activityreport.go`) — with matching shape and ordering, and is
covered by cross-backend parity tests. It is served through a huma OpenAPI route
(`internal/server/huma_routes_activity.go`) and an `activity` CLI subcommand
(`cmd/agentsview/activity.go`).

The dashboard refreshes on a fixed cadence without thrashing. A shared
`RefreshControl` component, used by the Activity, Analytics, and Usage
dashboards, owns a configurable refresh interval (default five minutes) plus a
manual refresh button and a relative "Updated Xm ago" label. SSE and sync events
only flag that newer data exists; the report refetches via the scheduler or the
button, not on every event, so a session actively writing files does not re-run
the aggregation continuously. A failed background refresh keeps the last good
report on screen instead of blanking it, while a first-load or range-change
failure still surfaces the error.

Generated insights can be produced with any supported agent CLI (claude, codex,
copilot, gemini, kiro). The inline Activity Insight panel and the standalone
Insights page share one agent selector.

A parser rule attributes CI worktree sessions (`ci-worktrees/<repo>/...`) to
their owning repository instead of the per-run scratch worktree name, so CI-run
reviews group under the repo in project breakdowns rather than fragmenting into
one project per run.

## Where to look

- `internal/activity/` — domain types, sweep-line concurrency math, and the
  automated/interactive split
- `internal/{db,postgres,duckdb}/activityreport.go` — backend-parity query
  implementations
- `internal/insight/` — insight generation: agent CLI invocation, prompt, and
  summary windowing
- `internal/server/huma_routes_activity.go`, `cmd/agentsview/activity.go` — API
  route and CLI subcommand
- `frontend/src/lib/components/activity/` — Svelte 5 page and components
  (toolbar, breakdowns, concurrency timeline, sessions table, insight panel)
- `frontend/src/lib/components/shared/RefreshControl.svelte`,
  `frontend/src/lib/utils/refresh.ts` — shared refresh control and scheduler
- `internal/parser/project.go` — ci-worktrees attribution rule

## Notes and limitations

- The automated/interactive segmentation is computed in the backend and returned
  by the API; the frontend can show the combined view or filter to one segment.
- Backend parity is maintained across SQLite, PostgreSQL, and DuckDB; the
  PostgreSQL parity tests run under the `pgtest` build tag.
- The insight agent choice is shared in memory between the Activity panel and
  the Insights page and resets to claude on reload.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,7 +10,9 @@ Instructions for autonomous coding agents working in this repository.
 
 ## Required Git Rules
 
-1. Commit every turn.
+1. Commit every turn that changes tracked files.
+1. Do not make empty commits. If a turn is read-only or only changes ignored
+   files, state that no commit was made.
 1. Do not amend commits.
 1. Do not change branches without explicit user permission.
 
--- a/Makefile
+++ b/Makefile
@@ -245,11 +245,11 @@ test-postgres: ensure-embed-dir postgres-up
 	@echo "Waiting for postgres to be ready..."
 	@sleep 2
 	TEST_PG_URL="postgres://agentsview_test:agentsview_test_password@localhost:5433/agentsview_test?sslmode=disable" \
-		CGO_ENABLED=1 go test -tags "fts5,pgtest" -v ./internal/postgres/... -count=1
+		CGO_ENABLED=1 go test -tags "fts5,pgtest" -v ./internal/postgres/... ./internal/activity/... -count=1
 
 # PostgreSQL integration tests for CI (postgres already running as service)
 test-postgres-ci: ensure-embed-dir
-	CGO_ENABLED=1 go test -tags "fts5,pgtest" -v ./internal/postgres/... -count=1
+	CGO_ENABLED=1 go test -tags "fts5,pgtest" -v ./internal/postgres/... ./internal/activity/... -count=1
 
 # Start test SSH container
 ssh-up:
--- a/cmd/agentsview/activity.go
+++ b/cmd/agentsview/activity.go
@@ -1,0 +1,239 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"text/tabwriter"
+	"time"
+
+	"go.kenn.io/agentsview/internal/activity"
+	"go.kenn.io/agentsview/internal/db"
+)
+
+// ActivityReportConfig holds the flags for `agentsview activity report`.
+type ActivityReportConfig struct {
+	Preset   string
+	Date     string
+	From     string
+	To       string
+	Timezone string
+	Bucket   string
+	Project  string
+	Agent    string
+	Machine  string
+	JSON     bool
+	NoSync   bool
+	Offline  bool
+}
+
+// runActivityReport syncs, resolves the range, runs the report, and prints it.
+func runActivityReport(cfg ActivityReportConfig) {
+	database, appCfg := openUsageDB()
+	defer database.Close()
+
+	ensureFreshData(appCfg, database, cfg.NoSync)
+
+	r, err := resolveActivityReportPriced(cfg, database)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
+	}
+
+	if cfg.JSON {
+		enc := json.NewEncoder(os.Stdout)
+		enc.SetIndent("", "  ")
+		if err := enc.Encode(r); err != nil {
+			fmt.Fprintf(os.Stderr, "error: %v\n", err)
+			os.Exit(1)
+		}
+		return
+	}
+
+	printActivityReport(r)
+}
+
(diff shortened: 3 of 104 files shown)


mjacobs (commented):
Really nice PR — built it and clicked through on a real archive; it's solid and demos beautifully. The pure aggregator + 3-backend split is carefully done, and the two follow-up fixes (usage dedup ordering, zero-cost primary model) are exactly the right cross-backend instincts. No blockers from me — just polish, a couple of parity/test things worth a look, and a few "confirm intent" questions.

**Worth a look (parity / behavior)**
- **DuckDB session Title diverges from SQLite/PG.** `internal/duckdb/activityreport.go:88` uses a bare `COALESCE(display_name, session_name, first_message, NULLIF(project,''), id)`, while SQLite (`db/activityreport.go:89-90`) and PG (`postgres/activityreport.go:84`) wrap the name columns in `NULLIF(...,'')`. Since those columns can hold `''` (the NULLIF guards in the other two backends exist for exactly that), a session with an empty `display_name` gets a different — possibly empty — title in DuckDB. Cosmetic, but a real parity gap; trivial fix is to mirror the shared `NULLIF(COALESCE(...),'')` shape.
- **Usage upper-bound operator differs:** SQLite `u.ts < ?` (`db/activityreport.go:220`) vs PG/DuckDB `u.ts <= ?` (`postgres:222`, `duckdb:368`). Benign today (it's the padded prefilter, and `dedupUsage` re-clips to the exact half-open range), but the operators should match or get a one-line "intentional, masked downstream" note.
- **Automated/interactive split is computed + serialized but the UI renders flat bars.** The Go comments say "rendered as a stacked bar," but `ConcurrencyTimeline.svelte` / `Breakdowns.svelte` never read the split fields. I see the PR notes already defer the per-segment viz — so this is just: consider softening those present-tense comments (or trimming the dormant fields) so the contract matches what v1 actually shows.
- **`SummaryCards` "Agent-minutes" renders a raw float** (`134.226` on my data) via `fmtInt` (no rounding), while `Breakdowns` rounds the same metric. One-line `Math.round`.

**Confirm intent (likely deliberate)**
- Single-day insights skip the new activity summary via the `req.DateTo > req.DateFrom` gate (`huma_routes_insights.go`) — intended? `daily_activity` insights are often single-day, and the validator permits `DateTo == DateFrom`.
- New `/ci-worktrees/` parser marker (`internal/parser/project.go:173`) is a bare segment, unlike its siblings which are anchored to a tool dir (`/.codex/worktrees/`, etc.) — it'll match any path containing a `ci-worktrees` dir. Anchor it, or confirm collisions are fine.
- `activityRangeSummary`'s comment says it matches "BuildPrompt's session set," but the two use different date/timezone semantics (half-open UTC window w/ `ended_at` fallback vs `ListSessions`' calendar-date BETWEEN on start date). Low stakes (LLM input) — reword or align.

**Test hardening (nits, not blocking)**
- The 3-way parity test is `//go:build pgtest`-gated and `t.Skip`s without `TEST_PG_URL`, so a local `make test` gets no cross-backend parity coverage (CI does run it). Its fixture also never exercises the `usage_events` / shared-dedup-key / same-second / zero-cost paths that the dedup fix touched, and there's no DuckDB-local dedup test. Extending `parityFixture` with those cases is the highest-value hardening here.

For what it's worth, I also chased and **cleared** a few things that turned out to be non-issues: the unparseable-`ts` zero-time sort fallback (dedup drops those rows before they claim a key), full-tie usage ordering (the accumulations are commutative), and the two `popstate` handlers (module-eval order guarantees the router registers first). Nice work overall.

mariusvniekerk:
Sample e2e fixture dashboard screenshots for the Activity report.

Overview of the populated May 22 week:



Token overlay with the Fri 06:00-07:00 concurrency slot selected, filtering the session table:

wesm:
This is getting close. Will just address roborev issues and one last item (configuring the insights agent from the Activity page) and then will merge

PR #731, opened 2026-06-18
feat(sync): local-first multi-machine artifact sync

### LINKED ISSUE #692: Local-first multi-machine sync
Progress branch: https://github.com/maphew/agentsview/tree/docs/local-first-multi-machine-sync

# Local-First Multi-Machine Sync: the Artifact Ledger

| | |
| ------- | ----------------------------------------------------------- |
| Status | Draft proposal |
| Date | 2026-06-12 |
| Authors | maphew, with Claude Fable 5 (multi-agent research workflow) |
| Inputs | `local-first-sync-research/` (audits, research, critique) |

## Summary

Make agentsview local-first with serverless multi-machine sync, in the
mold of fossil-scm: every machine holds the complete archive, sync is an
idempotent exchange of immutable artifacts over any dumb transport
(Syncthing folder, S3 bucket, or another agentsview instance over HTTP),
and no machine is architecturally privileged. No CRDT library is needed:
session content is single-writer-per-machine and append-mostly (a
grow-only set), and the thin layer of user-mutable metadata (renames,
trash, stars, pins) is handled by a fossil-ticket-style append-only
change log replayed deterministically with hybrid logical clocks.

SQLite remains exactly what the project already declares it to be: a
local, rebuildable derivation. The live database file never crosses the
wire. PostgreSQL push survives as an optional aggregation peer, demoted
from coordination point.

## Motivation

Today a user with a laptop and a desktop has three partial options, each
with a mandatory hub or a one-way arrow:

- `pg push` / `pg serve`: requires an always-on PostgreSQL server.
- `agentsview sync --host` (SSH pull): serverless and proven, but
  pull-only, manual/cron-only, re-downloads a full tar of every agent
  dir each run, only covers file-based agents, and never propagates
  user metadata (a rename or star made on one machine is invisible
  everywhere else).
- `duckdb push` / `quack serve`: a read mirror, beta, no FTS.

Upstream demand for something better is visible: issue 572 ("How to
synchronize data across multiple macOS devices"), issue 412 (periodic
SSH remote sync), issue 517 (multiple named pg targets), issue 484
(stars/pins in pg serve), issue 332 (pg push machine attribution), and
issue 655 (pg push same-id collision ping-pong).

### Goals

1. Every machine ends up with the full archive: all sessions from all
   machines, queryable and searchable locally (FTS5 intact).
2. User curation converges: renames, trash/restore, stars, pins made on
   any machine reach all machines.
3. No mandatory server. Any always-on peer, NAS share, or object-store
   bucket improves availability, but none is required by the
   architecture (fossil's "central server is a social convention").
4. Mixed app versions keep syncing: machine A on a newer release must
   interoperate with machine B on an older one.
5. The existing single-machine experience is unchanged when sync is not
   configured.

### Non-goals

- Real-time collaborative editing. Convergence latency is transport
  latency plus sync cadence (seconds to minutes).
- Partial/selective sync, subscriptions, or multi-tenant sharing. The
  trust model is a fully mutually trusted personal fleet (see Trust).
- Replacing `pg push`/`pg serve`; they remain as an optional hub.

## Why no CRDT engine

The data shape decides this. Audit of every table and column
(`local-first-sync-research/01-codebase-audits.md`) shows two classes:

1. **Bulk session content** (sessions, messages, tool_calls,
   tool_result_events, usage_events, secret_findings): derived
   deterministically from session files, created by exactly one
   machine, append-mostly. Merging two machines' archives is set-union
   of records that cannot conflict — the degenerate "grow-only set"
   CRDT that needs no library. Fossil's own docs describe its artifact
   bag the same way.
2. **User-mutable metadata** (sessions.display_name,
   sessions.deleted_at, starred_sessions, pinned_messages + notes,
   excluded_sessions tombstones, worktree_project_mappings): the only
   genuinely multi-writer data, edited rarely and by one human. An
   append-only log of timestamped change events with last-writer-wins
   replay — fossil's ticket model verbatim — is sufficient, auditable,
   and deterministic.

General-purpose CRDT machinery would solve a problem this data does not
have, while charging real costs (below).

## Alternatives considered and rejected

Full sourcing in `local-first-sync-research/03-technology-research.md`.
State of the ecosystem as of mid-2026:

- **Automerge**: `automerge-go` is effectively unmaintained (last
  commit Oct 2024, no tagged release, wraps a pre-3.0 core via cgo, no
  transport layer). Even Automerge 3 requires full in-memory document
  loads; the maintainers scope documents as "units of collaboration"
  and their own bulk-data research (sedimentree) moves large content
  out of the CRDT into content-addressed blobs — i.e. toward this
  design. Rejected for bulk and for metadata.
- **cr-sqlite**: upstream dormant since v0.16.3 (Jan 2024); the only
  maintained lineage is Fly.io's purpose-built fork for Corrosion. CRR
  constraints collide with this schema head-on: FTS5 virtual tables
  cannot be CRRs, the messages rowid PK and external-content FTS
  linkage break, triggers/CASCADE FKs are restricted, `__crsql_clock`
  shadow tables bloat a multi-GB DB, and Fly documented an ALTER TABLE
  metadata-backfill storm — this repo alters the sessions table
  routinely. Rejected.
- **SQLite session extension** (changesets): maintained forever as part
  of SQLite and accessible from Go via zombiezen/modernc (not via
  mattn/go-sqlite3 — issue 825 there, open since 2020) — but the binary
  changeset format is coupled to table column count, so every release
  that adds a sessions column (frequent here; dataVersion is at 36)
  bricks mixed-version sync. NDJSON's ignore-unknown-fields tolerance
  gives version skew handling for free. Rejected; its HLC/two-tier
  semantics are adopted, its codec is not.
- **Whole-DB replication** (Litestream v0.5 read replicas,
  sqlite3_rsync): healthy tools, but they produce N separate replica
  DBs. The entire Store interface, UI, and analytics assume one
  queryable DB; cross-replica fan-out would touch everything. Also
  one-way by design. Rejected as the architecture; fine as a backup
  strategy alongside.
- **Raw-session-file mirror over Syncthing** (sync the agent dirs
  themselves, let each machine parse everything): fastest to build and
  the best version-skew story, but structurally blind to non-file
  sessions — at least 7 agents in the registry are `FileBased:false`,
  plus uploads, claude.ai/ChatGPT imports, SSH-pulled sessions, and
  orphan-preserved sessions whose files are gone. It would also
  file-copy Zed's live `threads.db` SQLite database, the exact
  corruption hazard this design exists to avoid. Rejected as the sync
  unit; raw files return as optional fallback artifacts (see
  Invariants).
- **Server-light engines** (ElectricSQL, PowerSync, Evolu, Jazz,
  Ditto, Turso/libSQL embedded replicas, Marmot v2): all require a
  sync service, an always-on cluster, or have no Go story. Rejected.

### Overview

Each install maintains a write-once, content-addressed artifact store
alongside (never inside) the SQLite DB:

```text
$AGENTSVIEW_DATA_DIR/artifacts/<origin>/
  checkpoints/cp-<seq>.json        append-only numbered index files
  manifests/<hash>.json.zst        session manifests
  segments/<hash>.ndjson.zst       message segments
  meta/<hlc>-<hash>.json           user-edit change feed
  raw/<hash>                       optional raw source file fallback
```

A machine writes only under its own origin prefix (single-writer per
prefix means transports cannot conflict). Sync between any two stores —
or between a store and a folder/bucket/peer — is idempotent set-union
of immutable files. Ingestion derives the local SQLite rows from
foreign artifacts through the existing write paths.

```text
machine A                                            machine B
sync engine -> SQLite -> exporter -> artifacts/A --\
                                                    >-- transport --
ingester <- artifacts/B <---------------------------/   (folder/S3/
    |                                                    HTTP peer)
    v
SQLite (A + B merged, FTS5 maintained by normal triggers)
```

### Origin identity

Each install generates and persists an origin ID once: configured
machine name (default `os.Hostname()`, reusing the validation that
rejects the `local` sentinel, `internal/config/config.go`,
`internal/postgres/sync.go`) plus a short random suffix, e.g.
`thinkpad-x9k2`. The suffix survives hostname changes and distinguishes
restored/cloned machines; persistence copies the `EnsureAuthToken`
pattern (`internal/config/config.go`).

Global session identity is `(origin, native_session_id)`. Locally
produced rows are untouched (bare IDs, `machine='local'`). Foreign
sessions are stored as `id = origin + "~" + nativeID`, `machine =
origin` — byte-for-byte the proven SSH remote-sync convention
(`EngineConfig.IDPrefix`/`Machine` in `internal/ssh/sync.go`,
`applyRemoteRewrites` in `internal/sync/engine.go`, `StripHostPrefix`
in `internal/parser/types.go`), which every read path, the UI, and
`GetMachines` already render correctly. This avoids composite-PK
surgery across SQLite/PG/DuckDB under the Backend Parity rule.

### Artifact kinds

1. **Message segment**: canonical NDJSON (zstd) of N consecutive parsed
   messages keyed by natural coordinates — ordinal, source_uuid, role,
   content, tool_calls by (ordinal, call_index), tool_result_events by
   event_index, token fields. Natural-coordinate keying is already the
   schema's cross-copy convention (secret_findings; the orphan copier's
   (session_id, ordinal) joins). Message rowids are explicitly unstable
   (`nextMessageIDTx`) and never appear in artifacts.
2. **Session manifest**: small JSON with the parser-derived session
   header (the same field set `sessionPushFingerprint` enumerates), an
   ordered list of segment hashes, inline usage_events, the producer's
   data_version, and a generation counter. A newer manifest for the
   same session supersedes older ones, ordered by (data_version,
   generation). Steady-state appends emit one tail segment plus a new
   manifest reusing prior segment hashes. Superseded manifests/segments
   become unreferenced and GC-able after a grace window.
3. **Meta change event**: tiny JSON
   `{v, hlc, origin, session_gid, op, value}` with op in {rename,
   soft_delete, restore, star, unstar, pin, unpin, purge}; pins anchor
   by source_uuid with ordinal fallback (the existing
   `savePinsTx`/`restorePinsTx` logic). Append-only forever; the full
   edit history is retained.
4. **Checkpoint**: `cp-<seq>.json` mapping session_gid to current
   manifest hash plus the meta-feed high-water mark. Append-only
   numbered files keep the store fully write-once; discovery of changes
   is O(changed), not O(store).
5. **Raw source fallback** (optional, on by default for file-based
   agents): the original session file stored as a content-addressed
   blob and referenced from the manifest. See Invariants for why.

### Export

Export reads from the DB, not from source files, so non-file agents,
uploads, imports, SSH-pulled sessions, and orphan-preserved sessions
all publish. After each successful session write, the session is queued
and debounced by reusing the existing pg-watch loop: the artifact
exporter implements the same small target interface
(`cmd/agentsview/pg_watch.go`), so `agentsview sync --watch` is the
existing daemon with a different sink. Change detection reuses the
fingerprint-skip discipline (`sessionPushFingerprint` plus per-session
last-exported-manifest state, modeled on `pg_sync_state`). Export is
scoped to machine-owned rows; `machine='local'` is rewritten to the
origin ID at export time. Uploads (which default to `machine='remote'`)
are explicitly included — both prior designs fumbled this.

### Ingestion

A new `internal/artifact` importer per foreign origin: read the latest
checkpoint, diff manifest hashes against an `artifact_sync_state` table
(modeled on `pg_sync_state`), fetch and hash-verify missing segments,
assemble `db.Session` plus `[]db.Message`, apply the `origin~` prefix,
and write through the existing paths (`UpsertSession`,
`ReplaceSessionMessages`/`WriteSessionBatchAtomic`). That single
decision inherits, for free: FTS5 maintenance via the normal triggers
(including the bulk trigger-swap fast path), excluded/trashed tombstone
rejection, pin re-attachment by source_uuid, and stats triggers. The
importer then replays new meta events in HLC order and fires the SSE
broadcaster (closing the existing gap where non-engine writes never
emit `data_changed`).

Manifests that reference segments not yet delivered are recorded as
phantoms (fossil's term) and retried on the next pass, tolerating
arbitrary delivery order from dumb transports.

### Metadata ledger

Every user-mutation handler (rename, soft-delete/restore/permanent
delete, star, pin) additionally appends one meta event to the
machine's own feed. Replay is ordered by (HLC, artifact-hash tiebreak)
— a data-intrinsic ordering key, so every node derives identical state
from identical artifact sets regardless of local clocks. The HLC is
persisted across restarts, monotonic per node, with a bounded-drift
clamp (the Actual Budget pattern). Per-field last-writer-wins; when two
origins write the same field within clock-skew distance, the losing
value is appended to a `meta_conflicts` log and the UI shows a fork
badge with both values — converge automatically, never silently lose
(fossil's lesson). Applied events go through the existing DB mutators
in a suppress-re-export mode so no echo loops arise.

### Deletes

- Soft delete / restore: ordinary meta events; `deleted_at` converges.
- Permanent delete (`purge`): per-event opt-in only ("delete
  everywhere" confirmation). It propagates an `excluded_sessions`
  tombstone, which `UpsertSession` already enforces against
  resurrection, and peers locally shun the session's bulk artifacts.
  Default remains today's semantics: EmptyTrash is local-only.
- Checkpoint absence is never deletion (see Invariants).

### Transports

One verb, three target shapes, all the same set-union:

1. **Folder** — `agentsview sync /path/to/share` (Syncthing, Dropbox,
   NFS, rclone mount). Safe for dumb file sync because every file is
   immutable, written temp+rename, and single-writer-per-prefix. The
   live SQLite DB never crosses the wire — the documented corruption
   class (SQLite's "How To Corrupt", Zotero's KB, Syncthing forums)
   does not apply to write-once artifact files.
2. **HTTP peer** — `agentsview sync https://desktop:8080`. Four routes
   on the existing embedded server behind the existing Bearer-token
   middleware: list origins, get checkpoint, get artifact by hash, post
   artifact (hash-verified, write-once, idempotent). Stateless and
   resumable; fossil's igot/gimme reduced to HTTP GETs because
   content-addressing makes "have" a stat call. Any running agentsview
   is a rendezvous, like `fossil ui`.
3. **Object storage** — same layout under an S3/B2 prefix; rclone
   against the folder shape covers it until native support lands.

### Interaction with resync and dataVersion

The artifact store lives outside the DB file, so `ResyncAll`'s atomic
swap does not touch it; `artifact_sync_state` is carried across the
swap alongside the existing metadata copy (`CopySessionMetadataFrom`).
After a parser-version resync, changed sessions re-export with bumped
data_version manifests — the same "force full push after resync" rule
pg push uses. Segments whose canonical content is unchanged keep their
hashes; a parser change touching a common message field genuinely
re-ships content, which the raw-file fallback hedges (peers may
re-derive locally instead of re-downloading).

Mixed versions: bundles are NDJSON with ignore-unknown-fields and
skip-unknown-ops rules plus an explicit format version, so an older
reader skips fields it does not know and a newer reader tolerates their
absence. Each machine's own parser and dataVersion govern only its own
DB.

### What pg push becomes

Short term: unchanged (ingested peer sessions are ordinary rows, and
their machine column carries true origin). Medium term: extract the
small SessionSink interface latent in push.go's orchestration (which
contains no SQL); PG becomes one sink, the artifact exporter another.
Long term: PostgreSQL is an optional aggregation/analytics peer.
Prerequisite fixes regardless of this design: machine-scoped export
(upstream issues 332 and 655).

## Invariants (pinned before any code)

1. **Canonical serialization is a forever-contract.** Sorted keys,
   fixed number formatting, explicit format version; golden tests
   enforce byte-stability. Any silent change re-hashes every segment
   and triggers a fleet-wide reship. The raw-source fallback artifact
   exists so that even a broken contract degrades to local re-derive,
   not re-download.
2. **Checkpoint absence is never deletion.** Tombstone events are the
   only delete mechanism. A session vanishing from an origin's
   checkpoint (local EmptyTrash, export bug, truncation) must not
   propagate removal.
3. **`ErrSessionTrashed`/`ErrSessionExcluded` on import means retry
   later, never advance the watermark.** Meta events are tiny and
   segments large, so a soft-delete routinely arrives before content;
   if the watermark advanced anyway, a later restore would strand stale
   content with nothing to trigger a re-fetch.
4. **Single-writer-per-prefix is the only write rule on shared
   transports.** Colliding origin IDs (cloned machine, restored backup)
   must be detected (checkpoint seq conflict) and surfaced loudly, not
   merged.
5. **The live SQLite file never crosses the wire.** Documentation must
   say this explicitly and warn against syncing the data dir.

## Trust model

A fully mutually trusted personal fleet. Folder transports have no
per-writer identity (prefix discipline is convention; Syncthing has no
per-subdir ACL), and HTTP mode is one shared symmetric Bearer token —
any peer can forge any origin's metadata. That is acceptable for one
person's machines and must be documented as exactly that. Per-peer
tokens are the minimum follow-up before any sharing story; origin
signatures are the eventual answer.

Practical availability note, stated plainly in docs: two
intermittently-on laptops sync only when both are online. A NAS folder,
S3 bucket, or any always-on peer is the practical rendezvous — by
social convention, not architecture, exactly as in fossil.

## Migration

Fully additive. Upgrade generates an origin ID; no rewrite of existing
rows. `agentsview sync --init` backfills artifacts for the whole
existing DB (including orphans) and seeds the meta feed from current
display_name/deleted_at/stars/pins timestamped with local_modified_at.
Machines without sync configured behave exactly as today. New tables
(`artifact_sync_state`, `meta_clock`/conflicts) arrive via the existing
idempotent migration pattern; no dataVersion bump, no resync.

## Phasing

1. **Prereq fixes (days)**: machine-scoped pg push export; preserve
   per-session machine at push time (upstream issues 332, 655). Real
   bugs regardless of this design.
2. **Phase 1 (2-4 weeks)**: artifact store, canonical serializer with
   golden tests, exporter, folder-transport set-union, importer,
   `sync --init`. Delivers the headline want — every machine sees all
   sessions — read-only, over Syncthing/Dropbox/NFS, no schema surgery.
3. **Phase 2 (2-3 weeks)**: HLC, meta ledger, deterministic replay,
   fork badges, purge with confirm UX. Delivers converging curation.
4. **Phase 3 (1-2 weeks)**: HTTP peer endpoints behind existing auth,
   `sync --watch` via the pg-watch loop, peers page in the UI.
5. **Phase 4 (ongoing)**: GC of superseded artifacts, native S3 target,
   SessionSink refactor of pg push, two-instance E2E harness.

Estimated 6-10 weeks total for one developer; each phase ships value
alone.

## Risks

- Canonical-serialization drift (highest variance; mitigated by golden
  tests, format version, raw fallback).
- HLC/LWW edge cases: skewed clocks, restart persistence, tie
  determinism — table-driven tests required; replay must be idempotent
  under any feed permutation.
- Storage growth: a third local copy of the corpus (source files + DB +
  compressed artifacts) plus peers'. zstd gives 5-10x on JSONL; GC of
  superseded bulk artifacts is mandatory, with a grace window against
  slow peers.
- Meta feed file count: one small file per edit grows forever;
  personal-scale fine, needs a batching/compaction story eventually.
- FTS5 initial ingest at multi-GB scale re-tokenizes every peer's
  corpus; use the existing Drop/Rebuild bulk path for first ingest.
- N-times row counts make the machine filter load-bearing UX for the
  sidebar and analytics.
- Scope creep: this deliberately stops at set-union plus LWW. Partial
  sync, subscriptions, or content merging would erode the simplicity
  that makes it safe to own (~3-4k LOC).

## Open questions

- Should the raw-source fallback be mandatory rather than optional for
  file-based agents? (Cost: storage; benefit: version-skew immunity.)
- Per-agent or per-project export excludes (selective publish, fossil
  private-branch analog) in v1 or later?
- Dotfile-synced agent dirs produce visible duplicates under two
  origins (today: silent same-id merge). Document "pick one transport
  per agent dir", or attempt content-hash coalescing in the UI later?
- Does `worktree_project_mappings` (already machine-keyed) ride the
  meta ledger or stay local-only?

## Related upstream issues

- 332 — pg push overwrites original machine name on remote-synced
  sessions (open).
- 655 — pg push: sessions.id sole PG PK; same-id pushes from two
  machines silently merge and ping-pong (filed from this work).
- 412 — periodic SSH remote sync from serve (open feature request).
- 517 — multiple named pg targets (open feature request).
- 484 — stars/pins in pg serve (closed; metadata demand signal).
- 572 — multi-machine dashboard question (closed; demand signal).

## References

Full research underlying every claim here, including sources and
line-level code citations, lives in `local-first-sync-research/`:
codebase audits (01), the SSH remote-sync deep audit (02), technology
research with sources (03), the three competing design proposals (04),
and the adversarial critique that selected and hardened this design
(05).

_gpt-5.5 on behalf of maphew_.

### LINKED ISSUE #1034: artifact-sync: unparseable HLC accepted into metadata replay and wins LWW permanently
Found during hands-on multi-machine QA of the artifact-sync branch (#731). This is the same class as the open roborev finding on `internal/artifact/replay.go:147`, and it reproduces end-to-end with a real forged artifact.

### Summary

A metadata event whose HLC string does not parse is still accepted into replay, and because last-writer-wins ordering is a raw string comparison of `order_key`, an out-of-range HLC sorts above every real timestamp and **wins permanently** — no legitimate rename/star/delete/pin from any machine can ever displace it.

### Root cause

- `validateMetadataArtifactEvent` (`internal/artifact/replay.go:147-173`) checks that `event.HLC` equals the filename HLC prefix, matches the origin, has a non-empty session GID, and has a supported version. It never checks that the HLC actually parses.
- `internal/artifact/replay.go:47` only advances the clock `if perr == nil` (`ParseHLCTimestamp` succeeded) and otherwise silently falls through — an unparseable HLC skips clock handling but the event is still validated and applied.
- LWW ordering is a raw string compare: `internal/db/metadata_replay.go:330` does `if hasCurrent && ev.OrderKey <= current.OrderKey`, where `order_key = <hlc>-<hash>`. A forged HLC like `9999-99-99T999999.000000000Z` sorts lexicographically above every real `2026-...` key, so it wins and can never be out-ordered.

### Reproduction

1. Two converged instances sharing a folder; one real session owned by origin A.
2. Forge `<share>/<evil-origin>/meta/9999-99-99T999999.000000000Z-<sha256>.json` where the filename hash is the sha256 of the literal file bytes, and the body is a well-formed rename event with `hlc: "9999-99-99T999999.000000000Z"` and `value: {display_name: "PWNED"}`.
3. `agentsview sync <share>` on the other instance → `imported 1 metadata events`; the session's `display_name` becomes `PWNED`, and `metadata_replay_state.order_key` stores the unparseable key as the permanent winner. No later legitimate edit changes it.

### Impact

Even under the "fully trusted personal fleet" v1 threat model this is not purely an adversary case: a corrupt/truncated metadata event, or a future format skew, produces the same un-out-orderable poison for `display_name`, `starred`, `deleted_at`, or pin state.

### Suggested fix

Call `ParseHLCTimestamp` inside `validateMetadataArtifactEvent` (or `readMetadataArtifact`) and skip/quarantine events whose HLC does not parse — mirroring the malformed-payload quarantine that already exists a few lines up at `replay.go:65-77`.

### LINKED ISSUE #1035: artifact-sync: corrupt segment/manifest hash mismatch aborts sync permanently and spreads fleet-wide
Found during hands-on multi-machine QA of the artifact-sync branch (#731).

### Summary

A single hash-mismatched (corrupt) segment or manifest in a shared artifact folder aborts the entire sync with a fatal error, permanently, and propagates to every machine that touches the share — turning one bad file into a fleet-wide sync outage. A *missing* artifact is handled gracefully (skip-and-continue); a *corrupt* one is not.

### Root cause

- A missing segment returns `errIncompleteArtifact` (`internal/artifact/sync.go:1043`), which `importOrigin` swallows to skip-and-continue (`sync.go:715`, `sync.go:728`).
- A hash-mismatched segment instead returns a plain error (`sync.go:1050`), and a corrupt manifest does the same (`sync.go:1026`). Neither is `errIncompleteArtifact`/`errFutureArtifactVersion`, so it propagates: `importOrigin` → `importDetailed` → `sync` all return the error, the CLI prints `fatal: artifact sync: segment <hash> hash mismatch`, and exits 1.

Two aggravating factors:

1. **The corrupt file is mirrored locally before validation.** `folderTransport.Exchange` runs `CopyUnion(target, localRoot)` (`internal/artifact/transport.go:45`) *before* import validation, so the bad file lands in the local artifact store and every subsequent sync re-aborts forever.
2. **It spreads.** Once the corrupt file is in the shared folder, the next machine to sync copies it into *its* local store via the same `CopyUnion` and also begins exiting 1. In testing, after machine B pulled the poisoned share, machine A's own sync started aborting too.

### Reproduction

1. Two converged instances sharing a folder.
2. Under a fake origin dir in the share, build a valid checkpoint → manifest chain whose referenced segment file (`<hash>.ndjson.zst`) decompresses to tampered content (so the recomputed sha256 of the decompressed bytes no longer matches the filename hash).
3. `agentsview sync <share>` → `fatal: artifact sync: segment <hash> hash mismatch: got <other>` (exit 1). Re-run: same fatal error every time. Sync the other machine against the same share: it now also exits 1.

### Impact

Durability, unlike the adversary case, is squarely inside the v1 threat model: bit rot, an interrupted Dropbox/Syncthing write, or a half-flushed S3 upload all produce a hash mismatch and permanently break artifact sync for the whole fleet.

### Suggested fix

Treat a hash mismatch like a missing/incomplete artifact — skip that origin/manifest and continue rather than aborting the whole run — and ideally quarantine the offending file (or refuse to mirror it via `CopyUnion`) so it is not re-copied into local stores and re-encountered on every sync.

maphew:
Implements local-first multi-machine artifact sync from #692. Each machine retains its SQLite archive and exchanges immutable, content-addressed session artifacts plus a hybrid logical clock metadata ledger through shared folders, HTTP peers, or S3-compatible object stores.

The sync path preserves origin-aware identities through SQLite and optional PostgreSQL, treats imported artifacts as untrusted input, regenerates corrupt local objects from SQLite, and safely retries partially published metadata. Lifecycle operations serialize local trash, restore, and purge mutations with peer metadata replay, so pre-publication failures compensate safely and purge artifacts become durable before local deletion.

Transport security matches the project threat model: non-loopback HTTP requires explicit --allow-insecure consent, authenticated redirects are rejected, custom S3 endpoints require TLS outside loopback unless explicitly opted in, and insecure object-store cleanup cannot delete remote objects.

Automatic garbage collection prunes both the local store and target after folder sync. HTTP and S3 targets retain superseded history until transport-aware remote pruning is available, preventing remote-only artifacts from entering a recurring download-and-delete loop.

The branch also adds conflict and peer visibility, conservative manual garbage collection, and operational documentation. Legacy PostgreSQL identities converge without duplicating sessions or losing local curation, relationships, source state, or soft-delete state.

Closes #692.
Closes #1034.
Closes #1035.


maphew:
Feedback is welcome. Still in draft mode since work and testing to this point has been completely agent driven, a combination of GPT-5.5 and Claude 4.8. Next up is manually trying various distributed machine scenarios and seeing how well any of this works in practice. Assuming the idea eventually proves out, I'm happy to split into smaller manageable PR.

maphew:
Thanks for the review. Both findings were valid and are addressed in 28048707 and f252c35f.

**High — Windows-invalid `:` in metadata filenames.** Confirmed: filenames are `OrderingKey(hash)` = `HLCTimestamp.String()` + hash, and `String()` rendered the wall time with `15:04:05` colons. Since the DB mutation runs before the metadata write (e.g. `humaStarSession`), star/pin/rename/delete would mutate the DB and then fail the file write on Windows, diverging the DB and the ledger. Fix: dropped the `:` separators from the HLC wall layout (`2006-01-02T150405.000000000Z`). Keeping a single representation everywhere — content, filename, and the persisted `orderKey` cursor — preserves lexicographic ordering, parse round-trip, and the filename-equals-content integrity check with no separate encode/decode. Updated the canonical-format goldens, added an assertion that the filename has no Windows-invalid characters, and a parse-rejection case for the old colon form.

Note this changes the canonical on-disk HLC string (`event.HLC` no longer contains colons). That is fine here since the format is pre-release, but any artifacts written by an earlier draft build would be stale-format.

**Medium — divergent origin sources.** Confirmed: `runServe` never populated `cfg.ArtifactOriginID` before `server.New`, so an empty config origin made the recorder fall back to a random DB-state origin while CLI folder sync minted a machine-named origin in `config.toml`. Fix: ensure the config origin (authoritative) at serve startup before the recorder exists, and reconcile the DB sync-state to it via a new idempotent `artifact.AdoptOrigin`. The recorder, peer import, and folder sync now converge on one persisted origin. Added `AdoptOrigin` unit tests (persist, idempotent, overwrite-divergent, reject-invalid).

`go vet`, `gofmt`, and the `artifact`/`server`/`config`/`e2e` Go suites pass.

_Claude Opus 4.8 reasoning-medium on behalf of maphew_

maphew:
Thanks again. All three findings were valid and are addressed in 055f3b3c.

**High — local metadata events missing from the replay register (`sync.go:523`).** Confirmed: import skips the local origin and `MetadataRecorder.Append` only wrote the artifact file, so locally-originated edits never landed in `metadata_replay_state`/`metadata_applied_events`. A later peer event for the same field saw no current winner and could overwrite a newer local rename/star/delete/pin without recording a conflict. Fix: `Append` now records the event through a new `db.RecordLocalMetadataProjection`, which runs the same per-field LWW conflict/register/applied bookkeeping as replay but skips re-applying the mutation (the handler already applied it). `ApplyMetadataProjection` was refactored to share that path via an `applyMutation` flag. Added a test where a local rename with a higher HLC survives a later-imported lower-HLC peer rename and the peer is recorded as the conflict loser.

**Medium — remote HLCs not observed by the local clock (`replay.go:51`).** Confirmed. Replay now tracks the newest accepted remote HLC per origin and calls `HLCClock.Observe` once at the end, so the next local edit is causally ahead of imported peers. Advancing past the drift bound is best-effort and never fails an otherwise successful import. Import is routed through the recorder (`MetadataRecorder.Import`) so appends and imports share one clock instance rather than racing two clocks on the same persisted key. Added a test asserting a post-import local edit gets an HLC strictly after a peer HLC that was ahead of the local wall clock.

**Medium — one unavailable target aborted the rest of the origin (`replay.go:53`).** Confirmed. `ErrMetadataTargetUnavailable` now skips only that event (left unapplied, watermark not advanced) and replay continues with later events, which retries the deferred event on a subsequent run once its target exists. Added a test where an event for a missing session no longer blocks a later event for an existing session, and the deferred event applies after the session arrives.

`go vet`, `gofmt`, and the `artifact`/`db`/`server` Go suites pass.

_Claude Opus 4.8 reasoning-medium on behalf of maphew_

maphew:
Thanks. Both convergence gaps were valid and are fixed in 1d8d24c1 and 8cac9fff.

**Medium — usage-only sessions never exported (`sync.go:414`).** Confirmed: export paged through `ListSessions`, whose sidebar filter (`message_count > 0`) hid owned sessions that carry only usage events, so they never reached a checkpoint. Fix: export now enumerates owned, non-deleted sessions from a raw query (`db.ListOwnedSessionIDsForExport`) instead of the list API. Soft-deleted and foreign sessions stay excluded — deletes still propagate via metadata tombstone events, never checkpoint absence. Added a test that a zero-message session with a usage event lands in the checkpoint and imports into a peer with its usage events intact, plus one asserting deleted/foreign sessions stay out.

**Medium — bulk star emitted no metadata events (`huma_routes_starred.go:92`).** Confirmed: bulk star wrote the DB but appended nothing, so localStorage-migrated stars stayed local-only. Fix: `BulkStarSessions` now returns the IDs actually starred (sessions that exist), across the SQLite, PostgreSQL, and DuckDB backends to preserve parity, and the handler appends a `star` metadata event for each, matching single-session star. Stale IDs are still skipped and produce no event. Added a server test asserting one star event artifact per existing session and none for a missing id.

`go vet`, `gofmt`, and the `db`/`artifact`/`server`/`duckdb` Go suites pass; the PG curation pgtest compiles under the `pgtest` tag.

_Claude Opus 4.8 reasoning-medium on behalf of maphew_

maphew:
Thanks. Addressed in e77db3a9, 4110dce2, acbb7899, and 6ea6fa89.

**Medium — `--host` + artifact target silently ignored (`sync.go:118`).** Confirmed: `--host` dispatches to SSH remote sync and returns before artifact sync runs. Fix: `validateSyncConfig` now rejects the combination outright (ordered after the `--init`/`--watch` checks so their specific messages still win). Added accept/reject tests.

**Medium — unconditional S3 PUT violates write-once (`transport_s3.go:355`).** Confirmed: unlike the folder and HTTP transports, S3 blind-overwrote. Fix: uploads now send `If-None-Match: *`; on the 412 precondition failure the object is fetched and compared — identical content is an accepted duplicate, divergent content is a conflict, never a silent merge. The mock honors the conditional and a test covers both outcomes, and the MinIO integration test confirms a real server accepts it.

**Medium — `CopyUnion` deferred close (`sync.go:1415`).** This one is actually a false positive: the `defer from.Close()` was inside the per-entry `WalkDir` callback closure, so it ran after each file, not at the end of the traversal — no descriptor accumulation. That said, I removed the ambiguity entirely by switching to `os.ReadFile` (no manual open/defer), which reads cleaner regardless.

**Medium — remote events applied before the HLC advances (`replay.go:86`).** Confirmed and the most important one. Fix: replay now calls `Observe` on each remote event's HLC *before* applying it and defers any event whose wall time is beyond the drift bound, leaving it unapplied for a later run to retry once local wall time catches up. A future-skewed peer can no longer drag local state to a value a subsequent local edit couldn't out-order. Added a deferral test; the two-instance harness now orders concurrent edits with within-drift offsets.

`go vet`, `gofmt`, and the `artifact`/`db`/`server`/`duckdb`/`cmd` Go suites pass; the MinIO integration test (`make test-minio`) passes against real MinIO.

_Claude Opus 4.8 reasoning-medium on behalf of maphew_

wesm:
I will rebase this

wesm:
I'll continue to work a bit on this to see if I can get it into a state that I'm comfortable with

maphew:
@wesm thanks for taking a direct interest! It's appreciated. I'm largely away for this week plus a bit. I'm looking forward to exercising this first hand when I'm back : )

wesm:
I'll work on getting this landed after 0.35.0 is released, there is already a lot of stuff merged to main since the last release

wesm:
I still plan to work on this, bear with me. I'll work on getting it in after 0.37.0 releases

maphew:
thanks for the status update @wesm! I'm happy to spend time and tokens to help, just let me know what might be useful

maphew:
Finally got to the first-hand distributed testing I mentioned back in June. I built this branch and drove the binary through seven multi-instance scenarios by simulating separate machines on one host (isolated `HOME` + `AGENTSVIEW_DATA_DIR` + `CLAUDE_PROJECTS_DIR` per machine, a shared folder / live HTTP peer as the rendezvous). Full build + short test suite is green including `internal/artifact`.

**Core convergence works well.** Every scenario passed its main checks:

| Scenario | Transport | Result |
|---|---|---|
| Two-machine session + new-session propagation | folder | pass |
| Rename/star convergence, both directions | folder + server API | pass |
| Offline concurrent-rename LWW + conflict/fork badge | folder | pass |
| `--watch` continuous convergence + SIGINT final flush | folder | pass |
| Live HTTP peer + no-restart index refresh + auth | HTTP | pass |
| Soft-delete / restore / purge tombstone / GC | folder | pass |
| Forged/corrupt artifact robustness | folder | 2 bugs found |

Two things I'd flag before merge, both verified against the source and filed as separate issues:

- **#1034** — an unparseable HLC is accepted into metadata replay, and because LWW ordering is a raw string compare of `order_key = <hlc>-<hash>`, a forged `9999-...` HLC sorts above every real timestamp and wins permanently. This is the same class as the open roborev finding on `replay.go:147`; it now has an end-to-end repro.
- **#1035** — a corrupt (hash-mismatched) segment or manifest aborts the whole sync fatally, and because the folder transport mirrors files into the local store before validating, it re-aborts forever and propagates to every machine that touches the share. A *missing* artifact is handled gracefully; a *corrupt* one is not — so this is a durability bug (bit rot, interrupted Dropbox/S3 write), not only an adversary case.

Beyond those, a handful of non-blocking UX/docs rough edges I noticed (happy to file individually if useful):

- Sequential cross-machine edits get logged as "conflicts" and light the fork badge — e.g. renaming on machine B 38s *after* it imported A's rename still records a conflict on both machines. Convergence is correct; the docs' "close together" framing and the badge overstate it. Conflict rows are also per-machine/asymmetric, so `conflict_count` differs by machine and the machine whose edit lost may show nothing.
- `--watch` is silent after startup (no per-exchange/import/shutdown log lines), and peer artifacts in the share are only pulled on the `--interval` floor (default 15m), not on change — the watcher only watches local session dirs. Worth a sentence in `--help`/docs so users tune `--interval`.
- Plain `agentsview sync` prints `Sync complete: 0 sessions synced` on the very run that ingested new sessions (daemon ingests at startup, CLI reports the second pass), which reads as "nothing happened."
- Plain `sync` silently starts a daemon that then blocks `sync <folder>` with the write-lock error; nothing warns that `sync` itself started it.
- The HTTP peer `--token` is the peer's `config.toml` `auth_token`, but artifact-sync.md never says so; syncing without `--token` to an auth peer also reports "peer rejected the bearer token" when none was sent.
- Peer soft-delete uses import time rather than the event HLC for `deleted_at`, so trash-retention windows drift across machines.

Glad to open issues for any of the above, retest fixes, or run the same harness on real WSL↔Windows machines (the earlier HLC-filename fix suggests Windows paths are worth exercising for real). Just let me know what's most useful.

wesm:
Thanks for testing it out. I can just address these issues in the branch, I'll kick off a rebase now

wesm:
@maphew thanks for the distributed test harness — that scenario matrix is exactly the coverage this needed. Both of your blockers are fixed on the branch (head `06cb481`, CI green including a clean roborev pass), along with one item from your non-blocking list:

- **#1035 — corrupt artifacts abort sync fleet-wide.** Corrupt artifacts are now tolerated and healed instead of fatal. A hash-mismatched or undecodable segment/manifest/checkpoint/meta event is renamed to `<name>.corrupt` and skipped, so import continues with the remaining sessions and read paths fall back (import walks backward past a corrupt latest checkpoint, for example). Quarantine also stops the folder transport from propagating corruption: `CopyUnion` validates recognized artifacts by content before copying, skips corrupt sources, and repairs a corrupt destination from a valid local copy — a share with one bit-rotted file heals on the next exchange instead of poisoning every machine that touches it. HTTP/S3 exchanges skip invalid artifacts on both push and pull, and GC skips an origin it cannot fully read rather than collecting against a partial view. One asymmetry to know about: an object corrupt in place inside an S3 bucket is skipped with a log but not overwritten, because push only uploads names the remote is missing; folder shares do heal in place.
- **#1034 — forged HLC wins LWW permanently.** Metadata events with unparseable HLCs are rejected at all three doors: `WriteArtifact` (peer POST and local append), `CopyUnion` propagation, and replay, where a pre-existing bad event is skipped and marked applied so it cannot wedge its origin. Your forged `9999-…` repro is now a test case. Future format versions remain tolerated (deferred, not judged), so the upgrade path is unchanged.
- **`deleted_at` drift (your last bullet).** Peer soft-deletes now anchor `deleted_at` to the event's HLC wall time instead of import time, so trash-retention windows agree across machines.

Also addressed in the same batch: roborev's follow-up on the export side of the corrupt-checkpoint case — checkpoint sequence selection no longer parses the latest body; it derives the next sequence from checkpoint filenames (quarantined ones included), so a corrupt local checkpoint cannot block export and a sequence number that may already have been published is never reused (this one lands with the next push to the branch) — plus a stale-subagent-link repair in the PG push fast path and a Korean catalog sync.

For the rest of your list (conflict overstatement and asymmetric conflict rows, `--watch` silence plus interval-only peer pulls, the "0 sessions synced" message, the silent daemon/write-lock interaction, `--token` docs): yes, please file those individually — all real, none merge-blocking. A re-run of your scenario-7 harness against the current head would be very useful, and the real WSL↔Windows pass even more so given the HLC-filename history.

wesm:
I'm still working on this, there were a bunch of performance concerns. I am going to release 0.38.0 first and then see about this

wesm:
I am still tracking this and will return to it in the near future

maphew:
Thanks for the performance work, Wes. Since the 0.38 release is out, I can take the branch-maintenance pass: bring #731 up to current main, resolve the conflicts, address the two latest correctness findings, and post reproducible large-archive benchmark results. After that, would you prefer a staged PR series or to keep this together?

maphew:
Following up: I kept #731 untouched and split the assist into three draft side PRs, each targeting the feature branch in my fork:

- [Current-main catch-up](https://github.com/maphew/agentsview/pull/2) — resolves the current merge conflicts and preserves the artifact-sync behavior across the newer DB/CLI/frontend shapes.
- [Correctness fixes](https://github.com/maphew/agentsview/pull/4) — flushes pending watcher changes during shutdown and derives peer status from the exact checkpoint provenance.
- [Performance benchmarks](https://github.com/maphew/agentsview/pull/3) — adds reproducible export/import/no-op/incremental workloads plus the measured large-archive baseline.

All three are mergeable and green across CI and the benchmark gate. The documented 200-session × 80-message (~6.1 MB) run on an i9-13900 measured about 0.95s export, 0.87s import, 0.58s converged no-op, and 0.67s incremental sync.

Use, cherry-pick, or choose whatever is helpful. I will not push changes to #731.

wesm:
We've built a CAS system for msgvault that is in kenn-io/kit, I am working on generalizing that so that it can serve as the CAS layer for this PR, but it is going to take me some time, maybe another week. That will prevent creating an independent content-addressed storage system for this feature alone. Stay tuned

wesm:
https://github.com/kenn-io/docbank is now public so I'm looking at how we could base the artifact management on docbank to reduce code ownership for CAS in agentsview

PR #739, opened 2026-06-19
feat(session list): add --sort/--reverse (and ?order_by/?descending)

### LINKED ISSUE #729: Add --sort and --reverse/-r flags to `agentsview session list`
## Problem / motivation

`agentsview session list` already supports ~20 filter flags (`--project`, `--agent`, `--min-tool-failures`, `--has-secret`, `--outcome`, `--health-grade`, `--min-messages`, …), but the output order is hard-coded to "most-recently-active first" via the SQL `ORDER BY COALESCE(NULLIF(ended_at,''), NULLIF(started_at,''), created_at) DESC, id DESC` in `internal/db/sessions.go#buildSessionQuery`.

In practice the filters a user reaches for first — `--min-tool-failures`, `--has-secret`, `--health-grade=F`, `--outcome=failure`, `--min-messages` — are all oriented around finding *problem* or *notable* sessions, but the order forces them to scroll through everything newer first. With 100+ sessions in the local SQLite archive, the truly interesting matches get buried under routine work that just happened to finish.

## Proposed solution

Add `--sort <field>` and `--reverse`/`-r` flags to `agentsview session list`, mirroring the flag pattern already in `hf jobs ps` (proposed in huggingface_hub#4374) and `hf {models,datasets,spaces,repos} ls`:

- `--sort` accepts a small enum of column names that already exist on `sessions` and have the matching index in `internal/db/schema.sql`:
  - `recent` (default, current behavior — `ended_at DESC, id DESC`)
  - `started` (`started_at`, ascending)
  - `messages` (`message_count`)
  - `user-messages` (`user_message_count`)
  - `output-tokens` (`total_output_tokens`, requires `has_total_output_tokens=1`)
  - `peak-context` (`peak_context_tokens`, requires `has_peak_context_tokens=1`)
  - `failures` (`tool_failure_signal_count`)
  - `retries` (`tool_retry_count`)
  - `edit-churn` (`edit_churn_count`)
  - `compactions` (`compaction_count`)
  - `context-pressure` (`context_pressure_max`)
  - `health` (`health_score`, NULLs last)
  - `secrets` (`secret_leak_count`)
  - `id` (`id`)
- `--reverse`/`-r` flips the direction (the universal "flip sort order" convention across `ls -r`, `git log --reverse`, `docker logs --reverse`)
- The CLI flag threads through to a new `OrderBy` / `Descending` field on `service.ListFilter` and `db.SessionFilter`, then maps to a SQL fragment in `buildSessionQuery` — replacing the hard-coded `ORDER BY` clause
- `idx_sessions_message_count`, `idx_sessions_user_message_count`, etc. already exist (`internal/db/schema.sql:139,141`), so most sort orders won't need new indexes

The CLI list is currently documented as "the full set of HTTP query-param equivalents as CLI flags" (per `cmd/agentsview/session_list.go` ABOUTME), so the same `OrderBy`/`Descending` fields should also be exposed as `?order_by=...&descending=...` query params on the v1 list endpoint.

## Alternatives considered

- **`--sort=recent|messages|started|...` with a per-value default direction (e.g. `--sort=messages` is DESC, `--sort=started` is ASC)** — rejected; separating sort field from direction is more uniform and matches every modern CLI convention
- **Add a `--reverse` flag only without a `--sort` flag** — rejected; the current "most recent first" default already matches most users, so the gap is "give me a way to put X at the top", not "give me a way to flip the existing default"
- **Implement as a separate `agentsview session list --top-problems` preset** — rejected; with N possible sort axes, the preset model either explodes combinatorially or bakes in the maintainer's opinion of what "problems" means
- **Compute cost from `usage_events` and add a `cost` sort** — rejected for now; cost is a join/aggregate over `usage_events`, not a column on `sessions`, so it would need a different query path. If the demand is real we can add it as a follow-up, but the 14 in-row sort axes above cover the user need today
- **Add a SQL `ORDER BY` per existing filter (e.g. `--min-tool-failures` always sorts failures DESC)** — rejected; it locks in behavior the user might not want for that filter

## Use cases

- `agentsview session list --sort=failures --reverse --min-tool-failures 1` — surface the most broken sessions, fastest first
- `agentsview session list --sort=health --reverse --min-messages 50` — find low-health sessions with real substance
- `agentsview session list --sort=messages --reverse --limit 10` — top 10 longest sessions by message count
- `agentsview session list --sort=secrets --reverse --limit 5` — surface any session that has detected secret leaks
- `agentsview session list --sort=peak-context --reverse` — find sessions with the biggest context windows (proxy for expensive runs)
- `agentsview session list --sort=started --since 2026-06-01` — chronological review of this month's sessions

## Distinct from existing flags

- `--limit` (caps return count, doesn't reorder)
- `--cursor` (paginates a fixed order, doesn't reorder)
- `--outcome`, `--health-grade`, etc. (filter rows, don't order them)
- the HTTP `?sort=` (does not exist; the API endpoint in `internal/server/server.go` also has no order_by query param today)

## Implementation sketch

1. `service.ListFilter` + `db.SessionFilter`: add `OrderBy string` (column name) and `Descending bool` (default true for `recent`, false otherwise — or accept the small inconsistency)
2. `internal/db/sessions.go#buildSessionQuery`: replace the hard-coded `ORDER BY COALESCE(...) DESC, id DESC` with a switch on `f.OrderBy` that emits the matching column; append ` DESC` if `Descending` (or ` ASC` if not)
3. `cmd/agentsview/session_list.go`: add `--sort` (String) + `--reverse`/`-r` (Bool) flags, translate to filter, validate `OrderBy` against an allow-list
4. `internal/server/server.go` list handler: accept `?order_by=...&descending=...` query params (per the CLI/HTTP parity comment), validate against the same allow-list
5. Tests: add a few cases to `internal/db/sessions_test.go` covering each sort axis + cursor pagination under each sort, and to `cmd/agentsview/session_list_test.go` for the CLI flag wiring

The diff should be small (3-5 files, ~150-250 lines including tests), surgical, and backward-compatible (default `OrderBy="recent"` matches today's behavior).

mjacobs:
Adds `--sort <field>` and `--reverse`/`-r` to `agentsview session list`, with the matching `?order_by=…&descending=…` query params on `GET /api/v1/sessions`. Closes #729.

Sort keys: `recent` (default), `started`, `messages`, `user-messages`, `output-tokens`, `peak-context`, `failures`, `retries`, `edit-churn`, `compactions`, `context-pressure`, `health`, `secrets`, `id`. `recent` defaults to descending (today's behavior); the rest default to ascending. `--reverse` flips the key's default; on the HTTP side `descending` sets the direction absolutely (omitted → the key's default).

## How it works

Rather than special-casing each sort, this introduces a single sort registry (`internal/db/sort.go`) mapping each key to a dialect-aware, non-null ORDER BY expression and a typed keyset cursor value, and lifts `ORDER BY` + the keyset predicate into `QueryBuilder` (`OrderByClause` / `CursorPredicate`). SQLite, PostgreSQL, and DuckDB now render ordering from one definition instead of the three duplicated inline `ORDER BY COALESCE(...) DESC, id DESC` strings that were there before.

Pagination works under every sort: `SessionCursor` carries `(sort, direction, value)`, the value is re-typed per column kind (with `::bigint`/`::timestamptz`/`::double precision` casts on PostgreSQL), and nullable sorts (`health`, `context-pressure`) use a direction-aware sentinel `COALESCE` so NULLs sort last identically on every backend. A cursor reused under a different sort or direction is rejected as invalid; cursors minted before these fields existed still decode as recent-descending.

The `secrets` sort is version-gated to match the count the API actually displays (`hideStaleSecretCount`): when active scanner-rule versions are known, counts from stale versions are treated as 0, so a session shown with "0 secrets" can't rank above one with current findings.

## Scope / tradeoffs

- **No new indexes.** At local-archive scale a scan+sort is sub-millisecond; adding ~10 single-column indexes would tax every write and bloat the DB. Straightforward follow-up if profiling shows a need.
- **Sidebar index left unchanged** — this is scoped to `session list`, and the sidebar has its own recent-activity ordering.
- The diff is larger than a minimal flag patch because ordering/pagination is generalized across all three backends rather than bolted onto the default sort; the upside is the removal of the duplicated per-backend ORDER BY strings.

## Where to look

`internal/db/sort.go` (registry + cursor helpers), `internal/db/query_dialect.go` (`OrderByClause`/`CursorPredicate`, per-kind casts), the three `ListSessions` implementations (`internal/db`, `internal/postgres`, `internal/duckdb`), `internal/server/huma_routes_sessions.go` (query params), and `cmd/agentsview/session_list.go` (flags).

### DIFF
--- a/cmd/agentsview/session_list.go
+++ b/cmd/agentsview/session_list.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"go.kenn.io/agentsview/internal/db"
@@ -25,6 +26,8 @@ func newSessionListCommand() *cobra.Command {
 		hasSecret                               bool
 		cursor                                  string
 		limit                                   int
+		sort                                    string
+		reverse                                 bool
 	)
 	cmd := &cobra.Command{
 		Use:          "list",
@@ -58,9 +61,16 @@ func newSessionListCommand() *cobra.Command {
 				HasSecret:        hasSecret,
 				Cursor:           cursor,
 				Limit:            limit,
+				OrderBy:          sort,
 			}
 			if cmd.Flags().Changed("min-tool-failures") {
 				f.MinToolFailures = &minToolFailures
+			}
+			// --reverse flips the sort key's canonical direction; leave
+			// Descending nil otherwise so the default applies.
+			if cmd.Flags().Changed("reverse") {
+				d := db.SortDefaultDescending(sort) != reverse
+				f.Descending = &d
 			}
 
 			list, err := svc.List(cmd.Context(), f)
@@ -118,6 +128,10 @@ func newSessionListCommand() *cobra.Command {
 			"Maximum sessions to return (default %d, max %d)",
 			db.DefaultSessionLimit, db.MaxSessionLimit,
 		))
+	flags.StringVar(&sort, "sort", "recent",
+		"Sort by: "+strings.Join(db.SortKeys(), ", "))
+	flags.BoolVarP(&reverse, "reverse", "r", false,
+		"Reverse the sort direction")
 
 	return cmd
 }
--- a/cmd/agentsview/sort_test.go
+++ b/cmd/agentsview/sort_test.go
@@ -1,0 +1,65 @@
+package main
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.kenn.io/agentsview/internal/db"
+)
+
+// sessionListIDs decodes `session list --format json` output into ordered IDs.
+func sessionListIDs(t *testing.T, out string) []string {
+	t.Helper()
+	var got struct {
+		Sessions []struct {
+			ID string `json:"id"`
+		} `json:"sessions"`
+	}
+	require.NoError(t, json.Unmarshal([]byte(out), &got),
+		"stdout should be valid JSON: %q", out)
+	ids := make([]string, len(got.Sessions))
+	for i, s := range got.Sessions {
+		ids[i] = s.ID
+	}
+	return ids
+}
+
+func TestSessionList_SortAndReverse(t *testing.T) {
+	dataDir := t.TempDir()
+	t.Setenv("AGENTSVIEW_DATA_DIR", dataDir)
+	seedSessionWithOpts(t, dataDir, "lo", "p", func(s *db.Session) { s.MessageCount = 2 })
+	seedSessionWithOpts(t, dataDir, "mid", "p", func(s *db.Session) { s.MessageCount = 5 })
+	seedSessionWithOpts(t, dataDir, "hi", "p", func(s *db.Session) { s.MessageCount = 9 })
+
+	// --sort messages defaults to ascending.
+	out, err := executeCommand(newRootCommand(),
+		"session", "list", "--sort", "messages", "--format", "json")
+	require.NoError(t, err)
+	assert.Equal(t, []string{"lo", "mid", "hi"}, sessionListIDs(t, out))
+
+	// --reverse flips it to descending.
+	out, err = executeCommand(newRootCommand(),
+		"session", "list", "--sort", "messages", "--reverse", "--format", "json")
+	require.NoError(t, err)
+	assert.Equal(t, []string{"hi", "mid", "lo"}, sessionListIDs(t, out))
+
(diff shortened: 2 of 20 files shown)


cpcloud:
+1

Is there a way we can support sort-direction per key instead of forcing the same direction for all sort keys? I might've missed it, but it would be nice if we're going to build a `SessionFilter` abstraction to support per-key sort direction.

wesm:
Maybe we can look at that as follow up? Merging this so I can try to get a release out today!

cpcloud:
Yep, definitely not a blocker!

PR #740, opened 2026-06-19
feat(parser): add aider session parser

youdie006:
## Summary

Adds an `aider` session parser. aider stores per-repo chat history as Markdown at `<repo>/.aider.chat.history.md`, where each file accumulates multiple `# aider chat started at ...` runs.

## What it does

- Splits a history file into one session per run, using the existing virtual-path fan-out (`<historyFile>#<runIdx>`) already used by the Shelley and Zed parsers, so a physical file maps to multiple sessions rather than one flattened blob.
- Parses `####` user prompts, assistant prose, and `> ` tool/edit lines. aider has no tool role in agentsview, so tool/edit lines surface as assistant transcript content (the `> Applied edit to ...` lines are preserved verbatim), mirroring the gptme parser.
- Discovers history files with a bounded, time-budgeted (2s) rootless walk under a configurable root — default `$HOME`, override via `AIDER_DIR` / `aider_dirs` — with a skip-set and a depth cap. The agent uses a shallow file watch so it does not recursively watch the whole home directory; new runs are picked up by the periodic sync.

## Identity and limitations

- Session IDs derive from the history-file path, the run's header timestamp, and an ordinal among same-header runs, so appends and different-header edits never re-key existing sessions. Residual: runs that share a byte-identical header timestamp (same repo, same second) disambiguate by position, so removing an earlier same-header run re-keys its later same-header siblings — rare given aider's 1-second header resolution, and pinned by a test.
- Per the multi-session-per-file model (same as Shelley/Zed), a run removed from a still-present file is reconciled on the next full resync rather than incrementally.

## Where to look

`internal/parser/aider.go` (parser + discovery), `internal/sync/engine.go` (`processAider` fan-out and the virtual-path stat resolution that lets a single session re-sync via the live watcher), `internal/parser/types.go` (registry entry, shallow watch).

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ agentsview auto-discovers sessions from all of these:
 
 | Agent                 | Session Directory                                                                                                                                                       |
 | --------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Aider                 | `<repo>/.aider.chat.history.md` (per repo; bounded scan of `~`, set `AIDER_DIR` to scope)                                                                               |
 | Amp                   | `~/.local/share/amp/threads/`                                                                                                                                           |
 | Antigravity           | `~/.gemini/antigravity/`                                                                                                                                                |
 | Antigravity CLI       | `~/.gemini/antigravity-cli/` (see note below)                                                                                                                           |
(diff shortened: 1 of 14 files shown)


wesm:
working on this now

wesm (approved):

PR #742, opened 2026-06-19
fix(activity): align refresh control with filters

cpcloud:
## Summary

Aligns the Activity toolbar refresh control with the filter controls so it matches the Analytics and Usage views instead of sitting at the far right edge of the toolbar.

## Changes

- Removes the Activity-specific refresh wrapper that forced `margin-left: auto`.
- Keeps the shared `RefreshControl` directly in the toolbar flow after the session filter.
- Adds a regression test that guards against reintroducing the right-aligned wrapper.

## Visuals

Before:

before activity toolbar refresh placement

After:

after activity toolbar refresh placement

### DIFF
--- a/frontend/src/lib/components/activity/ActivityPage.svelte
+++ b/frontend/src/lib/components/activity/ActivityPage.svelte
@@ -163,14 +163,12 @@
       />
     </div>
 
-    <div class="refresh-slot">
-      <RefreshControl
-        lastUpdatedAt={activity.lastUpdatedAt}
-        busy={activity.loading}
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
So one thing i also spot that you may want to steal from middleman, is the alingment of that down arrow chevron, we've got a combobox / select component in there that makes that alignment a bit nicer

cpcloud:
> So one thing i also spot that you may want to steal from middleman, is the alingment of that down arrow chevron, we've got a combobox / select component in there that makes that alignment a bit nicer

Fixed in #741.

PR #744, opened 2026-06-19
Yoke date ranges across panels

cpcloud:
This PR adds persisted date yoking across date-aware panels so selecting a date range in sessions analytics, usage, activity, trends, or insights carries across the rest of the viewer. Date yoking is now unconditional: there is no toggle, no opt-out UI, and no persisted enabled flag.

The frontend now has a shared yoke-date store that persists the current range in localStorage, plus URL/yoke precedence rules per panel. Explicit URL date params still win over stored yoke state, bare routes seed from the stored yoke range, and analytics/usage initial fetches wait until URL/yoke date state has been applied.

The UI work removes the old link-button affordance from the top date controls and leaves the existing date pickers/presets as the only visible date controls. Shared local date helpers keep rolling and fixed range materialization consistent across panels, including sessions `window_days` URLs and activity week/month period bounds.

Activity adopts only yoked ranges its backend can represent exactly. Longer ranges such as a 1y analytics/usage yoke are left on Activity's current range instead of being silently clipped or causing Activity's one-year custom-range limit to reject the first load.

Reviewers should start with `frontend/src/lib/stores/yokedDates.svelte.ts`, `frontend/src/App.svelte`, and the page integrations under `frontend/src/lib/components/{analytics,usage,activity,trends,insights}/`.

### DIFF
--- a/frontend/e2e/insights-quality.spec.ts
+++ b/frontend/e2e/insights-quality.spec.ts
@@ -127,7 +127,17 @@ test.describe("Insights quality rollout", () => {
     await page.keyboard.press("Escape");
     await expect(savedInsight).toBeVisible();
     await savedInsight.click();
-    await expect(page).toHaveURL(/\/insights\?insight=42$/);
+    await expect(page).toHaveURL(/\/insights\?.*insight=42/);
+    const selectedInsightUrl = new URL(page.url());
+    expect(selectedInsightUrl.pathname).toBe("/insights");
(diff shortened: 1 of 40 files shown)


mariusvniekerk:
do we need a toggle?  Seems sensible enough to just DO?

cpcloud:
I can definitely remove the toggle.

wesm:
rebasing

PR #754, opened 2026-06-19
fix(postgres): batch pg push comparison reads (#331)

### LINKED ISSUE #331: pg push: slow for large databases due to per-session PG round-trips
## Problem

`pg push` is slow on large databases. With ~35K sessions, an incremental push that only needs to sync 8 changed sessions still evaluates ~54K candidate sessions against PG, each requiring multiple PG round-trips.

```
Pushing... 450/54038 sessions, 0 messages
```

Most sessions are no-ops (PG already matches local), but the current architecture checks each one individually inside a transaction.

## Root Causes

1. **Candidate window too wide** — `ListSessionsModifiedBetween` returns all sessions since `lastPush`, and fingerprint filtering only prunes if prior boundary state exists. When boundary state is empty or stale, all sessions pass through.

2. **Per-session PG round-trips** — `pushBatch` opens a transaction for each batch of 50 sessions, then for each session: upserts the session row, queries PG message count + content fingerprint, compares against local, and only then decides to skip. That's multiple PG queries per session even when nothing changed.

3. **Multi-row INSERT instead of COPY** — message inserts use `INSERT ... VALUES` in batches of 100 rows. PostgreSQL's `COPY FROM` protocol (supported natively by pgx's `CopyFrom`) is 5-10x faster for bulk loads.

## Potential Improvements

- **Batch fingerprint check** — query PG for all candidate session fingerprints in one query, filter locally, then only push sessions that actually differ
- **Use `COPY FROM`** — pgx supports `CopyFrom` natively; switch `bulkInsertMessages`, `bulkInsertToolCalls`, and `bulkInsertToolResultEvents` to use it
- **Narrow the candidate window** — investigate why boundary state is empty/stale after normal incremental pushes, causing the full session list to be evaluated
- **Parallel session processing** — process independent sessions concurrently (within connection pool limits)

rodboev:
## Summary
- Keep the candidate-window and boundary-session behavior in `internal/postgres/push.go` unchanged for this PR, and batch the PostgreSQL-side comparison reads used to decide whether a candidate session can be skipped.
- Implement new batched loaders in `internal/postgres/push_fingerprint.go` for message aggregates, message content hashes, role/time fingerprints, message flags, message system ordinals, token fingerprints, tool-call aggregates, tool-call fingerprints, and usage fingerprints, with chunking inside the helper when session counts exceed `ANY($1)` practicality.
- Use the preloaded message and tool-call aggregates on the hot no-op path, and retry any comparison-preload SQL failure in a fresh transaction without the batched preload instead of continuing inside an already-aborted transaction.
- Add targeted regression tests in `internal/postgres/push_test.go` and `internal/postgres/push_fingerprint_test.go` to cover the new batch-driven skip decision path and helper behavior with empty inputs.

## Scope
- Files changed are `internal/postgres/push.go`, `internal/postgres/push_fingerprint.go`, `internal/postgres/push_test.go`, and `internal/postgres/push_fingerprint_test.go`.
- No boundary/windowing semantics, no schema changes, and no changes to PR #731 or broader sync-work areas.

## Notes
- A focused PG comparison query-count assertion was not added because the existing harness does not expose a stable helper-call/query metric for this exact path without adding brittle test-only instrumentation.
- The review-driven follow-up keeps the existing non-batched fingerprint fallback, but now that fallback only runs from a clean transaction after preload failure instead of on the poisoned transaction that raised the preload error.

Fixes #331

### DIFF
--- a/internal/postgres/push.go
+++ b/internal/postgres/push.go
@@ -556,6 +556,10 @@ type batchResult struct {
 	skippedConflicts int
 }
 
+var errPushComparisonPreload = errors.New(
+	"push comparison preload failed",
+)
+
 // pushBatch pushes a slice of sessions within a single
 // transaction. On success it appends to pushed and returns
 // ok=true with session/message counts. On a session-level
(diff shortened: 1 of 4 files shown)


wesm:
awesome, thanks for working on this!

PR #791, opened 2026-06-21
feat(pricing): bundle LiteLLM snapshot as offline fallback

### LINKED ISSUE #315: Bundle LiteLLM pricing snapshot as offline fallback
## Problem

The current `pricing.FallbackPricing()` hardcodes ~8 Claude/Codex
variants. Any user who is offline, air-gapped, or simply hasn't
completed the async LiteLLM refresh yet sees \$0 cost for every
model not on that list — including most OpenAI, Bedrock, Vertex,
and Azure rates.

After #314 the server seeds the hardcoded fallback synchronously
and refreshes from LiteLLM in the background, so online users
converge quickly. Offline and air-gapped users remain stuck on
the hand-maintained list.

## Proposal

Bundle a snapshot of LiteLLM's `model_prices_and_context_window.json`
inside the repo and embed it via `go:embed`, replacing the
hardcoded rate table in \`internal/pricing/fallback.go\`.

1. Commit \`internal/pricing/litellm_snapshot.json\` (~500KB).
2. Change \`FallbackPricing()\` to call
   \`ParseLiteLLMPricing(embeddedBytes)\` so the existing pricing
   shape is reused.
3. Add a \`make refresh-pricing\` target that downloads the
   upstream JSON, parses it to validate, and writes the snapshot.
   Optional: monthly Dependabot-style automation opens a refresh
   PR.
4. Consider pruning the snapshot at build time to the providers
   agentsview actually supports, to keep the binary lean.

## Benefits

- Offline and air-gapped users get the full multi-provider
  catalog instead of \$0 rows.
- Fresh installs are instant and complete; the two-stage async
  refresh in runServe can collapse back to a single sync upsert.
- \`internal/pricing/fallback.go\` stops being a hand-maintained
  list that drifts out of sync with reality.

## Tradeoffs

- Binary grows by ~500KB. Pruning to relevant providers would
  reduce that.
- Snapshot goes stale between releases, but live LiteLLM fetch
  still runs when online and the refresh target / automation
  keeps the committed copy current.

## Related

- #314 — adds the two-stage seed this issue would simplify.

rodboev:
## Summary
- replace the hand-maintained pricing fallback slice with a sorted LiteLLM snapshot that is decoded through the existing `pricing.FallbackPricing()` and `FallbackVersion` API surface
- keep SQLite, PostgreSQL, DuckDB, CLI usage, and startup refresh consumers on the same fallback source while preserving the existing async refresh behavior
- move the generated `litellm_snapshot.json.gz` out of the normal branch history and store it on the `litellm-pricing-snapshot` orphan artifact branch
- add `make pricing-snapshot` to restore the ignored generated snapshot from that branch, and make `make build`, `make build-release`, `make install`, and release targets embed it before compiling
- update the GitHub tag release workflow to restore the pricing snapshot before its direct `go build` commands, so production releases fail early instead of silently shipping without embedded offline pricing
- keep the snapshot generator responsible for fetching LiteLLM, applying explicit legacy/current overlays, sorting output, computing a deterministic version hash, and writing the gzipped asset under `internal/pricing/snapshot/`
- batch DuckDB fallback pricing upserts so the larger offline fallback does not push Windows DuckDB package tests past their timeout

## Notes
- clean source checkouts stay small because the generated `.json.gz` remains ignored and lives outside the main branch history
- the orphan branch artifact is intentionally tiny and Git-addressable; normal build and release paths restore it automatically before embedding
- explicit overlays remain limited to known models or aliases absent from the current LiteLLM catalog but still needed for existing sessions, including newer Claude/OpenAI/Mistral names and `openrouter/owl-alpha`
- backend pricing behavior remains shared through `pricing.FallbackPricing()`, with no backend-specific fallback divergence
- a raw manual `go build` from a clean checkout can still compile without the snapshot, but supported Makefile and GitHub release build paths embed it

Closes #315

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,7 @@ jobs:
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
+          fetch-depth: 0
           persist-credentials: false
 
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c  # v6.4.0
@@ -129,6 +130,9 @@ jobs:
           update: false
           install: mingw-w64-x86_64-gcc
           path-type: inherit
+
+      - name: Restore pricing snapshot
+        run: go run ./internal/pricing/cmd/litellm-snapshot -restore
 
       - name: Run Go tests
         run: go test -tags "fts5" ./... -v -count=1
@@ -183,11 +187,15 @@ jobs:
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
+          fetch-depth: 0
           persist-credentials: false
 
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c  # v6.4.0
         with:
           go-version-file: go.mod
+
+      - name: Restore pricing snapshot
+        run: go run ./internal/pricing/cmd/litellm-snapshot -restore
 
       - name: Test with coverage
         run: go test -tags "fts5" -coverprofile=coverage.out ./...
@@ -224,11 +232,15 @@ jobs:
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
+          fetch-depth: 0
           persist-credentials: false
 
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c  # v6.4.0
         with:
           go-version-file: go.mod
+
+      - name: Restore pricing snapshot
+        run: go run ./internal/pricing/cmd/litellm-snapshot -restore
 
       - name: Run PostgreSQL integration tests
         run: make test-postgres-ci
@@ -272,6 +284,7 @@ jobs:
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
+          fetch-depth: 0
           persist-credentials: false
 
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c  # v6.4.0
@@ -294,6 +307,9 @@ jobs:
         run: |
           rm -rf internal/web/dist
           cp -r frontend/dist internal/web/dist
+
+      - name: Restore pricing snapshot
+        run: go run ./internal/pricing/cmd/litellm-snapshot -restore
 
       - name: Pre-build Go binaries
         run: |
--- a/internal/pricing/cmd/litellm-snapshot/main.go
+++ b/internal/pricing/cmd/litellm-snapshot/main.go
@@ -1,0 +1,573 @@
+package main
+
+import (
+	"bytes"
+	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io"
+	"net/http"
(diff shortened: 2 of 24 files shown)


wesm:
I'll probably want to put this on an orphan branch or something, will take a look!

.github/workflows/ci.yml
```diff
@@ -108,6 +109,9 @@ jobs:
           install: mingw-w64-x86_64-gcc
           path-type: inherit
 
+      - name: Restore pricing snapshot
```
mariusvniekerk, line 134: do we have a way to ensure these line up with those embedded in the Makefile?

internal/pricing/cmd/litellm-snapshot/main.go
```diff
@@ -0,0 +1,489 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"go.kenn.io/agentsview/internal/pricing"
+)
+
+var defaultOutputPath = filepath.FromSlash(
```
mariusvniekerk, line 23: we should probably use go:embed here.

.github/workflows/ci.yml
rodboev: Centralized the pinned ref, SHA256, branch, and file values as compiled defaults in the Go tool itself (`defaultSnapshotRef`, `defaultSnapshotSHA256`, etc.). All CI workflows, the Makefile recipe, and `prepare-sidecar.sh` now just call `go run ./internal/pricing/cmd/litellm-snapshot -restore` with no args, so there's exactly one place to update when the snapshot changes.

Also fixed the Windows CI failure in `TestPrepareSidecarRestoresPricingSnapshotBeforeBuild` (CRLF line endings on Windows caused the `strings.Index` match to miss).

internal/pricing/cmd/litellm-snapshot/main.go
rodboev: This file is the snapshot generator/restorer tool, not the consumer. It needs to write files to disk (for `go generate`, restore from git artifacts, etc.), so `go:embed` isn't applicable here. The consumer side (`internal/pricing/fallback.go`) does use `go:embed` to bundle the generated snapshot into the binary:

```go
//go:embed snapshot/*
var litellmSnapshotFS embed.FS
```

rodboev:
**Finding 1** (Makefile:146 — `make dev` missing `pricing-snapshot`): fixed in 7468691e. `dev` now depends on `pricing-snapshot` so a clean checkout restores the snapshot before Air starts.

**Finding 2** (`--depth=1` shallow fetch of pinned SHA): false positive. The pinned ref and the artifact branch tip are updated together in the same commit whenever the snapshot is regenerated. `--depth=1` fetches the branch tip, which *is* the pinned commit. The branch never advances past the pin without the pin constants being updated in the same change.

rodboev:
Same two findings as bdff8bb, addressed in the prior reply. Finding 1 is fixed by this commit (7468691e adds `pricing-snapshot` to `make dev`). The residual nil-on-bare-`go build` is by design: `.keep` satisfies the embed directive so the binary compiles, and the online API is the primary pricing path. Finding 2 (shallow fetch) was addressed previously; the pin and branch tip move together.

wesm:
I have a bunch of unpushed work here locally

wesm:
do you want to pause here so I can have my agent assess the state of things?

rodboev:
Sure, go ahead! I'll hold off until you've had a chance to look things over.

wesm:
Looks like all good, just give me a moment here to rebase and push

mariusvniekerk (approved):

PR #800, opened 2026-06-22
fix(sync): harden incremental JSONL resume state

### LINKED ISSUE #162: Incremental sync: DAG boundary validation, ordinal tracking, and atomicity
## Context

PR #160 introduced incremental JSONL parsing for Codex and Claude
sessions to prevent `SyncPaths` from blocking SSE live updates. The
incremental path works well for the common case (linear appends), but
several edge cases were identified during review that require schema
changes or larger refactors beyond that PR's scope.

### 1. DAG fork detection at sync boundaries

**Problem**: `hasDAGFork()` validates UUID linearity only within the
newly appended entries. The first appended entry's `parentUuid` is not
checked against the last UUID from the previous sync. If a user forks a
Claude session from an older node (not the tip), the incremental parser
treats it as linear and appends it incorrectly.

**Fix**: Add a `last_entry_uuid` column to the sessions table. Populate
it during full parses with the UUID of the last user/assistant entry.
Pass it into `ParseClaudeSessionFrom` to initialize `lastUUID` in
`hasDAGFork`, or fall back to a full parse when the first appended
entry's `parentUuid` doesn't match.

### 2. Ordinal drift when trailing messages are filtered

**Problem**: The next incremental parse starts from `MaxOrdinal()+1`,
but `toDBMessages` / `pairAndFilter` can drop messages (e.g. user
entries containing only `tool_result` blocks). If an incremental append
ends with only filtered messages, `file_size` advances but the DB's max
ordinal does not. The subsequent sync reuses ordinals, causing drift
versus a full parse.

**Fix**: Store `next_ordinal` in session metadata (separate from the
DB's max stored ordinal). Update it after each incremental sync to
reflect the raw parsed ordinal count, not the filtered count.

### 3. Non-atomic message + metadata writes

**Problem**: `writeIncremental` executes message insertion and
`UpdateSessionIncremental` in separate statements. If messages are
inserted but the metadata update fails, `file_size` stays stale while
`MaxOrdinal` has advanced, risking duplicate messages on retry.

**Fix**: Wrap both operations in a single database transaction so
message ordinals and file offsets advance atomically.

### 4. Codex exec events dropped in incremental path

**Problem**: `processCodex` hardcodes `includeExec=false` for the
incremental parse function. If a session was originally synced via
`SyncSingleSession` with `includeExec=true` (exec-originated), a
subsequent incremental sync silently drops newly appended exec events
while advancing the offset.

**Fix**: Store whether the session is exec-originated and pass it
through to the incremental parser, or skip incremental for exec sessions.

### 5. Cross-sync tool result pairing

**Problem**: The incremental parser pairs tool calls and results only
within the appended chunk. If an assistant `tool_use` was stored in a
previous sync and the matching user `tool_result` arrives in a later
append, `pairToolResults` can't see the earlier call. The result content
is never linked.

**Fix**: Look up unmatched tool calls by `tool_use_id` from the DB
during incremental sync, or fall back to a full reparse when unmatched
tool results are detected.

### 6. Subagent mapping lost in incremental Claude parsing

**Problem**: Full Claude parsing collects `queue-operation` and
`progress` events to populate `subagentMap`, which annotates Task/Agent
tool calls with `SubagentSessionID`. The incremental parser skips all
non-user/assistant lines, so newly appended Task/Agent tool calls lose
subagent linkage.

**Fix**: Mirror the subagent-mapping collection in
`ParseClaudeSessionFrom` and annotate incremental messages before
returning.

rodboev:
## Summary
- Persist local-only `last_entry_uuid` and `next_ordinal` so incremental sync resumes from the real JSONL boundary instead of `MaxOrdinal()+1`.
- Make incremental append writes atomic, and make Claude incremental parsing self-heal when the new tail needs prior-session state for DAG validation, cross-sync `tool_result` pairing, or queue and progress based subagent linkage repairs.
- Preserve Codex `function_call_output` tails by storing generic call results on full parse and forcing incremental Codex fallback when those outputs land after an already-synced tool call.

## Scope
- Keeps `last_entry_uuid` and `next_ordinal` inside local SQLite sync state; PostgreSQL push and read models stay unchanged.
- Uses full-parse fallback for ambiguous cross-sync `tool_result` tails instead of adding a broader DB lookup layer in this PR.
- Keeps the Codex runtime change narrow to the now-proved live gap, generic `function_call_output` tails that otherwise disappear on incremental exec sessions.

## Review Notes
- Start with `internal/db/schema.sql`, `internal/db/db.go`, and `internal/db/sessions.go`, where the new resume metadata and incremental update contract are introduced.
- Then review `internal/sync/engine.go` and `internal/parser/claude.go`, where the stored boundary UUID, next ordinal, atomic append path, and incremental subagent mapping are threaded through the Claude append flow.
- Focused validation is the parser, sync, and DB commands from the plan; the incremental review fixes add split-sync queue and progress linkage coverage plus the Codex exec `function_call_output` retention gate, while broader lint, PostgreSQL, frontend, integration, and desktop coverage stay in CI.

Fixes #162

### DIFF
--- a/internal/db/automated_backfill_test.go
+++ b/internal/db/automated_backfill_test.go
@@ -236,7 +236,12 @@ func TestIncrementalUpdateReclassifiesOnPatternChange(t *testing.T) {
 
 	// Incremental update with umc still <= 1.
 	err = d.UpdateSessionIncremental(
-		"changelog-inc", nil, 2, 1, 1024, 100, 0, 0, false, false,
+		"changelog-inc", IncrementalSessionUpdate{
+			MsgCount:     2,
+			UserMsgCount: 1,
+			FileSize:     1024,
+			FileMtime:    100,
+		},
 	)
 	require.NoError(t, err, "incremental update")
 
@@ -267,7 +272,12 @@ func TestIncrementalUpdateClearsWhenCountGrows(t *testing.T) {
 
 	// Incremental update pushes umc > 1 — must clear.
 	err = d.UpdateSessionIncremental(
-		"grew-past-one", nil, 7, 3, 2048, 200, 0, 0, false, false,
+		"grew-past-one", IncrementalSessionUpdate{
+			MsgCount:     7,
+			UserMsgCount: 3,
+			FileSize:     2048,
+			FileMtime:    200,
+		},
 	)
 	require.NoError(t, err, "incremental update")
 
@@ -292,7 +302,12 @@ func TestIncrementalUpdateLeavesNonMatching(t *testing.T) {
 	})
 
 	err := d.UpdateSessionIncremental(
-		"normal-single", nil, 2, 1, 1024, 100, 0, 0, false, false,
+		"normal-single", IncrementalSessionUpdate{
+			MsgCount:     2,
+			UserMsgCount: 1,
+			FileSize:     1024,
+			FileMtime:    100,
+		},
 	)
 	require.NoError(t, err, "incremental update")
 
@@ -328,7 +343,12 @@ func TestIncrementalUpdateClearsTerminationStatus(t *testing.T) {
 		"precondition: expected tool_call_pending")
 
 	err = d.UpdateSessionIncremental(
-		"stale-term", nil, 4, 2, 2048, 200, 0, 0, false, false,
+		"stale-term", IncrementalSessionUpdate{
+			MsgCount:     4,
+			UserMsgCount: 2,
+			FileSize:     2048,
+			FileMtime:    200,
+		},
 	)
 	require.NoError(t, err, "incremental update")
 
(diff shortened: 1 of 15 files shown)


rodboev:
`next_ordinal` and `last_entry_uuid` are intentionally local ingest cursors, not replicated session data.

They are derived from the current machine's on-disk JSONL boundary and are only read back by the local incremental scanner to decide where the next append resume starts. `pg push` is carrying session content into Postgres, not transporting scanner resume state between machines. If we copied these fields into the PG schema and push fingerprint, another machine could pick up a stale file offset or UUID tail that does not describe its local transcript file.

So the branch keeps the user-visible session fields in parity, but leaves these two resume markers SQLite-only on purpose.

wesm:
Thank you, I'll get a 0.34.2 release out today with this fix and the other fixes you have contributed

PR #804, opened 2026-06-22
fix(search): quote FTS terms so operator-char tokens don't 500

mjacobs:
Fixes a 500 on full-text search when a single query token contains FTS5 operator characters, e.g. `error-401` or `status:500`.

## Cause

`prepareFTSQuery` only quoted the query when it contained a space, so a single token containing `-` or `:` reached the FTS5 `MATCH` with those characters interpreted as operators, producing a malformed query.

## Fix

A shared `db.PrepareFTSQuery` quotes each term (doubling embedded quotes), combines terms with implicit AND, and supports an exact-phrase opt-in via a leading quote; `db.FTSTerms` / `db.StripFTSQuotes` reverse it for the name/branch LIKE matching and snippet centering. The server handler and the db content-search path both delegate to it. The Postgres ILIKE path and the DuckDB read store are aligned to the same per-term-AND semantics for backend parity.

## Behavior change

Multi-word session search is now **AND-of-terms** across all backends rather than a single exact phrase; exact-phrase matching is still available by leading the query with a quote. Frontend help text that describes search as phrase-by-default should be updated to match.

Reviewers: `internal/db/search.go` (the shared helpers), and the four delegating sites — `internal/server/search.go`, `internal/db/search_content.go`, `internal/postgres/messages.go`, `internal/duckdb/store.go`.

https://claude.ai/code/session_017qDj5kCrsxw1RgpLiLaZka

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -29,6 +29,11 @@ import (
 // trigger a non-destructive re-sync (mtime reset + skip cache
 // clear) so existing session data is preserved.
 //
+// Bumped to 50: parser-derived text is sanitized for PostgreSQL
+// parity and fingerprints. Existing rows need re-parsing so stored
+// message/session shape, timestamps, roles, token counts, and content
+// fingerprints are based on the sanitized parse output.
+//
 // Bumped to 49: incremental JSONL resume now persists next_ordinal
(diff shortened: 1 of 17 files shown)


wesm:
looking at this

PR #805, opened 2026-06-22
feat(parse-diff): classify live-write skew as raced, not changed

mjacobs:
`parse-diff` re-parses on-disk sources and compares them against stored rows, so a parser change's effect can be gated in CI with `--fail-on-change`. Two things made that gate untrustworthy; this PR fixes both, plus a Codex title-refresh gap the work surfaced.

## Live-write skew classified as `raced`

`parse-diff` snapshots stored rows, then re-parses sources; a daemon or an active session writing the archive mid-run produced torn comparisons reported as `changed`, tripping `--fail-on-change` on a write unrelated to the parser. After each per-file compare, the source mtime is re-stat'd (the same agent-aware nanosecond value sync stores in `file_mtime`) and compared against the snapshot's `Session.FileMtime`; a file written after the snapshot is classified `raced` instead of `changed`. Raced sessions are excluded from `--fail-on-change`, so a concurrent write no longer fails the run, while a genuine change on a demonstrably untouched file still does. Classification is conservative (unreadable source or missing stored mtime → raced) but never masks a real change.

## Report-only engine write guard

`NewDiffEngine` returns the full `*Engine`, which still exposed `SyncAll`/`ResyncAll`, so a report-only/force-parse engine could in principle be driven into a write path. `refuseWriteInForceParse` now guards those entrypoints as a no-op when `forceParse` is set (only `NewDiffEngine` sets it). Additive; no existing write behavior changes.

## Codex title refresh (surfaced by this work)

To keep the raced guard consistent, the incremental write now folds the `session_index.jsonl` mtime into the stored `file_mtime`, exactly as a full parse does. That raised the stored watermark on the append path and exposed a latent gap: a title-only index rename whose mtime lands at or below the stored value was filtered out by an `indexMtime > storedMtime` gate (in both the incremental and `SyncAllSince` quick-sync paths), leaving the title stale behind `shouldSkipCodex`'s fast path. Both paths now compare the index title to the stored `session_name` directly. The broader "mtime-watermark conflates transcript + title" class is tracked as a follow-up.

## Scope

Covers the parse-diff v2 sub-items for live-write skew and the write-path guard, plus the Codex title-refresh fixes above. Incremental-append skew detection and DB-backed coverage for Warp/Forge/Piebald remain for a follow-up.

Reviewers: `internal/sync/parsediff.go`, `internal/sync/parsediff_compare.go`, `internal/sync/parsediff_report.go`, and the Codex paths in `internal/sync/engine.go`.

### DIFF
--- a/cmd/agentsview/parse_diff.go
+++ b/cmd/agentsview/parse_diff.go
@@ -299,6 +299,7 @@ func renderParseDiffReport(
 	renderParseDiffFieldCounts(w, r.FieldCounts)
 	renderParseDiffParseErrors(w, r.Sessions)
 	renderParseDiffChanged(w, r.Sessions, verbose)
+	renderParseDiffRaced(w, r.Sessions, verbose)
 	if verbose {
 		renderParseDiffPendingResync(w, r.Sessions)
 	}
@@ -379,6 +380,8 @@ func renderParseDiffSummary(w io.Writer, t sync.ParseDiffTotals) {
 		"(no stored row; sync would add them)")
 	line("Skipped", t.Skipped,
 		"(source not re-parsed: missing, remote, trashed, or not sampled)")
+	line("Raced", t.Raced,
+		"(source changed mid-run; inconclusive, not counted as drift)")
 	line("Excluded by parser", t.ExcludedByParser,
 		"(parser intentionally drops these; sync would delete them)")
 	line("Parse errors", t.ParseErrors,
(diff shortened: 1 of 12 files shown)


wesm:
phew! merging

mjacobs:
Thanks @wesm! Sorry about the churn on this one (and #802), I'm doing a "micro-postmortem"(?) to understand + close process gaps.

PR #806, opened 2026-06-22
feat(sync): surface detailed resync progress

wesm:
Data-version upgrades can spend long stretches in full-resync work that is currently hard to see from either `agentsview serve` output or the desktop status bar. That makes healthy startup work look like a frozen process, especially while the FTS index is being rebuilt on large archives.

This change promotes those internal resync milestones into structured sync progress and exposes the current in-flight progress through the existing sync status polling path. Manual sync streams keep their existing progress contract while gaining detail and hint fields, and startup resyncs become visible to the Tauri UI without adding another event channel.

Reviewers should look at the engine progress lifecycle first: current progress is intentionally transient, separate from last completed sync stats, and cleared when the sync entrypoint exits. The FTS rebuild hint is emitted before the expensive call so the CLI and status bar can explain the wait while it is happening.


mariusvniekerk:
Potentially we could also start the application whilst the FTS index is being built and just have the search box disabled with the contents
"search index rebuilding..." (potentially even with progress information if we have anything)

wesm:
We could, that would be more work though — this at least lets the user have a better idea what's going on and more of an expectation that it takes a while to index for large databases

PR #810, opened 2026-06-22
ci: verify docs build

wesm:
The documentation site now lives in this repository, but the main CI workflow was not exercising the migrated Zensical build path. That made docs source issues, asset hydration failures, built-site regressions, and Vercel redirect mistakes easy to miss until deployment or a manual maintainer check.

This adds a dedicated docs CI job that runs the existing `make docs-check` entrypoint with uv available, so CI follows the same full path maintainers use locally. A workflow guard test keeps the job wired to that entrypoint and avoids silently falling back to the no-uv skip path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,28 @@ jobs:
         run: npm test
         working-directory: frontend
 
+  docs:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
We may want some skills /hooks to get triggered so that when agents do stuff they are forced to look at the docs too

wesm:
Yeah, this is just getting back to parity with where we were before in the docs repo pre-integration

PR #821, opened 2026-06-23
feat(frontend): enrich tool summaries and add skim layout

wesm:
Two related transcript-readability changes. Frontend only; no API, DB, or schema change.

## Enriched tool summaries

A new pure util `summarizeToolCall` (`frontend/src/lib/utils/tool-summary.ts`) builds a structured one-line header for each tool call and adds result-derived suffixes: Read `(N lines)`, Edit `(+A -B)`, Write `(+N)`, Grep `(N matches)`, Glob `(N files)`. Counts are emitted only when they parse with confidence — non-zero, one trailing newline ignored, and Grep `output_mode=count` suppressed. `ToolBlock.svelte` renders this summary whether the block is collapsed or expanded, and falls back to the prior first-body-line preview (collapsed-only) when the util returns null.

This changes one prior behavior: the structured summary now takes precedence over the display content's first line, so a Bash header shows `$ <input command>` even when display content exists. The one ToolBlock test that asserted the old precedence is updated.

## Skim layout

Adds a fourth `messageLayout`, `skim`, to the default/compact/stream cycle (store, settings, and header icon). `.layout-skim` (in `MessageList.svelte`) collapses each tool call to its one-liner, flattens tool-group and parallel-group chrome, hides nested subagent transcripts, and sets `pointer-events: none` on tool headers so a click cannot toggle hidden collapse state (row selection still works via the row wrapper). A pure helper `resolveMessageLayout` suspends skim back to the full layout while an in-session search highlight is active, so matches stay expanded.

## Where to look

- `frontend/src/lib/utils/tool-summary.ts` — summary construction and conservative counts (+ table-driven tests).
- `frontend/src/lib/components/content/ToolBlock.svelte` — ungated summary, collapsed-only legacy fallback.
- `frontend/src/lib/components/content/MessageList.svelte` — `.layout-skim` CSS and the effective-layout search guard.
- `frontend/src/lib/utils/message-layout.ts` — the skim-during-search guard.

The skim CSS hiding is not unit-tested (jsdom does not compute styles from `<style>` blocks, matching the existing untested compact/stream layouts); the search-guard logic is covered by `resolveMessageLayout` unit tests.

### DIFF
--- a/frontend/src/lib/components/content/MessageList.svelte
+++ b/frontend/src/lib/components/content/MessageList.svelte
@@ -20,6 +20,7 @@
     hasVisibleSegments,
   } from "../../utils/content-parser.js";
   import { isSystemMessage } from "../../utils/messages.js";
+  import { resolveMessageLayout } from "../../utils/message-layout.js";
   import { inSessionSearch } from "../../stores/inSessionSearch.svelte.js";
   import { sessionActivity } from "../../stores/sessionActivity.svelte.js";
   import SessionFindBar from "./SessionFindBar.svelte";
@@ -519,6 +520,10 @@
       ? inSessionSearch.query
       : "",
   );
+
+  let effectiveLayout = $derived(
+    resolveMessageLayout(ui.messageLayout, highlightQuery !== ""),
+  );
 </script>
 
 {#if !sessions.activeSessionId}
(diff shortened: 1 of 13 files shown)


mariusvniekerk:
Screenshots for the transcript readability changes:

Normal transcript with enriched tool summaries:


Skim layout with collapsed one-line tool summaries:


Focused transcript in skim layout:

PR #826, opened 2026-06-23
Make local CLI and desktop daemon-first

wesm:
Make AgentsView daemon-first for local CLI and desktop workflows.

This PR changes the local archive model from “every command opens SQLite, does work, then exits” to a restartable per-data-dir daemon that normally owns local writes. The goal is to remove repeated expensive SQLite startup/open work from common CLI paths while also preventing multiple processes from racing full resyncs, push state updates, pricing refreshes, and WAL/checkpoint activity against the same archive.

The core invariant is now: one writable owner per data directory. A background daemon holds the write-owner lock for its lifetime. Direct/offline commands either acquire that same lock before opening SQLite read-write or refuse with guidance when another process owns the archive. Read-only opens use an explicit read-only DB path and skip the write-owner lock.

What changes for CLI users:

- Read-only commands stay fast on a cold archive. Commands such as `session list`, `session get`, `session messages`, stats, health/projects, and explicit offline reports attach to a compatible daemon when one is already running; otherwise they read SQLite directly in read-only mode and return the latest indexed data.
- Freshness-sensitive or mutating commands use the daemon by default. Local `sync`, `session sync`, normal usage refreshes, `token-use`, `pg push`, `pg push --watch`, and `duckdb push` attach to or start the detached daemon so the daemon remains the write owner.
- `AGENTSVIEW_NO_DAEMON=1` remains the escape hatch for scripts and offline one-shots. With it set, write commands run directly only after acquiring the write-owner lock; if a daemon or offline writer already owns the archive, they refuse instead of becoming a hidden second writer.
- `--no-sync`/offline paths continue to work. No-sync daemons expose on-demand routes for manual sync/push work without starting a second writer, and local fallback paths preserve no-sync behavior when daemon auto-start is disabled.

Daemon lifecycle and compatibility:

- `serve --background` remains the detach primitive. CLI and desktop launch through the existing background path rather than spawning a foreground child that dies with the launcher.
- The daemon records runtime metadata, API/data compatibility, auth requirements, no-sync state, and read-only status. Transport discovery ignores incompatible daemons for normal attach, and startup reports incompatible launch winners instead of silently treating them as success.
- Older compatible daemons are replaced through the serialized background-start flow so replacement runs under the launch lock rather than stopping the old daemon before a new owner is ready.
- `agentsview update` stops running writable daemons before applying the update and restarts a daemon afterward when appropriate, preserving auth/no-sync details from the stopped runtime.
- The daemon self-exits after an idle timeout when there are no external requests and no active daemon-owned jobs. Periodic sync/file-watcher work prevents exit while it is running but does not keep the daemon alive forever by itself.

Write and query routing:

- New archive query/write backend abstractions centralize the daemon-vs-direct decision instead of scattering ad hoc `detectTransport` calls through activity, usage, token-use, sync, PG, and DuckDB commands.
- Daemon-owned HTTP routes now cover local sync, session sync, remote sync fan-out, PG push, DuckDB push, and usage/activity query surfaces needed by the daemon-first CLI paths.
- PG/DuckDB push and watch keep their small local state writes daemon-owned while still supporting direct offline operation under the write lock.
- Remote sync is restricted to configured remote hosts, and SSH target validation rejects option-shaped host/user values before calling `ssh`.

SQLite and data safety:

- Writable opens go through the write-owner lock. This protects the full-resync atomic DB swap and prevents long-lived processes from holding stale writer handles across a swapped archive.
- Read-only SQLite opens tolerate the daemon-owning case and avoid running migrations or taking the write-owner lock.
- Stale data-version archives still trigger the full resync path through daemon sync routes, including no-sync daemons handling manual sync requests.
- Read-only stats/usage/activity paths preserve pricing and default date-range semantics between daemon-backed and direct SQLite backends.

Desktop app behavior:

- The Tauri app no longer owns a sidecar backend child as the primary model. It starts or attaches to the detached daemon via `agentsview serve --background`, discovers the actual daemon URL, and navigates only to that discovered URL.
- Clean desktop exit drops the connection rather than killing the shared daemon. The idle reaper is responsible for cleanup when the daemon is no longer in active use.
- The desktop update path explicitly stops detached daemons before update/restart work so the happy-path upgrade does not leave an older daemon running behind the newer app/CLI.

Docs and cleanup:

- README, quickstart, configuration, command docs, and session API docs now describe daemon-first behavior, the no-daemon escape hatch, idle shutdown, local data-directory files, and updated command behavior.
- The temporary superpowers planning/design specs for the old docs-repository work were removed.
- Test fixtures were refactored where they reduced repeated setup, but the large stats/session scaffolding remains in the owning test files rather than being hidden in separate fixture files.

Tradeoffs and limitations:

- Cold read-only CLI commands prefer stale-but-fast local reads rather than auto-starting a daemon solely to answer a read.
- A detached daemon can outlive the desktop window by design. It is bounded by idle shutdown, explicit `serve stop`, and update-driven stop/restart behavior rather than by the desktop process lifetime.
- Direct offline write commands may now refuse when a daemon owns the archive. That refusal is intentional: it keeps the single-writer invariant instead of reintroducing competing SQLite writers.
- Auto-started daemons are constrained by the serve/auth configuration and should not silently create an unauthenticated non-loopback service from an ordinary local CLI command.

Review focus:

- Transport and lifecycle decisions: `cmd/agentsview/transport.go`, `serve_background.go`, `serve_lifecycle.go`, `update.go`, and `desktop/src-tauri/src/lib.rs`.
- Write ownership and direct fallback behavior: `cmd/agentsview/write_lock.go`, `archive_write_backend.go`, `archive_query_backend.go`, and `internal/db/db.go`.
- Daemon-owned command routes: `internal/server/huma_routes_sync.go`, `huma_routes_push.go`, `cmd/agentsview/sync.go`, `pg.go`/`pg_watch.go`, `duckdb.go`, `usage.go`, `session_usage.go`, and `token_use.go`.
- Compatibility/security edges: runtime metadata in `daemon_runtime.go`, remote sync validation in `internal/ssh/ssh.go`, and read-only daemon handling across usage/activity/session paths.


internal/db/db.go
```diff
@@ -503,6 +518,215 @@ func Open(path string) (*DB, error) {
...
+	return db, nil
+}
+
+func checkReadOnlySchemaCompatibility(path string) error {
+	conn, err := sql.Open("sqlite3", makeDSN(path, true))
+	if err != nil {
+		return fmt.Errorf("checking read-only schema: %w", err)
+	}
+	defer conn.Close()
+	required := map[string][]string{
```
mariusvniekerk: This seems very fragile....

mariusvniekerk (approved):

PR #828, opened 2026-06-23
fix(usage): dedupe replayed continued-session rows

### LINKED ISSUE #307: Deduplicate cross-file messages for accurate usage cost tracking
## Problem

When a Claude Code session is continued, the new JSONL file contains all
previous messages plus new ones. The same `message.id:requestId` pair
appears in multiple session files. ccusage deduplicates these using a
hash set; agentsview counts each session file independently.

This causes ~6% overcounting in `agentsview usage daily` compared to
ccusage for the same date range.

**Data from real usage:**
- 522K total usage entries across all JSONL files
- 8,185 cross-file duplicate `message.id:requestId` pairs
- 276K extra entries from duplicates (most are within-file streaming
  duplicates, already fixed in the parser)

## Proposed fix

1. Add `message_id` TEXT column to the `messages` table (nullable,
   extracted from JSONL `message.id` field)
2. Populate it in the claude and codex parsers during
   `extractClaudeTokenFields` / equivalent
3. Non-destructive migration: ALTER TABLE ADD COLUMN + backfill on
   next full resync
4. In the usage query, deduplicate by `message_id` using
   `ROW_NUMBER() OVER (PARTITION BY message_id ORDER BY ordinal DESC)`
   or similar, keeping only the last occurrence

## Context

This is follow-up work from the `feat/usage-cost-tracking` branch which
added `agentsview usage daily` and `agentsview usage statusline`
commands. The streaming duplicate fix (consecutive same-message.id
entries within a single file) was already implemented in the parser
via `collapseStreamingDuplicates()`.

rodboev:
Continued Claude sessions can still inflate usage totals when replayed history rows do not carry a complete `messageId:requestId` pair. Current main already deduplicates rows that have both Claude identifiers, but replayed rows that only preserve the shared source identity still fall through and count once per file.

This change keeps the existing Claude pair as the preferred usage key and adds a narrower fallback for replayed message rows that lack a complete pair. SQLite, PostgreSQL, and DuckDB now use the same precedence in usage aggregation and activity reports, so daily totals, top-session rankings, per-session usage, and activity-report cost totals stay aligned. The scope stays in usage accounting and activity-report dedup, without adding schema columns or parser changes.

Review the shared dedup-key selection first. The important behavior is that complete Claude pairs keep their current semantics, incomplete replay rows deduplicate by persisted source identity, and rows that still lack every stable key continue to count individually.

Fixes #307

### DIFF
--- a/internal/activity/activity.go
+++ b/internal/activity/activity.go
@@ -56,8 +56,10 @@ type UsageRow struct {
 	Timestamp       string // ts, RFC3339 or ""
 	OutputTokens    int
 	Cost            float64
+	Agent           string
 	ClaudeMessageID string
 	ClaudeRequestID string
+	SourceUUID      string
 	UsageDedupKey   string
 }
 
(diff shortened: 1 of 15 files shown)


rodboev:
Activity reports now carry the same fallback too. The shared activity dedup path uses the full Claude pair when it exists, falls back to `agent + source_uuid` when the pair is incomplete, and still leaves rows with no stable key alone.

SQLite, PostgreSQL, and DuckDB now pass the extra fields through their activity-report usage rows, so the report totals stay aligned with the daily, top-session, and per-session usage views on replayed continued-session rows.

wesm:
Great. Merging this

PR #829, opened 2026-06-23
fix(postgres): reset push watermarks on target changes

### LINKED ISSUE #186: feat: detect PG URL change and auto-reset watermarks
When the PostgreSQL connection URL changes (e.g. pointing to a different database), local push watermarks become stale and the incremental push skips most sessions.

The fix would be to store a hash or identifier of the target PG database alongside the watermark, detect when it changes, and automatically clear watermarks to trigger a full re-push.

Current workaround: run \`pg push --full\` after re-pointing.

rodboev:
`pg push` still treats its local incremental watermark as if it belonged to one timeless PostgreSQL destination. A brand-new target already forces a repush because its push marker is missing, but switching back to a previously used populated database or schema can still reuse the old watermark and silently skip sessions that never reached the current target.

This change stores a versioned fingerprint for the effective PostgreSQL target, including schema, next to the existing push watermark state. When the fingerprint changes, `pg push` clears only the persisted push watermark and boundary state before selecting sessions, so the new target gets a full repush instead of inheriting stale incremental state.

The scope stays inside `internal/postgres`. Status remains informational, config semantics stay singular, and the new fingerprint is designed to be reusable later if multi-target push work namespaces watermarks per destination. Focused local validation covered `go test ./internal/postgres -count=1`, the new fingerprint and target-state unit slice, tagged pgtest compile, and `go vet ./internal/postgres`. The env-backed pgtest regression for live target switching is in the diff but was not run locally because `TEST_PG_URL` is unset in this environment.

Fixes #186

### DIFF
--- a/internal/postgres/connect.go
+++ b/internal/postgres/connect.go
@@ -2,7 +2,9 @@ package postgres
 
 import (
 	"context"
+	"crypto/sha256"
 	"database/sql"
+	"encoding/hex"
 	"fmt"
 	"log"
 	"net"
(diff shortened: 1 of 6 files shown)


rodboev:
Fixed in e8565e8. The reset now treats filtered boundary state as local push state too, so a filtered push to one PostgreSQL target can't suppress a later push to a different target. I also added unit coverage for the reset helper and a pgtest that does a filtered push to schema A, then pushes to schema B to prove the second target gets the session.

rodboev:
Fixed in f8bfd13. pgTargetFingerprint now includes the full parsed endpoint sequence, including pgx fallback host and port entries, so a multi-host DSN change can't reuse the previous target's local watermark state. I also extended the fingerprint test to cover fallback host and fallback port changes without making password or application-name noise affect the hash.

PR #830, opened 2026-06-24
feat(postgres): add named push targets

### LINKED ISSUE #517: Support multiple named [pg.NAME] targets with per-target config
## Background

pg push currently assumes a single PostgreSQL target per machine. PGConfig is a flat struct field on Config.PG, and ResolvePG() produces exactly one resolved config. runPGPush loads it and constructs one postgres.Sync. There is no way to declare multiple targets in config.toml, no way to select among them on the CLI, and no shared default to fall back to.

This blocks a few real use cases:
- Pushing different project subsets to different remotes (e.g. work projects to a team PG, personal projects to a private one).
- Pushing the same data to multiple remotes for redundancy or distinct consumers.
- Running multiple cron jobs against different targets without juggling config files or env-var wrappers.

The PG schema itself already supports multi-source data (sessions.machine TEXT NOT NULL partitions by machine name). The gap is purely client-side: config, CLI, and per-target sync state.

## Proposal

Extend config.toml to allow named [pg.NAME] blocks, each carrying the full set of current PGConfig fields:

```toml
default_pg = "work"   # required when more than one target is defined

[pg.work]
url = "postgres://..."
schema = "agentsview"
machine_name = "laptop-work"
projects = ["acme", "infra"]

[pg.personal]
url = "postgres://..."
machine_name = "laptop-personal"
exclude_projects = ["acme"]
```

CLI shape:

- `pg push` → push to the default target.
- `pg push <name>` → push to a specific target.
- `pg push --all` → push to every defined target, sequentially.
- `pg status [<name>]` and `pg status --all` mirror the same shape.

_Backward compatibility_ - A bare [pg] block continues to work and is treated as the implicit single target. pg push with no positional argument behaves exactly as today. Mixing [pg] and [pg.NAME] blocks is rejected at config-load time with a clear error.

 ### Why this fits the existing shape

- The PG schema already partitions by machine, so two targets pointing at the same DB with different machine_name values coexist correctly.
- PGConfig's current fields (url, schema, machine_name, projects, exclude_projects, allow_insecure) just move into a per-target block.
- #412 establishes the precedent for multi-entry remote config ([[remotes]]) on the pull side. This issue is the push-side analog.
- postgres.Sync is already constructed per-call inside runPGPush / runPGStatus, so producing one instance per target requires no global-state surgery.

### Behavior details worth pinning down

- Per-target sync watermarks. The keys last_push_at and last_push_boundary_state are stored as single global SQLite rows. Each target needs its own watermark otherwise pushing to target B will skip work already counted as "done" for target A. Migration: rekey existing values to last_push_at:<default-name> on first run after upgrade. Related to #186, which already flags watermark fragility.
- Reserved names. local is already reserved as a machine_name value. Target names should get a similar reserved-word check (e.g. block local, possibly all since it collides with the flag).
- machine_name provenance. Per-target machine_name interacts with #332 (push currently stomps the original sess.Machine). Both issues should converge on a clear rule: per-target machine_name applies to originally-local sessions; remote-sourced sessions keep their original attribution.
- CLI exit semantics for --all. On a per-target failure: log, continue with remaining targets, exit non-zero at the end. Don't abort on first failure.
- Env-var overrides. Current AGENTSVIEW_PG_URL / AGENTSVIEW_PG_SCHEMA / AGENTSVIEW_PG_MACHINE need a defined target-resolution rule. simplest: they override the default target only.

### Out of scope for v1

- Auto-push from serve (analog of #412's auto-pull). Follow-up once the multi-target plumbing lands. Suggested shape: auto_push = true per [pg.NAME] with an optional push_interval.
- Hot-reload of config changes. A restart on config change is fine for v1.
- pg push --target=<name> flag in addition to the positional. Positional matches git remote ergonomics; a flag is duplicative.

### Alternatives considered

- [[pg]] array-of-tables with a name field. Matches #412's [[remotes]] shape. [pg.NAME] reads better when targets have stable identities used as CLI arguments.
- Profile-based config (single [pg], selected via AGENTSVIEW_PG_PROFILE). Works for the "switch between two configs" case but doesn't enable pg push --all and forces every invocation through a shell wrapper.
- Status quo + per-invocation env vars. Already possible, but doesn't compose with cron and surfaces no shared defaults.

### Open questions

1. Does this overlap with pg-sync refactor spec mentioned in #186?
2. Preference between [pg.NAME] and [[pg]] array-of-tables?
3. Should default_pg be required when multiple targets are defined, or should the config silently pick one (alphabetical, first-declared)?

### Related
  - #412 pull-side analog: periodic SSH remote sync from serve with [[remotes]]
  - #186 watermark fragility on PG URL change; directly relevant to per-target watermarks
  - #332 pg push machine-name stomping; interacts with per-target machine_name semantics
  - #108 original "multi-machine aggregation" ask (closed)

rodboev:
`pg push` and `pg status` still assume one global PostgreSQL destination, which makes multiple remotes clumsy to operate and unsafe to share because one target can advance the same local watermark another target depends on, and the default-target service status path still reports the unscoped legacy watermark.

This change adds named `[pg.NAME]` targets and `default_pg` while keeping the existing single `[pg]` shape working as it does today. `pg push` and `pg status` can now use the default target, select one named target, or run `--all` sequentially. Target selection now resolves only the target being used, so a broken sibling target no longer blocks `pg push archive`, `pg status archive`, or the per-target `--all` loop. `pg push --watch` follows the default target or one selected named target, and rejects `--all` so the long-running multi-target watcher problem stays out of scope. The existing `AGENTSVIEW_PG_*` overrides stay supported, but they apply only to the effective default target instead of trying to rewrite every named entry.

The sync layer now scopes push watermark state by target name in named-target mode and lazily migrates the legacy global watermark into the selected default target on first use. Follow-up review fixes keep `pg_push_marker_id` local to the SQLite database so `pg push --all` does not self-conflict across named targets, preserve `Last push: never` for healthy `pg service status` output, and document that legacy `[pg]` field names are unavailable as named target identifiers. `pg service status` now reads the effective default target scope, and the PostgreSQL docs now describe named targets, `default_pg`, default-target-only environment overrides, and the default-target-only service and watch behavior. `pg serve`, service install, and the broader long-running multi-target watch problem still stay out of scope here so the PR stays reviewable.

Fixes #517


rodboev:
Addressed in afc6ea0f.

rodboev:
Follow-up details for afc6ea0f:

- named-target validation now stays on PG resolution paths instead of config finalization, so unrelated commands can still load config
- status reads are side-effect free again; they read the scoped watermark first, then fall back to the legacy default watermark without migrating state
- the duplicate per-target classifier setup is gone
- target ordering now comes from shared config helper logic without eagerly resolving sibling targets

mjacobs (commented):
Nice work — the named-target design is clean and the genuinely tricky parts are handled carefully. Things I specifically liked:

- Validation is correctly lazy: `finalize()` (internal/config/config.go:1057) no longer validates PG targets, so a typo'd `default_pg` still lets purely-local commands load and only PG resolution errors — `TestLoadMinimal_DefersNamedPGValidationForNonPGCommands` pins it.
- Per-target watermark scoping is applied uniformly — every `last_push_at`/boundary/fingerprint access in `Push` moved from `s.local` to the scoped store from `effectiveSyncState()`, with `pg_push_marker_id` deliberately kept DB-global (push.go) so `pg push --all` doesn't self-conflict.
- The legacy→default migration (internal/postgres/sync.go:43-94) is idempotent and self-healing: a legacy key is cleared only after its value lands in scope, so a crash leaves the value in both places, never neither.
- The status read path is genuinely read-only: `ReadLastPushAt` (sync.go:342) builds its scoped store with `migrateLegacy=false` and only falls back to a plain read of the legacy key.

Two minor, non-blocking items:

1. (low — test placement) The three `TestScopedSyncStateStore*` tests in `internal/postgres/sync_test.go` cover the most safety-critical new logic (the copy-then-clear migration), but that file is `//go:build pgtest`, so the default `make test` skips them — `go test -tags fts5 -run TestScopedSyncStateStore ./internal/postgres/...` reports `[no tests to run]`. They need no PostgreSQL (`testDB` is just `db.Open(tmp)`). Moving them — and `testDB` — into the existing non-pgtest `internal/postgres/sync_unit_test.go` would let the default suite guard the migration loop. (Non-tagged files compile into the pgtest build too, so the remaining PG tests keep seeing `testDB`.)
2. (nit) Header output differs between sibling commands for a single named target: `pg status work` prints `Target: work` (pg.go:280) but `pg push work` prints nothing (pg.go:107). Cosmetic; aligning push to the status condition would let `pg push work` confirm which target it wrote to.

One question: is `pg push --all` against a legacy single `[pg]` block (currently accepted as a plain single push) the intended contract, or would you rather it error since there's nothing to fan out over? Harmless either way.

Also checked the earlier automated-review items against the latest commit: the validation-scoping and status-read fixes are both in `afc6ea0f` and tested, the per-target classifier double-call is gone, and the unused `ResolvePGTargets()` is at most a minor cleanup (the CLI fan-out shares one ordering source via `PGTargetNames()`).

wesm:
rebasing

PR #831, opened 2026-06-24
feat(ui): add UI text-size scaling and high-contrast mode

wesm:
Adds two Settings -> Appearance accessibility controls, both persisted in localStorage. Frontend-only.

## Text size (90-130%)

A uniform UI scale that reuses the existing root-zoom mechanism: one effect sets `document.documentElement.style.zoom` to the font scale composed with the desktop window zoom -- `(IS_DESKTOP ? zoomLevel/100 : 1) * (fontScale/100)`, rounded to 4 decimals. On the web only the font scale applies; on desktop the two multiply. `agentsview-font-scale` persists everywhere; the desktop `agentsview-zoom-level` stays desktop-only.

## High contrast

A `high-contrast` root class (composing with light/dark) drives `:root.high-contrast` and `:root.dark.high-contrast` overrides of the neutral text/border tokens and `--accent-blue`, plus a strengthened focus ring. Every overridden token clears WCAG AA against the real backgrounds. The shell and the token-driven majority of the transcript are covered by the token overrides; the three literal-colored transcript components (CallGroup, SubagentCalls, CallRow) get colocated `:global(.high-contrast)` overrides. Data-visualization palettes and category identity colors (`--cat-*`) are left unchanged so their encoded meaning is preserved.

## Where to look

- `frontend/src/lib/stores/ui.svelte.ts` -- state, setters, the composed-zoom and high-contrast effects, and the persistence split.
- `frontend/src/app.css` -- high-contrast token overrides and focus ring.
- `frontend/src/lib/components/settings/AppearanceSettings.svelte` -- the controls.

## Note

This branch and #821 both edit `ui.svelte.ts` in adjacent regions; whichever merges second may need a small conflict resolution.

### DIFF
--- a/frontend/e2e/appearance-a11y.spec.ts
+++ b/frontend/e2e/appearance-a11y.spec.ts
@@ -1,0 +1,91 @@
+import { test, expect } from "@playwright/test";
+import { SessionsPage } from "./pages/sessions-page";
+
+function readZoom(page: import("@playwright/test").Page): Promise<string> {
+  return page.evaluate(() =>
+    document.documentElement.style.getPropertyValue("zoom"),
+  );
+}
+
+test.describe("Appearance accessibility", () => {
+  test("text size scales the UI on web without horizontal overflow", async ({
+    page,
+  }) => {
+    await page.addInitScript(() => {
+      localStorage.setItem("agentsview-font-scale", "130");
+    });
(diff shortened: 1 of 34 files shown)


wesm:
tested locally, will merge when CI green

PR #832, opened 2026-06-24
docs(frontend): document i18n guidance

### LINKED ISSUE #824: Document frontend i18n guidance in DESIGN.md
PR #822 added initial frontend i18n support and follow-up design-system guardrails, including `DESIGN.md`. A remaining docs gap is that `DESIGN.md` does not yet explain the expected i18n workflow for frontend UI copy.

Add guidance covering when new UI strings should go through locale dictionaries, how to name keys, how to update all supported locale files, and how shared components should accept translated labels rather than hard-coded strings. This should make future frontend work treat i18n as part of the design-system/component contract, not an afterthought.

Reference: https://github.com/kenn-io/agentsview/pull/822

icatw:
Adds frontend i18n guidance to DESIGN.md so UI copy is treated as part of the design-system contract. The guidance covers locale key ownership, keeping supported locale files aligned, passing translated copy into shared components, and leaving technical identifiers unchanged.

Closes #824

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -19,6 +19,27 @@ control.
 - If a new control pattern is genuinely needed, add or extend a shared
   component first, then use it from feature components.
 
+## Internationalization
+
+- Treat user-facing frontend UI copy as part of the component contract. New
+  headings, labels, button text, placeholders, tooltips, aria labels, empty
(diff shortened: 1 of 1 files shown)


mariusvniekerk:
thanks

PR #833, opened 2026-06-24
feat(frontend): localize shell navigation

icatw:
Localizes the frontend app shell and high-frequency navigation surfaces using the existing svelte-i18n setup.

This covers the header navigation, command palette, status bar, sidebar session filters, shared range/project controls, and active filter chips. It keeps technical identifiers such as agent names, model names, file paths, commands, and export formats unchanged.

Session detail actions and analytics/report page content are intentionally left for follow-up translation PRs so this review stays focused on the shell/navigation layer.


mariusvniekerk:
Translated locale screenshot from Playwright:



mariusvniekerk:
@icatw took another look at the recommended i18n libraries for modern svelte so changed it out for paraglide as per https://svelte.dev/docs/cli/paraglide

functionally still behaves the same.

icatw:
Thanks for updating this to Paraglide. That makes sense to me, and I’ll use the Paraglide workflow for follow-up i18n PRs.

PR #834, opened 2026-06-24
refactor(service): add Search and UsageSummary to the SessionService interface (prep: MCP feature)

mjacobs:
The full-text session search and the aggregate usage summary were served
by HTTP handlers (bypassing the `SessionService` interface, i.e. directly
into referencing `db/server-local` helpers), unlike every other handler.

This is the groundwork for a native `agentsview mcp` subcommand (follow-up
PR), which serves these operations over MCP, but the interface additions
stand on their own.

This adds them to `SessionService` so search and usage-summary handlers
become thin delegating encoders like the rest.

What changed:

- `SessionService` gains `Search(SearchRequest)` and
  `UsageSummary(UsageRequest)`, implemented in both the direct and HTTP
  backends.
- `Search` normalizes the query via `db.PrepareFTSQuery`, clamps the
  limit, and returns `ErrSearchUnavailable` when the store has no FTS
  index. `humaSearch` now delegates to it.
- The usage fold/cache helpers, the per-total response types, and the
  summary response shape move from `internal/server` to
  `internal/service`. `service.BuildUsageFilter` becomes the single
  usage-filter validator, used by both the summary method on the
  interface and the server's comparison/top-sessions handlers.
  `humaUsageSummary` delegates to `s.sessions.UsageSummary`.

Backend parity: both new methods route through `db.Store`
(`db.Search`, `db.GetDailyUsage`), so SQLite and the PostgreSQL read
store behave identically — there is no `*db.DB`-only path. A read-only
store surfaces as `db.ErrReadOnly` over both transports (mapped to 501
by the HTTP layer), matching the existing usage handlers.

Scope notes: the prior-period `Comparison` stays in `internal/server`
(it backs a separate `/usage/comparison` endpoint and is never part of
`/usage/summary`), so the `/usage/summary` JSON is byte-for-byte
unchanged.

Where to look: `internal/service/{service,direct,http,usage}.go` for the
interface additions; `internal/server/{huma_routes_search,
huma_routes_usage,usage}.go` for the handlers becoming thin delegates.


wesm:
I need to land my daemon PR and then I'll rebase this

wesm:
rebasing

Issue #281, opened 2026-04-04
bulk delete sessions


younes-io:
is there a way to bulk delete sessions? (with filters, etc)

rodboev:
This looks fixed on current main by PR #835, which adds batch select and delete sessions. The issue can probably be closed.

PR #835, opened 2026-06-24
feat: batch select and delete sessions in sidebar

huaiyuWangh:
## Summary

Add multi-select batch delete functionality to the session sidebar, allowing users to select and delete multiple sessions at once instead of deleting them one by one.

Ref #281

## Changes

**Backend:**
- New `SoftDeleteSessions(ids []string)` DB method for batch soft-delete in a single transaction (batches of 500)
- New `POST /api/v1/sessions/batch-delete` endpoint accepting `{ session_ids: [...] }`, following the existing bulk-star pattern
- Read-only stubs added for PG and DuckDB stores
- Store interface updated with `SoftDeleteSessions`

**Frontend:**
- Multi-select toggle button in the session list header (checkmark icon)
- Per-item checkboxes when select mode is active; clicking items toggles selection instead of navigation
- Batch toolbar showing selection count with `Delete` and `Cancel` buttons, plus `All/Clear` for select-all
- `batchDeleteSessions(ids)` store method calls the new endpoint, updates local state, tracks undo toast entries, and exits select mode automatically

**Tests:**
- `TestSoftDeleteSessions` covering normal batch, pre-deleted skip, empty input, and non-existent IDs

### DIFF
--- a/frontend/src/App.svelte
+++ b/frontend/src/App.svelte
@@ -551,7 +551,7 @@
         if (!last) return;
         btn.disabled = true;
         try {
-          await sessions.restoreSession(last.id);
+          await sessions.restoreRecentlyDeleted(last);
         } catch {
           // restore failed — toast will remain
         } finally {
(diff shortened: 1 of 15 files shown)


wesm:
looking at this

wesm:
thank you!

PR #836, opened 2026-06-24
feat(config): add per-remote sync interval

### LINKED ISSUE #412: Feature: periodic SSH remote sync from `serve` (config-driven `[[remotes]]` list)

rodboev:
The serve daemon already runs a local sync every fifteen minutes, but the `[[remote_hosts]]` entries introduced in #600 are only ever pulled by the one-shot `agentsview sync` command, so a long-running daemon never refreshes data from configured remotes on its own. Users running the dashboard as a persistent service have to drive remote pulls out-of-band with cron or a wrapper script, which is exactly the kind of scheduling the daemon is otherwise responsible for.

This adds an optional `interval` field to each `[[remote_hosts]]` entry, expressed as a duration string such as `"5m"` or `"1h"`. The field is decoded natively the same way `events_coalesce_interval` is, and it defaults to zero, which means "do not periodically sync this host." Configs written before this change behave exactly as they do today, and nothing starts pulling from a remote unless the user explicitly sets an interval.

On the daemon side, the periodic sync routine now launches one lightweight ticker goroutine per remote host that has a positive interval, each calling the same single-host sync path the CLI uses, in incremental mode, logging any failure without disturbing the others or the local periodic sync. After a successful remote sync pass, the goroutine emits a `data_changed` SSE event through the serve broadcaster so connected UI clients pick up the new data immediately rather than waiting for the next local periodic sync. The emitter is threaded as a `sync.Emitter` interface to keep the dependency direction clean; the CLI path continues to run without one. The remote host list is validated before these goroutines start; if validation fails the daemon logs a warning and keeps serving with local sync intact rather than refusing to start. The existing fifteen-minute local sync loop is left untouched, and because per-host remote sync reuses the already-tested `runRemoteSyncOnce` path, the new surface area is the configuration field, its validation, the scheduling glue, and the emit-on-success behavior.

Tests cover loading the new field from a config file, validating the negative-interval case, and the emit-on-success/no-emit-on-error/nil-emitter behavior of the sync loop.

Closes #412

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -813,7 +813,9 @@ func startPeriodicSync(
 	if validRemotes {
 		for _, rh := range cfg.RemoteHosts {
 			if rh.Interval > 0 {
-				go startRemoteHostSync(cfg, database, engine, rh, emitter)
+				go startRemoteHostSync(
+					cfg, database, engine, rh, emitter, idleTracker,
+				)
 			}
 		}
 	}
@@ -834,6 +836,7 @@ func startRemoteHostSync(
 	engine *sync.Engine,
 	rh config.RemoteHost,
 	emitter sync.Emitter,
+	idleTracker *server.IdleTracker,
 ) {
 	syncFn := remoteHostSyncFunc(
 		cfg, database, engine, rh,
@@ -841,7 +844,7 @@ func startRemoteHostSync(
 			return rs.Run(ctx)
 		},
 	)
-	runRemoteHostSyncLoop(rh.Host, rh.Interval, syncFn, emitter, nil)
+	runRemoteHostSyncLoop(rh.Host, rh.Interval, syncFn, emitter, idleTracker, nil)
 }
 
 type remoteSyncExclusiveRunner interface {
@@ -882,7 +885,14 @@ func remoteHostSyncFunc(
 // runRemoteHostSyncLoop drives the per-host sync ticker. syncFn returns
 // the number of sessions synced so we only emit when data changed.
 // When done is non-nil, closing it stops the loop.
-func runRemoteHostSyncLoop(host string, interval time.Duration, syncFn func() (int, error), emitter sync.Emitter, done <-chan struct{}) {
+func runRemoteHostSyncLoop(
+	host string,
+	interval time.Duration,
+	syncFn func() (int, error),
+	emitter sync.Emitter,
+	idleTracker *server.IdleTracker,
+	done <-chan struct{},
+) {
 	ticker := time.NewTicker(interval)
 	defer ticker.Stop()
 	for {
@@ -892,7 +902,17 @@ func runRemoteHostSyncLoop(host string, interval time.Duration, syncFn func() (i
 		case <-ticker.C:
 		}
 		log.Printf("Running scheduled remote sync for %s...", host)
-		synced, err := syncFn()
+		finishWork, ok := idleTracker.BeginWork()
+		if !ok {
+			log.Printf("scheduled remote sync %s skipped: daemon is shutting down", host)
+			continue
+		}
+		var synced int
+		var err error
+		func() {
+			defer finishWork()
+			synced, err = syncFn()
+		}()
 		if err != nil {
 			log.Printf("scheduled remote sync %s: %v", host, err)
 			continue
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -21,6 +21,7 @@ import (
 	"go.kenn.io/agentsview/internal/secrets"
 	"go.kenn.io/agentsview/internal/server"
 	"go.kenn.io/agentsview/internal/signals"
+	"go.kenn.io/agentsview/internal/ssh"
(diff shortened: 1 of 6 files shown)


rodboev:
Addressed. `startRemoteHostSync` now takes a `sync.Emitter` and calls `Emit("remote")` after each successful remote sync pass. The broadcaster is threaded from the serve setup through `startPeriodicSync` into the per-host goroutines. Emit is skipped on sync failure and when the emitter is nil (CLI path). Three tests cover the success, error, and nil-emitter cases via an extracted `runRemoteHostSyncLoop` helper.

wesm (approved):
LGTM. I will merge (and rebase if necessary) after landing #826

wesm:
rebasing

wesm:
good, merging. thank you!

PR #837, opened 2026-06-24
fix(parser): compute per-turn Gemini context tokens as delta

### LINKED ISSUE #348: Context size for gemini thread not correct
For a gemini thread this is what I see in agentsview: 217.2k ctx / 28.2k out

But gemini cli reports that the context used is 10% (sadly it does not report actual token count but I presume that means it is something around 100k). Also when I do `/stats session` to get the output token count, I get report per model used but the sum does not equal to 28.2k but 290 + 7004 + 23340 = 30634.

```
Model                           Reqs  Input Tokens   Cache Reads Output Tokens 
───────────────────────────────────────────────────────────────────────────────
gemini-2.5-flash-lite              5        42,306             0           290 
  ↳ utility_router                 4        41,287             0           282 
  ↳ utility_summarizer             1         1,019             0             8 
gemini-3.1-pro-preview            22       396,626       264,423         7,004 
  ↳ main                          22       396,626       264,423         7,004 
gemini-3-flash-preview            47     2,499,408     1,930,600        23,340 
  ↳ main                          28     2,148,608     1,699,478        21,262 
  ↳ subagent                      19       350,800       231,122         2,078 
```

When I look into the json file, at the last message of the thread I see

```json
      "tokens": {
        "input": 109007,
        "output": 99,
        "cached": 108219,
        "thoughts": 0,
        "tool": 0,
        "total": 109106
      },
```

When I grep the file for the `input` field I can see it is monotonically increasing. Given the analysis in
- #106 

I think the value 217.2k for context size is not correct as it is said to be the max value of the `tokens.input` field.

Getting the sum of `tokens.output` I am getting the same value as agentsview

```
jq "[.messages[].tokens.output // 0] | add" "<session-path.json>"
```

so I am not sure here which one is more correct but they are definitely different.

The context size attached to the messages in the thread is also approximately double of what is in the json input field.

rodboev:
Gemini CLI records `tokens.input` as a monotonically increasing cumulative total, a running sum of every input token consumed up to that point in the session rather than the input for a single turn. The parser copies that raw cumulative value straight into each message's context size at `internal/parser/gemini.go:269` (`ContextTokens: tok.Input + tok.Cached`) and into the normalized `TokenUsage` JSON via `normalizedGeminiTokenUsage`, so the displayed per-turn context window and the per-message usage/cost data both inflate steadily across the session: a tenth message that actually consumed roughly fifteen thousand context tokens is shown using the full running total of every prior turn. The same value also feeds the session-level peak, which `accumulateMessageTokenUsage` derives as the maximum `ContextTokens` across messages, so the reported peak context is simply the final cumulative figure rather than the largest real per-turn window.

This computes the per-turn values as deltas across both surfaces. A single pass (`applyGeminiCumulativeDeltas`) walks each session's messages in order and, for every message that carries token counts, parses the cumulative `input` and `cached` values from `TokenUsage`, computes independent per-field deltas, rebuilds `TokenUsage` with the corrected values, and derives `ContextTokens` from the individual deltas. Output tokens are already per-turn in Gemini and stay untouched. Each field resets independently when its cumulative count decreases (context compaction), treating the current value as a fresh baseline rather than emitting a negative size. The pass runs as the first step of `buildGeminiSession`, before token aggregation, so per-message context sizes, per-message usage/cost JSON, and the session peak are all computed from the corrected per-turn values. The parser callers and `parseGeminiMessage` keep their existing shapes; the cumulative-to-delta conversion is centralized in one place that both the single-document and JSONL session paths already flow through.

A parser test synthesizes an inline Gemini session with increasing cumulative input and cached counts and asserts that each message reports per-turn deltas in both `ContextTokens` and `TokenUsage`, and that the session peak reflects the largest delta rather than the final running total.

Closes #348

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -237,9 +237,10 @@ import (
 // classification, so historical skill usage is backfilled on
 // re-parse.)
 //
+// (51: Gemini cumulative-to-delta token reparse.)
 // (17: Codex <skill> template filtering.)
 // (16: <turn_aborted> system messages.)
-const dataVersion = 50
+const dataVersion = 51
 
 const tokenCoverageRepairStatsKey = "token_coverage_repair_v1"
 
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -696,8 +696,8 @@ func TestMigration_ToolResultEventsTable(t *testing.T) {
 }
 
 func TestCurrentDataVersionSanitizedMessageShape(t *testing.T) {
-	assert.Equal(t, 50, CurrentDataVersion(),
-		"sanitized message and session shape requires a data version bump")
+	assert.Equal(t, 51, CurrentDataVersion(),
+		"Gemini cumulative-to-delta reparse requires a data version bump")
 }
 
 func TestInsertMessages_PreservesToolResultEvents(t *testing.T) {
--- a/internal/parser/gemini.go
+++ b/internal/parser/gemini.go
@@ -276,19 +276,46 @@ func parseGeminiMessage(
 	}, true
 }
 
-func applyGeminiContextDeltas(messages []ParsedMessage) {
-	prev := 0
+func applyGeminiCumulativeDeltas(messages []ParsedMessage) {
+	var prevInput, prevCached int
 	for i := range messages {
 		if !messages[i].HasContextTokens {
 			continue
 		}
-		cumulative := messages[i].ContextTokens
-		delta := cumulative - prev
-		if delta < 0 {
-			delta = cumulative
+
+		var usage struct {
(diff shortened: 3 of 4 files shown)


rodboev:
Addressed. The delta pass now tracks cumulative `input` and `cached` separately by parsing `TokenUsage` JSON, computes independent per-field deltas (with independent reset on compaction), rebuilds `TokenUsage` with the delta values, and derives `ContextTokens` from the individual deltas. `output_tokens` stays as-is since Gemini output counts are already per-turn. The existing fixture test asserts the corrected `TokenUsage` fields alongside `ContextTokens` and `PeakContextTokens`.

PR #841, opened 2026-06-24
feat(frontend): localize UI into Chinese and default to English

Junt184:
Title:
feat(frontend): localize UI into Chinese and default to English

Description:
Adds full Simplified Chinese (zh-CN) localization for the SPA and makes
English the default interface language.

## What changed

- Wires ~80 Svelte components and `App.svelte` through `svelte-i18n`
  (`$_`), replacing hard-coded English strings across the
  header/navigation, sidebar and session list, message/tool/thinking
  content, the session analysis panel, analytics dashboards (heatmaps,
  velocity, tool/skill usage), usage, trends, insights, activity, all
  modals, filters, and every settings section.
- Expands `en.json` and `zh-CN.json` to 869 keys each with full parity.
  Counts use ICU plurals and dynamic sentences use parameterized
  messages rather than concatenated fragments. Technical identifiers
  (agent/model names, CLI snippets, file paths, API paths) are left
  untranslated.
- Defaults the interface to English: `chooseInitialLocale` now honors
  only an explicit stored preference instead of auto-detecting the
  browser language, so the UI stays English until the user opts into
  another locale via Settings → Language.

## Fixes

- The language-switcher dropdown was clipped by the section box.
  `SettingsSection` gains an opt-in `allowOverflow` prop so a
  typeahead popover can escape the rounded container; the header keeps
  its own corner clipping.

## Notes

- `initI18n()` is now called in the vitest setup so component tests can
  format messages; tests resolve to the `en` locale, so existing
  English-facing assertions are unaffected.
- Shared controls (e.g. `OptionTypeahead`) continue to receive
  translated labels/placeholders from their callers rather than holding
  copy internally.

## Where to look

- `frontend/src/lib/i18n/index.ts` — locale selection / default.
- `frontend/src/lib/i18n/locales/{en,zh-CN}.json` — dictionaries.
- `frontend/src/lib/components/settings/SettingsSection.svelte` —
  `allowOverflow`.


mariusvniekerk:
we've decided to not use svelte-i18n.

PR #852, opened 2026-06-24
fix(insights): stop forcing Gemini sandboxed insight runs

### LINKED ISSUE #224: Bug: Failure to generate insights from Insights to Gemini
V.0.15.0
The following error occurs when generating insights with Gemini.

```
gemini generation failed
stderr
Fatal error: Failed to relaunch the CLI process.
stderr
Error: Sandbox image 'us-docker.pkg.dev/gemini-code-dev/gemini-cli/sandbox:0.31.0' is missing or could not be pulled. Please check the image name, your network connection, or notify gemini-cli-dev@google.com if the issue persists.
stderr
at start_sandbox (file:///C:/Users/user/AppData/Roaming/npm/node_modules/@google/gemini-cli/dist/src/utils/sandbox.js:191:19)
stderr
at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
stderr
at async relaunchOnExitCode (file:///C:/Users/user/AppData/Roaming/npm/node_modules/@google/gemini-cli/dist/src/utils/relaunch.js:12:30)
stderr
at async main (file:///C:/Users/user/AppData/Roaming/npm/node_modules/@google/gemini-cli/dist/src/gemini.js:342:13)
```

rodboev:
Gemini insight generation still fails because agentsview unconditionally passes `--sandbox` to the Gemini CLI. That forces a Docker-backed sandbox bootstrap path, which is exactly the missing-image failure the reporter kept confirming in #224 across multiple app versions, CLI versions, and machines.

This change narrows the fix to the Gemini invocation contract inside the insight runner. It stops forcing the sandboxed path, keeps the existing stream-json integration, and leaves the current model pin untouched so the review stays focused on the known failure mode instead of bundling a separate policy change. Other runners and other Gemini issues, including auth problems, stay out of scope here.

Review should focus on the Gemini CLI args and the fake-binary regression that proves base still sends the sandbox flag while head does not. Local validation stays on the insight runner and the focused Go checks; CI can cover the broader matrix.

Fixes #224

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -85,8 +85,8 @@ Key assumptions reviewers MUST account for:
     depends on it.
 
 16. INSIGHT AGENT SANDBOXING: Not all agent CLIs offer equal
-    sandboxing. Claude (--tools ""), Codex (--sandbox read-only), and
-    Gemini (--sandbox) have verified no-tools/read-only modes.
+   sandboxing. Claude (--tools "") and Codex (--sandbox read-only)
+   have verified no-tools/read-only modes. Gemini currently does not.
     Copilot only supports --disable-builtin-mcps (blocks MCP tools
     but not all tool use). Kiro CLI uses --trust-tools= and
     --no-interactive (disables tool trust prompts but does not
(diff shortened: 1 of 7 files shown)


wesm:
Since gemini is deprecated we should switch over to Antigravity anyway cc @mjacobs

PR #855, opened 2026-06-25
feat(frontend): complete zh-CN localization across remaining pages

Junt184:
Add Chinese translations for ~60 Svelte components that still used hard-coded English strings, bringing Paraglide message coverage from 176 to 825 keys across every locale.

Components now localised: analytics dashboards (heatmaps, velocity, tool/skill usage, grade/outcome distribution, project/agent breakdowns, session health, session shape, top sessions, active filters, activity timeline), usage pages (summary cards, cost attribution, cache efficiency, cost time series, treemap, filter dropdown, top sessions table), trends (term table, line chart, page), insights page (quality patterns, signals, generated archive), activity page, concurrency timeline (preserving #838 overlay axis), all six modals (confirm delete, about, publish, update, shortcuts, resync), every settings section (agent directories, appearance, terminal, GitHub, remote access, worktree mappings), import modal, command palette, pinned page, trash page, perf debug panel, system boundary card, status dot, range picker, session breadcrumb, content panels (signal, code block, tool call group, compact boundary, thinking block, subagent calls/inline, find bar, activity lane, parallel group, call row), and two shared controls.

Every locale stays at 825 keys with full parity. Parameterized messages use Paraglide's {param} syntax; static messages are rendered through m.key_name() calls as described in DESIGN.md.

Fix the language-switcher dropdown being clipped by the section box: SettingsSection gains an allowOverflow prop so a typeahead can escape the rounded container; LanguageSettings opts in.


Junt184:
Thanks for the review. Fixed the XSS regression: ConfirmDeleteModal and
UpdateModal no longer pass session/version data through {@html} — those
values now render via normal escaped interpolation. Removed the
remaining {@html} message calls (publish/import/update hints) too, so
there are no {@html m.*} usages left.

Also rebased onto the latest main and resolved conflicts (the new
session reading-control localizations and #838 overlay axis are
preserved). svelte-check and the build are clean.

PR #860, opened 2026-06-25
feat(insights): add export and publish for generated insights

### LINKED ISSUE #67: Feature: Export and programmatic access to insights
## Problem

The insights feature generates valuable analysis of session activity, but the generated insights are ephemeral — they exist only in the local SQLite database and can only be viewed through the web UI. This limits their utility for:

1. **Preservation** — insights cannot be exported, backed up, or shared
2. **Cross-session analysis** — no way to programmatically analyze insights across multiple time periods
3. **Integration** — insights cannot be consumed by other tools or workflows

For users who want to track patterns over time or use insights as input for other processes, the current architecture requires manually copying content from the UI.

## Use Case

A user working on a long-running project might want to:

- Generate weekly or monthly insights covering extended periods
- Export insights as files for backup or sharing
- Analyze multiple insights together to identify trends (e.g., "what patterns emerged across the last 4 weeks?")
- Feed insights into external tools for further analysis or documentation

Currently, insights are isolated within the database with no export path.

## Observations

Looking at the codebase:

1. **Sessions have export functionality** — `GET /api/sessions/:id/export` downloads HTML, `POST /api/sessions/:id/publish` publishes to GitHub gist
2. **Date range support exists** — the `insights` table has `date_from` and `date_to` fields, and the code supports multi-day ranges, though the UI may not fully expose this
3. **Insights are stored persistently** — they survive server restarts and have IDs for retrieval

This suggests the foundation for exporting/accessing insights already exists in the architecture.

## Possible Directions

Some options to consider (not prescriptive):

**1. Export API**

Add endpoints similar to session export:
- `GET /api/insights/:id/export` — download insight as markdown/JSON
- `POST /api/insights/:id/publish` — publish to gist

This would parallel the existing session export workflow.

**2. Bulk export**

A way to export all insights (or filtered by project/type) for backup:
- UI button: "Export all insights"
- CLI command: `agentsview insights export --output insights.json`

**3. Programmatic access**

For users who want to build on top of insights:
- REST API for listing/fetching insights (already exists for internal use)
- Documentation for external tool integration
- Possibility of an MCP server exposing insights to AI agents

## Workaround

Currently, users can manually copy insight content from the web UI, but this doesn't preserve metadata (date range, type, project) or support bulk operations.

rodboev:
Insights can be generated and reviewed today, but they still stop at a dead end once selected. The insight API only exposes list, get, delete, and generate, and the selected insight pane only shows metadata plus delete, so there is no built-in way to export an insight or publish it the way sessions already can.

This change adds the per-insight surface that sessions already have: HTML export, markdown export, and GitHub Gist publish for a single selected insight. The backend should reuse the existing export and gist helpers so token handling, public versus secret publish behavior, error mapping, and filename sanitization stay consistent, and the frontend should extend the current export helpers and shared publish modal instead of forking a second publish flow. Scope stays intentionally narrow to a selected insight, with no bulk export, CLI work, MCP exposure, or docs-only follow-up folded into the patch.

Reviewers should focus on the shared export and publish plumbing, especially that the new routes stay aligned with the existing session behavior and that the shared publish modal still behaves correctly when opened from both sessions and insights. Focused server coverage for the new routes and focused frontend coverage for the detail-pane actions and publish-target state should be enough here, and CI can carry the broader matrix.

Fixes #67

### DIFF
--- a/frontend/src/lib/api/client-markdown-export.test.ts
+++ b/frontend/src/lib/api/client-markdown-export.test.ts
@@ -1,6 +1,9 @@
 // @vitest-environment jsdom
 import { describe, it, expect, vi, beforeEach, afterEach } from "vite-plus/test";
-import { getMarkdownExportUrl } from "./client.js";
+import {
+  getInsightMarkdownExportUrl,
+  getMarkdownExportUrl,
+} from "./client.js";
 
 const storage = {
   getItem: vi.fn().mockReturnValue(""),
@@ -29,6 +32,12 @@ describe("markdown export URLs", () => {
     );
     expect(getMarkdownExportUrl("sess-123", 1)).toBe(
       "/api/v1/sessions/sess-123/md?depth=1",
+    );
+  });
+
+  it("builds markdown export URL for an insight", () => {
+    expect(getInsightMarkdownExportUrl(42)).toBe(
+      "/api/v1/insights/42/md",
     );
   });
 });
--- a/frontend/src/lib/api/client.ts
+++ b/frontend/src/lib/api/client.ts
@@ -315,6 +315,10 @@ export function getExportUrl(sessionId: string): string {
   return `${getBase()}/sessions/${sessionId}/export`;
 }
 
+export function getInsightExportUrl(insightId: number): string {
+  return `${getBase()}/insights/${insightId}/export`;
+}
+
 /** Get markdown export URL for a session, with optional child depth. */
 export function getMarkdownExportUrl(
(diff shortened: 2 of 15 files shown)


rodboev:
Addressed the roborev note by keeping `watchEvents` available in the `client.js` mock for `InsightsPage.test.ts`. The test still stubs the export call, but it no longer shadows the event-stream export that the real events store imports.

wesm:
looking at this

PR #861, opened 2026-06-25
feat(analytics): add dashboard model filter

### LINKED ISSUE #633: Model filter on the main dashboard
Would it be possible to add a filter on model name on the main dashboard? 

I would even like a dashboard that shows model usages in sessions if possible. I see this filter only exists on "Usage" page which is useful, but since I am using subscription-based Claude Code I don't get to see any selection there though

rodboev:
The analytics dashboard can already narrow by project, machine, agent, and time,
but still rolls every chart and summary together across models. This adds a
model-name filter to the main dashboard. A comma-separated `model` value is
threaded through the analytics route layer, the shared `AnalyticsFilter`, the
analytics/trends/filtered-model lookup helpers, and the dashboard request state.

When a model filter is active, the dashboard first scopes to sessions that contain
at least one matching model message, then derives message, token, tool-call,
trend-term, session-shape, and velocity metrics from the matching model rows within
that session set. Summary totals, activity, heatmap, projects, tools, skills, top
sessions, signals, trend terms, session shape, and velocity stay aligned across the
SQLite, PostgreSQL, and DuckDB backends, including mixed-model sessions and hour/day
slices.

`model` is the first message-grain analytics filter — a session spans many models —
so it can't ride the session-grain WHERE builders the other filters use. Each panel
instead needs a user->assistant pairing pass over candidate message rows. Rather
than copy that pairing into every panel, the model membership, user-turn pairing,
and day/hour match now live in one shared streaming reducer (the `internal/db`
message-scope helpers); each backend keeps its own candidate-row SQL (dialect,
placeholders, driver) and feeds rows through the shared reducer and its stats/timing
projections. Analytics and trends across all three backends share one implementation
instead of six near-duplicate pairing loops. The reducer requires only that rows
arrive grouped by session with ascending per-session ordinal — what
`ORDER BY session_id, ordinal` yields under any collation — so PostgreSQL's
collation-dependent ordering stays correct alongside SQLite and DuckDB.

On the frontend, the analytics store gains model-filter state, request params,
clear/toggle helpers, and active-filter chips, and the toolbar gains a model
dropdown that keeps known models stable across refreshes and filtered reloads.

One known limitation: the summary aggregation (median, p90, concentration) is still
computed per backend rather than shared; folding it into the same shared path is a
pre-existing cleanup left out here to keep this change scoped to model filtering.

Reviewers: the shared reducer (`internal/db/messagescope.go`,
`messagescope_reducer.go`) and the per-backend candidate-row resolvers
(`internal/{db,postgres,duckdb}/analytics_scope.go`) are the core; the rest is
route, filter, and frontend wiring. New analytics tests cover the filter across all
three backends.

Fixes #633

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -80,6 +80,52 @@ navigating to explicit range URLs also pins the range while the
 bare page URL returns to rolling mode. The `All` preset always
 pins to `(earliest_session, today)`.
 
+### Model Filter
+
+The dashboard toolbar includes a **Model** dropdown that scopes
+every panel to one or more AI models. By default the button reads
+**Model: All** and nothing is filtered.
+
+<!-- Screenshot placeholder: capture the dashboard Model dropdown
+open with a model selected, then add it here as
+/assets/generated/screenshots/analytics-model-filter.png after
+release. See docs/screenshots for the capture setup. -->
+
+Open the dropdown for a searchable list of the models found in
+your sessions, then click models to include them. The button then
+shows the chosen model — for example **Model: gpt-4o** — or
+**Model: 3 selected** once several are active. Click the
+**All models** row at the top of the list to clear the filter and
+return to every model. Selected models also appear as removable
+chips beneath the toolbar.
+
+While a model filter is active, every dashboard panel reflects
+only the selected model(s): the summary cards, activity chart,
+heatmap, hour-of-week grid, projects, session shape, velocity,
+tools, skills, top sessions, and the Session Health rollup.
+
+Model filtering is **message-grain**, unlike the session-grain
+project and agent filters, because a single session can switch
+models across turns. A session is included when it has at least
+one message from a selected model, and most panels count only the
+matching messages. The user turn paired with a matching assistant
+turn is kept alongside it — even though a user message carries no
+model of its own — so prompts and their responses stay aligned in
+the counts and in the top-session evidence.
+
+**Session Health** is the exception. It is scoped to whole
+sessions that used the selected model, but its health scores,
+outcomes, tool-failure rates, and compaction counts stay
+whole-session aggregates — they are not recomputed from only that
+model's messages.
+
+!!! note "Dashboard-only scope"
+    The model filter applies only to the analytics dashboard. The
+    [Session Insights](/insights/) page and the session list are
+    not scoped by it, so a model selected here does not silently
+    narrow those views. The [Usage](/token-usage/) page keeps its
+    own separate model filter.
+
 ### Activity Heatmap
 
 A GitHub-style contribution graph showing daily activity. Toggle
--- a/frontend/src/lib/api/generated/models/DbAnalyticsSummary.ts
+++ b/frontend/src/lib/api/generated/models/DbAnalyticsSummary.ts
@@ -10,6 +10,7 @@ export type DbAnalyticsSummary = {
   avg_messages: number;
   concentration: number;
   median_messages: number;
+  models: any[] | null;
   most_active_project: string;
   p90_messages: number;
   token_reporting_sessions: number;
--- a/frontend/src/lib/api/generated/services/AnalyticsService.test.ts
+++ b/frontend/src/lib/api/generated/services/AnalyticsService.test.ts
@@ -1,0 +1,48 @@
+import {
+  describe,
+  expect,
+  it,
+  vi,
+  beforeEach,
+} from "vite-plus/test";
+
+const { request } = vi.hoisted(() => ({
+  request: vi.fn(),
+}));
+
+vi.mock("../core/OpenAPI", () => ({
+  OpenAPI: {},
+}));
+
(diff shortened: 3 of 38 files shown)


rodboev:
Applied the chunking fix in `d5134519` for both SQLite and PostgreSQL model-list lookups, so the summary path no longer builds one oversized `IN (...)` clause.

I left the mixed-model behavior session-scoped on purpose. The issue asks for filtering the main dashboard by model, and specifically calls out seeing model usage in sessions. This PR keeps the dashboard scoped to sessions that contain at least one matching model, which is the contract already described in the PR body. That keeps session-level cards, projects, tools, skills, and top sessions on one consistent population instead of partially rewriting some of them into message-level metrics.

A message-scoped model-usage dashboard still seems useful, I just see that as a separate slice from this session dashboard filter.

rodboev:
The trends paths now clear `sessionFilter.Model` before building the shared session predicate, then apply the selected model directly to outer `m.model` in both SQLite and PostgreSQL. That keeps mixed-model sessions from leaking non-selected messages into trend `message_count` and term totals.

I also added mixed-model coverage for both backends, plus a PostgreSQL query-shape test that keeps the model predicate on the outer message rows.

rodboev:
`AnalyticsService.getApiV1AnalyticsSignalSessions()` now accepts `model` and includes it in the query object, so signal evidence requests stay aligned with the active analytics model filter.

I also added a focused frontend test for that request path, and `npm test -- --run AnalyticsService.test.ts` passed locally.

rodboev:
Fixed on 8a237056.

Session-shape now derives length buckets and autonomy from the same filtered per-session message stats that the model-scoped analytics views already use in internal/db/analytics.go, internal/postgres/analytics.go, and internal/duckdb/analytics_usage.go, so mixed-model sessions stop leaking off-model message volume and off-model user turns into the selected model's shape view.

Velocity complexity now uses those filtered message counts as well, which keeps model-filtered sessions in the correct 1-15 or 16-60 bucket instead of classifying them by the full session totals.

I also added mixed-model regressions for SQLite and DuckDB, plus matching PostgreSQL pgtests for the same shape and complexity cases.

rodboev:
I traced the last roborev report back to the model-scoped message loaders rather than any single panel. The shared filtered-message helpers, the velocity loaders, and the trend scans were all treating `messages.model` as a row-local gate, which drops parser-realistic user turns when only the assistant row carries the selected model.

This rework keeps the model filter anchored on the selected assistant turn, but it now buffers empty-model user rows and attributes them to that assistant exchange when the reply matches the active model. I’m applying that rule in SQLite, PostgreSQL, and DuckDB, and I switched the existing mixed-model regressions over to empty-model user rows so the tests exercise the real parser shape that roborev flagged.

rodboev:
Follow-up pushed in

rodboev:
Applied the tool analytics fix. Model-scoped tool counts now resolve each call against its own message timestamp, with session fallback only when the tool row has no message timestamp, so `Model + Hour` and `Model + DayOfWeek` stop pulling in off-window tool calls from the same session. I added regression coverage for SQLite, PostgreSQL, and DuckDB, and reran `CGO_ENABLED=1 go test -tags "fts5,kit_posthog_disabled" ./internal/db ./internal/duckdb ./internal/postgres -count=1`.

wesm:
Looking at this. Trying to figure out why this PR is so big so will see if there is some refactoring that can be done to make the change simpler

rodboev:
Tests are more than half the diff. Most of the rest comes from three backends (SQLite/PostgreSQL/DuckDB) that don't share a query abstraction, so the model filter had to be threaded through each one independently. The rest grew from iterative edge-case fixes during review.

I can split future work like this into stacked PRs if that's easier to review. A shared analytics query layer would also cut this kind of cross-backend feature down significantly; that felt like a bigger refactor to take on inside this PR.

wesm:
No worries, I get it! I have agents looking to see what refactoring could be done now to help with this

wesm:
I'm working on this, will update the PR in place

rodboev:
Sorry, just saw the comment. I'll leave it for you to work on.

The force push was just to replace the old head with the rebased one. I resolved conflicts in `internal/db/analytics_test.go`, `internal/db/analytics.go`, and `internal/duckdb/analytics_usage.go`, then reran the focused backend tests before pushing.

wesm:
Okay, I'll force push over what is here once my agents are done working, they are in the middle of a large superpowers plan

rodboev:
Thanks for this. The single-owner approach with thin adapters eliminates the scope leaks at the source. I'll dive deeper into this to make sure future work is aligned along these principles so extraction opportunities are caught before the code gets written. Appreciate you taking the time to do it right.

wesm:
All good! I was hoping it might be possible to make the change smaller but at least architecture wise this appears to be a tidier way to go about it. I'll work on getting this merged here today

PR #862, opened 2026-06-25
feat(parser): record Antigravity producing-version from schema fingerprint

mjacobs:
Antigravity sessions carry no literal product version, so a new dot release silently misparses until a symptom surfaces. This records a producing-version label per Antigravity session, derived from the SQLite schema fingerprint (a sha256 over the database's CREATE statements plus `user_version`), stored in the existing `sessions.source_version` column. One schema spans several agy dot releases, so the fingerprint maps to a release-range label; an unrecognized (newer) schema is tagged `agy-schema:<prefix>` so it stays visible instead of being silently decoded with heuristics tuned for an older payload. Even without gating decode on it, having the version recorded turns "which release broke this" from archaeology into a query.

The fingerprint algorithm is byte-for-byte identical to agy-reader's format-audit recorder, so labels seeded from agy-reader's compatibility notes match real databases. Both the IDE (`.db`) and CLI (`.db`) parse paths share one helper so they classify identically; legacy `.pb` and sidecar-only sessions leave `source_version` empty rather than fabricating a label. The label is computed as soon as the `.db` opens, so a readable schema still carries its marker even when the step query fails and the parser falls back to the trajectory sidecar.

`dataVersion` is bumped to 54 so existing Antigravity rows get `source_version` backfilled through the normal non-destructive resync, not just newly synced sessions.

Where to look:
- `internal/parser/antigravity_version.go` — schema fingerprint + label mapping (known release ranges vs the `agy-schema:` unknown marker).
- `internal/parser/antigravity.go`, `internal/parser/antigravity_cli.go` — IDE and CLI wiring through the shared helper.
- `internal/db/db.go` — `dataVersion` bump and bump-log entry.

Version-gating of decode heuristics and a decode-confidence signal that keys off the unknown-schema marker are deferred follow-ups.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,7 +151,7 @@ jobs:
         run: go run ./internal/pricing/cmd/litellm-snapshot -restore
 
       - name: Run Go tests
-        run: go test -tags "fts5" ./... -v -count=1
+        run: go test -tags "fts5" ./... -v -count=1 -timeout=20m
         env:
           CGO_ENABLED: "1"
 
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -1,8 +1,10 @@
 import { defineConfig } from "@playwright/test";
 
+const isCI = process.env.CI === "true";
+
 export default defineConfig({
   testDir: "e2e",
-  timeout: 20_000,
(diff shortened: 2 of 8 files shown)


wesm:
fixing the CI failures

PR #863, opened 2026-06-25
fix(sync): detect Codex title-only renames in full sync regardless of index mtime

mjacobs:
A Codex session's title lives in `session_index.jsonl`, separate from its transcript, and the index mtime is folded into the stored `file_mtime` watermark. `shouldSkipCodex`'s `storedMtime == effectiveMtime` fast path skipped a file without ever consulting the title, so a later title-only rename whose folded index mtime landed at or below the stored watermark was masked on a full sync — the renamed title never resolved in the archive until the transcript itself changed.

This checks the live index title against the stored `session_name` directly before deciding to skip, then skips only when the bare transcript mtime is unchanged. Comparing the bare file mtime rather than the folded effective mtime means an unrelated bump of the shared `session_index.jsonl` (another session's rename touches the same file) no longer forces a needless reparse once the title is confirmed unchanged.

The incremental and `SyncAllSince` quick-sync paths already detect title renames this way; this closes the remaining full-sync entry point.

Where to look:
- `internal/sync/engine.go` — `shouldSkipCodex`.

Follow-up (not in this PR): `codexIndexSessionNameChanged` loads the full session row via `GetSessionFull` on the skip path; a cheaper stored-name-only lookup would keep the title check fully off the hot path.

### DIFF
--- a/internal/sync/engine.go
+++ b/internal/sync/engine.go
@@ -4956,14 +4956,26 @@ func (e *Engine) shouldSkipCodex(
 		db.CurrentDataVersion() {
 		return false
 	}
-	fileMtime := info.ModTime().UnixNano()
-	effectiveMtime := parser.CodexEffectiveMtime(path, fileMtime)
-	if storedMtime == effectiveMtime {
-		return true
+	// A Codex title lives in session_index.jsonl, not the transcript, so a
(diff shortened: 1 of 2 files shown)


wesm:
fixing ci failures

PR #867, opened 2026-06-25
fix(analytics): surface active duration in top sessions (#83)

### LINKED ISSUE #83: Top session duration metric total vs API
Currently the top session by duration uses "total session time" at least for Copilot.

That is not a very useful metric since many users will start agents before they go to lunch or bed, or keep terminal windows open for days and hours without active usage. So the top entries for me are many hours but are not indicative of whether it's really a long or short session (my number 1 was a big one, my number 2 tiny "hello" test).

When I exit the Copilot CLI it gives me a second "API time spent" which seems more useful:
```
 API time spent:         29m 56s
 Total session time:     37m 48s
```

Not sure if it's an issue with other agents or just something that would need adjustment with the copilot agents view to get a useful metric? Maybe show both in agents view as well?

Image

rodboev:
Top Sessions currently ranks by wall-clock duration, computed as the gap between `started_at` and `ended_at`. That overstates sessions left idle, a terminal left open overnight dominates the ranking even if actual tool work lasted only a few minutes. The underlying per-turn timing data already exists and powers the per-session timing endpoint, but the analytics aggregation path never uses it.

This change adds an explicit `active_duration_min` field to the Top Sessions contract, computed from the same tool-turn timing semantics the repo already uses elsewhere, and switches the duration ranking to use that active value across the SQLite, PostgreSQL, and DuckDB analytics backends. The existing `duration_min` field stays populated so total session time remains visible to users as secondary context, which keeps the “show both” part of the issue intact.

On the frontend, the duration view now reads as active or tool time, uses the active value as the primary metric, keeps total duration available alongside it, and routes the new tooltip text through the existing Paraglide message surface. Focused backend regression tests cover the case where wall-clock duration and active duration disagree, the SQLite DST fallback path, DuckDB parity, PostgreSQL running-session null handling, and the ranking helper, while the Top Sessions frontend test proves the new display contract.

Fixes #83

### DIFF
--- a/frontend/messages/en.json
+++ b/frontend/messages/en.json
@@ -391,6 +391,8 @@
   "analytics_top_sessions_by_messages": "By Messages",
   "analytics_top_sessions_by_duration": "By Duration",
   "analytics_top_sessions_by_output_tokens": "By Output Tokens",
+  "analytics_top_sessions_active_duration": "Active duration",
+  "analytics_top_sessions_total_duration": "{duration} total",
   "analytics_projects_title": "Projects",
   "analytics_project_messages": [
     {
--- a/frontend/messages/zh-CN.json
+++ b/frontend/messages/zh-CN.json
@@ -385,6 +385,8 @@
   "analytics_top_sessions_by_messages": "按消息数",
   "analytics_top_sessions_by_duration": "按时长",
   "analytics_top_sessions_by_output_tokens": "按输出 token",
+  "analytics_top_sessions_active_duration": "活跃时长",
+  "analytics_top_sessions_total_duration": "总计 {duration}",
   "analytics_projects_title": "项目",
   "analytics_project_messages": [
     {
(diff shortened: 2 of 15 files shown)


wesm:
Good catch, I was just thinking about fixing this myself

mariusvniekerk:
I'm taking a look here, this feels like it will undercount heavily for sessions dominated by agent reasoning.

mariusvniekerk:
Should we make this reuse the "active minutes" concept from the velocity calculation for analytics?

rodboev:
Do you want me to hold? I traced both paths and they are different:

- current PR code: sums tool-use turn durations
- velocity: sums every positive gap between consecutive messages, capped at 5 minutes per gap

That means the current branch can undercount reasoning-heavy, low-tool sessions. I was about to patch top-session duration to reuse the velocity calculation across SQLite, PostgreSQL, DuckDB, and the SQLite fallback and see how the totals net out, but I can switch things up based on your direction.

rodboev:
Even with this change, the top results are still very long sessions, and the new active-duration values stay close to total duration on my data. It does knock out the worst idle outliers, so it is moving in the right direction, but it still is not measuring activity the way the velocity view does. I'm reworking this to reuse the "active minutes" concept from velocity so long idle gaps are capped consistently and the Top Sessions ranking lines up better with actual active work.

rodboev:
Withdrawing in favor of #869.

PR #873, opened 2026-06-26
fix(analytics): Count subagent sessions in analytics totals

### LINKED ISSUE #871: Analytics dashboard undercounts tokens and sessions by excluding subagents
## What I was doing

I run Claude Code Workflows a lot, and a run spawns many subagent transcripts.
agentsview parses and parent-links them fine. I opened the Insights dashboard to
see what all that workflow activity cost.

## What happened

The dashboard's token and session totals are far lower than what I've actually
used. The `usage daily` view, looking at the same window, reports much higher
numbers. The two disagree because the analytics aggregates drop subagent
sessions while the usage view counts them.

It's easy to confirm:

- Compare `agentsview usage daily` against the Insights summary for the same
  window. Usage is higher.
- `SELECT relationship_type, count(*), sum(total_output_tokens) FROM sessions
  GROUP BY relationship_type;` shows a large `subagent` bucket that the
  analytics summary never counts.

On my data the summary misses about a quarter of my output tokens. On a
workflow-heavy project it's closer to half.

## Why it matters

A subagent is real, independent spend. Its transcript lives in its own file, and
the parent only ever sees the synthesized result, not the subagent's tokens. For
anyone running Workflows or Task subagents, the dashboard's headline cost is just
wrong, and it contradicts the usage view sitting one tab over. The cost number is
the one number on that page you actually trust, and right now you can't.

jesserobbins:
The Insights dashboard undercounts tokens and sessions because the analytics aggregates exclude subagent sessions. Workflow runs spawn many subagent transcripts; agentsview parses and parent-links them, but their tokens never reach the summary. The cost/usage view already counts them, so the two surfaces disagree. Closes #871.

Two filters hid them, and both had to change for the totals only:

- The analytics WHERE clause excluded `relationship_type IN ('subagent', 'fork')`, the same predicate the session list uses to nest subagents under their parent. Right for navigation, wrong for token and session sums.
- The summary defaults `ExcludeOneShot` on, and workflow subagents are all one-shot (one orchestrator prompt, one result), so they were re-excluded even once the relationship filter let them through.

This adds an opt-in `IncludeSubagents` flag on `AnalyticsFilter`, set only on `GetAnalyticsSummary` and `GetAnalyticsProjects`. Those surfaces count subagents and exempt them from the one-shot filter. Forks and continuations stay excluded everywhere, since their tokens overlap their root and would double-count. The flag defaults off, so distribution surfaces (session-shape, velocity, timing) keep the root-only filter unchanged; subagents average ~8 minutes against ~103 for a root session and would flatten any duration or shape histogram. Mirrored across SQLite, PostgreSQL, and DuckDB.

The `stats` command (v1 window analytics) had the same gap, and it's fixed here too. That pipeline loads sessions once and feeds both totals and distributions from them, so it loads a second subagent-inclusive set and overrides only the additive totals. The shape, velocity, timing, and human-vs-automation breakdowns stay root-only. Counting subagents in the session total while keeping them out of the human and automation buckets would have broken the `sessions_all == human + automation` invariant, so a third bucket, `sessions_subagent`, keeps the partition whole: the stats line now reads `14,205 (human 8,269, automation 1,117, subagent 4,819)`.

The navigation count and the dashboard count now differ on purpose. The sidebar shows the sessions you started; Insights counts the work the subagents did under them.

One gap left: `cache_economics` dollar figures in the stats output are still root-only. They come from a separate per-message query rather than a row sum, so folding subagents in cleanly is its own change. The token and session totals are the headline numbers and they're correct now; the cache dollars are a smaller follow-up.

Where to look: `internal/db/analytics.go` (`RelationshipExclusionSQL`, `OneShotExclusionSQL`), `internal/db/session_stats.go` (the two-row-set load and `applySubagentInclusiveTotals`), mirrored in `internal/postgres/analytics.go` and `internal/duckdb/analytics_usage.go`. The cross-surface split is asserted in `internal/db/session_stats_test.go`, `internal/db/store_contract_test.go`, and `frontend/e2e/session-count-consistency.spec.ts`.


wesm:
thank you!

PR #874, opened 2026-06-26
fix(sync): repair persisted Codex goal-context rows

tpn:
Refinement on top of #886 / 9bee8a3c, which already taught the Codex parser to classify `/goal` continuation envelopes as system content for newly parsed sessions.

This PR handles the archive and compatibility side of that fix. It bumps the parser data version to 56 on top of main's version-55 Kimi usage-event bump, so existing source-backed Codex sessions are non-destructively re-parsed and old stored `/goal` rows are removed from message lists and user-turn counts instead of remaining in persistent SQLite archives.

It also adds read-path fallbacks for legacy rows that may still exist before reparse, or in rows that cannot be re-emitted from source files. Backend system-prefix filtering now uses a dedicated Codex goal-context predicate plus SQLite/PostgreSQL/DuckDB SQL variants, while frontend system-message detection and HTML/focused export use the same wrapper semantics. Both the current `<codex_internal_context ... source="goal">` form, including extra attributes, and the older `<goal_context>` wrapper stay out of search, transcript filtering, and exported sessions.

The main tradeoff is the data-version bump: multi-host PostgreSQL deployments should upgrade `pg serve` and all pushers together, because older agentsview binaries reject rows written by a newer parser data version.

### DIFF
--- a/frontend/src/lib/utils/messages.test.ts
+++ b/frontend/src/lib/utils/messages.test.ts
@@ -58,6 +58,11 @@ describe("isSystemMessage", () => {
     ["command-name", "<command-name>/commit</command-name>"],
     ["local-command", "<local-command-output>ok</local-command-output>"],
     ["stop hook", "Stop hook feedback: blocked"],
+    ["legacy goal context", "\n\t<goal_context>state</goal_context>"],
+    [
+      "codex internal goal context",
+      '  <codex_internal_context source="goal">state',
+    ],
   ])("detects prefix-based system message: %s", (_label, content) => {
     expect(isSystemMessage(msg({ content }))).toBe(true);
   });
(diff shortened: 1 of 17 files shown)


wesm:
This was probably fixed by #886, let me rebase this and see

tpn:
Addressed the RoboRev export finding in `f1b131ae`: `filterExportHTMLMessages` and Focused export now use a Codex goal-context-specific predicate instead of `db.IsSystemPrefixed`, so non-goal system-prefixed rows such as continuation/task/stop-hook messages remain in HTML exports. Added DB and export regression coverage for the narrowed behavior.

tpn:
Addressed the latest RoboRev goal-context consistency finding in `5826ee7e`: `IsSystemPrefixed`, SQLite/PostgreSQL/DuckDB `SystemPrefixSQL` paths, the frontend system-message helper, and HTML/focused export now share the same Codex goal-wrapper semantics for `<codex_internal_context ... source="goal">` with extra attributes plus legacy `<goal_context>` rows. Rebased onto `origin/main` at `e3a0af35`; the parser data-version bump is now `56` because main already uses `55` for the Kimi usage-event reparse.

PR #876, opened 2026-06-26
feat(parser): provider facade core

mariusvniekerk:
Establishes the provider facade contract the rest of this stack builds on: the `Provider` and `ProviderFactory` interfaces, the facade value types (`SourceRef`, `ParseRequest`/`ParseOutcome`, `SourceFingerprint`, capabilities), and `ProviderBase` as the single embedded no-op surface. No providers migrate here — this branch is the foundation plus its design docs.

Where to look: `internal/parser/provider.go` and the provider-facade design docs.


mariusvniekerk:

- #876 ◀
    - #877
        - #878
            - #879
                - #880
                    - #881
                        - #882
                            - #883
                                - #884
                                    - #885

mariusvniekerk:
**Re: roborev High — `internal/sync/engine.go:4463` cached `skip` results bypassing `data_version` resync**

Not a bug. The new skip-cache write in collectAndBatch (engine.go:4463-4466) cannot strand sessions behind a stale data_version. db.CurrentDataVersion() is a compile-time constant (db.go:2035-2036), and every per-session/per-path data_version is stamped to that same constant (parsediff.go:59, dataVersionForWrite, SetSessionDataVersion). A stored data_version can therefore only fall below CurrentDataVersion() after a binary upgrade bumps the `dataVersion` constant — and that exact condition sets dataStale in probeDatabase (`version < dataVersion`, db.go:918), making NeedsResync() return true (db.go:2024). NeedsResync() forces resyncAllLocked before any skip-cache-consuming sync runs: main.go:172-173 runs the initial resync before the watcher starts (main.go:907), and SyncAll gates on `didResync := full || e.db.NeedsResync()` (engine.go:2852). resyncAllLocked clears e.skipCache (engine.go:2347-2349) and rebuilds into a fresh temp DB with an empty skipped_files table; CopyOrphanedDataFromExcluding (orphaned.go) never copies skipped_files and only carries id-diff orphan sessions, not reparsed ones. The data_version doc comment (db.go:25-30) confirms the intended behavior: a lower stored version triggers a "non-destructive re-sync (mtime reset + skip cache clear)". There is no in-place data_version revalidation path that bumps a per-session version without either bumping the global constant or going through resync. Within one binary version data_version is constant, so a cached mtime-keyed skip — only written when shouldSkipFile/shouldSkipByPath already observed stored data_version >= current — is always correct. The optional belt-and-suspenders guard (have shouldUseCachedSkip also require GetDataVersionByPath(path) >= CurrentDataVersion()) would be harmless but is redundant today since every bump already clears the cache via resync.<sup>generated by a clanker</sup>

mariusvniekerk:
**Re: roborev Medium — `desktop-artifacts.yml` push trigger skips Go-only changes**

Confirmed the path-filter mechanics: desktop-artifacts.yml's push-to-main trigger filters on desktop/**, frontend/**, go.mod, go.sum, .github/workflows/desktop-*.yml, and prepare-sidecar.sh does build the sidecar from `go build ./cmd/agentsview` plus `go run ./internal/pricing/...`. So a Go-only push to main does skip this workflow. However, this is the intended, guard-test-pinned design and causes no incorrect outcome:

1. Backend correctness is fully covered: ci.yml triggers on `push: branches:[main]` with no path filter, so every Go-only main push runs the full lint/test/e2e suite against the exact code that becomes the sidecar.
2. Nothing consumes the skipped output: the push-to-main run only uploads ephemeral actions/upload-artifact artifacts. The single workflow_run consumer (desktop-release-health.yml) keys off "Desktop Release", not "Desktop Artifacts" — there is no downstream job that depends on these push artifacts being current.
3. Releases never go stale: real desktop bundles come from desktop-release.yml on `push: tags v*`, which rebuilds the sidecar fresh from source.
4. The narrow filter is deliberate and documented ("Only build the expensive desktop/tauri bundles ... only when desktop-relevant files change") and pinned by TestCIRunsOnAllPullRequestsWhileDesktopBuildsTargetMain.

Adding cmd/** and internal/** would force expensive 5-platform tauri builds on every backend change merged to main, contradicting that intent, with no correctness benefit since ci.yml already validates the backend and releases rebuild from source. Closing as working-as-intended.

PR #877, opened 2026-06-26
feat(parser): reusable source-set framework

mariusvniekerk:
Adds the reusable source-set bases and functional options that concrete providers compose for discovery, lookup, fingerprinting, and changed-path classification. The framework symbols are exported so they can be consumed across the package without tripping unused-symbol lint at intermediate branches in the stack.

Also adds the `WithCompanionFiles` option, unifies the path-under-root helper onto a single root-first signature, and removes the dead `SQLiteFanoutSourceSet`.

Where to look: `internal/parser/jsonl_source_set.go`, `single_file_source_set.go`, and the source-set base types.


mariusvniekerk:

- #876
    - #877 ◀
        - #878
            - #879
                - #880
                    - #881
                        - #882
                            - #883
                                - #884
                                    - #885

PR #883, opened 2026-06-26
feat(parser): migrate Codex + editor-extension providers

mariusvniekerk:
Migrates codex, gemini, copilot, the Copilot IDE variants (VS Code and Visual Studio), and positron onto facade providers.

Includes the Codex incremental-fingerprint expectation update: a Codex re-sync is a full re-parse that stores the raw file size and hash, with the parsed-prefix versus partial-tail distinction enforced at parse-diff time via `CodexTranscriptConsumedSize` rather than in the stored fingerprint.

PR #882, opened 2026-06-26
feat(parser): migrate bespoke providers (openhands, cursor, vibe, hermes, claude, cowork)

mariusvniekerk:
Migrates the providers with bespoke source shapes onto concrete facade providers: openhands (directory snapshot), cursor, vibe, hermes, claude (recursive transcripts with incremental append parsing and subagents), and cowork (reuses the Claude transcript format).

Because Claude's migration routes parse-diff through the provider path, this branch also folds in the parse-diff raced-source reclassification — gating reliability on `parseDiffAgentDiscoverable` — so a concurrent daemon write is classified Raced rather than Changed and cannot trip `--fail-on-change`.

Where to look: `claude_provider.go` (the largest change) and the parse-diff change in `internal/sync/parsediff.go`.

PR #881, opened 2026-06-26
feat(parser): migrate opencode-family providers (opencode, kilo, mimocode)

mariusvniekerk:
Migrates opencode, kilo, and mimocode — which share the OpenCode source format — onto a shared concrete provider, folding their discovery/lookup/parse and dropping the legacy sync dispatch.

PR #880, opened 2026-06-26
feat(parser): migrate OpenClaw + QClaw providers

mariusvniekerk:
Migrates the OpenClaw and QClaw providers onto the facade, preserving discovery and parse behavior.

PR #884, opened 2026-06-26
feat(parser): migrate SQLite/container providers

mariusvniekerk:
Migrates the container and DB-backed providers — zed, shelley, kiro, antigravity, and the db-backed family — onto facade providers.

Includes the Kiro SQLite fan-out fix so each row a store produces counts toward `TotalSessions`, matching the per-session tally the legacy sync phase reported.


mariusvniekerk:

- #876
    - #877
        - #878
            - #879
                - #880
                    - #881
                        - #882
                            - #883
                                - #884 ◀
                                    - #885

PR #879, opened 2026-06-26
feat(parser): migrate Qwen + QwenPaw providers

mariusvniekerk:
Migrates Qwen and QwenPaw onto facade providers, preserving their nested session-directory layouts and raw-session lookup behavior.

PR #878, opened 2026-06-26
feat(parser): migrate CLI/JSONL providers

mariusvniekerk:
Migrates the directory-JSONL providers — commandcode, iflow, gptme, deepseek-tui, amp, zencoder, pi, kimi, cortex, workbuddy — onto concrete facade providers. Each folds its discovery, raw-session lookup, fingerprinting, and parse normalization onto the provider and drops the legacy sync dispatch, becoming provider-authoritative. Symlinked project-root discovery and existing parse output are preserved.

Where to look: the per-agent `*_provider.go` files and their source sets.

PR #885, opened 2026-06-26
refactor(parser): require explicit provider factories + migration cleanup

mariusvniekerk:
Closes out the migration. Replaces the legacy default factory with an explicit per-agent factory registry that panics on a missing factory, and routes the import-only ChatGPT and Claude.ai export providers through dedicated export interfaces.

Removes the migration scaffolding now that every provider is folded: the anti-shim guard test and the dead `DiscoverFunc`/`FindSourceFunc` registry hooks. Also includes the source-pinning and lookup hardening tail.

Where to look: the explicit `providerFactoryForDef` switch and the import-only provider.


mariusvniekerk:

- #876
    - #877
        - #878
            - #879
                - #880
                    - #881
                        - #882
                            - #883
                                - #884
                                    - #885 ◀

PR #892, opened 2026-06-26
Preserve XML-style prompt tags in markdown rendering

### LINKED ISSUE #336: when there's xml tagging in the prompt those get stripped, but are highly relevant and typical to use
When a prompt has xml style tag wrapping, we can't see that in the session viewer ui. 

see the attached first message txt which corresponds to the image screenshotted of agentsview

Image

[session-aad36855-first-message.txt](https://github.com/user-attachments/files/26766669/session-aad36855-first-message.txt)

rodboev:
## Summary
- escape non-HTML XML-style prompt tags before markdown parsing
- keep real HTML tags and the existing bash pseudo-tags on the current sanitize/render path
- add coverage for literal XML prompt tags and sanitized HTML

Closes https://github.com/kenn-io/agentsview/issues/336

## Testing
- npm test -- markdown.test.ts

### DIFF
--- a/frontend/src/lib/utils/markdown.test.ts
+++ b/frontend/src/lib/utils/markdown.test.ts
@@ -348,6 +348,32 @@ describe("renderMarkdown", () => {
       expect(img).not.toBeNull();
       expect(img!.hasAttribute("onerror")).toBe(false);
     });
+
+    it("does not escape custom tags inside inline code spans", () => {
+      const dom = parseHTML(
+        renderMarkdown("`<policy>keep tags</policy>`"),
+      );
+      const code = dom.querySelector("p > code");
+      expect(code).not.toBeNull();
+      expect(code!.textContent).toBe("<policy>keep tags</policy>");
+    });
+
+    it("does not escape custom tags inside fenced code blocks", () => {
+      const dom = parseHTML(
+        renderMarkdown("```\n<policy>keep tags</policy>\n```"),
+      );
+      const code = dom.querySelector("pre > code");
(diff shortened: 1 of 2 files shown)


rodboev:
Fixed the two cases Roborev called out. The XML-tag prepass now skips fenced code, inline code, bash wrapper bodies, and angle autolinks before escaping unknown tags, so literal prompt tags stay intact where markdown already treats them as code or links. I also added focused coverage for the code-span, fenced-block, bash-output, and autolink cases.

rodboev:
Tightened this to the lexer boundary instead of trying to pre-scan markdown text. Unknown XML-style tags are now escaped from `marked` tokens after code blocks, code spans, bash wrappers, and real angle autolinks have already been classified, which covers the indented-code, uneven-fence, unclosed-fence, and namespaced-tag cases Roborev called out. I added regressions for each of those cases in `markdown.test.ts`.

rodboev:
Narrowed the `link` special-case to angle-autolink-shaped tokens only. Normal markdown links now keep their anchor structure and only recurse into label tokens, so custom-tag labels like `[<policy>read</policy>](...)` still render as links while literal angle-tag autolinks stay escaped. I added a regression for that label case alongside the existing markdown-tag coverage.

rodboev:
Extended the recursion into plain nested token-container objects, which covers Marked table cells under `header` and `rows` instead of only top-level typed tokens. I added a GFM table regression for custom tags inside a table cell so that surface now stays literal too.

PR #898, opened 2026-06-27
feat(postgres): implement PG serve session curation and insight persistence

### LINKED ISSUE #183: feat: write support in pg serve
The `pg serve` read-only PostgreSQL store does not support write operations. Implementing stars, pins, insights, rename, delete, and trash operations against PostgreSQL would allow full functionality when serving from PG.

The `db.Store` interface already defines the boundaries for these operations — the PG store just needs concrete implementations.

rodboev:
PG serve already supports shared stars and pins, but the rest of the dashboard curation surface still behaves as read-only. Session rename, trash, restore, empty trash, insight delete, and generated-insight persistence all route through the existing `db.Store` seam, yet the PostgreSQL adapter still returns `db.ErrReadOnly` for those methods, and the insight-generation route stops on a coarse read-only check before it can save anything.

This change fills in the remaining PG dashboard-write slice without widening scope to local-only features or other read-only backends. PostgreSQL gets the missing insight storage plus Store implementations for insight CRUD and session-management methods, and the server now treats generated-insight writes as an explicit capability: pg serve advertises it, the frontend enables the Generate controls in that mode, and plain read-only stores such as duckdb still fail fast before any generator work starts. Settings remain blocked in pg serve through the existing read-only contract, while local `serve --no-sync` stays writable because it still uses a local Store.

The tests cover the PG round trips for insight CRUD and session management, the read-only capability split on the insight route, the pg serve and local no-sync settings behavior, the pg serve version metadata that enables the frontend, and the two frontend insight entry points. Stars and pins are left untouched, and ingestion-oriented methods such as `WriteSessionBatchAtomic` remain read-only.

Closes #183

### DIFF
--- a/cmd/agentsview/pg.go
+++ b/cmd/agentsview/pg.go
@@ -418,6 +418,14 @@ func runPGServe(appCfg config.Config, basePath string) {
 	); err != nil {
 		fatal("pg serve: %v", err)
 	}
+	if err := store.DetectInsightGenerationAvailability(
+		ctx,
+	); err != nil {
+		fatal(
+			"pg serve: probing insight generation capability: %v",
+			err,
+		)
+	}
 
 	rtOpts := serveRuntimeOptions{
 		Mode:          "pg-serve",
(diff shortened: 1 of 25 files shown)


wesm:
looking at this

PR #899, opened 2026-06-27
feat(config): add substrings and exact_matches to automated session config

### LINKED ISSUE #370: Feature: user-defined automated session patterns in config.toml
## Background

The automated-session classifier in `internal/db/automated.go` ships with a hardcoded list of patterns (roborev reviews, Claude Code title generator/warmup, etc.). Users with their own tooling that issues recognizable prompts — e.g. a custom changelog generator that always starts with \`You are generating a changelog for ...\` — currently have no way to tell agentsview to treat those as automated without forking the binary.

## Proposal

Add a \`[automated]\` section to \`config.toml\` that lets users append their own patterns to the classifier:

\`\`\`toml
[automated]
prefixes = [
  \"You are generating a changelog for my-project\",
  \"You are a release notes summarizer for\",
]
substrings = [
  \"invoked by my-internal-tool\",
]
exact_matches = [
  \"PingOnly\",
]
\`\`\`

These would be loaded once at startup, merged with the built-in lists, and used by \`IsAutomatedSession\`. The single-turn gate (or whatever gating model lands from the related fix) applies the same way.

## Out of scope

- Per-agent pattern overrides (only worth it if users hit it).
- Regex patterns (start with literal strings; revisit if real users need regex).
- Hot-reload on config changes (a restart is acceptable for now).

## Related

- Built-in patterns added in PR #369.
- Follow-up: a separate change is needed to revisit the \`user_message_count <= 1\` gate, which currently excludes many legitimate automated sessions when synced from a machine whose Codex parser counts user messages slightly differently. That fix is independent of this issue.

rodboev:
Issue `#370` asks for the full literal-pattern config surface on the automated-session classifier. Prefixes were already made configurable in PR `#383`, building on the built-in classifier expansion from PR `#369`, but substrings and exact-message matches still stay hardcoded-only, which leaves users without a config-driven way to model embedded tool markers or whole-message pings.

This change adds `substrings` and `exact_matches` alongside the existing `prefixes` field, wires all three categories through the existing classifier-config entrypoint, and includes each user-configured category in `ClassifierHash()` so behavior changes still trigger the expected rebuild path. Built-in patterns remain active, and the single-turn gate stays unchanged.

The tests cover the new config round-trip, substring and exact-match semantics, hash invalidation for each new category, and preservation of the current multi-turn gate.

Fixes #370

### DIFF
--- a/cmd/agentsview/classifier_test.go
+++ b/cmd/agentsview/classifier_test.go
@@ -33,7 +33,7 @@ type classifierFixture struct {
 // newClassifierFixture prepares a temp data dir with a minimal
 // config.toml carrying the given user prefixes, loads a minimal
 // config pointed at that dir, applies any option mutators, and
-// installs the classifier prefixes into the db singleton.
+// installs the classifier patterns into the db singleton.
 func newClassifierFixture(
 	t *testing.T, prefixes []string, opts ...func(*config.Config),
 ) classifierFixture {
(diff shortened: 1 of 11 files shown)


wesm (approved):
i'm adding docs now

PR #903, opened 2026-06-28
feat(cli): standardize --format/--json output flags across commands

### LINKED ISSUE #902: CLI machine-readable output flags are inconsistent across commands
## Problem

There is no single way to ask an agentsview command for machine-readable
output. Depending on which command you run, you get one of three different
conventions:

- `--format human|json` (with `--json` accepted too): `session *`
- `--format human|json` only, no `--json`: `stats`, `secrets *`
- bare boolean `--json` only, no `--format`: `projects`, `health`,
  `usage daily`, `activity report`, `parse-diff`
- JSON only, no flag at all: `token-use`, `openapi`

Because the form is per-command, you cannot predict it. The same flag spelling
works on one command and errors on another:

```
$ agentsview session list --format json
# works

$ agentsview usage daily --format json
error: unknown flag: --format
# but `agentsview usage daily --json` works

$ agentsview stats --json
error: unknown flag: --json
# but `agentsview stats --format json` works
```

## Why it matters

Anyone scripting against the CLI has to memorize which form each command takes,
or probe each one by trial and error. A natural script that pipes several
commands into a JSON processor breaks on whichever command happens to use the
other convention, and the failure (`unknown flag`) gives no hint that a
different spelling would have worked.

## Proposed direction

Standardize on `--format human|json` for every command that emits
machine-readable output, with `--json` kept as a universal alias for
`--format json` so existing invocations keep working. Commands that stream a
fixed format and have no human rendering (for example `session export`, which
streams raw bytes, and `session watch`, which streams NDJSON) should reject the
flag rather than silently ignore it.

prateek:
Fixes #902

Every command that emits machine-readable output now accepts the same flag
pair: `--format human|json`, with `--json` as a boolean alias for
`--format json`. Before this change the surface was split three ways, so the
same flag spelling worked on one command and failed on another:

- `--format`-only: `stats`, `secrets`
- bare `--json`-only: `projects`, `health`, `usage daily`, `activity report`,
  `parse-diff`
- JSON-only with no flag: `token-use`, `openapi`

A scripter can now use one convention everywhere, and every previously-working
`--json` invocation still works.

## How it fits together

Registration and reading are centralized so the convention can't drift again:

- `registerFormatFlags` installs the `--format`/`--json` pair. Every
  machine-readable command calls it instead of declaring its own flag.
- `formatFlag` (a `pflag.Value` enum) restricts `--format` to `human` or
  `json`.
- `outputFormat` is the single reader; `--json` wins when set, otherwise the
  `--format` value, defaulting to `human`.
- Streaming commands that can't honor a format choice reject the inherited pair
  through a shared `rejectFormatFlags` helper: `session export` (raw bytes) and
  `session watch` (NDJSON).

`token-use` (deprecated) and `openapi` (a spec dump with no human form) stay
JSON-only.

## Tradeoffs and limitations

- One behavior change goes beyond pure consistency: a mistyped `--format` value
  now errors at parse time. Previously, on the commands that already had
  `--format`, an unrecognized value silently fell back to human output, which
  is a quiet trap for a script expecting JSON. The enum surfaces it instead.
- `session export` and `session watch` inherit `--format`/`--json` from the
  `session` group's persistent flags and so still list them in `--help`, even
  though they reject them at runtime. That is a cobra inheritance limitation,
  not a new contract; the rejection and the docs make the behavior explicit.

## Where to look

- `cmd/agentsview/session.go` — the shared helpers (`registerFormatFlags`,
  `formatFlag`, `outputFormat`, `rejectFormatFlags`).
- `cmd/agentsview/cli.go`, `parse_diff.go`, `stats.go`, `secrets.go` — the
  migrated commands.
- `cmd/agentsview/session_export.go`, `session_watch.go` — the streaming
  rejections.
- `cmd/agentsview/output_format_test.go` — `TestFormatAndJSONFlagsArePaired`
  walks the command tree and fails the build if any command ever registers one
  flag without the other.
- `docs/` — command reference tables updated to show the flag pair.

### DIFF
--- a/cmd/agentsview/cli.go
+++ b/cmd/agentsview/cli.go
@@ -375,7 +375,6 @@ func newImportCommand() *cobra.Command {
 }
 
 func newProjectsCommand() *cobra.Command {
-	var jsonOutput bool
 	cmd := &cobra.Command{
 		Use:          "projects",
 		Short:        "List projects with session counts",
@@ -383,10 +382,10 @@ func newProjectsCommand() *cobra.Command {
 		SilenceUsage: true,
 		Args:         cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
-			runProjects(jsonOutput)
+			runProjects(outputFormat(cmd) == "json")
 		},
 	}
-	cmd.Flags().BoolVar(&jsonOutput, "json", false, "Output as JSON array")
+	registerFormatFlags(cmd.Flags())
 	return cmd
 }
 
@@ -403,11 +402,11 @@ func newHealthCommand() *cobra.Command {
 		SilenceUsage: true,
 		Args:         cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
+			cfg.JSON = outputFormat(cmd) == "json"
 			runHealth(args, cfg)
 		},
 	}
-	cmd.Flags().BoolVar(&cfg.JSON, "json", false,
-		"Output as JSON")
+	registerFormatFlags(cmd.Flags())
 	cmd.Flags().IntVar(&cfg.Limit, "limit",
 		defaultHealthLimit,
 		"Number of sessions to list (max 500)")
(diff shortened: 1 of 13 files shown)


wesm:
looks good here, will merge once CI is green (i have an agent fixing now)

PR #905, opened 2026-06-28
fix(usage): accept duration syntax for usage daily --since/--until

### LINKED ISSUE #904: usage daily --since silently returns a wrong/empty window for duration inputs like 7d
### agentsview version

Built from source at `main` @ `634d930b` (`agentsview version` reports `agentsview dev` for source builds).

### Install method

Built from source

### OS / platform

macOS 26.4.1 (Darwin 25.4.0, arm64)

### Which agent and version

n/a (CLI usage-reporting bug, agent-independent)

### Which model(s)

n/a

`agentsview stats --since` accepts a duration (`28d`) or a date (`YYYY-MM-DD`). `agentsview usage daily --since` accepts only `YYYY-MM-DD`, but passing a duration does **not** error — it flows into the query as a malformed date and silently produces a wrong or empty window with exit code 0. Silent wrong output with a success exit is the real defect: a script reading the result sees "no spend" rather than a failure.

Observed on the direct-SQLite path:

- `agentsview usage daily --since 7d` → exit 0, `$0.00` / 0 rows ("looks like no spend")
- `agentsview usage daily --since 14d` → exit 0, returns a window that does not match 14 days
- `agentsview usage daily --since 2026-06-13` → correct

Expected: `usage daily --since/--until` should accept the same duration-or-date syntax as `stats`, and any unparseable value should fail loudly with a non-zero exit, not a silent empty result.

**Root cause:** `runUsageDaily` passed the raw `--since`/`--until` strings straight into `db.UsageFilter`. The query builder appends `T00:00:00Z` and pads, so `7d` becomes `7dT00:00:00Z`, fails to parse, and yields an unbounded/garbage window. The daemon HTTP path happens to reject it with a 400; the direct SQLite path fails silently.

### Steps to reproduce

```
1. Build/run a recent agentsview with some indexed usage data.
2. agentsview usage daily --since 7d --offline --no-sync   # exit 0, $0.00 (wrong)
3. agentsview usage daily --since 14d --offline --no-sync  # exit 0, window != 14 days
4. agentsview usage daily --since 2026-06-13 ...           # correct, for contrast
```

prateek:
`agentsview usage daily --since/--until` accepted only `YYYY-MM-DD`. A `stats`-style duration like `7d` did not error; it flowed into the query as a malformed date and silently produced a wrong or empty window with exit 0. `stats` already accepted durations, so the two commands diverged and the `usage` path failed silently.

This resolves both bounds through a shared `db.ResolveWindowDate` (reusing the `stats` `parseWindowPoint` grammar) before the filter is built, so durations and dates work on both the direct-SQLite and daemon-HTTP paths. Unparseable input and an inverted explicit window (`--since` after `--until`) now fail loudly with a non-zero exit instead of returning silent empty output. Flag help and the `usage daily` docs are updated to match.

**Where to look:**

- `internal/db/session_stats.go` — `ResolveWindowDate`, the shared duration-or-date resolver built on the existing `stats` parser.
- `cmd/agentsview/usage.go` — `resolveUsageWindow` resolves `--until` first, anchors a duration `--since` to it, and rejects bad/inverted input; `runUsageDaily` exits non-zero on that error.
- `cmd/agentsview/cli.go`, `docs/commands.md`, `docs/token-usage.md` — help text and reference rows.

**Notes / limitations:**

- Resolution happens once at the CLI layer, so both storage backends get the same concrete date — no per-backend logic, preserving backend parity.
- A duration `--since` anchors to the resolved `--until` (or to `now` when `--until` is open), and dates stand alone — matching how `stats` resolves a window. So `--until 2026-04-10 --since 14d` yields the 14 days ending April 10.
- Scope is the `usage daily` CLI flags. The HTTP API, MCP `get_usage_summary`, and `usage cursor` still take dates only; extending duration sugar there is a separate API-design call.

Fixes #904

### DIFF
--- a/cmd/agentsview/cli.go
+++ b/cmd/agentsview/cli.go
@@ -443,8 +443,8 @@ func newUsageDailyCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().BoolVar(&cfg.JSON, "json", false, "Output as JSON")
-	cmd.Flags().StringVar(&cfg.Since, "since", "", "Start date (YYYY-MM-DD)")
-	cmd.Flags().StringVar(&cfg.Until, "until", "", "End date (YYYY-MM-DD)")
+	cmd.Flags().StringVar(&cfg.Since, "since", "", "Start of window (duration like 28d, or YYYY-MM-DD)")
+	cmd.Flags().StringVar(&cfg.Until, "until", "", "End of window (duration like 28d, or YYYY-MM-DD)")
 	cmd.Flags().BoolVar(&cfg.All, "all", false, "Include all history (overrides default 30-day window)")
 	cmd.Flags().StringVar(&cfg.Agent, "agent", "", "Filter by agent name")
 	cmd.Flags().BoolVar(&cfg.Breakdown, "breakdown", false, "Show per-model breakdown rows")
--- a/cmd/agentsview/usage.go
+++ b/cmd/agentsview/usage.go
@@ -66,15 +66,46 @@ type UsageDailyConfig struct {
 	Timezone  string
 }
 
+// resolveUsageWindow resolves the raw --since/--until flags into concrete
+// inclusive YYYY-MM-DD bounds — a duration like 28d or a date, the same
+// syntax as `stats --since` — and rejects an inverted explicit window so a
+// reversed range fails loudly instead of returning an empty result.
+func resolveUsageWindow(
+	since, until string, now time.Time,
+) (string, string, error) {
+	from, err := db.ResolveWindowDate(since, now)
+	if err != nil {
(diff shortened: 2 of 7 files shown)


prateek:
Thanks @roborev-ci — fixed in a89688bd.

`resolveUsageWindow` now resolves `--until` first and anchors a duration `--since` to the resolved `--until` (falling back to `now` only when `--until` is open), matching stats' `windowBounds`. A date `--since` still stands alone. `usage daily --until 2026-04-10 --since 14d` now yields the 14-day window ending 2026-04-10 instead of tripping the inverted-window guard. Added table cases for duration-since-against-explicit-until and duration-since-against-duration-until.

wesm:
looking

PR #909, opened 2026-06-29
Add HTTP daemon remote sync

wesm:
Some users cannot rely on passwordless SSH on fleet nodes, but still need collector-owned ingestion of raw session files. This adds a daemon-backed HTTP transport for persisted remote hosts so Tailscale-secured nodes can expose authenticated remote-sync targets and archive endpoints while SSH remains the default and ad hoc host sync stays SSH-only.

The implementation keeps the collector-owns-the-archive model: configured API sync requests resolve transport, URL, and token from server config instead of trusting request payloads; daemon endpoints authenticate before host and CORS checks; and HTTP archive import reuses the extraction, remap, and skip-cache protections shared with SSH.

Reviewers should focus on the shared remotesync package, the server remote-sync routes and auth middleware, and the three dispatch paths for CLI, API-triggered, and interval remote sync. The detached daemon idle-timeout setting applies to serve --background; supervised daemons remain always-on under their process manager.


wesm:
roborev follow-up on the d7a893b comment: this one is a false positive. The source uses fmt.Fprintf, where %% renders a single %, so the generated shell uses ${dir%/*}/session_index.jsonl and resolves the parent directory as intended. Verified with CGO_ENABLED=1 go test -tags fts5 ./internal/ssh -run TestResolveScriptIncludesCodexIndex -count=1 -v.

PR #920, opened 2026-06-29
feat(sessions): fork Claude sessions from a message point

### LINKED ISSUE #107: feat: fork/resume conversation from any message point
## Context

Hey Wes; first off, I want to say you've done a brilliant job putting together AgentView. I've been silently PRing features without much discussion, and I realize I should be more communicative. I was actually building something similar on my own and ended up merging features I had that were missing here. But the foundation you built is solid, and I think there's room for much more.

AgentView is becoming a multi-purpose tool; everyone uses it differently depending on how they work with their AI sessions. Some people want analytics, some want search, some want to revisit and learn from past conversations. This feature idea comes from that realization.

## The Problem

How many times have you looked at a past conversation and thought: "it went sideways at message #15; if only I could go back to that point and take a different approach"?

Currently:
- Claude Code has some ability to revert, but it's unclear how it works or whether it can be done at any arbitrary point
- There's no way to visually inspect a conversation, identify the decision point, and branch from there
- Once a conversation goes in a bad direction, the context is polluted and recovery is costly

## Proposed Feature: Fork/Resume from Message

Allow users to **right-click on any message** in the session viewer and:

1. **"Resume conversation from here"**; Start a new Claude Code session with the conversation context up to that message, effectively rewinding
2. **"Fork conversation from here"**; Same as above, but preserves the original session and creates a new branch, with the relationship tracked in AgentView

### What this enables

- **Branching workflows**: Try different approaches from the same decision point
- **Error recovery**: Go back to before things went wrong without losing the original session for reference
- **Experimentation**: "What if I had asked differently at this point?"
- **Learning**: Compare how different prompts/approaches from the same starting point lead to different outcomes

### Implementation considerations

- AgentView already tracks `parent_session_id` and `relationship_type` (subagent, fork); the data model partially supports this
- The fork would need to serialize messages up to the chosen ordinal into a format Claude Code (or other agents) can consume as context
- Each agent has different mechanisms for session resumption; this might start as Claude Code-only and expand
- The forked session's relationship back to the parent (and the fork point ordinal) should be stored so the UI can visualize conversation trees

### UI ideas

- Right-click context menu on any message: "Fork from here"
- Session detail view showing a tree/timeline of forks
- Visual diff between forked branches (stretch goal)

## Broader vision

I think AgentView is evolving beyond a viewer into a session management tool. People use it in their own way; some for debugging, some for auditing, some for learning. Features like forking lean into the "power user" direction and could set AgentView apart from simple log viewers.

Would love to hear your thoughts on this direction. Happy to discuss scope, priorities, and implementation approach.

rodboev:
AgentsView can resume or fork a whole Claude session, but it cannot start from the message where the user actually wants to branch. The session detail view already exposes message ordinals, and the backend already owns resume command construction, so the missing piece is a message-point contract between the message UI and the resume handler.

This adds a Claude-only message-point fork path that accepts a session ID plus message ordinal, validates that the ordinal belongs to the local Claude session, and builds a deterministic context prefix for a new Claude run. The existing whole-session resume and fork behavior stays unchanged when no ordinal is supplied, and storage remains a data source rather than the place where fork policy lives. Embedded child-session transcripts keep their own session identity, local read-only mode falls back to a copyable command, remote read-only mode hides the action, and each generated prompt file is isolated per launch and self-cleaning when run.

The frontend adds the affordance at the message level instead of overloading the breadcrumb session menu. The focused coverage should prove whole-session preservation, unsupported-agent handling, missing-ordinal handling, and the message-level request body.

Fixes #107

### DIFF
--- a/cmd/agentsview/daemon_runtime_test.go
+++ b/cmd/agentsview/daemon_runtime_test.go
@@ -293,6 +293,67 @@ func startExternalDaemonStartLockHelper(t *testing.T, dataDir string) io.Closer 
 	return stdin
 }
 
+func holdExternalBackgroundLaunchLock(t *testing.T, dataDir string) func() {
+	t.Helper()
+	stdin := startExternalBackgroundLaunchLockHelper(t, dataDir)
+
+	var once sync.Once
+	unlock := func() {
+		once.Do(func() { _ = stdin.Close() })
+	}
+	t.Cleanup(unlock)
+	return unlock
+}
+
+func startExternalBackgroundLaunchLockHelper(
+	t *testing.T,
+	dataDir string,
+) io.Closer {
+	t.Helper()
+	cmd := exec.Command(
+		os.Args[0],
+		"-test.run=^TestHoldExternalBackgroundLaunchLockHelperProcess$",
+	)
+	cmd.Env = append(
+		os.Environ(),
+		"AGENTSVIEW_HOLD_EXTERNAL_BACKGROUND_LAUNCH_LOCK_HELPER=1",
+		"AGENTSVIEW_HOLD_EXTERNAL_BACKGROUND_LAUNCH_LOCK_DATA_DIR="+dataDir,
+	)
+	stdin, err := cmd.StdinPipe()
+	require.NoError(t, err)
+	stdout, err := cmd.StdoutPipe()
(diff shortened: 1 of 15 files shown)


wesm:
looking at this

PR #925, opened 2026-06-29
feat(usage): add pairwise cost comparisons

### LINKED ISSUE #361: Feature: comparative cost analysis
I'd love the ability to look at session aggregates sliced by model and project.
An example: after Opus 4.7 was released ,there's been a lot of noise about whether the new tokenizer made it more expensive. Certainly, it uses more (input) tokens. However, Opus 4.7 seems to be less wordy and able to accomplish similar tasks at lower reasoning thresholds.

It would be really useful to be able to slice the agentsview data to be able to answer questions like "how do models x vs y compare for cost per turn" or "how do models x vs y compare for cost per session", where cost is averaged $ and/or token counts .

rodboev:
The usage dashboard has enough aggregate data to show which models and projects cost the most, but it does not let users compare two choices directly. Issue #361 asks for questions like model A versus model B cost per session or token profile, which currently requires manual filtering and side-by-side arithmetic outside AgentsView.

This adds a pairwise usage comparison endpoint and dashboard panel. The comparison logic lives in the service layer so the direct backend, HTTP backend, server route, and frontend all use one metric definition. Each side applies the same usage filters and the same model or project filter path as the existing summary endpoint, so timezone handling, deduplication, pricing, and eligibility stay aligned with the rest of the usage dashboard.

The UI is additive: existing summary cards, time series, attribution, top sessions, and cache efficiency remain in place. The new panel focuses on model and project comparisons first, with cost, token, and session-based metrics returned by the backend instead of recomputed in the browser.

Fixes #361


wesm:
looking

wesm:
looking

PR #930, opened 2026-06-30
Reuse DuckDB path for Quack sync

cpcloud:
Quack sync now uses the existing DuckDB configuration and command/API paths instead of a parallel Quack config surface.

`AGENTSVIEW_DUCKDB_URL` and `AGENTSVIEW_DUCKDB_TOKEN` drive remote Quack pushes through `agentsview duckdb push`, and `duckdb push --watch` preserves continuous sync on the same backend path. The daemon route remains `/api/v1/push/duckdb`; there is no separate `/push/quack`, `[quack]`, or `AGENTSVIEW_QUACK_*` path.

Remote URL-backed DuckDB/Quack pushes now use a per-target sync-state scope derived from a non-secret URL fingerprint, so switching remotes does not reuse another target's `duckdb_last_push_at` watermark. Local DuckDB file mirrors keep the historical unscoped watermark for compatibility, and `duckdb status` reads whichever scope matches the configured target.

Quack-specific code remains only at the protocol boundary: DuckDB remote attach/opening and `agentsview duckdb quack serve`, which starts DuckDB `quack_serve`. Serve startup projects `quack_serve` results down to listen fields and all serve startup paths now report auth state without printing bearer token values, because stdout and service-manager output can become logs. `duckdb quack serve` requires an explicit token from `--token`, `AGENTSVIEW_DUCKDB_TOKEN`, or `[duckdb].token` rather than generating a hidden or persisted token; Quack URL redaction also strips URL userinfo and secret-like query values before attach errors can print them.

The auth recovery path now also treats `AGENTSVIEW_AUTH_TOKEN` as an actual token override, so supervised deployments can recover browser auth without relying on printed startup tokens.

`duckdb status` also uses the config-backed DuckDB/Quack opener while preserving the existing machine-scoped DuckDB session and message counts, so shared mirrors and remote Quack endpoints do not report global totals for the local machine view.

The tradeoff is that Quack is treated as DuckDB remote transport rather than a new storage backend. That keeps query behavior aligned with the DuckDB mirror while still supporting remote Quack push/read workflows through the existing DuckDB surface.


wesm:
looking

PR #933, opened 2026-06-30
fix(cli): require daemon transport for read commands

mariusvniekerk:
Read commands still had compatibility paths that could open SQLite directly when daemon discovery did not yield HTTP. That undermined the daemon-first migration and left upgraded users exposed to old schema/read-only failures instead of letting the writable daemon own migrations.

This makes daemon-capable reads resolve through daemon transport: they use reachable daemons, start or replace the local daemon where appropriate, and fail with restart guidance when the daemon is unreachable or incompatible instead of falling back. The remaining direct DB access is limited to commands whose contract is explicitly local, offline, or writer-owned, such as raw source export and offline archive queries.

### DIFF
--- a/cmd/agentsview/archive_query_backend.go
+++ b/cmd/agentsview/archive_query_backend.go
@@ -73,7 +73,9 @@ func resolveArchiveQueryBackendWithConfig(
 			case policy.ReadOnlyDaemon == archiveQueryRejectReadOnlyDaemon:
 				return nil, nil, readOnlySessionUsageDaemonError(tr.URL)
 			case policy.ReadOnlyDaemon == archiveQuerySkipReadOnlyDaemon:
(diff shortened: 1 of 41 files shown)


mariusvniekerk:
For the latest roborev finding on `agentsview stats --include-github-outcomes`: accepting this as an intentional daemon credential-boundary tradeoff.

Daemon-backed stats should use daemon-side GitHub credentials. Forwarding a per-invocation CLI token through `/api/v1/session-stats` would expand the credential surface for additive outcome metadata, including potential exposure through request logging, tracing, debugging proxies, or future middleware. The core stats remain correct; only GitHub PR outcome enrichment depends on the daemon having its own token configured.

So the current behavior is acceptable: CLI-local GitHub tokens are not forwarded over the daemon API. Users who need daemon-backed GitHub outcome stats should configure the daemon environment/token.

PR #413, opened 2026-04-29
feat(cursor): ingest sessions from state.vscdb

wesm:
## Summary

Adds Cursor support backed by Cursor's global `state.vscdb` SQLite database in addition to the existing JSONL transcript ingestion. The vscdb source is the canonical one for live Cursor sessions; the parser reconstructs message threads from cursorDiskKV bubble headers and bubbles, persists tool calls with their results, and resolves project names from the adjacent `workspaceStorage/` directories.

When both vscdb and JSONL data exist for the same session, vscdb wins. SyncAll, SyncPaths, and the explicit per-session resync path all consult the stored file_path so a JSONL change does not overwrite the richer vscdb messages, and a `state.vscdb#<id>` virtual path now resyncs through `syncSingleCursorVscdb` when no JSONL fallback exists.

Other notable changes:
- Default `state.vscdb` location is platform-aware (`Library/Application Support/Cursor/...` on macOS, `AppData/Roaming/Cursor/...` on Windows, `.config/Cursor/...` on Linux), overridable via `CURSOR_STATE_DB`.
- SQLite paths are opened through `file:` URIs so reserved characters (`?`, `#`, `%`, spaces) in the filesystem path don't get parsed as DSN parameters.
- `extractWorkspaceProject` decodes `file://` workspace folder URLs with Windows drive-letter and UNC handling.
- vscdb change detection now reparses sessions when the stored `data_version` falls behind the parser's current version, matching the file-sync invariant.

## Origin and follow-ups

Rebased from @badliveware's original branch; review iterations addressed platform handling, dedup completeness across non-SyncAll paths, DSN safety, and data-version reparsing.

This still needs validation by Cursor users:
- Live updates from vscdb depend on the 15-minute scheduled SyncAll today. The file watcher observes only agent dirs; adding `CursorStateDB` to the watch set so vscdb writes propagate immediately is a known follow-up.
- The schema work in `internal/parser/cursor_vscdb.go` was developed against synthetic fixtures. Real-world Cursor schemas across versions need verification against actual installs. Goal is to be a drop-in replacement for the standalone vscdb parser in `~/code/tkmx-client`.

PR #936, opened 2026-06-30
feat(stats): add Cursor attribution parity

### LINKED ISSUE #358: tkmx-client integration: missing coverage vs. retired legacy collectors
## Context

tkmx-client just retired four legacy collectors (`workflow.js`, `codex.js`, `outcomes.js`, and is preparing to retire `cursor.js`) in favor of `agentsview stats --format json`. Most of the coverage is clean — the session_stats v1 schema matches what tkmx-server renders. But a handful of gaps surfaced during the cutover that agentsview needs to close before the last legacy collector can be removed.

Filed as a single tracker issue so we can split into follow-ups as needed.

## Gap 1 — Cursor `ai-tracking.db` parity (blocks retiring `reporter/cursor.js`)

Cursor's primary value prop is per-commit **line-level attribution** (tab completion vs composer vs human). That data lives in `~/.cursor/ai-tracking/ai-code-tracking.db`, not in Cursor's session logs. `agentsview` currently only reads the session logs, so on this user's real data the Cursor session count was 8 total (most recent: 2026-02-27), while the legacy `cursor.js` reports active commit-level usage.

Fields tkmx-client's retired `cursor.js` emitted (shape for reference):

- `scored_commits` — count of commits the `scored_commits` table has attribution for
- `tab_lines_added`, `tab_lines_deleted`
- `composer_lines_added`, `composer_lines_deleted`
- `human_lines_added`, `human_lines_deleted`
- `ai_authored_pct` — (tab + composer added) / (total added)
- `conversations` — `{ \"<model>/<mode>\": count }` from `conversation_summaries`

Source schema:

- `~/.cursor/ai-tracking/ai-code-tracking.db` → `scored_commits` (columns include `commitDate`, `tabLinesAdded`, `composerLinesAdded`, `humanLinesAdded`, and the corresponding `*Deleted`)
- Same DB → `conversation_summaries` (columns include `model`, `mode`, `updatedAt`)

### What would land the gap

- Add a Cursor syncer that ingests `ai-code-tracking.db` alongside the existing session log sync
- Surface the line-attribution aggregates as either:
  - an extension to `session_stats.agent_portfolio.cursor` (e.g. a `cursor_attribution` sub-object), or
  - a new top-level `code_attribution` block that could also accommodate GitHub Copilot's similar telemetry in the future
- Model+mode conversation counts can fold into `session_stats.model_mix.by_tokens` (if token counts are available) or a separate `cursor.conversations` breakdown

The first shape is probably best — it keeps Cursor's quirks scoped.

## Gap 2 — `outcome_stats.commits = 0` on real repos with activity

`agentsview stats --format json --since 28d` reports `outcome_stats: {\"repos_active\": 50, \"commits\": 0, \"loc_added\": 0, \"loc_removed\": 0, \"files_changed\": 0}` on a machine with 462 commits across 19 active repos (confirmed via the legacy `outcomes.js` git-log path).

`repos_active` working correctly (repos are being discovered from session cwds), but the git-log aggregation isn't producing commit/LOC totals. Likely a data-path issue rather than a logic bug — maybe the author-email scoping or the `--author` anchor isn't matching once you're aggregating across worktrees with different git configs. Legacy `outcomes.js` applied a recent fix for this (\"scope author email to --local and anchor git log --author\") worth diffing against.

### What would land the gap

- Verify agentsview's git-log invocation matches the behavior `reporter/outcomes.js` lands at by v1.3.0 of tkmx-client (specifically commits `bbbf908` and `6b5be35` in that repo)
- Likely relevant: author-email discovery per-repo, de-duping commits by SHA across sibling worktree checkouts, `--no-merges` convention

## Gap 3 — `tool_mix.by_category` casing convention

`agentsview stats` emits TitleCase tool names (`Bash`, `Edit`, `Read`, `Grep`, `Glob`, `Write`, `Task`, `Other`, `Tool`). tkmx-server's renderer was originally built against lowercase (`read`, `edit`, `shell`, ...) and has been updated to accept both, but worth locking down a convention upstream.

### What would land the gap

- Document the canonical casing in the session_stats v1 schema (or add a `/docs/session-stats-v1.md` if it doesn't exist)
- Decide whether `Bash` vs `shell` is the preferred key for shell-type tools — legacy clients used `shell`, agentsview uses `Bash`

## Gap 4 — `reporter_timezone: \"Local\"` sentinel

On this user's machine, `temporal.reporter_timezone` came back as the string `\"Local\"` (from the Go `time.Local` sentinel) rather than an IANA zone like `America/Los_Angeles`. Downstream consumers (tkmx-server's canonical-TZ picker) treat this as a valid timezone and pass it to Luxon, which can cause subtle bugs in streak / heatmap computation.

### What would land the gap

- Resolve `time.Local` to its canonical IANA name before emitting the blob, or
- Emit an empty string in that case so the downstream picker's fallback chain (owner TZ → reporter TZ → UTC) kicks in correctly

## Priority

1. **Gap 2** (outcome_stats) blocks the cross-agent outcome display that's already shipping in tkmx-server's profile page — highest priority
2. **Gap 1** (Cursor) blocks retiring the final tkmx-client legacy collector
3. **Gap 4** (timezone) is a correctness issue for any non-IANA resolver
4. **Gap 3** (casing) is cosmetic / documentation

## Cross-references

- tkmx-client retirement: https://github.com/srosro/tkmx-client/commit/1c669a6 (workflow/codex/outcomes)
- tkmx-server branch waiting on these: `session-stats-profile`

rodboev:
tkmx-client still needs Cursor attribution parity before it can retire the last legacy Cursor collector. AgentsView already ingests Cursor transcripts, but it still ignores `~/.cursor/ai-tracking/ai-code-tracking.db`, so the stats payload has no commit-level tab, composer, human, or conversation-count attribution from Cursor itself.

This keeps the scope to the surviving Cursor attribution slice from #358. The new reader lives beside the existing Cursor parser, can be redirected with `AGENTSVIEW_CURSOR_ATTRIBUTION_DB`, follows Cursor's live schema by windowing scored commits on `scoredAt` while leaving the text `commitDate` column as metadata, and leaves transcript discovery untouched. Aggregation stays owned by service stats code, and the result lands in a dedicated top-level `cursor_attribution` block rather than being mixed into git-derived `outcome_stats` or folded into the session-derived `agent_portfolio` maps. That preserves the meaning of the existing stats owners while giving downstream consumers one explicit Cursor attribution surface to read.

Project include and exclude filters still suppress the block because Cursor's local attribution DB has no project key, and non-Cursor agent filters keep it nil for the same reason. The human-readable stats output only renders the section when the service-owned block is present.

PR #413 stays separate. It covers Cursor `state.vscdb` session ingestion, which is a different Cursor data source from `ai-code-tracking.db` and does not provide the line-attribution totals required here.

Closes #358

### DIFF
--- a/internal/parser/cursor_attribution.go
+++ b/internal/parser/cursor_attribution.go
@@ -160,9 +160,12 @@ func cursorAttributionDBPath() string {
 }
 
 func openCursorAttributionDB(path string) (*sql.DB, error) {
+	// The file: prefix is required for go-sqlite3 to honor mode=ro;
+	// without it the query string is dropped and the live Cursor db
+	// would be opened read-write.
 	conn, err := sql.Open(
 		"sqlite3",
-		path+"?mode=ro&_busy_timeout=3000",
+		"file:"+path+"?mode=ro&_busy_timeout=3000",
 	)
 	if err != nil {
 		return nil, fmt.Errorf("opening cursor attribution db: %w", err)
--- a/internal/parser/cursor_attribution.go
+++ b/internal/parser/cursor_attribution.go
@@ -93,8 +93,8 @@ func LoadCursorAttribution(
 			COUNT(*)
 		FROM conversation_summaries
 		WHERE updatedAt >= ? AND updatedAt < ?
-		GROUP BY model, mode
-		ORDER BY COUNT(*) DESC, model, mode`,
+		GROUP BY COALESCE(model, ''), COALESCE(mode, '')
+		ORDER BY COUNT(*) DESC, COALESCE(model, ''), COALESCE(mode, '')`,
 		timeToMillis(from), timeToMillis(to),
 	)
 	if err != nil {
--- a/internal/parser/cursor_attribution.go
+++ b/internal/parser/cursor_attribution.go
@@ -1,0 +1,186 @@
+package parser
+
+import (
+	"database/sql"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"time"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+type CursorAttribution struct {
+	ScoredCommits        int64                     `json:"scored_commits"`
+	LinesAdded           int64                     `json:"lines_added"`
+	LinesDeleted         int64                     `json:"lines_deleted"`
+	TabLinesAdded        int64                     `json:"tab_lines_added"`
+	TabLinesDeleted      int64                     `json:"tab_lines_deleted"`
+	ComposerLinesAdded   int64                     `json:"composer_lines_added"`
+	ComposerLinesDeleted int64                     `json:"composer_lines_deleted"`
+	HumanLinesAdded      int64                     `json:"human_lines_added"`
+	HumanLinesDeleted    int64                     `json:"human_lines_deleted"`
+	BlankLinesAdded      int64                     `json:"blank_lines_added"`
+	BlankLinesDeleted    int64                     `json:"blank_lines_deleted"`
+	AIAuthoredPct        float64                   `json:"ai_authored_pct"`
+	ConversationCounts   []CursorConversationCount `json:"conversation_counts,omitempty"`
+}
+
+type CursorConversationCount struct {
+	Model string `json:"model"`
+	Mode  string `json:"mode"`
+	Count int64  `json:"count"`
+}
+
+func LoadCursorAttribution(
+	from, to time.Time,
+) (*CursorAttribution, error) {
+	dbPath := cursorAttributionDBPath()
+	if dbPath == "" {
+		return nil, nil
+	}
+	if _, err := os.Stat(dbPath); err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("stat cursor attribution db: %w", err)
+	}
+
+	conn, err := openCursorAttributionDB(dbPath)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	attr := &CursorAttribution{}
+	if err := conn.QueryRow(
+		`SELECT
+			COUNT(*),
+			COALESCE(SUM(linesAdded), 0),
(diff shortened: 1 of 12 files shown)


wesm:
fixing windows CI

wesm:
This PR has cleared review but I think we need to take a step back and think about how this API is realistically going to evolve to support many harnesses and providers. Since I need to get a release out to fix bugs in 0.35.0 I'm going to cut that now and then take time to give this appropriate consideration

internal/parser/cursor_attribution.go
```diff
@@ -0,0 +1,194 @@
...
+	"os"
+	"path/filepath"
+	"sort"
+	"time"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+type CursorAttribution struct {
+	ScoredCommits        int64                     `json:"scored_commits"`
```
mariusvniekerk, line 15: Given that none of this data is used anywhere by agentsview why bother?

wesm: There are some downstream consumers of this data like https://github.com/srosro/tkmx-client that need it I think

wesm: looking more closely

wesm:
"  The `code_attribution` block is an export surface rather than something
  agentsview consumes internally — the point of the PR is to let downstream
  consumers (tkmx-client is the immediate one) retire their ad hoc readers of
  `ai-code-tracking.db` and take this from `stats --format json` instead.
  tkmx-client ships `scored_commits` in its report payload today and gates the
  whole commit block on it (`reporter/cursor.ts:100-101`), and logs it per run
  (`reporter/report.ts:368`), so dropping the field would block that migration.
  It also does work on its own terms: it's what distinguishes
  `status: "empty"` from `"available"`, and it's the record count that makes
  the line totals interpretable — "40 AI lines across 2 commits" and "40 AI
  lines across 400 commits" are different signals. Cursor only attributes
  lines on commits it has scored, so the count tells a consumer how much of
  the window's activity the attribution actually covers.
"

PR #943, opened 2026-07-01
Improve Windows test suite fixture reuse

wesm:
Windows CI has been spending a growing amount of time in unit-test setup rather than in the assertions themselves. This branch consolidates compatible read-only or sequential fixture scenarios so repeated SQLite archive, sync engine, HTTP server, and provider setup is reused where the tests are already isolated by IDs, projects, dates, or source paths.

The behavior checks remain in place as subtests, while mutation-heavy and failure-path cases stay separate. The main tradeoff is that a few test functions are larger, but they now make the shared fixture boundaries explicit and avoid paying the same setup cost over and over on Windows.

The branch also adds repo-level LF normalization so Git for Windows no longer falls back to contributor-local autocrlf behavior and produces line-ending-only churn during this kind of test work.

Reviewers should look closely at the fixture consolidations in internal/sync, internal/db, cmd/agentsview, and internal/server to confirm each merged case still preserves the original preconditions and does not mask unintended state coupling.


wesm:
This does appear to shave a couple minutes off the Windows test suite runtime

wesm:
Roborev's note about `windows.SetInformationJobObject` appears to be a false positive on the current dependency set. This repo is using `golang.org/x/sys v0.46.0`, where `SetInformationJobObject` is generated as `func SetInformationJobObject(...) (ret int, err error)`, so the two-value assignment is correct. The current head `21426b6` also has passing Windows checks for both `Go Test (windows-latest)` and `Desktop Build (Windows)`.

PR #945, opened 2026-07-01
Fix DuckDB push schema setup through Quack

cpcloud:
Pushing to a Quack-backed DuckDB was broken in layers: schema setup issued `CREATE INDEX` against proxy tables and failed instantly; one poisoned row (a NUL byte in stored tool output) aborted the entire single-transaction push, forever; every statement paid its own HTTPS round trip (a measured 15,000x below line rate); and any server restart permanently stranded every connected client. This PR makes Quack push and serve correct first, then fast.

**Correctness**

- Schema setup, status/store reads, and mutations now use call shapes Quack attachments can actually execute; index and migration work runs only on the real server database.
- NUL bytes are stripped from persisted tool result content at ingest, with a data-version bump so existing rows are re-ingested. DuckDB VARCHARs cannot hold them, and one such row poisoned every push.
- Session batches commit in explicit server-side transactions. A failed batch retries per session; a persistently failing session is skipped and counted instead of aborting the run. Context cancellation stays fatal, and the sync watermark only advances on clean runs.
- Clients recover from server restarts: stale or detached attachments are detected and re-attached on one pinned connection, with a single retry for self-contained calls. Previously every client failed with `Invalid connection id` until process restart. Attach errors scrub credentials.

**Performance**

- Batches coalesce into one multi-statement `BEGIN`/`COMMIT` `query()` call instead of one round trip per statement.
- Batches are bounded by rendered bytes (2 MiB) so each POST executes within the Quack client's fixed 120 s deadline on small servers. Oversized single sessions leave the coalesced path and execute rendered mutation groups inside one remote transaction, preserving session atomicity without returning to per-row round trips.
- Adjacent single-row INSERTs render as bounded multi-row `VALUES` statements (~10x less server-side parse/dispatch), and tool rows push as read-free delete+insert instead of per-row existence checks that previously cost hundreds of thousands of serial reads per catch-up.
- A timed-out coalesced batch may still be executing server-side, so timeouts back off and retry once instead of immediately grinding per-session retries against the abandoned writer.

**Tradeoffs**

- The 120 s HTTP deadline is hard-coded in the Quack extension (its ATTACH `TIMEOUT` option is ignored), so this PR sizes normal coalesced work under the deadline rather than raising it.
- Oversized sessions can still be slower than normal coalesced batches, but they keep delete-and-replace work atomic while using rendered INSERT groups to avoid per-row round trips.
- The per-session retry path deliberately keeps statement-by-statement execution so a failing session still reports the exact statement that rejected it.

Core changes in `internal/duckdb/{sync,push,connect}.go`; real-server coverage, including restart recovery and rollback behavior, in `internal/duckdb/quack_smoke_duckdbtest_test.go`.

### DIFF
--- a/internal/duckdb/push.go
+++ b/internal/duckdb/push.go
@@ -458,18 +458,17 @@ func appendDuckRemoteMutationBatch(
 	if maxBytes <= 0 {
 		maxBytes = duckRemoteMutationCoalesceMaxBytes
 	}
-	if next.transactionBytes() > maxBytes {
+	nextBytes := next.transactionBytes()
+	if nextBytes > maxBytes {
 		if err := execDuckRemoteMutationBatch(
 			ctx, execCoalesced, label, current, true,
 		); err != nil {
 			return current, err
 		}
-		if err := execDuckRemoteMutationBatchChunks(
-			ctx, execCoalesced, label, next, maxBytes,
-		); err != nil {
-			return &duckRemoteMutationBatch{}, err
-		}
-		return &duckRemoteMutationBatch{}, nil
+		return &duckRemoteMutationBatch{}, fmt.Errorf(
+			"duckdb remote mutation batch exceeds remote mutation coalesce budget: %d > %d",
+			nextBytes, maxBytes,
+		)
 	}
 	if current.Len() > 0 && current.combinedTransactionBytes(next) > maxBytes {
 		if err := execDuckRemoteMutationBatch(ctx, execCoalesced, label, current, true); err != nil {
@@ -549,22 +548,7 @@ func execDuckRemoteMutationBatchWithStatementFallback(
 	return nil
 }
 
-func execDuckRemoteMutationBatchChunks(
-	ctx context.Context,
-	exec func(context.Context, string) error,
-	label string,
-	batch *duckRemoteMutationBatch,
-	maxBytes int,
-) error {
-	for _, chunk := range batch.chunks(maxBytes) {
-		if err := execDuckRemoteMutationBatch(ctx, exec, label, chunk, true); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-func execDuckRemoteMutationBatchChunksWithStatementFallback(
+func execDuckRemoteMutationBatchOversizeWithStatementFallback(
 	ctx context.Context,
 	execCoalesced func(context.Context, string) error,
 	execStatement func(context.Context, string) error,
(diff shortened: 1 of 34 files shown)


internal/duckdb/push.go
```diff
@@ -44,7 +44,24 @@ func (s *Sync) syncModelPricing(ctx context.Context) error {
 
 	for i := 0; i < len(prices); i += duckPricingUpsertBatch {
 		end := min(i+duckPricingUpsertBatch, len(prices))
-		query, args := duckPricingUpsertStatement(prices[i:end])
+		batch := prices[i:end]
+		if s.connectionKind == duckDBQuackClientConnection {
```
cpcloud: Why are we branching here?

internal/duckdb/push.go
```diff
@@ -373,7 +481,9 @@ func upsertSession(
 			?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,
 			?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,
 			?, ?, ?, ?, ?, ?, ?, ?
-		)
+		)`
+	if upsert {
```
cpcloud: it seems super weird that you're not just using upsert.

cpcloud (commented):
overall not entirely following why pushing to the remote requires all this special casing.

internal/duckdb/push.go
cpcloud: Resolved in `b90a4ae`: the connection-kind branch now selects how mutation SQL is executed, not a separate write implementation. Local DuckDB keeps normal transaction execution, while Quack sends the same mutation SQL through `agentsview_remote.query(?)` so DuckDB plans it against the base tables instead of proxy tables.

internal/duckdb/push.go
cpcloud: Resolved in `b90a4ae`: sessions, model pricing, and cursor usage now keep their `ON CONFLICT` upserts. For Quack those upserts are executed server-side through the query hook, which preserves the intended conflict semantics while avoiding DuckDB proxy-table limitations. The dependent row sets still use the update/insert/delete path because DuckDB rejects delete-plus-reinsert of the same unique keys within one transaction.

cpcloud:
Closing the loop on the review question about the Quack push special-casing: the branch is not trying to create a second write implementation. It keeps the same mutation SQL and conflict semantics, but changes where DuckDB plans mutation-shaped statements when the target is a Quack attachment.

The client attachment exposes proxy tables, and DuckDB rejects index DDL and mutation/conflict shapes such as `DELETE`, `UPDATE`, and `ON CONFLICT` when those are planned against proxy storage. Local/base DuckDB still uses the normal `tx.ExecContext` path. Quack-backed push uses the connection-kind branch to send those same mutation statements through `agentsview_remote.query(?)`, so the Quack server plans them against the real base tables.

The final smoke coverage now mutates and re-pushes a session, verifies timestamp round-tripping, verifies tool row updates, then removes the tool call and verifies stale dependent rows are deleted. That is the behavior the branch needs beyond the initial index-skipping schema fix.

mariusvniekerk (approved):

PR #947, opened 2026-07-01
fix(usage): surface unsupported Copilot usage filters

### LINKED ISSUE #349: Usage dashboard is silently empty for Copilot CLI sessions
## What I saw

With only `copilot` and `vscode-copilot` agents selected over a 17-day window (13 and 139 sessions respectively), the Usage dashboard shows zeros and "No data for this period" everywhere — no explanation, no hint that it's an unsupported agent.

Image

The Model filter dropdown is also empty of Copilot's models — only Claude Code models appear, even though Copilot CLI sessions exist and report models like `gpt-5.3-codex` in their session events:

Image

The CLI has the same silent behavior:

```text
$ agentsview usage daily --agent copilot --since 2026-04-01
DATE   INPUT  OUTPUT  CACHE_CR  CACHE_RD  COST   MODELS
----   -----  ------  --------  --------  ----   ------
----   -----  ------  --------  --------  ----   ------
TOTAL  0      0       0         0         $0.00
```

With no message, "$0.00" is easy to misread as "I didn't spend anything". I'm opening this to track the problem — not because I know how it should be resolved.

## Why it's empty

The [token-usage docs](https://www.agentsview.io/token-usage/) note that only Claude Code, Codex, OpenCode, and Pi record per-message `token_usage`. For Copilot this isn't just a parser gap — Copilot CLI session files in `~/.copilot/session-state/<uuid>/events.jsonl` do not contain input/output/cache token counts or cost information at all. The event stream is limited to model names, session/user/assistant/tool events, and context metadata.

Copilot also doesn't use per-token billing. Subscribers have a monthly quota of "premium interactions", so even a populated `token_usage` wouldn't map onto the existing cost model cleanly. That's part of why a long-term answer is non-obvious.

## Near-term: make the gap visible

Even without a full cost story, some signal that these agents aren't in the tokenized set would remove the "did I misconfigure something?" confusion. A few things that could help:

- A note on the Usage page when the active filter is an agent that doesn't contribute `token_usage` (e.g., "Copilot CLI sessions don't record per-message token data — see [doc link]").
- Same note from the CLI when `--agent <unsupported>` is passed, or when the filtered result is entirely empty because of agent selection.
- Optionally, visually mark these agents in the agent filter itself (e.g., a "no token data" tag next to the entry).

## Longer-term: could quota data be the answer?

For users whose primary agent is Copilot CLI — for example, roborev code-review workflows that invoke Copilot CLI per review — premium-interaction quota is the binding constraint, and there is currently no way to see how much is being consumed over time.

GitHub exposes an undocumented endpoint that returns per-quota snapshots for the authenticated user. It is already used by Microsoft's Copilot extension in VS Code and by `oh-my-posh`, which suggests it is de facto stable even though it isn't in the public REST API reference:

```bash
gh api copilot_internal/user
```

Mid-cycle example (values illustrative):

```json
{
  "login": "<user>",
  "copilot_plan": "business",
  "quota_reset_date": "2026-05-01",
  "quota_snapshots": {
    "premium_interactions": {
      "entitlement": 300,
      "remaining": 128,
      "percent_remaining": 42.7,
      "overage_permitted": true,
      "quota_reset_at": 0
    }
  }
}
```

Over-quota example (what the same shape looks like when overage is active):

```json
{
  "login": "<user>",
  "copilot_plan": "business",
  "quota_reset_date": "2026-05-01",
  "quota_snapshots": {
    "premium_interactions": {
      "entitlement": 300,
      "remaining": -6,
      "percent_remaining": -2.0,
      "overage_permitted": true,
      "quota_reset_at": 0
    }
  }
}
```

`remaining` goes negative while `overage_permitted` is true, so the shape is expressive enough to show both "you have X left" and "you're N over".

Known references for the endpoint:
- `microsoft/vscode` Copilot extension: https://github.com/microsoft/vscode/blob/main/extensions/copilot/CONTRIBUTING.md
- `oh-my-posh` copilot segment: https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/segments/copilot.go

The call can piggyback on `gh auth token` (no new PAT required).

## Open questions

I don't know how this should integrate with the existing Usage model. Some things I'm not sure about:

- Should quota live in Usage at all, or in a separate "Copilot" view?
- How should it be presented when multiple agents are mixed in the same filter?
- Is the undocumented-endpoint dependency an acceptable tradeoff, and if so, gated behind an opt-in flag or on by default?
- Is there a way to show quota in terms comparable to cost (e.g., "interactions per day trend") so it at least fits visually next to the token charts?

Happy to iterate on any of this based on what makes sense for the project.

## Constraints worth weighing

- The quota endpoint is undocumented. It could change without notice. A graceful "quota unavailable" fallback would be important.
- Agentsview already makes outbound calls (LiteLLM pricing, update checks), but those are unauthenticated. A quota fetch would introduce the first authenticated-call pattern: either shelling out to `gh` (new runtime dependency) or reading a token from config.
- Quota data is user-global, not per-session, so it lives alongside the existing token aggregates rather than inside them — a real conceptual mismatch with the current Usage model.

## Related

- Token-usage doc (lists supported agents): https://www.agentsview.io/token-usage/
- #35 — Insights for Copilot
- #67 — Export and programmatic access to insights
- #229 — Add token usage metrics (originating design; "allow blanks / partial values when a source does not provide token usage")

rodboev:
GitHub Copilot sessions really do arrive without per-message token counters or per-message cost, so the Usage dashboard's current zero-state is misleading rather than accurate when a user filters down to Copilot-only agents. The page shows zero-value cards and generic empty panels even when matching Copilot sessions exist in the selected range, which reads as "nothing was spent" instead of "this agent family does not expose token usage."

PR #934 already fixed the CLI half of this issue and explicitly scoped the dashboard half out because it needs a filter-faithful matching-session count from the sessions table, not a usage-row count. Current summary assembly still has that exact gap: the dashboard only receives totals plus usage-row-gated session counts, so it cannot tell the difference between unsupported Copilot usage and a truly empty filter result. This change adds that missing cross-backend session-presence helper behind the shared store interface, then exposes one narrow machine-readable summary signal only when the filter is Copilot-only, the totals still have no token data, and matching sessions do exist. The frontend uses that signal to render a short note above the summary cards and leaves the existing charts, tables, and generic empty states unchanged.

Scope stays tight. There are no CLI changes here, because #934 already shipped them, and the separate model-picker complaint in the issue remains out of scope because it is a catalog-population problem, not a usage-visibility problem. Reviewers should look first at the new matching-session helper and its backend parity tests, then at the small summary-contract addition that the dashboard consumes.

Fixes #349

### DIFF
--- a/frontend/messages/en.json
+++ b/frontend/messages/en.json
@@ -686,6 +686,7 @@
   "usage_refresh": "Refresh usage data",
   "usage_summary_total_cost": "Total Cost",
   "usage_summary_copilot_ai_credits": "Copilot AI Credits",
+  "usage_summary_unsupported_copilot_no_token_data": "Copilot sessions matched this range, but Copilot does not record per-message token usage.",
   "usage_summary_input_tokens": "Input Tokens",
   "usage_summary_daily_burn": "Daily Burn",
   "usage_summary_peak_day": "Peak Day",
--- a/frontend/messages/zh-CN.json
+++ b/frontend/messages/zh-CN.json
@@ -1551,6 +1551,7 @@
   "usage_refresh_data": "刷新用量数据",
   "usage_vs_prior": "{pct}% 较上期",
   "usage_total_cost": "总成本",
+  "usage_summary_unsupported_copilot_no_token_data": "此范围内匹配到了 Copilot 会话，但 Copilot 不会记录每条消息的 token 用量。",
   "usage_copilot_credits": "Copilot AI 积分",
   "usage_input_tokens": "输入 Token",
   "usage_cached_suffix": "+{tokens} 已缓存",
--- a/frontend/messages/zh-TW.json
+++ b/frontend/messages/zh-TW.json
@@ -1551,6 +1551,7 @@
   "usage_refresh_data": "刷新用量數據",
   "usage_vs_prior": "{pct}% 較上期",
   "usage_total_cost": "總成本",
+  "usage_summary_unsupported_copilot_no_token_data": "此範圍內匹配到了 Copilot 會話，但 Copilot 不會記錄每條訊息的 token 用量。",
   "usage_copilot_credits": "Copilot AI 積分",
   "usage_input_tokens": "輸入 Token",
   "usage_cached_suffix": "+{tokens} 已緩存",
--- a/frontend/src/lib/api/generated/index.ts
+++ b/frontend/src/lib/api/generated/index.ts
@@ -148,6 +148,7 @@ export type { SyncSyncStats } from './models/SyncSyncStats';
 export { TerminalConfigBody } from './models/TerminalConfigBody';
 export type { TerminalResponse } from './models/TerminalResponse';
 export type { TrashResponse } from './models/TrashResponse';
+export type { UnsupportedUsage } from './models/UnsupportedUsage';
 export type { UpdateCheckResponse } from './models/UpdateCheckResponse';
 export type { UploadSessionResponse } from './models/UploadSessionResponse';
 export type { UsageSummaryResponse } from './models/UsageSummaryResponse';
--- a/frontend/src/lib/api/generated/models/UnsupportedUsage.ts
+++ b/frontend/src/lib/api/generated/models/UnsupportedUsage.ts
@@ -1,0 +1,7 @@
+/* generated using openapi-typescript-codegen -- do not edit */
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+export type UnsupportedUsage = {
+  kind: string;
+};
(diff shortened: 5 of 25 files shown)


wesm:
looking

wesm:
Merging as is. One follow-up worth filing separately rather than blocking this PR:

The Copilot-only trigger for the no-token-data hint relies on agent-name string matching (`IsCopilotAgent` / `IsCopilotAgentFilter` in `internal/db/usage.go`), now shared by the CLI (#934) and this dashboard path. If a fourth Copilot variant or another agent without per-message token data is added, someone has to remember to update that function. The deeper home would be capability flags on the agent registry in `internal/parser/types.go` — note it likely needs more than one flag, since existing call sites use `IsCopilotAgent` both for "doesn't record tokens" and for "cost is denominated in AI credits".

Related sharp edge for the same issue: the frontend renders nothing for an `unsupportedUsage.kind` it doesn't recognize, so a future second kind is invisible until the frontend and all three locale files ship in lockstep. A generic fallback message would let new kinds degrade visibly.

PR #948, opened 2026-07-01
feat(frontend): show parser malformed-lines badge on session detail

mjacobs:
#866 started surfacing parser anomaly signals in the CLI sync summary; the per-session `parser_malformed_lines` count has been serialized in the session API since then, but nothing in the UI rendered it. This adds the missing read side: a warning badge in the session-detail breadcrumb, shown only when the count is non-zero, with a tooltip explaining that N lines in the source file could not be parsed.

The badge reuses the local badge vocabulary already in `SessionBreadcrumb.svelte` — it sits next to the summary-mode badge from #914 and follows the same conditional-pill pattern — rather than introducing a new shared component. The hand-maintained `Session` type gains the already-serialized field, and plural-aware message keys are added for en, zh-CN, and zh-TW.

Frontend-only; no backend or schema changes. Scope is deliberately the session-detail surface — the session list is untouched.

Where to look: `SessionBreadcrumb.svelte` for the badge and its gating, `core.ts` for the type addition.

### DIFF
--- a/frontend/messages/en.json
+++ b/frontend/messages/en.json
@@ -172,6 +172,36 @@
   "session_breadcrumb_failed_to_open": "Failed to open",
   "session_breadcrumb_summary_mode": "Summary mode",
   "session_breadcrumb_summary_mode_tooltip": "Tool results, thinking, and diffs are missing from this transcript. Full fidelity needs the agy-reader sidecar — install it, or if installed, let it catch up. Click to learn how.",
+  "session_breadcrumb_malformed_lines": [
+    {
+      "declarations": [
+        "input count",
+        "local countPlural = count: plural"
+      ],
+      "selectors": [
+        "countPlural"
+      ],
+      "match": {
+        "countPlural=one": "{count} malformed line",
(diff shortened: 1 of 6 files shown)


mjacobs:
The `Go Test (windows-latest)` failure is unrelated to this diff (frontend-only). It's a cold-runner flake in the shared test-DB template from #943: `buildTestDBTemplate` runs a WAL checkpoint under a 1s context deadline inside a `sync.Once` that caches the error, so one slow checkpoint (`TestResolveActivityReport_BadBucket (1.31s)` is the once-holder) failed all 599 subsequent `testDB` openings in that binary with the same cached `checkpointing db template: wal checkpoint truncate: context deadline exceeded`. The same SHA passed on main's push run. Hardening proposed in #951; a re-run of the job should meanwhile go green.

cpcloud:
@mjacobs can you throw up a screenshot for this?

wesm:
looking

wesm:
Screenshot with fake data for the malformed-lines badge on the session detail breadcrumb:

PR #949, opened 2026-07-01
feat(parse-diff): cover DB-backed Warp, Forge, and Piebald providers

mjacobs:
parse-diff v1 (#662) rejected Warp, Forge, and Piebald because their sync phases had woven-in change detectors and no `DiscoverFunc`. The provider facade (#876–#885, #924) made that reason obsolete — all three now have a unified `Discover()` via `dbBackedProviderFactory` — but both parse-diff gates still keyed on `FileBased`, which is false for these shared-SQLite stores.

This relaxes the two gates (`parseDiffAgentDiscoverable`, `parseDiffAgentSupported`) to admit provider-authoritative agents with registered factories, instead of flipping `FileBased`: that flag is read at nine-plus other sites (watcher, sync engine, token accounting, remote sync, SSH resolve, settings) and genuinely means "reads a literal per-session file", so flipping it would change unrelated behavior.

Two hazards the wider gate exposed are handled in the same change:

- Virtual `<db>#<sessionID>` sources cannot be stat'd, so the raced-skew reclassification from #805 would have masked every real drift on these agents as "raced". `stripVirtualSourceSuffix` now knows their DB filenames and `parseDiffSourceReliableForRaced` re-requires `FileBased`, so DB-backed agents fail closed toward reporting a change rather than masking one.
- These providers discover one source per session (like OpenCode, unlike per-DB Kiro), so `--limit` presence-keying moved to a shared per-session base list; sessions cut by `--limit` now report "not sampled" instead of false-positive presence findings.

Quack (#930) is out of scope: it is the DuckDB remote-sync transport, not a registry agent with session sources to re-parse.

`--limit` ordering: the provider stamps each session's real mtime onto an additive, advisory `SourceRef.DiscoveryMTimeNS` while `Discover()` already has the store's session metas in hand, and the parse-diff sampler prefers it over the failed-stat fallback — so limited runs sample these virtual sources newest-first like file-based agents. The field is ordering-only metadata; skip-cache and data-version freshness still resolve through `Fingerprint`, and the raced-guard is unchanged.

Where to look: `internal/sync/parsediff.go` (gate, raced predicate, per-session base keying) and `internal/sync/parsediff_dbbacked_test.go` for the end-to-end coverage.

### DIFF
--- a/internal/parser/provider.go
+++ b/internal/parser/provider.go
@@ -176,6 +176,16 @@ type SourceRef struct {
 	FingerprintKey string
 	// ProjectHint is advisory metadata for UI grouping and may be empty.
 	ProjectHint string
+	// DiscoveryMTimeNS is an optional per-source modification time in Unix
+	// nanoseconds captured at discovery. Providers whose sources are virtual --
+	// a shared store fanned out to one source per session, where DisplayPath is
+	// "<db>#<sessionID>" and os.Stat cannot resolve a real mtime -- set it so
+	// ordering consumers (parse-diff's --limit sampler) can rank sources by each
+	// session's real mtime instead of a failed stat that collapses to 0. Zero
+	// means unset. It is advisory ordering metadata only: it is never persisted
+	// and must not be used for skip-cache or data-version freshness, which go
+	// through Fingerprint.
+	DiscoveryMTimeNS int64
 	// Opaque is in-memory-only source state: never persisted and never required
 	// for lookup from persisted rows, so any source that must survive a restart
 	// has to be recoverable from Key, DisplayPath, FingerprintKey,
--- a/cmd/agentsview/parse_diff.go
+++ b/cmd/agentsview/parse_diff.go
@@ -83,7 +83,7 @@ func newParseDiffCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringArrayVar(&cfg.Agents, "agent", nil,
-		"Restrict to these agents (repeatable; default: all file-based agents)")
+		"Restrict to these agents (repeatable; default: all re-parseable agents)")
 	cmd.Flags().IntVar(&cfg.Limit, "limit", 0,
 		"Re-parse only the N most recently modified source files (0 = all)")
 	cmd.Flags().BoolVar(&cfg.FailOnChange, "fail-on-change", false,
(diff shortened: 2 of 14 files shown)


mjacobs:
The single `Go Test (windows-latest)` failure (`TestEnsureBackgroundServeLaunchLoserIgnoresReadOnlyRuntimeDuringReplacement`: "server did not become ready within 2s") is a pre-existing readiness-timeout flake on slow Windows runners — this PR doesn't touch the serve/daemon path. The #948 run hit a different Windows flake in the same job; diagnosis in #948's thread and proposed hardening in #951.

mjacobs:
Fixed in `1a889f01` + `1989daae`. The per-session mtime was already in hand where `Discover()` enumerates the store's session metas, so it's stamped onto an additive, advisory `SourceRef.DiscoveryMTimeNS` there and the `--limit` sampler prefers it over the failed-stat fallback — O(N), no per-source re-fingerprinting. Ordering-only: skip-cache and data-version freshness still resolve through `Fingerprint`, and the raced-guard is untouched. New tests cover a newer session with a lexicographically earlier virtual path winning the sample, and the follow-up commit exports the three store filenames so parse-diff keys off the parser constants instead of duplicated literals. The PR description's "Known limitation" paragraph is updated accordingly.

internal/parser/provider.go
```diff
@@ -176,6 +176,16 @@ type SourceRef struct {
...
+	// DiscoveryMTimeNS is an optional per-source modification time in Unix
+	// nanoseconds captured at discovery. Providers whose sources are virtual --
+	// a shared store fanned out to one source per session, where DisplayPath is
+	// "<db>#<sessionID>" and os.Stat cannot resolve a real mtime -- set it so
+	// ordering consumers (parse-diff's --limit sampler) can rank sources by each
+	// session's real mtime instead of a failed stat that collapses to 0. Zero
+	// means unset. It is advisory ordering metadata only: it is never persisted
+	// and must not be used for skip-cache or data-version freshness, which go
+	// through Fingerprint.
+	DiscoveryMTimeNS int64
```
mariusvniekerk, line 188: if its a time we should use a time type for it?

wesm: [Claude responding on Wes's behalf]

Considered that, but int64 UnixNano is the established convention for mtimes across the provider boundary — `SourceFingerprint.MTimeNS` in the same file, `dbBackedSessionMeta.FileMtime`, and `discoveredFileMtime` on the consumer side are all int64 nanos. A `time.Time` here would be the only one in the pipeline and would need conversion at both the producer (assigned from `meta.FileMtime`) and the consumer (compared against other int64 mtimes in the `--limit` sort). The `NS` suffix and the "zero means unset" sentinel follow the existing `MTimeNS` pattern. Happy to do a pipeline-wide `time.Time` conversion as a separate change if we want that convention instead.

PR #951, opened 2026-07-01
fix(test): keep a slow template checkpoint from poisoning the Windows suite

mjacobs:
The shared test-DB templates from #943 checkpoint their WAL under a 1-second context deadline inside a `sync.Once` that caches the error. On a cold windows-latest runner, one checkpoint that misses the deadline permanently poisons the template, and every subsequent test in that binary fails with the same cached `checkpointing db template: wal checkpoint truncate: context deadline exceeded`. That is what happened on #948's CI run — a frontend-only diff — where the `cmd/agentsview` and `internal/db` binaries each hit the deadline once and 599 tests failed in cascade, while the identical SHA passed on main's push run.

Two changes, by blast radius:

- The two base templates (`internal/dbtest/dbtest.go`, `internal/db/db_test.go`) get a best-effort checkpoint with a generous once-per-binary deadline: the copy step already carries the `-wal`/`-shm` files along, so an uncheckpointed template is still a consistent database, and losing the checkpoint only costs compactness. Independently, a failed template build now degrades to creating the per-test database from scratch (the pre-#943 path) instead of failing the test — the template is a setup-cost optimization and should never be a single point of failure for the whole binary.
- The five seeded fixture builders (store contract, chunked analytics, daily usage, server analytics, cmd stats golden) keep their fatal checkpoint semantics — their consumers' copy semantics vary — but move from the 1s deadline to the same 30s one, which is effectively free since each runs once.

Two regression tests cover the new fallback, including cleanup of a partial template copy left behind by a mid-copy failure and the existing-file short-circuit.

Not addressed here: the unrelated `TestEnsureBackgroundServeLaunchLoser…` "server did not become ready within 2s" readiness flake that hit #949's run.

Where to look: `buildTestDBTemplate` in `internal/dbtest/dbtest.go` and `copyTestDBTemplate`/`openTestDBWithTemplate` in `internal/db/db_test.go`.

### DIFF
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -351,8 +351,23 @@ func TestMain(m *testing.M) {
 }
 
 func openCopiedTestDB(path string) (*DB, error) {
-	if err := copyTestDBTemplate(path); err != nil {
-		return nil, err
+	return openTestDBWithTemplate(path, copyTestDBTemplate)
+}
+
+func openTestDBWithTemplate(
+	path string, copyTemplate func(string) error,
+) (*DB, error) {
+	if err := copyTemplate(path); err != nil {
+		// The shared template is only a setup-cost optimization.
+		// Never let a template failure poison every test in the
+		// binary; build this database from scratch instead.
+		fmt.Fprintf(os.Stderr,
+			"db test: template unavailable, creating %s from scratch: %v\n",
+			path, err)
+		for _, suffix := range []string{"", "-wal", "-shm"} {
+			_ = os.Remove(path + suffix)
+		}
+		return Open(path)
 	}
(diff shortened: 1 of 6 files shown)


mjacobs:
Validated on a physical Windows 11 box (Go 1.26.4, mingw-w64 gcc 14.2 UCRT): with this branch, `internal/dbtest`, `internal/db`, `internal/db/git`, `internal/server`, and `cmd/agentsview` all pass under `CGO_ENABLED=1 -tags fts5`. The two new fallback tests exercise the poisoned-template path against real Windows file semantics.

wesm:
Thank you!

wesm:
rebasing

PR #952, opened 2026-07-01
fix(cli): avoid legacy Copilot billing wording

mjacobs:
After rebasing on the upstream dashboard fix in #947, the remaining change here is the CLI wording from #934. That note still described GitHub Copilot as billing via a monthly request quota, but Copilot billing has moved on and that claim is now too specific for what agentsview can infer from local records.

The CLI still warns when an all-Copilot `usage daily --agent ...` filter returns zero token and cost totals, but the message now describes only the data limitation agentsview observes: these Copilot records do not include token or cost data that can be totaled. That keeps the user-facing explanation accurate without trying to summarize GitHub's current billing model.

Scope is intentionally small. The dashboard behavior is already handled by #947 through the `unsupportedUsage` summary signal, so this PR no longer changes frontend copy, storage interfaces, or cross-backend session counting. Reviewers should look at `cmd/agentsview/usage.go` and the matching assertions in `cmd/agentsview/usage_test.go`.

Follow-up to #934 and #947.

### DIFF
--- a/cmd/agentsview/usage.go
+++ b/cmd/agentsview/usage.go
@@ -192,18 +192,18 @@ func runUsageDaily(cfg UsageDailyConfig) {
 }
 
 // noTokenDataNote returns a one-line stderr note for a zero usage result when
-// the user has filtered to a Copilot agent, which bills via a monthly request
-// quota rather than per-token and records no per-message token usage. It
-// returns "" when the filter is not a Copilot agent or real token/cost data
-// exists. This is an agent-property statement (issue #349) shown in response to
-// an explicit --agent the user typed, so it needs no session-presence check;
-// it is appropriate even for an empty window.
+// the user has filtered to a Copilot agent whose records do not include token
+// or cost data that agentsview can total. It returns "" when the filter is not
+// a Copilot agent or real token/cost data exists. This is an agent-property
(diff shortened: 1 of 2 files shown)


mjacobs:
Fixed in `19009003`. Both bound clauses now compare via `julianday()` on both sides (the repo's existing sub-second precedent in `internal/db/activity.go`), so RFC3339Nano session timestamps land on the correct side of whole-second bounds in both directions. Added `TestCountSessionsForUsage_FractionalSecondBoundary` covering both edges (lo+0.5s counted, hi+0.5s excluded; one non-UTC case) — both failed before the fix. PostgreSQL/DuckDB are untouched (real timestamp columns, already numeric), so this converges SQLite onto their behavior.

frontend/messages/en.json
```diff
@@ -729,6 +729,7 @@
   "usage_uncached_input": "Uncached Input",
   "usage_output": "Output",
   "usage_no_token_data": "No token data",
+  "usage_no_token_data_hint_copilot": "GitHub Copilot bills via a monthly request quota and doesn't record per-message token usage.",
```
mariusvniekerk: So copilot billing no longer matches this legacy behavior. So we probably need something a bit different?

wesm:
looking

PR #953, opened 2026-07-01
feat(antigravity): flag sessions decoded from unrecognized schemas

mjacobs:
Antigravity is decoded by reverse-engineered heuristics, and #862 started recording which schema produced each session: `sessions.source_version` carries either a known release-range label ("1.0.7-1.0.10") or an `agy-schema:<12hex>` marker for fingerprints the parser has never seen. An unrecognized marker means a newer Antigravity build changed the schema and the heuristics may be silently wrong — but nothing surfaced it. This PR turns that stored signal into a visible one.

A single classifier, `parser.DecodeConfidence(agent, sourceVersion)`, derives a confidence for the two Antigravity agents: `low` for the unknown-schema marker, `high` for a recognized label, nothing otherwise. It is deliberately agent-gated because `source_version` is a generic field — Piebald and CommandCode write their own values there — so non-empty alone proves nothing. This is a separate axis from `transcript_fidelity` (#914): fidelity measures sidecar coverage, confidence measures schema recognition; a session can be full-fidelity yet low-confidence and vice versa.

The value is computed at read time in `SessionDetail.MarshalJSON` rather than persisted. Confidence is a pure function of an already-stored column, so a persisted copy would refresh on exactly the same re-parse events while costing a dataVersion bump, a three-backend migration, and duplicate parse-diff drift whenever the schema map learns a new fingerprint. Derive-on-read gets identical freshness for none of that, and works unchanged across SQLite, PostgreSQL, and DuckDB read paths.

Three surfaces consume it: a warning badge in the session-detail breadcrumb (only for `low`; localized en/zh-CN/zh-TW), an `agentsview doctor` count of sessions on unrecognized schemas (folded into the existing Antigravity session scan, one extra aggregate rather than a second query), and a per-run `UnknownSchemaSessionsByAgent` counter in the sync anomaly summary so a new Antigravity release becomes visible on the first sync after it appears.

One test-visible consequence: the sync integration fixtures build minimal synthetic Antigravity databases, which are legitimately unrecognized schemas, so their expected `SyncStats` now include the anomaly count.

Where to look: `DecodeConfidence` in `internal/parser/antigravity_version.go`, the `MarshalJSON` hook in `internal/service/service.go`, and the recording seam in `internal/sync/engine.go`.

### DIFF
--- a/cmd/agentsview/doctor.go
+++ b/cmd/agentsview/doctor.go
@@ -32,22 +32,27 @@ type doctorAgentRoot struct {
 	Err            error
 }
 
+type doctorDBInspection struct {
+	DBExists                 bool
+	DBReadable               bool
+	DBError                  error
+	UserVersion              *int
+	SessionCounts            []doctorDataVersionCount
+	SessionCountsErr         error
+	AntigravityCLITotal      int
+	AntigravityCLISummary    int
+	AntigravityUnknownSchema int
+	AntigravityCountsErr     error
+}
+
 type doctorSyncReport struct {
-	Config                config.Config
-	DBExists              bool
-	DBReadable            bool
-	DBError               error
-	UserVersion           *int
-	SessionCounts         []doctorDataVersionCount
-	SessionCountsErr      error
-	AntigravityCLITotal   int
-	AntigravityCLISummary int
-	SummaryModeErr        error
-	TempFiles             []string
-	AgentRoots            []doctorAgentRoot
-	DebugLines            []string
-	DebugLogErr           error
-	HasResyncFailureLog   bool
+	Config config.Config
+	doctorDBInspection
+	TempFiles           []string
+	AgentRoots          []doctorAgentRoot
+	DebugLines          []string
+	DebugLogErr         error
+	HasResyncFailureLog bool
 }
 
 func newDoctorCommand() *cobra.Command {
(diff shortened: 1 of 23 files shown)


mjacobs:
Fixed in `375f9cb3`. `DecodeConfidence` is now a reflected `SessionDetail` field populated once at construction, round-tripped through Unmarshal for the HTTP-backed service, and picked up by Huma reflection — the generated TS model gains `decode_confidence?: string` on `ServiceSessionDetail` (and deliberately not on `DbSession`). Added `TestOpenAPIEndpointDocumentsDecodeConfidence` asserting the spec documents the field on the detail schema only. One note: regenerating the client surfaced pre-existing drift in 11 unrelated generated files (identical at this branch's base and upstream/main); that churn is left out of this PR and is worth a separate chore pass.

cmd/agentsview/doctor.go
```diff
@@ -117,29 +119,30 @@ func inspectDoctorDB(
...
 	antigravityCLISummary int,
-	summaryModeErr error,
+	antigravityUnknownSchema int,
+	antigravityCountsErr error,
 ) {
 	info, err := os.Stat(path)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return false, false, nil, nil, nil, nil, 0, 0, nil
+			return false, false, nil, nil, nil, nil, 0, 0, 0, nil
```
mariusvniekerk: this really looks like this function should return a struct

frontend/messages/en.json
```diff
@@ -172,6 +172,8 @@
   "session_breadcrumb_failed_to_open": "Failed to open",
   "session_breadcrumb_summary_mode": "Summary mode",
   "session_breadcrumb_summary_mode_tooltip": "Tool results, thinking, and diffs are missing from this transcript. Full fidelity needs the agy-reader sidecar — install it, or if installed, let it catch up. Click to learn how.",
+  "session_breadcrumb_decode_confidence_low": "Unverified schema",
```
mariusvniekerk: antigravity specific translation string should contain the antigravity key in it?  worse for the next one.

wesm:
looking

cmd/agentsview/doctor.go
mjacobs: Addressed on the current head `1ee181f9` (folded in via the review-feedback pass during the branch rewrite): `inspectDoctorDB` now returns a `doctorDBInspection` struct embedded in `doctorSyncReport`, and each failure path just sets `DBError` and returns. The inline diff above shows the old 10-value signature because the force-push orphaned the thread anchor.

PR #956, opened 2026-07-02
[codex] Prevent OpenCode WAL watcher feedback loop

### LINKED ISSUE #955: OpenCode SQLite reads trigger a WAL watcher feedback loop
## agentsview version

`agentsview v0.34.5-103-g9d4e1885 (commit 9d4e1885, built 2026-07-02T01:21:31Z)`

## Install method

Built from source

## OS / platform

Ubuntu 24.04.4 LTS, Linux 6.17.0-1025-oem, x86_64

## Which agent and version

OpenCode 1.15.13

## Which model(s)

n/a

Reading an idle OpenCode SQLite database can produce a self-sustaining watcher loop.

With OpenCode stopped and the main `opencode.db` unchanged, an AgentsView sync opens the WAL-mode database and transient `opencode.db-wal` / `opencode.db-shm` files appear. The OpenCode provider watches `opencode.db-*` and currently classifies every matching event as a change to every SQLite-backed OpenCode session.

Closing the read connection removes the transient sidecars, producing more watcher events. Those events enumerate and sync every OpenCode SQLite session again, recreate the sidecars, and continue the loop.

On a database containing 131 sessions, the debug log repeated roughly once per second:

```text
watcher: 3 file(s) changed, triggering sync
sync: 131 file(s) updated
watcher: 2 file(s) changed, triggering sync
sync: 131 file(s) updated
```

It eventually produced:

```text
watcher error: fsnotify: queue or buffer overflow
```

During the loop:

- no OpenCode process was running;
- `opencode.db` remained unchanged;
- `opencode.db-wal` was transient and zero bytes;
- `opencode.db-shm` was transient and 32 KiB.

Expected behavior: reading an idle OpenCode database must not recursively trigger new syncs. Real changes written by OpenCode to the main DB or WAL must still trigger a bounded update.

`openOpenCodeDB` currently opens the source using:

```text
?mode=ro&_journal_mode=WAL&_busy_timeout=3000
```

The SQLite driver executes `PRAGMA journal_mode=WAL` for `_journal_mode=WAL`. A read-only connection to an already-WAL-mode database may still require transient WAL-index sidecars, so removing that option alone is insufficient. Changed-path classification must also avoid treating `-shm` and empty/transient `-wal` lifecycle events as source-data changes.

## Steps to reproduce

1. Configure a SQLite-backed OpenCode root containing multiple sessions.
2. Stop OpenCode and confirm the main `opencode.db` is idle.
3. Start AgentsView.
4. Trigger a sync with `POST /api/v1/sync`.
5. Tail `~/.agentsview/debug.log` and observe repeated watcher-triggered all-session updates.
6. Observe transient `opencode.db-wal` and `opencode.db-shm` creation/removal.

## Acceptance criteria

- An AgentsView read of an idle OpenCode-family SQLite DB cannot sustain a watcher loop.
- `*-shm` and empty/transient `*-wal` events do not fan out to every session.
- Real main-DB or data-bearing WAL writes still trigger synchronization.
- OpenCode, Kilo, MiMoCode, and IcodeMate retain SQLite discovery and parsing behavior.
- Regression coverage uses WAL-mode sidecar events and proves the callback/sync sequence is bounded.

## Checklist

- [x] I searched existing issues.
- [x] No private session data is attached.

tpn:
Closes #955.

OpenCode-family SQLite readers now use a real read-only file URI and no longer request a journal-mode change while inspecting source databases. Changed-path classification ignores WAL-index (`-shm`) events and empty, header-only, removed, or otherwise non-data WAL lifecycle events, while retaining main-database and data-bearing WAL updates.

This prevents AgentsView's own read connections from recursively fanning one sidecar event out to every SQLite-backed OpenCode session. Checkpointed WAL removal relies on the corresponding main-database write, and live uncheckpointed WAL frames remain visible to the parser.

The provider watch plan now declares only the main database and WAL data file. Regression coverage uses a real writer-held WAL to verify that committed frames are classified and readable, while transient sidecars remain bounded no-ops.

### DIFF
--- a/internal/parser/antigravity.go
+++ b/internal/parser/antigravity.go
@@ -81,7 +81,7 @@ func (p *antigravityProvider) parseSession(
 
 	// Open read-only; SQLite session files have WAL/SHM
 	// sidecars that the driver expects in the same dir.
-	dsn := "file:" + path + "?mode=ro&immutable=0"
+	dsn := "file:" + sqliteURIPath(path) + "?mode=ro&immutable=0"
 	db, err := sql.Open("sqlite3", dsn)
 	if err != nil {
 		return nil, nil, nil, fmt.Errorf(
--- a/internal/parser/antigravity_cli.go
+++ b/internal/parser/antigravity_cli.go
@@ -310,7 +310,7 @@ func (p *antigravityCLIProvider) parseSessionWithStatus(
 func loadAntigravityCLIDBSteps(
 	path string,
 ) (antigravityStepLoadResult, error) {
-	dsn := "file:" + path + "?mode=ro&immutable=0"
+	dsn := "file:" + sqliteURIPath(path) + "?mode=ro&immutable=0"
 	db, err := sql.Open("sqlite3", dsn)
 	if err != nil {
 		return antigravityStepLoadResult{}, fmt.Errorf(
--- a/internal/parser/hermes.go
+++ b/internal/parser/hermes.go
@@ -553,7 +553,7 @@ func hermesStatePaths(root string) (stateDB, sessionsDir string, ok bool) {
 func (p *hermesProvider) parseStateDB(
 	stateDB, sessionsDir, project, machine string,
 ) ([]ParseResult, error) {
-	conn, err := sql.Open("sqlite3", "file:"+stateDB+"?mode=ro")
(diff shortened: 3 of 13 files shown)


wesm:
looking

PR #961, opened 2026-07-03
feat(parser): add Devin CLI session support

aaronflorey:
## Add Devin CLI as a supported local session source.

This teaches AgentsView to discover Devin roots from the local-share directory, parse session metadata from cli/sessions.db, read transcript content from cli/transcripts, and surface Devin sessions in the UI and settings docs. Devin is integrated through the provider facade because its session identity, freshness, and watch behavior come from a composite DB-plus-transcript source rather than a single transcript file.

The parser now handles Devin transcript messages, tool calls/results, token usage, privacy-safe error reporting, and fallback parsing from message_nodes when transcript JSON is missing. Sync and watcher integration were updated so Devin sessions use provider-owned virtual source paths, composite freshness, and provider watch plans without widening remote sync or other non-Devin behavior.

Reviewers should focus on:
- internal/parser/devin.go
- internal/parser/devin_provider.go
- internal/sync/engine.go
- cmd/agentsview/main.go

We intentionally narrow earlier provider-wide fallout so SSH remote sync, parse-diff, token-use probing, and resync safety counting stay scoped to existing behavior, with Devin as the only new non-file-backed exception.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,13 @@ On first run, agentsview discovers sessions from every supported agent on your
 machine, syncs them into a local SQLite database, and serves a web UI at
 `http://127.0.0.1:8080`.
 
+For Devin CLI, point `DEVIN_DIR` or `devin_dirs` at the local root that contains
+`cli/` — for example `~/Library/Application Support/devin` on macOS,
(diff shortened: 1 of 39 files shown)


wesm:
looking

PR #970, opened 2026-07-03
feat(parse-diff): classify incremental-append skew and recommend a resync baseline

mjacobs:
parse-diff v2 follow-up (after #805 sub-items 1+4 and #949 sub-item 3), addressing incremental-append skew.

`parse-diff` compares stored session rows against a fresh full re-parse. Sessions last written through the incremental JSONL append path (Claude/Codex) are not rewritten through full normalization, so a fresh full re-parse can legitimately differ on the per-message metadata / ordinal-set shape even at the current data_version and without a pending resync. Those sessions were reported as changed — noise that makes `--fail-on-change` untrustworthy in CI against a non-resynced archive.

This adds a SQLite-only `sessions.last_write_incremental` marker, set inside the single incremental-update seam. It is reset to false only by a genuine full message re-normalization (`ReplaceSessionContent`, `ReplaceSessionMessages`, and the batch `ReplaceMessages` branch) via a shared `resetIncrementalMarkerTx`, and seeded false on fresh INSERT — not on a bare `UpsertSession`. That distinction matters: the routine full-parse batch path takes `ReplaceMessages=false` for the append-only agents, upserting the session row while appending only new messages and leaving earlier incrementally written rows in place, so clearing the marker on every upsert would have made a single routine sync report still-present benign skew as real drift. The marker is therefore per-session ground truth (not inferred) and self-heals on the next full resync.

A new `DiffIncrementalSkew` class — mirroring `DiffRaced` — reclassifies a would-be change on such a row: it is listed for drill-down but excluded from failure counting, while still counting toward Examined. Precedence is raced > incremental-skew > changed. The CLI additionally prints a resync-baseline recommendation whenever skew is present.

The reclassification is confined to the incremental-artifact surface, not applied session-wide. Because the marker is session-level while a regression is field-level, the marker only reclassifies when every non-informational diff is on an allow-list holding just the ordinal-shape / per-message metadata field (`message_metadata`, via `diffsConfinedToIncrementalArtifacts`). A non-informational diff on any other field — head-derived `first_message`/`started_at`, message content, usage totals, tool_calls, or the recomputed aggregates, all of which both write paths normalize identically — keeps the session `DiffChanged` and still trips `--fail-on-change`. The session fields the incremental path actually freezes (`termination_status`, `session_name`, `cwd`, …) are already marked informational and never reach this test. Characterized against the real archive: `parse-diff` over the full Claude/Codex corpus shows 0 changed and 0 skew for quiescent current-version rows; the only per-message metadata drift observed was on a live-write raced session, which already outranks skew. The class is thus a conservative safety net for the one surface a whole-file re-parse can legitimately reshape, without masking regressions elsewhere.

Backend note: the marker is SQLite-only by design, like `next_ordinal`/`last_entry_uuid`/`file_hash`. parse-diff opens the local SQLite archive directly and is never reachable through the PostgreSQL/DuckDB read paths; the column is `json:"-"` and the push paths use explicit column lists that already omit the sync-bookkeeping cluster. The migration is a non-destructive ADD COLUMN with no dataVersion bump.

Where to look: `internal/db/sessions.go` / `messages.go` / `session_batch.go` (set, scan, and the full-rewrite-only reset), `internal/sync/parsediff_compare.go` and `parsediff.go` (classify, compute, and artifact confinement), `internal/sync/parsediff_report.go` (the class), `cmd/agentsview/parse_diff.go` (render + resync note).

### DIFF
--- a/cmd/agentsview/parse_diff.go
+++ b/cmd/agentsview/parse_diff.go
@@ -62,9 +62,11 @@ func newParseDiffCommand() *cobra.Command {
 			"Use it to vet parser changes against the real archive before\n" +
 			"bumping the data version, or to detect upstream format drift.\n" +
 			"Run it against a quiescent archive: sessions still being\n" +
-			"written, and sessions last written through the incremental\n" +
-			"append path, can show benign drift against a full re-parse. A\n" +
-			"freshly resynced archive gives the cleanest baseline.",
+			"written can show benign drift against a full re-parse and are\n" +
+			"classified as raced. Sessions last written through the\n" +
+			"incremental-append path are detected and classified as\n" +
+			"incremental_skew (both excluded from --fail-on-change); a\n" +
+			"freshly resynced archive still gives the cleanest baseline.",
 		GroupID:      groupData,
 		SilenceUsage: true,
 		Args:         cobra.NoArgs,
@@ -315,8 +317,26 @@ func renderParseDiffReport(
 	renderParseDiffParseErrors(w, r.Sessions)
 	renderParseDiffChanged(w, r.Sessions, verbose)
 	renderParseDiffRaced(w, r.Sessions, verbose)
+	renderParseDiffIncrementalSkew(w, r.Sessions, verbose)
 	if verbose {
 		renderParseDiffPendingResync(w, r.Sessions)
+	}
+
+	// Incremental-skew sessions are suppressed from --fail-on-change but
+	// still signal that the comparison basis is compromised for those
+	// rows. A full resync rewrites them through normalization, clears the
+	// marker, and restores full diff scrutiny -- so recommend it, mirroring
+	// the vacuous-resync warning above. Totals.IncrementalSkew is already
+	// in the JSON, so no new bool method is needed to gate the note.
+	if r.Totals.IncrementalSkew > 0 {
+		fmt.Fprintf(w,
+			"Note: %d session(s) were last written through the "+
+				"incremental-append path, so a full re-parse legitimately "+
+				"differs from the stored rows. They are classified "+
+				"incremental_skew and excluded from --fail-on-change. Run a "+
+				"full resync for a clean parse-diff baseline that restores "+
+				"drift detection on these sessions.\n\n",
+			r.Totals.IncrementalSkew)
 	}
 
 	fmt.Fprintf(w, "%d sessions changed, %d identical.\n",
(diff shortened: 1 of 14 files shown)


mjacobs:
Both findings addressed in be92ec51 and fbc60831.

**`sessions.go` — marker cleared before rows were re-normalized:** Confirmed. The marker was cleared on every `UpsertSession`, but a bare upsert rewrites only the session row. The routine full-parse batch path takes `ReplaceMessages=false` for the append-only agents (Claude/Codex): it upserts the session and appends new messages while leaving earlier incrementally written rows in place, so clearing the marker there surfaced still-present benign skew as real drift on the next sync. The marker is now reset only by a genuine full message re-normalization (`ReplaceSessionContent`, `ReplaceSessionMessages`, the batch `ReplaceMessages` branch) and seeded false on fresh INSERT — never on a bare upsert or an append-only write.

**`parsediff.go` — reclassification masked non-artifact fields:** Confirmed. The reclassification is now confined to the incremental-artifact surface (`diffsConfinedToIncrementalArtifacts`): the marker only downgrades a change to skew when every non-informational diff is on the allow-list, which holds just `message_metadata` (the ordinal-shape field). A non-informational diff on `first_message`/`started_at` (head-derived, byte-stable), message content, usage totals, tool_calls, or the recomputed aggregates keeps the session `DiffChanged` and still trips `--fail-on-change`. Characterized against the real archive before narrowing: `parse-diff` over the full Claude/Codex corpus shows 0 changed and 0 skew for quiescent current-version rows (the only per-message metadata drift observed was on a live-write raced session, which already outranks skew), so the class is a conservative safety net rather than a routinely hit path.

wesm:
looking

wesm:
Re: the incremental-skew over-suppression finding (`internal/sync/parsediff.go:760`)

The description is accurate: the marker is session-level and the confinement is field-level, so a `message_metadata` diff on a row the incremental path never wrote — or on subfields like `thinking_text`/`has_tool_use` — is reclassified as `incremental_skew` on a marked session. Declining the row-level tracking, because it would not change what `--fail-on-change` catches:

- A full-parser regression on the metadata surface hits the corpus, not just marked sessions. Markers exist only on sessions live-appended since their last full rewrite, so the archive bulk is unmarked, stays `DiffChanged`, and still fails the run. Row-level scrutiny of pre-append rows inside marked sessions is redundant with that coverage.
- An incremental-writer regression lands exactly on the rows that per-row provenance or an ordinal watermark would allow, so it is suppressed either way. Its signal is the skew count itself: the empirical baseline over the real Claude/Codex corpus is zero (see fbc60831), so any nonzero `incremental_skew` total is the indicator, and those sessions render with full field drill-down plus the resync-baseline note.
- Narrowing the allow-list to specific metadata subfields would be guesswork against zero observed skew. The tail parse lacks earlier-file context, so any of the metadata subfields can legitimately reshape on appended rows; excluding some risks reintroducing the false `--fail-on-change` failures this class exists to remove.

The tradeoff is now documented at the classification site in `parsediff.go` (commit on its way with the next push).

PR #971, opened 2026-07-03
fix(postgres): scope alias backfill marker per target

### LINKED ISSUE #939: 0.35.0: session-alias backfill re-forces a full push per project-filter set and won't complete on a shared hub (follow-up to #891)
Follow-up to #891. That fix (per-filter-scoped push watermarks, shipped in 0.35.0) made `pg push --projects` / filtered pushes incremental. On 0.35.0 that win is repeatedly undone by the new session-alias backfill (#898): filtered / shared-hub pushes keep getting forced back into a full ~30-minute sweep, and `pg status` shows `Last push: never` for the affected scope.

## Symptoms

- Every `agentsview pg push` (manual filtered push, and the background `pg service`) does a full ~40–54 min sweep of ~4,400 sessions even when only a handful changed.
- The daemon log repeatedly prints:
  ```
  pgsync: session alias backfill marker missing; forcing full push
  ```

### 1. The one-time backfill marker is stored per project-filter scope

`applySessionAliasBackfillRequirement` forces `full` until `pg_session_alias_backfill_v1` is set, and that key is read/written through the **scoped** sync-state store (`Sync.effectiveSyncState()`), so it is suffixed with the project-filter scope (`...:project-filter:<hash>`). Consequences:

- Each distinct `--projects` / `exclude_projects` set gets its own backfill marker and must run its own full sweep.
- **Changing the filter set at all mints a brand-new scope with no marker, re-triggering a full sweep** — which is exactly the workflow #891 is about (tuning an allow/deny list under a default-deny model).

Evidence from a live 0.35.0 hub after editing `exclude_projects`: local `pg_sync_state` holds `pg_session_alias_backfill_v1` markers for two *older* filter scopes (both `= 1`), but the *current* config's computed scope has no `pg_session_alias_backfill_v1` / `last_push_at` row at all — so every scheduled push since the edit re-forces full.

### 2. The marker only persists on a perfectly clean push

`completeSessionAliasBackfill` records the marker only when `result.Errors == 0 && result.SkippedConflicts == 0`. `SkippedConflicts` counts sessions owned by **another** machine (`errSessionOwnershipConflict`) — a permanent, expected condition on a shared hub. So on any hub where this host doesn't own 100% of the matching sessions, the one-time backfill can never be recorded, and every push re-forces full. (Stopping the local `agentsview serve` daemon to get one contention-free run is the only current workaround.)

## Repro

0.35.0, PG hub, ≥2 machines (or a concurrent `agentsview serve` causing contention):

1. `agentsview pg push --projects <allow-list>` → completes.
2. Change the allow/deny list, or run again while another pusher is active.
3. Each subsequent push logs `session alias backfill marker missing; forcing full push` and does a full sweep; `pg status` shows `Last push: never`.

## Proposed fixes

- **(cause 2)** Don't treat skipped ownership conflicts as an incomplete backfill — gate the marker on `Errors` only. Skipped conflicts are other machines' sessions this host can never push, so they don't indicate an incomplete backfill of *this host's own* sessions. Small and unambiguous; PR attached.
- **(cause 1)** Consider keying the one-time backfill marker by PG target rather than by project-filter — it is a per-DB → per-target migration, not a per-filter concern. This needs a maintainer call on the transition-window case (an earlier, broader-filter push can leave already-pushed sessions that a later, narrower backfill filter would not re-touch), so it is raised here rather than in the PR.

Environment: agentsview 0.35.0, PostgreSQL hub over Tailscale.

rodboev:
Filtered PostgreSQL pushes can still fall back to repeated full sweeps after the project filter set changes, because the one-time session-alias backfill marker is stored in the same project-filtered sync-state scope as the incremental push watermarks. The filtered scope is correct for `last_push_at`, target fingerprints, and boundary fingerprints, but it is the wrong scope for a migration marker that only needs to complete once per PG target.

This moves the alias-backfill marker onto target-scoped sync state while leaving the ordinary filtered push state in the existing project-filtered store. Changing `--projects` or `--exclude-projects` will no longer re-arm the alias-backfill full-push gate after the same target has already completed it, but different filter sets will still keep their own incremental watermarks and reset behavior exactly as before.

The change stays entirely inside `internal/postgres`, and it keeps the current one-time marker model rather than widening the fix into per-session tracking. One bounded limitation remains: the first filtered push that sets the target-scoped marker only backfills sessions inside that filter. Sessions outside that filtered run keep their old alias state until a later push includes them or they change locally, so this narrows the repeated full-push bug without claiming full per-session coverage.

This follows the narrowed analysis in [#939](https://github.com/kenn-io/agentsview/issues/939) from leejuhanKr and the split left by merged sibling [#940](https://github.com/kenn-io/agentsview/pull/940).

Closes #939

### DIFF
--- a/internal/postgres/push.go
+++ b/internal/postgres/push.go
@@ -70,6 +70,7 @@ func (s *Sync) Push(
 	start := time.Now()
 	var result PushResult
 	state := s.effectiveSyncState()
+	aliasBackfillState := s.aliasBackfillSyncStateOrDefault()
 
 	if err := CheckDataVersionCompat(ctx, s.pg); err != nil {
 		return result, err
@@ -132,9 +133,12 @@ func (s *Sync) Push(
 	legacyMarkerMachines := pushMarkerLegacyMachines(
 		markerMachine, markerMachineAliases,
 	)
+	// Keep the backfill marker scoped to target only; all other push
+	// state remains scoped by full effective sync state (including filter
+	// fingerprint when present).
 	aliasBackfillNeeded := false
 	full, aliasBackfillNeeded, err = applySessionAliasBackfillRequirement(
-		state, full,
+		aliasBackfillState, full,
 	)
 	if err != nil {
 		return result, err
@@ -339,7 +343,7 @@ func (s *Sync) Push(
 			return result, err
 		}
 		if err := completeSessionAliasBackfill(
-			state, aliasBackfillNeeded, result,
+			aliasBackfillState, aliasBackfillNeeded, result,
 		); err != nil {
 			return result, err
 		}
@@ -443,7 +447,7 @@ func (s *Sync) Push(
 		return result, err
 	}
 	if err := completeSessionAliasBackfill(
-		state, aliasBackfillNeeded, result,
+		aliasBackfillState, aliasBackfillNeeded, result,
 	); err != nil {
 		return result, err
 	}
--- a/internal/postgres/push_test.go
+++ b/internal/postgres/push_test.go
@@ -194,6 +194,115 @@ func TestCompleteSessionAliasBackfillMarksDoneUnlessErrors(t *testing.T) {
 	}
 }
 
+func TestSessionAliasBackfillRequirementUsesSyncAliasStateAcrossFilters(t *testing.T) {
+	filterScopeA := pushSyncStateScope(
+		"work",
+		[]string{"alpha"},
+		nil,
+	)
+	filterScopeB := pushSyncStateScope(
+		"work",
+		[]string{"beta"},
+		nil,
+	)
+
+	store := &syncStateStoreStub{}
+	syncA := &Sync{
+		syncState:          newScopedSyncStateStore(store, filterScopeA, false),
+		aliasBackfillState: newScopedSyncStateStore(store, "work", false),
+	}
+	syncB := &Sync{
+		syncState:          newScopedSyncStateStore(store, filterScopeB, false),
+		aliasBackfillState: newScopedSyncStateStore(store, "work", false),
+	}
+
+	full, needed, err := applySessionAliasBackfillRequirement(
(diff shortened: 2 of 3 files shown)


wesm:
looking at this

wesm:
Re the roborev finding on the target-scoped alias backfill marker: the mechanics are correct, but this is the transition-window case that was explicitly flagged and accepted when the per-target marker was proposed in #939, not an oversight. I'm accepting the limitation and documenting it in the code rather than changing the scoping.

Why not the suggested fix: both variants (writing the target marker only from unfiltered pushes, or keeping filter-scoped markers for filtered pushes) re-force a full push per filter scope for anyone who only runs filtered pushes. `full=true` bypasses the message-comparison shortcut in `pushBatch` and rewrites message content, so that's the ~42 min sweep vs sub-second incremental measured in #939 — reintroducing the regression this PR fixes, on exactly the varying-filter workflow it targets.

The gap is also narrower than the finding implies:

- `replacePGSessionAliases` runs for every session included in a batch, even when message writes are skipped. A new filter scope has no stored fingerprints, so it re-pushes everything in scope and refreshes aliases — new scopes self-heal.
- The affected population is sessions in scopes whose fingerprints were stored before the alias feature (0.35.0, #898), that never change again, and that were outside the filter active when the marker got set. Aliases only exist for vibe-agent sessions in `session_*` directories, so the consequence is missing alias rows for that slice (alias-based trash exclusion, tombstone following, and lookups), shrinking to zero as sessions change.

A code comment documenting this tradeoff at the marker scoping site in `internal/postgres/sync.go` will land on this branch. If closing the gap ever matters in practice, the right shape is a target-wide alias-only sweep (upsert alias rows for local vibe sessions already present in PG when the marker is missing, independent of the project filter) rather than re-scoping the marker — that can be a follow-up issue.

PR #972, opened 2026-07-03
fix(usage): wire agent exclusions through usage filters

### LINKED ISSUE #976: Usage page agent exclusions are a silent no-op
Clicking an Agent tile in the Usage Attribution panel appears to work: the tile dims, the exclusion lands in localStorage, and it survives reloads. But outgoing usage queries never change — `baseParams()` never copies `excludedAgents` into the request, `hasActiveFilters` ignores it, and the Usage toolbar has no Agent dropdown that would make the stale state visible.

Reproduce: open /usage, switch Attribution grouping to Agent, click any agent row, and watch the network tab. No `exclude_agent` parameter is sent and no numbers change.

The `exclude_agent` parameter already exists through the HTTP route, service layer, all three storage backends, and generated client. The frontend just never sends it.

prateek:
Fixes the pre-existing silent no-op where clicking an agent in Usage attribution updated local exclusion state, but the next usage requests did not send `excludeAgent`. The store now includes agent exclusions in the shared request params and counts them in `hasActiveFilters`, so the API sees the same filter state the UI shows.

The Usage toolbar also gets the Agent dropdown next to Project and Model. It uses the existing dropdown pattern, which makes agent exclusions visible and reversible without relying on attribution-row clicks. This stays scoped to Usage agent filtering and does not depend on the branch-dimension work.

Agent dropdown on the usage toolbar (synthetic data)

Reviewers should start with `frontend/src/lib/stores/usage.svelte.ts`; `UsagePage.svelte` is the toolbar wiring, and the added component/store tests cover the request params plus the attribution-click path.

Fixes #976.

### DIFF
--- a/frontend/src/lib/components/filters/SessionActiveFilters.svelte
+++ b/frontend/src/lib/components/filters/SessionActiveFilters.svelte
@@ -16,6 +16,7 @@
     onClearProjects?: () => void;
     onRemoveModel?: (model: string) => void;
     onClearModels?: () => void;
+    onClearAgents?: () => void;
   }
 
   let {
@@ -25,6 +26,7 @@
     onClearProjects,
     onRemoveModel,
     onClearModels,
+    onClearAgents,
   }: Props = $props();
 
   const selectedAgents = $derived(
@@ -87,6 +89,7 @@
     });
     onClearProjects?.();
     onClearModels?.();
+    onClearAgents?.();
   }
 </script>
 
--- a/frontend/src/lib/components/filters/SessionActiveFilters.test.ts
+++ b/frontend/src/lib/components/filters/SessionActiveFilters.test.ts
@@ -1,0 +1,49 @@
+import {
+  afterEach,
+  describe,
+  expect,
+  it,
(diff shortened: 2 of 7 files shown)


prateek:
Addressed the automated review finding: agent exclusions now round-trip through the URL (`exclude_agent` param), matching project exclusions, so a shared or reloaded /usage URL preserves them.

prateek:
Addressed the round-2 finding: UsagePage now calls `sessions.loadAgents()` on mount, so the Agent dropdown is populated on a direct /usage visit instead of depending on the sidebar having loaded agent metadata first. Covered by a mount test.

prateek:
Addressed the round-3 finding: `SessionActiveFilters` gained an `onClearAgents` hook invoked by Clear all, and UsagePage wires it to `usage.selectAllAgents()`, so agent exclusions now reset with the rest. Kept to the clear-all hook to match how excluded projects are handled on this surface (they don't render chips here either); per-exclusion chips would be a separate change across both dimensions.

Issue #928, opened 2026-06-30
Add git branch as an analysis dimension


prateek:
agentsview already stores `git_branch` on sessions, but the product still treats project as the main work dimension. For git worktree users, branch is often the useful unit: what happened on this branch, which branches used the most cost, and how activity compares across project branches.

Proposal: expose git branch as a scoped `(project, branch)` dimension.

Planned scope:

- Add backend/API support for scoped branch filtering.
- Expose branch filters in CLI, MCP, and the web session filter.
- Add branch rollups to activity reports.
- Add branch grouping to usage and cost views.

Review graph:

- PR 1, foundation: https://github.com/kenn-io/agentsview/pull/927
- PR 2, filter surfaces: https://github.com/prateek/agentsview/pull/1
- PR 3, activity rollups: https://github.com/prateek/agentsview/pull/2
- PR 4, usage grouping: https://github.com/prateek/agentsview/pull/3

PRs 2-4 are fork-internal for now because the upstream PR comes from a fork and GitHub/Git Spice do not publish that full cross-repo stack as upstream PRs. They all target PR 1's branch. After PR 1 lands, each follow-up can be restacked onto `main` and opened upstream independently.

wesm:
Notes from reviewing PR 1 (branch filter foundation), for the follow-up PRs. None of these block PR 1.

- **CLI/MCP ergonomics (PR 2):** the `git_branch` wire format is opaque tokens (`project\x1fbranch`, `\x1e`-joined) that come from the `/branches` endpoint verbatim. That works for the web UI, but nobody can type `\x1f` on a command line. The CLI/MCP surface should resolve human input (e.g. `--project X --branch Y`, or `project/branch` strings) into tokens at the edge rather than introducing a second wire format.
- **`GetBranches` is unbounded and alphabetical:** a heavy worktree user could get hundreds of `(project, branch)` pairs. The PR 2 dropdown may want recency ordering or client-side search; either would mean extending the query (non-breaking).
- **Filtering only, no grouping primitive:** PR 1 adds the filter dimension but no GROUP BY helper. Grouping is naturally per-view in PRs 3–4, and `BranchInfo`/the token is ready to serve as the group key.
- **Token count is uncapped** (2 bind params per pair). This matches the exposure of the existing CSV filters, so it's consistent rather than a new risk, but worth remembering if a surface ever passes user-controlled lists of arbitrary length.

prateek:
[...]

PR #973, opened 2026-07-03
chore(frontend): regenerate API client and protect the hand-written analytics test

prateek:
The committed generated client had drifted from the Go API structs:
several merged changes (remote host transport/url, session resolve
endpoints, pairwise comparison shapes, the stats models) landed without
re-running openapi-typescript-codegen. Regenerating also surfaced a
landmine: AnalyticsService.test.ts, a hand-written test from #861, lived
inside the generated directory and was silently deleted by every regen.

Three commits: move that test out to generated-client-analytics.test.ts
(matching the existing generated-client-usage.test.ts naming), fix its
import paths, and regenerate the client. The regen commit is entirely
mechanical output of `npm run generate:api` on current main.

Found while preparing the #928 branch-dimension stack, but independent
of it; landing this first keeps codegen noise out of those PRs.


prateek:
Closing in favor of #977. On reflection this bundles a mechanical catch-up regen with a real workflow question (commit the generated client vs. generate it), and the question should be settled first. The test relocation and catch-up are easy to resubmit to match whichever direction the issue lands on.

PR #974, opened 2026-07-03
feat(filter): expose branch filters in client surfaces

prateek:
Adds branch filtering to the session-browsing surfaces. The CLI accepts `--branch` on `session list`, `session search`, and `activity report`; each requires `--project` because branch names are project-scoped. MCP tools accept the same human inputs: a plain `project` plus a plain `git_branch`. Both surfaces encode the internal `(project, branch)` token at the edge, so callers never need to see or construct token separators.

On the web, the shared sidebar filter gets a Branch section with multi-select, client-side search, selected branches floated to the top, removable chips, a numeric active-filter badge, and URL round-trip. The same selection flows into Analytics and usage filtering.

Sidebar Branch filter: selection floated to top, count badge, and the Analytics dashboard filtered by that branch (synthetic data)

Branch metadata now sorts by most recent session activity in SQLite, PostgreSQL, and DuckDB, with project and branch tiebreakers for deterministic lists. `cmd/testfixture` seeds synthetic branch data so screenshots and e2e fixtures exercise the branch surfaces.

Known limitation: branch filters still travel as an uncapped token list, matching the existing CSV-style filters rather than adding a separate cap in this PR.

Part of #928; parallel to the activity-rollups and usage-grouping PRs.


prateek:
Addressed the automated review findings: `git_branch` was missing from the session-param allowlists in `sessionRouteParams.ts`, `router.svelte.ts`, and `UsagePage.svelte`, so a shared URL with a branch filter would drop it on load — all three now include it, with an allowlist test. Also reworded the `git_branch` row in docs/session-api.md to document the opaque token format (the CLI `--branch` flag encodes it for you).

wesm:
I'll take a closer look at this. My agentsview archive has thousands of branches so this selector has a lot higher cardinality than any of the other selectors in the project at this point

wesm:
Took a closer look. In my agentsview database I have 1066 branches — we don't want to stuff this much data into the DOM for the branch selector control on page render so we will need to do some virtualization work cc @mariusvniekerk

wesm:
superseded by #1017

PR #975, opened 2026-07-03
fix(sync): reparse same-size same-mtime in-place Claude rewrites

mjacobs:
The single-session freshness fast path skips re-parsing a Claude JSONL source when its size, mtime, and file identity (inode/device) all match the stored row. A same-size, same-mtime, same-inode **in-place** rewrite matches all three signals, so the engine keeps the stale rows instead of re-parsing. This happens in the wild when two writes land in one filesystem mtime granule, or on a coarse-mtime filesystem. The inode net (#357) only catches atomic-rename replacements (the inode changes); it is blind to in-place rewrites. The same gap produced a nondeterministic CI flake in `TestIncrementalSync_ClaudeSameSizeFileReplaceUsesFullParse` on filesystems where the inode signal is unavailable and the check degrades to mtime alone.

Fix: a content-hash tie-breaker that runs **only** once size + mtime + identity already match — it compares the on-disk prefix hash (over the stored `file_size`, which `shouldSkipFile` has confirmed equals the current size) against the stored `file_hash`, and forces a full parse on mismatch. The common path stays cheap: a real append changes the size and returns before hashing, so the hash is paid only for sources that already look byte-for-byte unchanged — the same skip-path-hashing the codebase already uses for coarse-mtime sources (S3/Aider/Shelley). Rows without a stored hash fall back to the prior size/mtime/identity behavior. Claude's incremental path now refreshes `file_hash` (previously only Codex did), so the stored hash stays a valid current fingerprint.

Covers local **and** remote (path-rewritten) sync: for remote sources the physical materialized file is hashed (the logical rewritten key is not openable), and the stored hash is computed over those same materialized bytes on both the full-parse and incremental paths, so the comparison is directly comparable — an unchanged re-download still hashes equal and skips, while a genuine rewrite re-parses.

Tradeoff: the content hash is a recurring read on the freshness fast path for otherwise-unchanged Claude sessions, bounded by the sync cutoff window. It is the deliberate cost of closing a correctness gap (silently serving stale content) that stat signals cannot detect.

Where to look: `internal/sync/engine.go` — `providerIncrementalContentChanged` and the `statPath` capture in `providerSingleSessionFresh`, plus the Claude `file_hash` refresh in `tryIncrementalJSONL`/`writeIncremental`. Tests in `internal/sync/engine_integration_test.go` cover the local in-place rewrite, the remote path-rewritten rewrite, and the unchanged-remote-re-download skip.

### DIFF
--- a/internal/sync/engine.go
+++ b/internal/sync/engine.go
@@ -4495,6 +4495,11 @@ func (e *Engine) providerSingleSessionFresh(
 	if e.providerIncrementalIdentityChanged(lookupPath, info) {
 		return 0, false, true
 	}
+	if e.providerIncrementalContentChanged(
+		e.idPrefix+sessionID, lookupPath, info,
+	) {
+		return 0, false, true
+	}
 	sess, _ := e.db.GetSession(ctx, e.idPrefix+sessionID)
 	return info.ModTime().UnixNano(), sess != nil &&
 		sess.Project != "" &&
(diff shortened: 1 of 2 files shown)


wesm:
Thank you, merging this

PR #978, opened 2026-07-03
feat(activity): group activity by branch

prateek:
Activity reports now include an activity-specific branch breakdown, filling the grouping piece that the shared branch filter foundation left to each view. The new `by_branch` rows are keyed by typed `project` and `branch` fields, so same-named branches in different projects stay separate and the API does not expose the internal filter token as a grouping key.

The CLI report prints a `By branch` section. The web Activity page adds a branch typeahead and a Branch breakdown panel, with the activity store loading branch options itself so the page can deep-link and refresh without depending on another view.

Activity page filtered to one branch, with the Branch breakdown panel (synthetic data)

SQLite, PostgreSQL, and DuckDB all carry the same `by_branch` shape. Empty branches remain distinct from a literal branch named `unknown`; reviewers should look at the activity aggregation path, the three read backends, and the Activity page breakdown wiring.

Part of #928


prateek:
Addressed the automated review finding: the `git_branch` row in docs/session-api.md now documents the opaque (project, branch) token format from `GET /api/v1/branches` instead of implying a plain branch name.

wesm:
superseded by #1017

PR #979, opened 2026-07-03
feat(usage): group usage by branch and make branches excludable

prateek:
Adds branch grouping to the usage view, building on the branch filter foundation. Usage summaries now return range-wide `branchTotals` and per-day `branchBreakdowns` for each `(project, branch)` pair across SQLite, PostgreSQL, and DuckDB. The cost-over-time chart and Attribution panel get a Branch group-by option with human `project/branch` labels.

Usage cost chart and attribution grouped by branch (synthetic data)

Branch can now be excluded like Project, Agent, and Model. Clicking a branch tile or row hides that branch across usage panels; the toolbar shows `Branch: N hidden` with unhide controls, and `exclude_git_branch` is stored in the URL and localStorage so reloads and prior-period comparisons keep the filter.

One branch hidden, "Branch: 1 hidden" dropdown open (synthetic data)

For reviewers, the branch grouping is intentionally implemented at the usage-view layer rather than as a shared grouping primitive. The exclude filter flows through HTTP, service, and backend usage filters as a NOT-wrapped composite-pair predicate; empty or malformed token lists exclude nothing, and cursor-usage rows are bypassed when branch filters are active. The usage store fetches branch options from `/branches`, so the usage page does not depend on session-page state.


prateek:
Addressed the automated review finding: the HTTP service backend was dropping `exclude_git_branch` when serializing usage requests, so daemon-mode CLI clients would silently ignore branch exclusions. Both the summary and pairwise query maps now carry it, with serialization tests for each path.

prateek:
Addressed the round-2 finding: the generated-client mock in `usage.test.ts` now includes `MetadataService.getApiV1Branches` (the store's new import), with a test covering `loadBranches()` caching. For the record, the previous state didn't fail module loading — vitest resolves a missing named export from a mocked module as `undefined` — but any test exercising `loadBranches()` would have crashed, so the mock gap was real.

prateek:
Addressed the round-3 finding: `FilterDropdown` now takes a `separator` prop (default `,`), and the Branch dropdown passes `BRANCH_LIST_SEP` directly instead of adapting the token list through commas — a comma inside a project or branch name no longer mis-splits the checked state. Covered by a component test with a comma-containing token.

wesm:
superseded by #1017

PR #989, opened 2026-07-03
feat(usage): show session context and token breakdown (#982)

### LINKED ISSUE #982: Context & token usage in sessions
### Problem / motivation

One of the things I've gotten use to using another Agent session tool is the ability to review my sessions token & context usage, not only overall but on a prompt by prompt basis. This has helped me identify places I can optimize how I'm working from tool usage (including specific tools vs other tools) along with understanding what I ask and how it affects the context.

### Proposed solution

Adding context & token details, especially in a session view:
* Add token usage for individual tool calls
* Display token usage for output blocks
* Display high level token usage (by type) and expandable break down of usage (ie each tool call) for the context at transition points (ie what is in the context going into a prompt, ie "new context injected this prompt"
* Final token usage (input, output, cached read, cached write) and break down of last "visible" context (CLAUDE.md, tool outputs, task coordinations, user messages etc)

The other tool is under an MIT license so I suspect I could add some small screenshots here but will hold off until requested

### Alternatives considered

_No response_

rodboev:
Session usage currently stops at aggregate token and cost totals. That is useful in the breadcrumb, but it does not show which prompts or steps consumed the context and token budget inside a session.

This extends the existing session usage path instead of adding a parallel endpoint. `GetSessionUsage` keeps computing the current totals, and it also returns ordered breakdown rows from the same deduplicated usage stream. The Huma response maps those rows into `/api/v1/sessions/{id}/usage`, the generated TypeScript client is refreshed, and the session-detail UI renders the breakdown while preserving the existing totals.

This is stacked on the tool usage analysis branch because both changes regenerate the committed frontend client. The generated output in this branch includes that API surface plus the new session usage breakdown field.

Depends on #988.

Closes #982

### DIFF
--- a/cmd/agentsview/archive_query_backend.go
+++ b/cmd/agentsview/archive_query_backend.go
@@ -263,7 +263,7 @@ func (b localArchiveQueryBackend) SessionUsage(
 		engine.Close()
 	}
 
-	u, err := b.database.GetSessionUsage(ctx, resolvedID)
+	u, err := b.database.GetSessionUsage(ctx, resolvedID, true)
 	if err != nil {
 		return nil, tokenUseExitErr,
 			fmt.Errorf("querying session usage: %w", err)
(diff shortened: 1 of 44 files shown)


wesm:
Thank you, taking a look at this shortly!

wesm:
image

PR #1004, opened 2026-07-06
feat(sync): add sync_include_cwd_prefixes ingestion filter

RobSchilderr:
Adds an opt-in `sync_include_cwd_prefixes` config option that restricts local session ingestion to sessions whose working directory lives under one of the configured path prefixes.

Motivation: on a machine shared across multiple clients or workspaces, the archive currently indexes every session it can find, so transcripts from unrelated (possibly confidential) workspaces end up side by side in one database and UI. The DuckDB/PG mirror `projects` filters can carve out a scoped read-only view, but that requires a second serving process and an exact project-name list that goes stale as projects are added. A cwd allow-list at ingestion solves this at the source, works uniformly across agents (including agents like Codex whose session files are date-organized rather than per-project), and keeps everything in the normal `serve` flow.

```toml
sync_include_cwd_prefixes = [
  "/home/me/work/client-a",
  "/home/me/oss",
]
```

## What changed

- **`internal/sync/cwd_filter.go`** (new): `cwdPrefixFilter` with normalization (trim, drop blanks, strip trailing separators) and path-boundary-aware matching (`/a/b` matches `/a/b/c` but not `/a/bc`; both `/` and `\` boundaries).
- **`internal/sync/engine.go`**: `EngineConfig.IncludeCwdPrefixes`; the filter is enforced as a veto in `prepareSessionWrite`, before the preserve/merge handling, so every full-write path across all agents is covered by the existing central seam. Callers already treat the veto as an intentional skip.
- **`internal/config/config.go`**: `SyncIncludeCwdPrefixes` (`toml:"sync_include_cwd_prefixes"`, config-file only, not exposed via the settings API).
- **`cmd/agentsview/*.go`, `internal/server/huma_routes_sync.go`**: the option is wired into every local engine construction site. The remote-sync engine (`internal/remotesync`) is deliberately left unwired — the prefixes describe local paths, so applying them to remote hosts' cwds would silently drop remote sessions.
- **`cmd/agentsview/cli.go`**: `--help` section documenting the option.
- **`docs/configuration.md`**: new "Restricting Ingestion by Working Directory" section under Sync Behavior.
- **Tests**: unit table for the matcher (`cwd_filter_test.go`), config load tests, and an engine integration test (`cwd_filter_integration_test.go`) syncing one in-prefix, one out-of-prefix, and one boundary-sibling (`/Users/alice/workspace` vs prefix `/Users/alice/work`) Claude session.

## Semantics / decisions

- Empty list (default): no behavior change.
- Sessions without a recorded cwd are skipped while the filter is set — they can't be attributed to a workspace, and an allow-list that silently admits unattributable sessions wouldn't be an allow-list. Documented in the config docs and `--help`.
- The filter gates ingestion only. Sessions already in the archive are preserved, consistent with the persistent-archive policy in `AGENTS.md`; the docs point at `agentsview prune` for cleaning up previously ingested sessions.
- Matching is case-sensitive and `~` is not expanded (documented).
- The incremental append path is unaffected by design: it only appends to sessions that already exist in the archive, which is exactly the pre-existing-data behavior above.

## Where to look

The veto placement in `prepareSessionWrite` is the main review surface — it sits after the worktree-project resolution and before `shouldPreserveOpenCodeFormatArchive`, so a filtered session is never written by any downstream path. The other question worth a look is the empty-cwd decision; happy to flip it to allow-when-missing if you'd rather bias toward completeness.


RobSchilderr:
Addressed the roborev medium finding in 368b2c7f:

- `tryIncrementalJSONL` now vetoes sessions whose stored `cwd` is outside the allow-list right after the incremental lookup, falling back to the full parse path so `prepareSessionWrite` applies the same veto (this also re-derives the cwd from the whole file, covering rows that predate cwd capture).
- `writeIncremental` refuses filtered updates as a defense-in-depth guard at the write seam, so no future producer can bypass the filter.
- Regression test `TestSyncEngineCwdPrefixFilterBlocksIncrementalAppend`: archives an outside-prefix session unfiltered, enables the filter, appends to the file, and asserts no new messages land. Verified red without the engine change, green with it.

wesm:
looking

wesm:
Makes sense. Merging

PR #1006, opened 2026-07-06
fix(activity): count subagent sessions in activity report cost

wesm:
## Summary

The Activity dashboard's Total Cost diverged from `agentsview usage daily` (the authoritative daily figure) whenever subagent sessions carried usage. The report's candidate-session query used the default analytics relationship exclusion, which drops `relationship_type = 'subagent'` rows, while `GetDailyUsage` never filters by relationship type — it relies on per-row usage dedup instead. On a real archive, one day showed ~$177 on /activity vs ~$206 from usage daily, with whole models missing from the activity breakdown.

All three backends (SQLite, PostgreSQL, DuckDB) now set `IncludeSubagents = true` in `GetActivityReport`, the same opt-in that `GetAnalyticsSummary` already uses for token/cost aggregates. Fork rows stay excluded because they replay a root session's messages and would double-count; the new tests pin both behaviors, and the cross-backend parity fixture now includes a subagent and a fork session.

Verified end-to-end against a copy of a real 22 GB archive: the activity report and `GetDailyUsage` now agree exactly (cost and output tokens) for the same day and timezone.

One observable side effect: the report's `sessions` count and `by_session` table now include subagent rows. This restores the intended cost semantics rather than changing row meaning, so `ActivityReportSchemaVersion` stays at 1; the docs note the subagent-counting behavior.

Where to look: the one-line filter change in each backend's `GetActivityReport` (`internal/db/activityreport.go`, `internal/postgres/activityreport.go`, `internal/duckdb/activityreport.go`); everything else is tests and docs.

### DIFF
--- a/docs/activity.md
+++ b/docs/activity.md
@@ -47,6 +47,11 @@ The summary cards show:
 - **Projects** and **Models** — distinct counts in the range
 - **Total Cost** — estimated cost attributed to activity in the range
 
+The report counts subagent sessions (for example Claude Code Task-tool agents)
+alongside their parent sessions, so **Total Cost** lines up with
+`agentsview usage daily` for the same day and timezone. Fork sessions are
+excluded because they replay a parent session's activity.
+
 If the selected range reaches into the future, the page marks it as partial and
 shows the report's current **as of** time.
 
--- a/docs/activity.md
+++ b/docs/activity.md
@@ -48,9 +48,10 @@ The summary cards show:
 - **Total Cost** — estimated cost attributed to activity in the range
 
 The report counts subagent sessions (for example Claude Code Task-tool agents)
-alongside their parent sessions, so **Total Cost** lines up with
-`agentsview usage daily` for the same day and timezone. Fork sessions are
-excluded because they replay a parent session's activity.
+and fork sessions (rewound conversation branches) alongside their parent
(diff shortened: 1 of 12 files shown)


wesm:
Addressed the roborev finding in ac81a6c9 — and it turned out to be a real bug, not just a docs problem.

Investigated how fork sessions are actually built: both fork producers (the claude and piebald parsers) split conversation DAG branches into disjoint message sets — a fork session holds only its own rewound-branch entries, never replayed copies of the parent's. Verified on a real archive: all 6,062 eligible usage rows in fork sessions were unique-keyed (no Claude message/request pair or source_uuid overlap with any non-fork session), so `GetDailyUsage` counts every one of them, and the activity report was dropping them (~1M output tokens concentrated in Jan–Apr 2026 data).

The fix adds `IncludeForks` to `AnalyticsFilter`, threads it through the shared `RelationshipExclusionSQL` helper, and sets it alongside `IncludeSubagents` in all three `GetActivityReport` backends. Any usage rows that do recur across related sessions still collapse in the aggregator's first-seen dedup — the same guarantee `GetDailyUsage` relies on — so parity holds by construction. The backend tests now cover both cases (a unique-usage fork that must count, and a replaying fork that must contribute a session row but no cost), and the cross-backend parity fixture exercises both.

PR #1007, opened 2026-07-07
fix(sync): skip local git discovery for foreign-machine sessions

wesm:
The sync engine's project-identity discovery (`cachedProjectIdentity` → `discoverLocalGitIdentity`) ran `filepath.EvalSymlinks` plus a git-root walk on every session's `cwd`, including sessions recorded on other machines. A remote machine's cwd names a path on that machine, so probing it on the local filesystem is wrong on any OS.

On macOS it is also expensive: `/home` is an automounter trigger, so merely stat'ing a synced Linux session's `/home/...` cwd wakes `automountd`, which resolves the auto_home map through `opendirectoryd`, and negative results are not cached. On an archive with tens of thousands of sessions synced from a remote Linux machine (thousands of distinct `/home/...` cwds) and a one-minute identity-cache TTL, any sustained session-write activity produced a continuous syscall storm — measured live as automountd pinned at ~100% and opendirectoryd at 1.3–3.6 cores, dropping to exactly zero when the daemon stopped. `fs_usage` showed four agentsview threads issuing `lstat64`/`readlink` on `/home` at roughly 1,500 calls/sec.

The fix gates discovery on the observation's machine matching the engine's own machine identity, next to the existing `idPrefix`/`pathRewriter` remote-import gate. Foreign-machine observations keep their raw cwd as the root path with no git enrichment, which is what they should have carried all along. With the fix deployed on the same archive, automountd fell from ~175% CPU to ~10% and opendirectoryd from multi-core to ~22% under identical daemon load.

Rows already written with locally resolved roots for foreign sessions are left as-is; they age out through normal observation upserts.

### DIFF
--- a/internal/sync/engine.go
+++ b/internal/sync/engine.go
@@ -6662,7 +6662,13 @@ func (e *Engine) cachedProjectIdentity(machine, rootPath string) projectIdentity
 		return cached
 	}
 	identity := projectIdentityCacheEntry{rootPath: rootPath}
-	if e.idPrefix == "" && e.pathRewriter == nil {
+	// Only probe the local filesystem for sessions recorded on this
+	// machine: another machine's cwd (e.g. /home/... from a synced Linux
+	// host) is meaningless here, and on macOS merely stat'ing such paths
+	// wakes the /home automounter — with tens of thousands of remote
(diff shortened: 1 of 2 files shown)


wesm:
Superseded by #1008 with a scrubbed description and test fixture.

PR #1017, opened 2026-07-07
feat: branch filtering, activity rollups, and usage grouping

wesm:
Adds Git branch as a first-class filter and grouping dimension across session
discovery, Activity, and Usage. This combines and supersedes #974, #978, and
#979 as part of #928.

Branch identity remains project-qualified through opaque tokens, so same-named
branches in different repositories stay independent. The stable
`/api/v1/branches` contract continues to return project, branch, and token;
bounded, recency-ordered picker search now uses the additive
`/api/v1/branch-names` endpoint.

Session search, the sidebar, HTTP clients, and MCP tools can filter by branch.
Picker suggestions follow the active project and session scope, retain the
branchless option outside bounded results, and recover safely from stale
archive-specific identities.

Activity now provides branch rollups across SQLite, PostgreSQL, and DuckDB. Its
branch panel shows the 39 highest-value rows and combines the remainder into an
Other row, keeping large archives responsive without misrepresenting totals.

Usage can group charts and attribution by branch with selection shared across
rows, URLs, and saved preferences. Branch data is loaded lazily without
interrupting comparisons; totals-only comparison queries skip branch buckets,
and unattributed usage remains visible so percentages reconcile with summary
totals.

Branch selections are include-based and intersect with the sidebar branch
filter when both are active. Large explicit selections can still increase URL
and request size, while branch breakdown payloads grow with per-day branch
churn; both costs are opt-in.


wesm:
This branch needs more work (I have over 1000 branches in my session database, and there are performance sharp edges) so I'm going to ship 0.37.0 and then get this merged when I can

wesm:
Will do a human QA pass on this before merging

wesm:
returning to this

PR #1024, opened 2026-07-08
feat(cli): render session search results as an aligned table

mjacobs:
`session search` human output printed two lines per match (a metadata line plus an indented snippet), which was hard to scan. Flat results now render as one aligned row per match — ID, MATCH, SCORE (only for semantic/hybrid), PROJECT, LOCATION, SNIPPET — in the same hand-rolled style as the `session list` table, with no new rendering dependency.

**Before:**

before

**After:**

after

On a TTY the snippet column expands to fill the remaining terminal width (floored at 40 cells on narrow terminals); PROJECT and LOCATION are capped so oversized values cannot starve it. Piped/non-TTY output prints all values untruncated so grep workflows keep working, and `--json` is unchanged. `--context` requests keep the record-style output, since context lines cannot live inside table rows.

Sizing, padding, and truncation are done in terminal display cells via `go-runewidth` (promoted from an existing indirect dependency), so full-width CJK and emoji content stays aligned.

Limitations: the row identifies the match, not the session — surfacing session names alongside matches is left for a follow-up. Reviewers should look at `cmd/agentsview/session_search.go`: the new `printContentMatchesTable` and its dispatch; the record-style printer is untouched apart from being demoted to the `--context` path.

### DIFF
--- a/cmd/agentsview/session_search.go
+++ b/cmd/agentsview/session_search.go
@@ -6,11 +6,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
+	"github.com/mattn/go-runewidth"
 	"github.com/spf13/cobra"
 	"go.kenn.io/agentsview/internal/db"
 	"go.kenn.io/agentsview/internal/service"
+	"golang.org/x/term"
 )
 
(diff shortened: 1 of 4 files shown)


wesm:
thank you — rebasing!

PR #1031, opened 2026-07-08
feat(sync): incremental HTTP remote sync via persistent per-host mirror

wesm:
## Summary

HTTP remote sync previously downloaded the entire remote session corpus on every sync. This change makes transfer, server-side reads, and client-side extraction O(changed files) while keeping the sync protocol free of per-agent knowledge.

**How it works**

- A new `POST /api/v1/remote-sync/manifest` route returns a gzipped JSON listing of every regular file under the allowed target dirs (path, size, nanosecond mtime).
- The client keeps a persistent byte-for-byte mirror of the remote agent dirs under `<data_dir>/remote-mirrors/<host>-<hash>/`, diffs it against the manifest via a stat walk (microsecond-truncated mtime comparison, a deliberate tradeoff for coarse-timestamp filesystems), and requests only changed files through a new delta mode on the archive endpoint (`files` list in the request body). The importer then runs over the complete mirror tree, so parsers that read sibling files keep working unchanged.
- Both archive writers now emit PAX tar headers so sub-second mtimes survive the round trip; archive responses are gzip-compressed when the client advertises it.
- Deletions and type-conflict cleanup run before extraction; a flock next to the mirror root is held from diff through import to serialize concurrent syncs.

**Compatibility**

- Old daemons have no manifest route and their SPA catch-all answers 200 + HTML, so the client gates on the response Content-Type and falls back to the legacy full-archive flow. Delta rejection mid-rollout retries once as a full archive. Old clients against new servers keep working (PAX extracts transparently; Go's transport handles the gzip).
- The mirror is a disposable transfer cache: deleting it forces a full re-download; DB sessions are never deleted.

**Security**

Delta file requests are validated fail-closed: absolute remote paths only (OS-neutral), path dialect (POSIX / drive / UNC) must match the allowed root to prevent archive-namespace collisions, `..` and relative paths rejected, and any symlinked root or symlinked path component is refused so delta mode cannot stream anything the manifest walk would not list. Mirror deletions and conflict removal are confined to the mirror root.

**Tradeoffs and limitations**

- First sync with a new client/server pair is still a full download (mirror bootstrap); a delta covering half or more of the corpus also falls back to a full archive rather than uploading a huge file list.
- The mirror doubles local disk usage for remote hosts (it is the transfer cache that makes deltas possible).
- One-time costs during rollout: a mirror bootstrapped from an old server re-fetches broadly once when the server upgrades (mtime precision), and old clients see a one-time re-parse against new servers.

**Where to look**

- `internal/remotesync/mirror.go` — mirror layout, diff, deletions, lock
- `internal/remotesync/resolve.go` — delta file validation
- `internal/remotesync/http.go` — client flow (manifest probe, delta/full download, fallbacks)
- `internal/server/huma_routes_remote_sync.go` — manifest route, delta archive mode, gzip negotiation
- `docs/superpowers/specs/2026-07-07-incremental-remote-sync-design.md` — design spec


wesm:
This is tested and working really well

Issue #1032, opened 2026-07-08
New 30+ second sync every 2 minutes consumes lots of CPU


colinmollenhour:
### agentsview version

agentsview v0.36.1 (commit 4c4bb56d, built 2026-07-03T19:22:00Z)

### Install method

Homebrew cask

### OS / platform

Ubuntu 24.04

### Which agent and version

n/a

### Which model(s)

n/a

### What happened, and what did you expect

Upon upgrading Agentsview (I can't remember previous version, I think 0.26 or so) Agentsview now consumes a lot more CPU than it did before. Of course an initial sync is to be expected but this goes on perpetually even when there are no or very few active agents.

<img width="2061" height="1045" alt="Image" src="https://github.com/user-attachments/assets/3693ceae-845d-478e-91d7-2750143088d1" />

### Sample session file or snippet

```text
2026/07/08 14:17:32 Polling unwatched directories...
2026/07/08 14:17:38 sync: 4555 file(s) updated
2026/07/08 14:17:38 discovery: opencode returned 4793 sources in 111ms
2026/07/08 14:17:38 discovered 4856 files (0 claude, 0 codex, 0 copilot, 0 gemini, 1 cursor, 0 amp, 0 zencoder, 0 iflow, 0 vscode-copilot, 0 visualstudio-copilot, 27 pi, 0 omp, 0 kiro, 0 zed, 0 vibe) in 118ms
2026/07/08 14:17:38 counted 0 db-backed sessions in 0s
2026/07/08 14:17:39 watcher: 9 file(s) changed, triggering sync
2026/07/08 14:18:03 GET /api/ping
2026/07/08 14:18:03 GET /api/v1/sessions
2026/07/08 14:18:12 file sync: 4713 synced, 143 skipped in 33.144s
2026/07/08 14:18:12 warp sync: 0s
2026/07/08 14:18:12 forge sync: 0s
2026/07/08 14:18:12 piebald sync: 0s
2026/07/08 14:18:12 persist skip cache (75 entries): 0s
2026/07/08 14:18:41 sync: 4555 file(s) updated
2026/07/08 14:18:41 watcher: 4 file(s) changed, triggering sync
2026/07/08 14:19:11 sync: 4555 file(s) updated
2026/07/08 14:19:32 Polling unwatched directories...
2026/07/08 14:19:32 discovery: opencode returned 4793 sources in 123ms
[...]

PR #1037, opened 2026-07-08
perf(sync): release large backfill heaps after signal recompute

### LINKED ISSUE #1022: high ram usage, about 1.5GB
### agentsview version

v0.36.1

### Install method

GitHub release binary

### OS / platform

26.5.2 (25F84)

### Which agent and version

n/a

### Which model(s)

n/a

it should only use about 300mb ram

```text

```

### Steps to reproduce

_No response_

### Checklist

rodboev:
Large local archives can finish initial sync and then keep the backend daemon's heap inflated while the background signal and secret backfill walks archived sessions. That path loads each session's full messages, tool calls, and tool result event content, computes signals and secret findings, persists the derived rows, then moves on to the next session. With thousands of archived sessions, those per-session allocations can be dead while the runtime still holds the expanded heap.

This change adds a narrow memory-pressure release to the signal backfill compute path. The sync engine now estimates the loaded content bytes for each successful recompute and accumulates those bytes across the backfill walk. When the accumulated payload crosses a 256 MiB threshold, it calls the runtime memory release hook and resets the counter. Direct delayed signal recomputes keep their existing behavior and do not force a release.

The scope is intentionally small. This does not rewrite message storage APIs, stream secret scanning, change parser behavior, alter remote or artifact sync, or touch the frontend WebView cache work from #261. It also keeps the CPU loop from #1032 separate; that issue was fixed in v0.37.1, while this path addresses memory growth during archived signal backfill.

Focused coverage proves the byte accounting includes hydrated message, tool, result event, and finding payloads, the backfill closure releases after accumulated bytes cross the threshold, below-threshold recomputes do not release, and direct `RecomputeSignals` stays outside the forced-release path.

Closes #1022

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -237,9 +237,7 @@ func runServe(cfg config.Config, opts serveOptions) {
 		go idleTracker.Do(func() {
 			if err := database.BackfillSignals(
 				ctx,
-				func(bCtx context.Context, id string) error {
-					return engine.RecomputeSignals(bCtx, id)
-				},
+				engine.BackfillSignalComputer(),
 			); err != nil && ctx.Err() == nil {
 				log.Printf("signals backfill: %v", err)
 			}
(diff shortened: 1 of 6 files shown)


wesm:
Thank you — this is one of the other pieces of the performance / memory use puzzle

PR #1042, opened 2026-07-09
fix(parser): attribute skill name for OpenCode skill tool calls

### LINKED ISSUE #1040: Top Skills analytics empty for OpenCode sessions (skill inference never runs)
# Background

The "Top Skills" analytics panel and its underlying skill-usage data are
empty for OpenCode sessions. Skill attribution relies on every parser
populating `ParsedToolCall.SkillName`, but the OpenCode parser never does,
so OpenCode tool calls never count toward Top Skills.

Root cause: `extractOpenCodeToolCall` in `internal/parser/opencode.go:782`
sets only `ToolUseID`, `ToolName`, `Category`, and `InputJSON`. It never runs
skill inference. Every other parser does:

- Codex (`internal/parser/codex.go:450`) calls
  `inferCodexSkillNameWithBase(name, inputJSON, b.cwd)`.
- Cursor (`internal/parser/cursor.go:258`) and Kimi
  (`internal/parser/kimi.go:350`) call `inferToolSkillName`.
- Claude JSONL (`internal/parser/content.go:87-96`) reads
  `input.skill`/`input.name` for the Skill tool, then falls back to
  `inferToolSkillName`.

OpenCode sessions contain the same kinds of tool calls that other agents
record: a read-file tool, a shell/bash tool that cats/seds/greps a
SKILL.md, and OpenCode's own skill tool. The existing heuristics in
`internal/parser/skill_inference.go` would attribute these, but they are
never invoked from the OpenCode parser.

# Version

agentsview 0.37.1.

# Acceptance Criteria

## Must

- [ ] OpenCode tool calls that read a SKILL.md (read-file tool, or a
      bash/cat/sed/grep shell command) are attributed a skill name so they
      count toward Top Skills.
- [ ] OpenCode's own skill tool is attributed a skill name when its input
      carries a skill path or name.
- [ ] Relative SKILL.md paths in OpenCode tool calls resolve against the
      session worktree (the parser already exposes it as the session Cwd).
- [ ] Resyncing existing OpenCode sessions populates skill usage without
      losing any session data.
- [ ] A unit test covers an OpenCode tool call inferring a skill name.

---

Created with [create-issue](https://github.com/tomzx/agents/blob/f51eabc/skills/create-issue/SKILL.md) via glm-5.2 (`f51eabc`)

tomzx:
# What

- `extractOpenCodeToolCall` in `internal/parser/opencode.go` now sets `ParsedToolCall.SkillName` for OpenCode's dedicated `skill` tool, reading the name directly from the tool part's stored `state` with a fallback chain: `state.input.name`, then `state.metadata.name`, then the basename of `state.metadata.dir`.
- Added a focused unit test (`TestExtractOpenCodeToolCall_SkillNameResolution`) covering the fallback chain and the non-skill-tool negative case, plus a DB-level green-path test (`TestParseOpenCodeDB_SkillTool`) seeded with real session-data shapes.
- No changes to other tools: read/bash/edit and the rest are left untouched, and only the `skill` tool is attributed.

## Before
Screenshot_2026-07-08-21-41-43-693_com android chrome

## After 
Screenshot_2026-07-08-21-42-53-137_com android chrome

# Why

Top Skills analytics was empty for OpenCode sessions because `extractOpenCodeToolCall` never set `ParsedToolCall.SkillName`, so OpenCode tool calls never reached the `tool_calls.n` column that the `BySkill` (Top Skills) query aggregates. The original acceptance criteria assumed path-based inference (read-file / bash-cat-sed-grep against a SKILL.md) would be needed, but inspecting real OpenCode session data showed the skill name is already recorded explicitly in the `skill` tool's state, so inference is unnecessary and noisier. Reading the name directly is more accurate and avoids mis-attribution from incidental `SKILL.md` mentions in shell commands. See #1040.

# How to test

1. `CGO_ENABLED=1 go test -tags fts5 ./internal/parser/ -run 'OpenCode'` - the `SkillTool` and `SkillNameResolution` tests assert the name is attributed.
2. Inspect a real session: `sqlite3 ~/.local/share/opencode/opencode.db "SELECT json_extract(data,'$.state.input.name') FROM part WHERE json_extract(data,'$.tool')='skill' LIMIT 3"` confirms the name field the parser now reads.

# Acceptance criteria coverage

- [x] OpenCode's own skill tool is attributed a skill name when its input carries a skill path or name. - Met: `openCodeSkillName` reads `state.input.name` first.
- [x] Resyncing existing OpenCode sessions populates skill usage without losing any session data. - Met: the parser now emits `SkillName`; the existing non-destructive sync repopulates `tool_calls.n` on resync.
- [x] A unit test covers an OpenCode tool call [attributing] a skill name. - Met: `TestExtractOpenCodeToolCall_SkillNameResolution` and `TestParseOpenCodeDB_SkillTool`.
- [ ] OpenCode tool calls that read a SKILL.md (read-file tool, or a bash/cat/sed/grep shell command) are attributed a skill name so they count toward Top Skills. - Not applicable under the revised approach: OpenCode records skill invocations through its dedicated `skill` tool (which carries the name directly), so path-based inference over read/bash commands is not needed. The underlying goal (Top Skills populated for OpenCode sessions) is met via the `skill` tool.
- [ ] Relative SKILL.md paths in OpenCode tool calls resolve against the session worktree. - Not applicable: with no path-based inference, there are no relative paths to resolve; the skill name comes straight from the stored state.

# References

- Closes #1040

---

SDLC phase: pr (FEAT-1040 #1040)
Created with [create-pr](https://github.com/tomzx/agents/blob/f51eabc7f4e3b3d338b351ceafb519b8be4049eb/skills/create-pr/SKILL.md) via glm-5.2 (`f51eabc`)

### DIFF
--- a/internal/parser/opencode.go
+++ b/internal/parser/opencode.go
@@ -776,7 +776,17 @@ type openCodeToolData struct {
 
 // openCodeToolState holds the nested state of a tool call.
 type openCodeToolState struct {
-	Input json.RawMessage `json:"input"`
+	Input    json.RawMessage       `json:"input"`
+	Metadata openCodeSkillMetadata `json:"metadata"`
+}
+
+// openCodeSkillMetadata captures the metadata OpenCode records for
+// its own skill tool. The resolved skill name and directory are
+// stored directly, so attribution does not rely on path-based
+// inference the way other agents do.
+type openCodeSkillMetadata struct {
+	Name string `json:"name"`
+	Dir  string `json:"dir"`
 }
 
 func extractOpenCodeToolCall(data string) ParsedToolCall {
@@ -785,12 +795,21 @@ func extractOpenCodeToolCall(data string) ParsedToolCall {
 		return ParsedToolCall{}
 	}
 
-	var inputJSON string
+	var (
+		inputJSON string
+		skillName string
+	)
 	if len(d.State) > 0 {
 		var state openCodeToolState
 		if err := json.Unmarshal(d.State, &state); err == nil {
 			if len(state.Input) > 0 {
 				inputJSON = string(state.Input)
+			}
+			// OpenCode's skill tool records the skill name
+			// directly in its input and metadata, so the name
+			// is read from there rather than inferred.
+			if d.ToolName == "skill" {
+				skillName = openCodeSkillName(state, inputJSON)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


wesm:
This was already fixed in #1041

PR #1051, opened 2026-07-09
feat(analytics): add skill usage over-time trend with granularity picker

tomzx:
Adds a dedicated SkillTrend analytics panel that charts per-skill usage over time as a multi-series SVG line chart, with a crosshair that snaps to the nearest bucket and a shared tooltip reading out every visible series at that point. Legend chips toggle individual series without repainting the survivors. The top six skills by volume get fixed categorical series colors (new `--chart-series-*` tokens with light/dark steps), and the remaining tail folds into a gray "Other" bucket.

The trend bucket size is now selectable via a new shared GranularityPicker (kit-ui SegmentedControl with localized labels) wired through the analytics store. The skills analytics endpoint accepts `granularity=day|week|month` (default `week`, the previous hardcoded behavior) across SQLite, DuckDB, and PostgreSQL, threading the parameter through `BuildSkillsAnalytics` and every backend's `GetAnalyticsSkills`.

The aggregate weekly sparkline that previously lived inside Top Skills is removed, since the dedicated panel supersedes it with per-skill detail. Message catalogs are updated for all locales (en, ko, zh-CN, zh-TW).

Screenshot_2026-07-09-05-44-59-214_com android chrome


wesm:
superseded by #1070 since I couldn't push here

PR #1061, opened 2026-07-09
fix(parser): strip Codex recommended plugin context

wesm:
Recent Codex builds prepend a `<recommended_plugins>` envelope to the synthetic `role=user` context at the start of sessions. Because it appears before the existing AGENTS.md and environment envelopes, agentsview mistakes the whole item for user-authored content, polluting message history and previews while inflating user-message counts and disrupting automated-session classification.

Strip the envelope only while locating the first genuine user turn, preserving later user-authored mentions and the prompt when Codex places it in the same item. Data version 60 triggers the existing non-destructive resync so already indexed Codex sessions are repaired.

### DIFF
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -833,31 +833,6 @@ func TestMigration_ToolResultEventsTable(t *testing.T) {
 		"expected tool_result_events table after reopen")
 }
 
-func TestCurrentDataVersionKimiUsageEvents(t *testing.T) {
-	assert.GreaterOrEqual(t, CurrentDataVersion(), 55,
-		"Kimi persisted usage events require a data version bump")
-}
-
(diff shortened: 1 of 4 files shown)


internal/db/db_test.go
```diff
@@ -853,6 +853,11 @@ func TestCurrentDataVersionToolInputNULSanitization(t *testing.T) {
 		"tool-call input_json NUL sanitization requires a data version bump")
 }
 
+func TestCurrentDataVersionRecommendedPluginsFiltering(t *testing.T) {
```
cpcloud: No need to clean up the pre-existing tests like this one, but probably good to not add to the pile.

wesm:
That's very pedantic. This fix is fine

Issue #1057, opened 2026-07-09
"Last seen" markers


jklap:
### Problem / motivation

I often use Agentsview to monitor/read what is happening in my sessions -- I often have multiple agents working in terminals and struggle to follow exactly what the agents are doing as they work, complicated by the fact that different vendor's terminal based tools have different levels of functionality to move around, following activity, etc.

This ask is to tweak Agentsview for real-time session's work (and so I and stop using another tool that I've been using only for that purpose)

### Proposed solution

Suggestions:
* Being able to jump from user prompt to user prompt within a session's view
* "Last seen" markers -- when I am reading through a session and get to the end of the current capture (or just stop scrolling) and then go away and come back to the tool or switch to a different session and then come back to this session, it would be nice if it had a clear delineation in the UI for the session on what I'd seen previously and what is new
* An indicator in the session list there is new output yet unseen by the user

Also, it seems that Agentsview can sometimes take a while pick up new output from the session -- is this something that can be improved/eliminated for the session that is currently open and the user is watching?

PR #1064, opened 2026-07-10
feat(frontend): navigate between user prompts

rodboev:
Long sessions currently require readers to move through every assistant and tool row to revisit the human instructions. This adds `Shift+J` and `Shift+K` to move among the visible user prompts while reusing the existing selection and virtualized-scroll path.

The change stays in the frontend. Plain `j` and `k`, arrow navigation, transcript filters, find-in-session behavior, live refresh, APIs, and persistent read-progress state keep their current behavior. The shortcuts modal documents the commands, and focused keyboard plus browser coverage exercises mixed-content navigation.

Refs #1057


wesm:
Super useful

PR #1065, opened 2026-07-10
feat(sessions): show read progress

### LINKED ISSUE #1057: "Last seen" markers

rodboev:
Session updates currently reload the transcript without retaining where the reader stopped, so later output has no boundary in the session view and no signal in the sidebar.

This change adds a read-only `latest_display_ordinal` to session detail and sidebar metadata in SQLite, PostgreSQL, and DuckDB. All three backends derive it with the same transcript visibility rules, so hidden system rows do not affect unread state. Browser storage persists one per-session `seenOrdinal`; the first visit establishes a baseline, visible rows advance it, and sidebar unread state compares the two cursors directly.

Message counts remain loading metadata. Progressive loading, equal-count rewrites, and sparse or one-based ordinals no longer infer transcript position from count values. Existing newest-first boundaries, divider placement inside tool groups, and exact tool-group child visibility remain intact.

The SSE watch path, parser behavior, database schema, and prompt-navigation scope remain unchanged. Prompt navigation continues separately in #1064.

Closes #1057

### DIFF
--- a/frontend/e2e/helpers/mock-sessions.ts
+++ b/frontend/e2e/helpers/mock-sessions.ts
@@ -12,6 +12,7 @@ interface MockSession {
   started_at: string;
   ended_at: string;
   message_count: number;
+  latest_display_ordinal: number | null;
   user_message_count?: number;
   created_at: string;
   file_path: string;
@@ -35,6 +36,7 @@ export function createMockSessions(
     started_at: now,
     ended_at: now,
     message_count: 10,
+    latest_display_ordinal: 9,
     user_message_count: 5,
     created_at: now,
     file_path: `/tmp/${prefix}-${i}.json`,
@@ -154,6 +156,7 @@ function toSidebarIndexRow(session: MockSession) {
     created_at: session.created_at,
     termination_status: session.termination_status ?? null,
     message_count: session.message_count,
+    latest_display_ordinal: session.latest_display_ordinal,
     user_message_count:
       session.user_message_count ?? session.message_count,
     is_automated: session.is_automated ?? false,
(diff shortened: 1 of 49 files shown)


rodboev:
Pulling this PR back for broader rework. The repeated review bounces trace to the same architectural mismatch: read progress currently reconciles backend totals, loaded display counts, and source ordinals, which have different semantics. The rework will move to a single canonical read cursor shared by session metadata, sidebar unread detection, and transcript visibility, then rerun focused coverage before updating the PR.

wesm:
Noted

mariusvniekerk:
This needs some changes since it does unbound growth of LocalStorage

mariusvniekerk (changes requested):
I'm not actually sure what this gets us?  This is a fair amount of code to support drawing a read till here marker without the means to jump to that point?

rodboev:
Good catch on both counts, and you're right, this doesn't carry its weight. The session row already has a change token that sync bumps on ingest (`file_hash` / `local_modified_at`), and the frontend should work around the existing token rather than a computed "unread" value.

Passing that token through the sidebar does the same job without per-backend derivation. That rework is now #1100.

PR #1072, opened 2026-07-10
feat(frontend): make linked date ranges optional

wesm:
Sessions can query a one-year window cheaply, while carrying that same range into Usage or Activity can trigger substantially more expensive aggregation. Cross-page date linking is now a browser-local, off-by-default preference, leaving Sessions, Usage, and Activity on their useful one-year, 30-day, and current-day defaults.

Sessions and Insights retain independent date selections despite sharing the analytics data store. Restored rolling Sessions selections and rolling URLs preserve rolling route intent while refreshing concrete request bounds before the initial list load, including direct session-detail entries, without consuming message deep links. Explicit detail URL dates replace the shared range before loading. Explicit retained Insights selections restore fixed or rolling parameters and can establish an empty enabled shared range. Returning to bare Usage refreshes its rolling bounds before the first request, including after midnight. Bare defaults remain implicit across reloads and history until a picker, dated URL, or shared seed establishes date intent.

When linking is enabled, the latest explicit selection made while opted out becomes the initial shared range, and subsequent selections carry across Sessions, Usage, Activity, Trends, and Insights. Legacy yoke state migrates toward disabled and empty, Activity keeps its maximum-span protection, and the Settings and user guides describe the behavior. The preference remains last-writer-wins between open tabs.

The shared date picker is pinned to the reviewed kit-ui update: completed custom ranges use canonical YYYY-MM-DD bounds, retain a stable overflow-safe toolbar trigger, and keep responsive endpoint readouts in narrow layouts.


mariusvniekerk (approved):

wesm:
This is confirmed working, will merge when roborev is happy

PR #1074, opened 2026-07-10
feat: add config-driven daemon lifecycle commands

wesm:
Background daemon management was coupled to `serve` flags and best-effort lifecycle behavior, which made automation difficult to reason about and allowed unsafe states to look successful. This adds an idempotent, config-driven `agentsview daemon start|status|stop|restart` surface for the writable SQLite daemon, with lifecycle serialization, complete writer-set validation, ownership revalidation, health and compatibility checks, and configuration preflight before replacement.

`serve --background` remains available for flag-driven compatibility cases such as `--no-sync` and one-off bind overrides. `serve restart` bridges to the canonical writable-only restart path, while the existing broader `serve stop` behavior for read-only PostgreSQL and DuckDB servers remains intentionally distinct.

The README and Zensical command, quickstart, configuration, remote-access, and homepage docs now make the canonical workflow, config-only boundary, authentication requirement for persistent non-loopback binds, and compatibility semantics explicit.


wesm:
Addressed the combined review in 13d54e66: background start and auto-start now reject a non-unique writable runtime set, and replacement validates the effective configuration before signaling the incumbent. The Windows failure was a path-quoting assertion mismatch; the test now compares the same quoted representation emitted by the error.

wesm:
Addressed the authenticated legacy-daemon discovery finding in `93557f67`. Canonical status, stop/restart, launch-contention inspection, and background writer-uniqueness checks now pass the configured auth token into legacy runtime migration. Added regressions covering authenticated legacy status, stop, and multi-writer startup refusal.

PR #1075, opened 2026-07-10
feat(export): add stable session evidence contract

mariusvniekerk:
Activity, usage, and session consumers need durable source facts without treating mutable directory labels, private paths, credential-bearing Git URLs, or floating prices as evidence.

This adds one versioned, privacy-bounded export contract across those surfaces. It emits archive-scoped project keys with credential-scrubbed repository and worktree context, exact fixed-scale money with pricing provenance, and bounded JSON or NDJSON session summaries with opaque cursor and generation semantics that work without a running daemon. Resolved project labels remain exact across SQLite, PostgreSQL, and DuckDB usage filters and pairwise comparisons, including labels that contain commas and projects represented only by fork or subagent sessions. Per-session repository and checkout evidence, including historical unknown and machine-root observations, remains immutable across resyncs; root-derived evidence upgrades only when remote resolution becomes resolved or ambiguous. Legacy archives repair missing snapshots through a resumable background migration, and export status initializes compatible export schema, reports failures, and remains incomplete until repair converges. Full resync preserves real historical snapshots while leaving legacy orphaned or trashed sessions without source evidence eligible for repair; this reconciliation is indexed and limited to the copied batch.

PostgreSQL and DuckDB publication use revision-indexed deltas and durable tombstones with filter-scoped cursors, so routine pushes scale with changed identity rows instead of total archive size. Explicit full pushes and database-generation changes retain full repair behavior. Legacy shared stores without source archive rows use a deterministic response scope so unrelated project totals and catalog entries remain distinct without migrating or scanning sessions. Revision-trigger initialization is idempotent across concurrent archive opens, preventing schema migration races. Daemon-started sync commands announce their intent before startup, avoid redundant local work, and retain the initial local population pass for remote-only requests. A bounded daemon-side fallback performs a skipped local sync if the initiating request never arrives or is canceled. Startup signal and identity backfills wait for sync completion and hold the same exclusion lock as later sync and resync operations, so an archive swap cannot overlap maintenance.

The Zensical session-export reference documents the public contract and the limits of legacy evidence reconstruction. It exports source observations only and excludes organization mappings, accounting policy, prompts, transcripts, diffs, file contents, review bodies, and agent logs. The format is unshipped, so this establishes its canonical shape without a legacy compatibility layer.


mariusvniekerk:
Exact-head disposition for `b16f24b2c`:

- The project-label collision finding is fixed. Attribution rows, treemap identity, pairwise selections, and key-based exclusions now use `project_key`; duplicate sanitized labels have regression coverage.
- The legacy shared-mirror migration suggestion is intentionally not implemented. These source-scoped identity tables and their v1 contract are unshipped, and the maintainer chose not to add compatibility scaffolding. Existing development mirrors must be republished by a current writer before being served; serve-before-republish for the superseded shape is unsupported and is now stated explicitly in the public design.

wesm (approved):

wesm (approved):

PR #1085, opened 2026-07-10
feat(grok): add a metadata-first session provider

### LINKED ISSUE #1066: Feature request: support Grok CLI / Grok Build sessions
## Summary

AgentsView does not appear to detect or record sessions from the [Grok CLI](https://grok.com/download) (also referred to as "Grok Build" in dev channels). In a local environment running Grok CLI v0.2.93 with an active session, the AgentsView `sessions.db` contains 0 rows with `agent = 'grok'` while other CLI agents (claude, codex, opencode, kimi, etc.) are tracked normally.

## Environment

- AgentsView CLI: v0.37.3
- Grok CLI: v0.2.93 (macOS aarch64, installed via official installer)
- OS: macOS 26.5.2
- Project cwd: `/Users/dinoel/1/appdb`
- Active Grok session PID: 96724 (comm=`grok`, cwd in project)

## Observed behavior

- `~/.agentsview/sessions.db` has 2,658 tracked sessions but `SELECT agent, COUNT(*) FROM sessions GROUP BY agent` shows no `grok` row.
- Grok CLI stores session metadata under `~/.grok/sessions/<cwd>/<session_id>/signals.json` and `~/.grok/active_sessions.json`.

## Request

Please add native support for Grok CLI sessions so that:
1. AgentsView recognizes an active `grok` process/session.
2. Session metadata (model, message counts, tool failures, etc.) is ingested into `sessions.db` under `agent = 'grok'`.
3. If applicable, supports the separate "Grok Build" channel/binary if it uses a different process name or session layout.

## Reference files

- Grok active sessions: `~/.grok/active_sessions.json`
- Grok session signals: `~/.grok/sessions/<enc_cwd>/<session_id>/signals.json`
- Grok config: `~/.grok/config.toml`

rodboev:
Grok Build stores headless sessions under `~/.grok/sessions`, but agentsview had no Grok entry in the supported-agent registry or provider factory dispatch, so those sessions never entered the sync pipeline and never produced `agent = 'grok'` rows. The initial provider slice fixed discovery and parsing, but sync still overwrote the parser's summary-derived counts with `0`, and search still missed those sessions because the metadata-first provider treated `firstPrompt` as metadata only.

This adds a Grok provider rooted at the official `.grok/sessions` tree and uses the vendor-proven session files to emit stable `grok:<session_id>` rows. `summary.json` is the primary session source, because the official Grok binary describes it as the session index entry carrying the summary, first prompt, model ID, timestamps, and message counts. Optional `signals.json` data is used only for counters that are actually present. Freshness follows the parser's semantic inputs, `summary.json` and optional `signals.json`; the provider watches, classifies, and fingerprints only those parsed files. The parser treats a non-empty `firstPrompt` as proof of one user turn, floors `MessageCount` at that authoritative user-turn minimum, emits that prompt as one synthetic user message for search, and the sync engine preserves those parser-owned summary counts instead of recomputing them from an empty message list.

The scope stays inside `internal/parser` plus the sync write seam: one new `AgentGrok` registry entry, one provider factory case, one provider-authoritative migration mode, a Grok provider or parser pair that own discovery, lookup, and metadata decoding, and a parser-owned count-preservation flag that keeps summary-derived counts intact through sync. It does not decode `updates.jsonl` into a full transcript; it emits only the summary-proven first prompt as a synthetic user message. xAI's own docs and binary strings describe Grok Build and the `grok` CLI as one shared `.grok` session tree, so one provider covers both names here. A second persisted format remains out of scope unless a future artifact proves it exists. Focused parser and sync tests cover discovery, lookup, composite fingerprinting, malformed-artifact boundaries, prompt and summary searchability, and persisted count preservation using synthesized vendor-shaped fixtures built from the official storage contract.

Fixes #1066


wesm:
thank you!

PR #1093, opened 2026-07-11
perf(parser): reuse JSONL reader workspaces

mariusvniekerk:
JSONL-backed parsers allocated two 64 KiB buffers for every source file before doing useful parsing work, making reader setup dominate allocation volume for small and incremental imports. This reuses the composite reader workspace through `sync.Pool`, avoids allocating scratch storage for normal single-fragment lines, and applies the lifecycle consistently across all JSONL-backed agents.

Representative Codex benchmarks reduce allocated bytes from 139–142 KiB to 8–11 KiB per parse (92–94%) and improve the tool-call workload median from 48.7 µs to 34.1 µs. Scratch buffers larger than 256 KiB are discarded on release so exceptional long records do not remain retained.

@wesm, would appreciate your review of the pooling lifecycle and retention tradeoff.

### DIFF
--- a/internal/parser/claude.go
+++ b/internal/parser/claude.go
@@ -106,6 +106,7 @@ func claudeParseWithExclusions(
 	parentSessionID = claudeCompanionParentSessionID(path, sessionID)
 
 	lr := newLineReader(f, maxLineSize)
+	defer releaseLineReader(lr)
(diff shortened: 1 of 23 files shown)


wesm:
rebasing

wesm (approved):

PR #1100, opened 2026-07-11
feat(frontend): show transcript read progress

### LINKED ISSUE #1057: "Last seen" markers

rodboev:
Adds browser-local transcript read progress: sessions that change after being read show an unread indicator in the sidebar, and reopening one places a last-seen divider at the stored ordinal. Markers remain bounded in localStorage, and first visits establish a baseline instead of marking every existing session unread.

A backend-neutral `transcript_revision` now advances transactionally with visible message and tool content, while metadata-only edits such as renaming a session leave read state unchanged. SQLite owns the revision, DuckDB and PostgreSQL mirror it, both mirror backends perform target-scoped one-time full backfills, and full resyncs preserve the prior revision for unchanged transcripts while advancing changed content and retaining archived orphans.

Read acknowledgement is coupled to the refreshed message window: a fetched revision is published only after its messages arrive, and newest-first traversal must cross the unread endpoints before the session is marked read. Precise in-session change ordinals preserve narrow boundaries; reopened revisions without durable change metadata conservatively begin at the earliest displayed message. This keeps the indicator accurate for appended messages, earlier rewrites, grouped tool output, filtered transcript views, and either sort direction.

Closes #1057


wesm:
looking

wesm:
LGTM!

PR #1101, opened 2026-07-12
perf(sync): keep full HTTP remote sync incremental

### LINKED ISSUE #1094: Improve full remote sync performance and phase visibility
## Problem

`agentsview sync --full` can be disproportionately slow for large remote archives.

The current flow appears to make `--full` force two expensive operations:

- Download and extract the complete remote mirror archive, even when the local mirror manifest is already current.
- Bypass the persistent skip cache and reprocess every discovered session.

These are separate concerns. A full parser/database refresh does not inherently require retransferring and rewriting every unchanged mirror file.

The progress label `Downloading session archive` also covers several different operations: remote directory traversal, tar creation, gzip compression, network transfer, local gzip decompression, and tar extraction. A single aggregate timer makes it difficult to identify whether CPU, network, or disk work is dominant.

## Proposed direction

- Decouple full session reprocessing from full mirror repair.
- Let `sync --full` use manifest-based incremental mirror updates while still reprocessing all sessions.
- Provide a separate explicit repair option when the entire remote mirror must be fetched again.
- Avoid extracting or overwriting unchanged mirror files during ordinary full syncs.
- Report separate progress and elapsed time for archive generation/transfer, decompression/extraction, discovery, and session processing.
- Include compressed bytes transferred and processing throughput where available.

## Acceptance criteria

- A full sync against an unchanged remote archive does not transfer the complete archive unless repair is explicitly requested.
- Full sync still reparses all sessions when requested.
- Progress output distinguishes transfer from local extraction and processing.
- Regression coverage demonstrates that mirror transfer work is bounded by changed files rather than total archive size for the ordinary full-sync path.
- Existing deletion, tombstone, and persistent-mirror behavior remains intact.

rodboev:
`agentsview sync --full` currently treats "reparse every HTTP-remote session" and "repair the persistent mirror cache" as the same operation, so unchanged remotes still download and re-extract the whole dir-scoped archive before parsing. This change keeps ordinary full sync on the manifest-backed mirror path, reparses every remote session as requested, and moves whole-mirror repair behind the explicit `--repair-mirror` path instead of baking it into every full sync. Explicit repair still reparses every imported remote session after refreshing the mirror bytes.

The transfer decision stays bounded by the existing mirror rules: bootstrap, half-or-more changed files, or a rejected delta still fall back to a full archive, file-scoped exports still fetch as their own small full archives, and mirror deletions remain non-destructive to the local database. This follows the scoped direction in #1094 from @wesm, keeping ordinary full sync work proportional to changed mirror files rather than total archive size while preserving repair as an explicit action.

Progress output is also split into archive transfer, archive extraction, discovery, and session processing so the elapsed timers line up with the actual work. Remote progress continues to show compressed transfer bytes, and the processing phase now reports throughput when the session and message counters make that meaningful for operators watching long remote syncs.

Closes #1094

### DIFF
--- a/cmd/agentsview/cli.go
+++ b/cmd/agentsview/cli.go
@@ -322,6 +322,10 @@ func newSyncCommand() *cobra.Command {
 		&cfg.Full, "full", false,
 		"Force a full resync regardless of data version",
 	)
+	cmd.Flags().BoolVar(
+		&cfg.RepairMirror, "repair-mirror", false,
+		"Refresh configured HTTP remote mirrors from a full archive",
+	)
 	cmd.Flags().StringVar(
 		&cfg.Host, "host", "",
 		"SSH hostname for remote sync",
--- a/cmd/agentsview/cli_test.go
+++ b/cmd/agentsview/cli_test.go
@@ -325,7 +325,33 @@ func TestRootHelpDocumentsRemoteHosts(t *testing.T) {
 func TestSyncHelpMentionsConfiguredHosts(t *testing.T) {
 	help, err := executeCommand(newRootCommand(), "sync", "--help")
 	require.NoError(t, err, "Execute")
-	for _, want := range []string{"remote_hosts", "--host", "passwordless"} {
+	for _, want := range []string{"remote_hosts", "--host", "--repair-mirror", "passwordless"} {
 		assert.Contains(t, help, want, "sync help missing %q", want)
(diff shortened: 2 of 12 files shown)


wesm:
I am working on an alternate solution for this

rodboev:
Sounds good, closing.

Issue #1097, opened 2026-07-11
serve skips runtime record when data dir is Administrator-owned → serve status blind, desktop app hangs


rodboev:
## Summary

On Windows, when the data dir (`~/.agentsview`) is owned by `BUILTIN\Administrators` instead of the current user, `serve` skips writing the daemon runtime record and falls back to the start lock. The server binds and serves normally, but `serve status` then reports `No agentsview server is running`.

This is a CLI-level bug and reproduces with the CLI alone, no desktop app involved. The desktop app is where it surfaces for users: it polls `serve status` to detect readiness, so it hangs at "Preparing your workspace" with "The background backend disappeared before it became ready." The directory ends up Administrator-owned whenever it was first created by a process running with an elevated token (e.g. a CLI launched from an elevated Task Manager), which on Windows defaults new objects' owner to `BUILTIN\Administrators`.

## Root cause

`debug.log` during a failing start:

```
Starting server at http://127.0.0.1:8080
warning: could not write daemon runtime record: prepare runtime dir:
C:\Users\...\.agentsview is not owned by current user or token owner
(keeping start lock as fallback)
```

The server is up (`/api/ping` returns 200), but with no runtime record written, `serve status` finds no daemon and reports "not running." The fallback keeps serving but leaves the daemon undiscoverable.

This then cascades into two more misleading symptoms. Restarting hits the write lock the still-running (but undiscoverable) daemon holds:

```
fatal: opening writable database: write lock ...\db.write.lock is held by
another process; run `agentsview serve stop`, ...
```

And the remedy that error suggests is a dead end: `serve stop` uses the same discovery as `serve status`, so it prints "No agentsview server is running" and never stops the daemon.

## Reproduction (CLI only, no desktop app)

```
> (Get-Acl ~/.agentsview).Owner
BUILTIN\Administrators                  # not the current user

> agentsview serve --background --no-sync
agentsview starting in background (pid N)

> agentsview serve status
No agentsview server is running.        # but curl http://127.0.0.1:8080/ -> 200
```

Fix the owner and it works immediately:

```
> icacls ~/.agentsview /setowner "<user>"
> agentsview serve --background --no-sync
agentsview running at http://127.0.0.1:8080 (pid N)   # "running", writes daemon.<pid>.json

> agentsview serve status
agentsview running at http://127.0.0.1:8080  pid: N  version: v0.37.5  uptime: 8s
```

The desktop app then loads past "Preparing your workspace" normally.

## Suggested handling

[...]

PR #1103, opened 2026-07-12
fix(daemon): recover recordless writable daemons after startup-state lock acquisition

rodboev:
Kit v0.7.1 and agentsview #1107 resolve the Administrator-owned-directory trigger from #1097 through the normal runtime-record path. This PR covers a separate failure mode: a writable daemon has already acquired the trusted startup-state lock and bound successfully, but publishing its runtime record then fails.

The daemon now records its bound endpoint, process identity, compatibility versions, launch options, and managed-Caddy identity in startup state when that post-lock write fails. Runtime records remain authoritative when present; otherwise discovery accepts the fallback only while the lock is held and the live ping matches the recorded process. If runtime-store enumeration itself fails, discovery and lifecycle commands use the same confirmed fallback; without one, they preserve the original inspection error. Stale state, incompatible daemons, and read-only servers continue to fail closed.

Status and transport discovery can use the recovered endpoint, while stop, restart, and upgrade paths can safely target the confirmed recordless writer. This preserves the existing policy of keeping an already-bound server alive after runtime-record publication fails instead of leaving that server running but undiscoverable.

Refs #1097

### DIFF
--- a/cmd/agentsview/daemon.go
+++ b/cmd/agentsview/daemon.go
@@ -450,6 +450,12 @@ func runDaemonStatus(w io.Writer, deps daemonCommandDeps) error {
 	}
 	records, err := deps.statusRecords(cfg.DataDir, cfg.AuthToken)
 	if err != nil {
+		if rt := writableRuntimeFallbackForCommand(cfg, deps); rt != nil {
+			for _, line := range serveStatusLines(rt) {
+				fmt.Fprintln(w, line)
+			}
+			return nil
+		}
 		return fmt.Errorf("daemon status: inspecting runtime store: %w", err)
 	}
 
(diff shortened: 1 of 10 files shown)


wesm:
Looking at this. Might be better to address this partly in kenn-io/kit, will follow up with work tomorrow

wesm:
I agree that the Administrator-owned-directory trigger should be fixed in kit, not treated as an agentsview lifecycle problem. [kenn-io/kit#37](https://github.com/kenn-io/kit/pull/37) aligns Windows private-directory owner validation with the protected DACL policy that already trusts the current user, token owner, LocalSystem, and built-in Administrators.

I still think this fallback has value as defense-in-depth for failures that happen after the start lock was successfully acquired but runtime-record publication later fails. However, it is not the root fix for #1097 as currently structured: kit gates both `RuntimeStore.LockPath` and `RuntimeStore.Write` through the same private-directory validation, so the Administrator-owned case can fail before there is a trusted start lock for the fallback to observe.

rodboev:
Makes sense. I retitled and rebodied this PR so it stays on the post-lock fallback only.

The Administrator-owned-directory trigger belongs to the kit-side owner-validation fix in https://github.com/kenn-io/kit/pull/37. This branch just keeps later runtime-record publication failures discoverable and stoppable after a trusted startup-state lock already exists.

PR #1106, opened 2026-07-12
fix: import current Grok Build sessions with full transcripts

arbing:
## Summary

Grok Build sessions under `~/.grok/sessions` were either missing or summary-only in AgentsView.

- Accept current snake_case `summary.json` / `signals.json` fields while keeping legacy camelCase support
- Resolve project labels from `git_root_dir` / `info.cwd` and URL-encoded cwd keys
- Parse `chat_history.jsonl` into full transcripts (user turns, assistant replies, thinking, tool calls)
- Fall back to summary-only mode when `chat_history.jsonl` is missing
- Update docs to describe the current Grok import surface

## Notes

- Grok still does not persist per-turn output tokens or cost in local session files, so breadcrumb `—out`, steps, and estimated cost remain unavailable when those fields are absent.
- Focused parser and sync coverage exercises the current schema and chat history path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,12 +357,12 @@ thread JSON files.
 | Zed                   | `~/Library/Application Support/Zed/` (macOS)                                                                                                                            |
 | Zencoder              | `~/.zencoder/sessions/`                                                                                                                                                 |
 
-Grok support is summary-only: AgentsView reads the summary,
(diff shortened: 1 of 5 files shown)


wesm:
thank you!

PR #1118, opened 2026-07-13
fix(desktop): retry backend stop before update install

### LINKED ISSUE #1117: Desktop in-app update reliably fails: 10s backend-stop timeout races serve stop's own 10s grace and daemon startup sync
### agentsview version

agentsview v0.37.1 (commit 9621af85, built 2026-07-08T14:36:14Z) — desktop app (also reproduced with v0.37.5 sidecar; the relevant code is unchanged on `main`)

### Install method

GitHub release binary (desktop NSIS installer)

### OS / platform

Windows 11 Pro (10.0.26200)

### Which agent and version

n/a

### Which model(s)

n/a

### What happened

Accepting the desktop app's update prompt reliably ends in an **"Update Failed"** dialog ("Failed to install the update. Please try downloading manually…"), so in-app updates never complete. It looks like a download failure to the user, but the download actually succeeds.

Captured stderr from the desktop app during the failure:

```
[agentsview] timed out waiting for serve stop before update install
[agentsview] serve stop did not report success, but no detached daemon port is known
[agentsview] update install failed: backend did not stop before update install
```

I verified the release artifacts themselves are fine: the `updater/latest.json` manifest, asset URLs, and minisign signature for `AgentsView_0.37.5_x64-setup.nsis.zip` all check out, and `update.download()` completes. The failure is in the stop-backend-before-install phase.

### Root cause

Two nested timing problems in `desktop/src-tauri/src/lib.rs` since the daemon-first change (#826):

1. `UPDATE_SIDECAR_STOP_TIMEOUT` is 10s for the whole stop phase (spawn `serve stop` + graceful daemon shutdown + port-free wait), but `serve stop` itself waits up to `serveStopGraceTimeout` = 10s for a graceful shutdown before escalating. The outer deadline always expires first whenever shutdown isn't near-instant (e.g. WAL checkpoint / active sync on a large archive — mine has ~4.6k sessions).
2. While the daemon is still running its startup sync (~34s on my archive), `serve stop` exits non-zero with `serve stop: a server is starting; retry once it is ready`. `stop_detached_backend_for_update_with_port` treats one failed attempt as fatal and aborts the update. Since `schedule_auto_update_check` fires 5s after launch — exactly inside the startup-sync window — accepting the update prompt shown at startup fails essentially every time.

### Steps to reproduce

1. Have a reasonably large archive so daemon startup sync takes >15s.
2. Launch the desktop app on a version with an available update.
3. Accept the "Update Available" prompt when it appears shortly after launch.
4. Download progresses, then "Update Failed" appears; stderr shows the lines above.

### Fix

PR incoming: give the stop phase a larger budget (120s), retry `serve stop` until the daemon is ready to stop instead of aborting on the first non-zero exit, and show a distinct dialog message for the backend-stop-timeout case so it is no longer reported as a download/install problem.

Leuconoe:
Fixes #1117

Accepting the desktop update prompt shortly after launch could finish the download but fail before installation. The updater allowed only ten seconds for the complete backend-stop phase even though `serve stop` can spend that entire interval on graceful shutdown, and it treated the daemon's temporary startup refusal as a terminal failure.

The updater now gives backend shutdown a 120-second budget shared across launcher termination, retry attempts, and detached-daemon shutdown. It retries only the explicit "server is starting" response; command errors and other non-zero exits still fail immediately. A backend-stop failure also gets a distinct dialog instead of being presented as a download problem.

### DIFF
--- a/desktop/src-tauri/src/lib.rs
+++ b/desktop/src-tauri/src/lib.rs
@@ -35,7 +35,14 @@ const STATUS_PROBE_TIMEOUT: Duration = Duration::from_millis(1250);
 const STATUS_PROBE_FAILURE_NOTICE_AFTER: u32 = 10;
 const STATUS_PROBE_FAILURE_FAIL_AFTER: u32 = 30;
 const LOGIN_SHELL_ENV_TIMEOUT: Duration = Duration::from_secs(3);
-const UPDATE_SIDECAR_STOP_TIMEOUT: Duration = Duration::from_secs(10);
+// Stopping the detached daemon before an update install must outlast
+// both a daemon that is still mid-startup (serve stop refuses with
+// "retry once it is ready" while the startup sync runs, which takes
+// tens of seconds on large archives) and serve stop's own 10s graceful
+// shutdown window before it escalates to a forced kill.
+const UPDATE_SIDECAR_STOP_TIMEOUT: Duration = Duration::from_secs(120);
+const UPDATE_STOP_RETRY_INTERVAL: Duration = Duration::from_secs(2);
+const SERVE_STOP_STARTING_RETRY_HINT: &str = "a server is starting; retry once it is ready";
 const DATA_VERSION_TOO_NEW_EXIT_CODE: i32 = 3;
 const DESKTOP_LOG_FILE_NAME: &str = "agentsview-desktop.log";
 const DESKTOP_LOG_QUEUE_CAPACITY: usize = 64;
@@ -2401,13 +2408,21 @@ async fn check_for_updates(handle: &AppHandle, silent: bool) {
         |bytes| update.install(bytes),
     ) {
         eprintln!("[agentsview] update install failed: {err}");
+        let message = match err {
+            InstallDownloadedUpdateError::BackendStopTimedOut => {
+                "The update was downloaded, but the local backend \
+                 could not be stopped in time. Please try again in \
+                 a moment."
+            }
+            InstallDownloadedUpdateError::Install(_) => {
+                "Failed to install the update. \
+                 Please try downloading manually from the releases page."
+            }
+        };
         let h = handle.clone();
         handle
             .dialog()
-            .message(
-                "Failed to install the update. \
-                 Please try downloading manually from the releases page.",
-            )
+            .message(message)
             .title("Update Failed")
             .show(move |_| restore_webview_focus(&h));
         return;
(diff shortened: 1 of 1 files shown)


Leuconoe:
The `e2e` failure here is a pre-existing regression on `main`, not caused by this PR (which only touches `desktop/src-tauri/src/lib.rs`; the desktop/Go/lint jobs all pass).

`main` fails the same two tests since 5682fc24 (#1100, transcript read progress):

| run | head | e2e |
|---|---|---|
| [29211839821](https://github.com/kenn-io/agentsview/actions/runs/29211839821) | fb760961 (before #1100) | ✅ |
| [29213539491](https://github.com/kenn-io/agentsview/actions/runs/29213539491) | 5682fc24 (#1100) | ❌ |
| [29218019875](https://github.com/kenn-io/agentsview/actions/runs/29218019875) | cc7559bc (latest main, this PR's base) | ❌ |

Identical failure in all red runs, on main and here:

```
[chromium|webkit] › e2e/navigation.spec.ts:34:3 › Navigation › Shift+J and Shift+K navigate visible user prompts
Error: expect(locator).toHaveCount(expected) failed
Locator:  locator('.virtual-row')
Expected: 6
Received: 14
```

`navigation.spec.ts:40` asserts the "project-beta / 3" fixture session renders exactly 6 `.virtual-row` rows; after the MessageList/windowing changes in #1100 it renders 14, so the expectation (or the row locator, if the new read-progress rows should be excluded) needs updating on main. Happy to rebase once main is green.

wesm:
looking at this

wesm:
Thanks for the detailed failure analysis. I tightened the desktop change, then squashed and rebased the branch while preserving Leuconoe as the commit author:

- The desktop stop phase now uses one 120-second deadline across launcher termination, retries, and detached-daemon shutdown. Only the explicit "server is starting" response is retried; command errors and other failures return immediately. The retry, fatal-error, and deadline paths now have unit coverage.
- The E2E failure was indeed pre-existing on `main`, but the 14 rows came from the test selecting `test-session-fork-1` through incidental text rather than from new read-progress rows. An exact-child-selector fix has since landed on `main` in #1115, so I dropped our redundant E2E change during the rebase.

I also updated the PR description to reflect the final desktop-only diff. Thanks again for tracking down the main-branch boundary and for the original fix.

Please hold off on any further pushes to this branch while maintainer review and CI finish. The concurrent merge raced the requested history rewrite and would have caused the squashed branch to be overwritten without the lease check.

PR #1128, opened 2026-07-13
Reject and repair invalid embedding vectors

mariusvniekerk:
Embedding providers can return dimensionally valid data that is still unusable for cosine search. AgentsView now rejects null components, non-finite values, and zero-norm vectors at the endpoint and persistence boundaries so faulty output is never stamped as complete.

Add an explicit embeddings build --repair-invalid recovery mode that scans one existing generation, durably queues only affected document keys, and resumes them after endpoint failures or cancellation. Repair status preserves exact remaining work when possible, repair scans keep document-content allocation bounded by the scanned batch, and failed targets remain available for later retries.

Repair runs on an isolated path: it does not refresh or expand the mirror, create or reset generations, change generation state, or fill unrelated missing work. Full rebuild and backstop modes remain mutually exclusive with repair.

Daemon compatibility now covers the expanded repair status contract. Embeddings commands resolve only the writable daemon and report restart guidance when an older writable daemon is still running, including when a compatible read-only server is also available.

Document standard Ollama configuration and the cache constraints required when using a direct multi-slot llama-server endpoint.

### DIFF
--- a/cmd/agentsview/daemon_runtime.go
+++ b/cmd/agentsview/daemon_runtime.go
@@ -20,12 +20,15 @@ import (
 	"github.com/gofrs/flock"
 	"github.com/shirou/gopsutil/v4/process"
 	"go.kenn.io/agentsview/internal/db"
+	"go.kenn.io/agentsview/internal/server"
 	"go.kenn.io/kit/daemon"
 )
 
 const (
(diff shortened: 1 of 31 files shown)


wesm:
fixing

PR #1130, opened 2026-07-13
fix(parser): keep encrypted Codex collab payloads out of transcripts

mjacobs:
Follow-up to #1125. Codex multi-agent v2 encrypts inter-agent task payloads server-side (openai/codex#26210, reported in openai/codex#32753): `spawn_agent`/`send_message` arguments arrive as opaque Fernet tokens. #1125 suppressed the child session's inbound ciphertext, but the parent side still exposed it — in formatted `spawn_agent` message bodies, collab tool argument previews, and stored `tool_calls.input_json` — all of which reach the transcript UI, markdown export, resume rendering, and FTS.

This PR removes the ciphertext at ingest, repairs existing archives on every backend, and blocks outdated binaries from writing it back.

## Parser (ingest)

- Redaction is scoped to the three collab tools that carry encrypted arguments (`spawn_agent`, `send_message`, `followup_task`). A legitimate Fernet token anywhere else — a shell command decrypting one, a user quoting one — is preserved.
- Redaction runs before preview truncation: a clipped token no longer validates and would leak a ciphertext prefix.
- Encrypted child sessions record a placeholder user turn naming the sending agent instead of silently dropping the task delivery. Replay detection digests the original ciphertext, so distinct tasks with identical placeholder text cannot shadow each other.

## SQLite archives (upgrade path)

- dataVersion 69 triggers the standard non-destructive resync and exhaustive certification without trusting either the cached tool-use flag or a surviving tool row. Versions 67 and 68 belong to Antigravity parent links and Hermes skill metadata, respectively, so the Codex watermark sits above both and neither can be mistaken for certified. (64 briefly existed with ingest redaction but no copy repair; 65–66 were unrelated Claude bumps.)
- Orphaned rows with no source file are scrubbed by a copy-time pass that mirrors the ingest scope, including tokens clipped by preview truncation.
- The watermark is stamped only on sessions the scrub can vouch for: token-free rows, plus token-bearing rows whose content every applicable redactor already treats as final (deliberately preserved literals). Rows the scoped scrub cannot examine — a lost `tool_calls` row, a child never linked to its parent — stay below the watermark so push-time verification keeps re-checking them.

## PostgreSQL

- The same historical repair runs in place for previews, message roles/content/lengths, and tool names/inputs. It persists UTF-8/control-byte normalization before redaction and preserves SQL `NULL` for absent tool inputs, guarded by write triggers that reject legacy ciphertext from outdated binaries.
- Version 69 rechecks every older row. After the scoped destructive repair, a metadata-independent certification pass examines previews, all token-bearing messages, and collab tool inputs. Token-free and verified-literal sessions advance to the watermark; anything an applicable redactor would still rewrite remains legacy, and read-only serving fails closed rather than exposing it.
- The repair transaction marks its own compatibility version so installed guards allow the scoped rewrites while sessions are still legacy. Promotion happens only afterward and only for sessions that pass exhaustive certification.
- Certification examines token-bearing formatted blocks even when `has_tool_use` is false and no `tool_calls` row survives. A block is safe only when its header cannot be produced by a canonical collab tool, covering both the mixed-tool residue case and the combined stale-flag/missing-row case.
- A collab `tool_calls` row is authoritative even when a legacy message's cached `has_tool_use` flag is false. Archive and mirror repair predicates follow the collab row, and shared-storage normalization does the same.
- Legacy write guards use the same conservative evidence boundaries: encrypted previews and user turns are rejected regardless of relationship metadata, formatted collab headers are derived from the canonical tool list and need neither metadata signal, and PostgreSQL rechecks `has_tool_use` transitions.
- Vector documents get a one-time sweep with a completion marker. A per-ensure scan still catches stale vectors pushed independently of session versions (a compatibility-marked writer can do this); a partial index keeps that scan proportional to the normally empty suspicious set.
- Compatibility probes count only guards enabled for ordinary writes (O or A); disabled and replica-only triggers fail closed, and writable migrations reinstall them.
- Read-only `pg serve` fails closed until a writable current build has repaired and certified the schema.

## CockroachDB

- Same guards, staged install: `autocommit_before_ddl` makes an atomic migration impossible, so triggers install as autocommitted DDL and the repair follows in a serializable retry transaction.
- Because CockroachDB cannot narrow triggers with `UPDATE OF`, its session guard explicitly admits trash/restore updates only when payload and relationship evidence is unchanged; an upgrade-tested `v7` trigger replaces the earlier body.
- Servers without trigger support (pre-24.3) can still use scan mode for repair verification, but direct `--pg` reads and `pg serve` fail closed because their later queries do not share the scan snapshot. Trigger-capable CockroachDB is required to serve the archive.

## DuckDB mirrors over Quack

- DuckDB has no triggers, so reads gate instead of writes: every remote store read embeds the legacy-version check in the same statement, which executes under one remote snapshot — a legacy push cannot land between a separate check and the read it was meant to cover.
- The mirror uses the same v69 repair-then-certify policy as PostgreSQL, including persisted UTF-8/control-byte normalization and preservation of nullable tool inputs. Rows that cannot be certified remain below the watermark, so Quack and local DuckDB reads reject the mirror instead of treating the unrelated v67/v68 parser versions as certification.
- A transient legacy-Codex preflight read failure is counted as a session push error and pins the watermark. The session retries on the next push instead of falling through to a later read that could publish it without durable local certification.
- Guard failures map back to the repair sentinel on every error surface (query, scan, iteration, close), including single-row reads. Two reads that previously bypassed the gate — branch listing and usage session counts — now go through it.
- Startup compat and post-repair checks probe the indexed version watermark instead of running content scans whose outcome the probe already decides; over Quack that is one remote statement instead of five.
- Additively migrated certification columns are read with safe defaults so older compatible writers cannot make typed session reads fail by omitting them.

## Cross-cutting

- Push-time verification (`PrepareSessionForSharedStorage`) applies the same metadata-independent checks as copy-time and mirror certification: a user turn is checked as an inter-agent delivery even when the session was never linked to its parent, and a formatted tool message is checked even when both its cached tool-use flag and collab `tool_calls` row were lost. Sessions that fail are withheld from PG and DuckDB mirrors — logged, skipped without counting as push errors (a permanent condition must not pin the push watermark), and their vector documents are deferred via a durable local-version gate that outlives the incremental push window.
- Certification evaluates the exact `SanitizeUTF8` form that PostgreSQL publishes, including preview, role, message content, collab tool name, and tool input. Preserved-row repair sanitizes before redaction and exhaustively checks every copied legacy Codex session, so control or invalid bytes cannot hide a token until after certification.
- The collab tool list is rendered from `parser.CodexCollabToolsSQL()` at every SQL site, and the watermark is one exported constant (`db.CodexRedactionDataVersion`) aliased by both mirror backends, so neither can drift.
- PostgreSQL and DuckDB push keep the exact message/tool snapshot that passed certification, bind it to the transcript revision, and release snapshots after each bounded batch. Vector documents carry the same revision and defer when it differs from the locked remote session.
- The local vectors.db mirror schema version bumps so stale embeddings are rebuilt.
- Fixes a pre-existing bug that broke every Quack push carrying project identity snapshots (named string types rejected by the remote literal renderer).

## Limitations

- PG trigger bodies are frozen at first install; changing the guard predicates requires renaming the trigger constants so existing schemas reinstall (the session guard is at `_v7`, the message guard at `_v5`, and the tool and vector guards at `_v4`; the DDL drops prior names).
- Uncertified orphan or mirror rows keep their content and stay below the watermark. Local push-time verification withholds them and mirror read gates reject them rather than certifying potentially encrypted content.

## Where to look

- Parser redaction and formatted-block proof: `internal/parser/codex.go` (`RedactCodex*`, `CodexStoredToolContentIsProvablyNonCollab`, `CodexStoredToolContentNeedsCollabRedaction`, `CodexCollabToolHeaderLikePatterns`)
- Orphan copy scrub and certification: `internal/db/orphaned.go`, push-time verification in `internal/db/shared_storage.go`
- PG/CRDB migration, guards, vector sweep: `internal/postgres/codex_redaction.go`, `internal/postgres/crdb.go`
- Quack read gate: `internal/duckdb/codex_redaction.go` (`codexGuardedQuackReadSQL`) and the gated reads in `internal/duckdb/store.go`


mjacobs:
fixing the last findings, will update soon

wesm:
rebasing

wesm:
looking at this, the multiple dataVersion bumps maybe need to be cleaned up, rebasing and doing that

mjacobs:
I'll take a fresh look at this tomorrow. Branch may be too messy at this point, a fresh start may be best.

PR #1133, opened 2026-07-14
Fix project selector crash on empty project names

### LINKED ISSUE #1123: Project selector crashes (each_key_duplicate) when the database contains a project with an empty name (e.g. Antigravity CLI summary-mode sessions)
### agentsview version

agentsview v0.37.5 (commit 8bd41f0b, built 2026-07-10T03:17:25Z)

### Install method

install.sh / install.ps1

### OS / platform

macOS 26.5.2 (arm64)

### Which agent and version

Antigravity CLI (agy) 1.1.1

### Which model(s)

n/a (UI regression; any Antigravity CLI summary-mode session triggers it regardless of model)

Clicking the project selector in the header shows the focus ring but the popover never opens; the control is effectively dead. Expected the dropdown to open and list projects. Worked on v0.34.5 with identical data; broke immediately after upgrading to v0.37.5. Reproduced in a clean incognito profile with all extensions disabled, on two different origins (direct IP and reverse-proxied HTTPS).

The console shows an uncaught Svelte error on every click:

```
Uncaught Error: https://svelte.dev/e/each_key_duplicate
    at index-client-*.js:30 ...
```

#### Root cause

kit-ui's `Typeahead.svelte` renders its options with a keyed each-block keyed on `option.name`:

```svelte
{#each rows as row, i (row.option.name)}
```

`ProjectTypeahead.svelte` prepends an "All projects" option with `name: ""` to the project list:

```ts
const allOption = { name: "", label: m.shared_all_projects(), ... };
```

When the API's project list itself contains a project named `""`, two rows share the key `""` and Svelte 5 throws `each_key_duplicate` mid-render, killing the popover.

On <= v0.36.0 the selector was the local `OptionTypeahead.svelte`, whose each-block was unkeyed (`{#each filtered as option, i}`), so the duplicate was harmless. This is why the regression first appears in v0.36.1 (the #957 migration to shared @kenn-io/kit-ui components).

#### How an empty-name project arises

Google Antigravity CLI (`agy`) sessions parsed in summary mode (plaintext `history.jsonl` / `brain/`, no trajectory sidecar) carry no cwd, so they sync with an empty project name. In my case, 53 `agy` sessions (created via roborev's Antigravity backend, see kenn-io/roborev#951) produced a "(none)" project with `name: ""`, and the selector has crashed on every load since the upgrade. Any database with such sessions reproduces this.

#### Suggested fix

Any of:

- Key the Typeahead rows by something unique (e.g. index, or a caller-supplied `key` falling back to index) in kit-ui;
- Give the injected "All projects" option a sentinel key that cannot collide with a real project name;
- Normalize empty project names at sync/API time (e.g. to a reserved `(none)` name), which would also make the "(none)" bucket selectable/filterable in the UI.

#### Aside: password managers grab the same input

While in that component: the Typeahead's `<input type="text">` carries no `autocomplete` attribute, so password-manager inline menus (Bitwarden confirmed, others likely) classify it as a fillable login/identity field and overlay their menu on the project filter. Adding `autocomplete="off"` plus the opt-out attributes (`data-bwignore`, `data-1p-ignore`) to kit-ui's Typeahead input would fix this for all consumers.

```text

```

### Steps to reproduce

1. Sync at least one session that yields an empty project name (e.g. any Antigravity CLI summary-mode session), or insert a project row with `name = ''`. `agentsview projects` will show a "(none)" entry.
2. Load the web UI and click the project selector in the header.
3. The selector highlights but never opens; the console shows `each_key_duplicate`.

### Checklist

rodboev:
Opening the header project selector crashed with Svelte's `each_key_duplicate` when the database contained a project with an empty name (e.g. Antigravity CLI summary-mode sessions): kit-ui's Typeahead keys options on `option.name`, and the empty-name project collided with the built-in All projects option, which also uses `""`.

Omit empty-name projects from the option list. A row for them was never functional because `""` already means the unfiltered state in the app-wide filter contract, and their sessions stay reachable under All projects. Named projects and the selection contract are unchanged.

Fixes #1123.

### DIFF
--- a/frontend/src/lib/components/layout/ProjectTypeahead.svelte
+++ b/frontend/src/lib/components/layout/ProjectTypeahead.svelte
@@ -19,12 +19,15 @@
   };
 
   const options = $derived.by(() => {
-    const items = projects.map((p) => ({
-      name: p.name,
-      label: `${p.name} (${p.session_count})`,
-      displayLabel: p.name,
-      count: p.session_count,
-    }));
+    // An empty-name project would duplicate the all-projects key, and "" already means unfiltered.
+    const items = projects
+      .filter((p) => p.name !== "")
+      .map((p) => ({
+        name: p.name,
+        label: `${p.name} (${p.session_count})`,
+        displayLabel: p.name,
+        count: p.session_count,
+      }));
     return [allOption, ...items];
   });
 
--- a/frontend/src/lib/components/layout/ProjectTypeahead.test.ts
+++ b/frontend/src/lib/components/layout/ProjectTypeahead.test.ts
@@ -1,0 +1,64 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from "vite-plus/test";
+import { fireEvent, screen } from "@testing-library/svelte";
+import { mount, unmount } from "svelte";
+import type { ProjectInfo } from "../../api/types/core.js";
+import { m } from "../../i18n/index.js";
+import ProjectTypeahead from "./ProjectTypeahead.svelte";
+
+const projects: ProjectInfo[] = [
+  { name: "", session_count: 53 },
+  { name: "repo-a", session_count: 2 },
+];
+
+describe("ProjectTypeahead", () => {
+  let component: Record<string, unknown> | null = null;
+
+  afterEach(() => {
+    if (component) {
+      unmount(component);
+      component = null;
+    }
+    document.body.innerHTML = "";
+  });
+
+  it("opens without crashing when a project has an empty name", async () => {
+    const onselect = vi.fn();
+    component = mount(ProjectTypeahead, {
+      target: document.body,
+      props: { projects, value: "", onselect },
+    });
+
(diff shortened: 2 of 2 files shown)


rodboev:
Doing a broader architectural rework here instead of another round of spot fixes. The recent review rounds kept finding new places where an explicit empty-project filter leaks, and that's a signal this PR drifted well past what #1123 actually reports: the selector popover crashing on a duplicate key when a project has an empty name.

Making `""` a first-class filterable value turned out to be a cross-stack contract change (three storage backends, transports, insights, secrets, usage), which is a bigger design decision than a crash fix should smuggle in. The rework lands the PR back on the reported bug: the selector opens, named projects filter as before, and empty-name projects are omitted from the dropdown, since `""` already means "all projects" in the existing filter contract. Their sessions stay reachable under All projects. That's also the omission alternative roborev suggested back in its second review.

Force-pushing the reworked version shortly, with the title and description updated to match. If end-to-end filterability for empty-name projects is wanted, that seems better done as its own change on top of a presence-carrying API contract.

PR #1135, opened 2026-07-14
Preserve Claude session identity metadata

### LINKED ISSUE #1089: Support custom agent labels for Claude SDK sessions
We run different Claude SDK agents (triage, review, code, fix) via CI and collect their transcripts into AgentsView. Works great for browsing runs, but every session shows up as "claude" — there's no way to tell which agent type ran.

The JSONL already has an `agentSetting` field per session (e.g. `{"type":"agent-setting","agentSetting":"triage"}`), and `entrypoint` distinguishes `cli` from `sdk-cli`. Neither is used for the agent label today.

Two asks:

1. Let `agentSetting` (or a similar field) override the agent display name — e.g. show "fs-triage" instead of "claude" when present.
2. Surface `entrypoint` somewhere (filter, badge, column) so users can distinguish CLI from SDK from web sessions even without a custom label.

rodboev:
Claude session identity can appear on appended SDK transcript lines that are not typed `agent-setting`, so incremental sync could miss the metadata and keep the generic Claude label.

This change treats top-level `agentSetting` and `entrypoint` as session identity metadata wherever they appear on appended Claude rows, forces that incremental path through a full parse, and keeps the existing plain-Claude fallback when metadata is absent. The identity fields still flow through SQLite, PostgreSQL, DuckDB, sidebar payloads, and breadcrumb/session labeling without fragmenting the stable `claude` agent filter.

The added coverage exercises absent metadata, late metadata on appended rows, and cross-backend persistence.

Closes #1089


wesm:
looking at this

wesm:
There is work needed here, I have Fable working on it, I'll push my changes in the morning

wesm:
Pushed a9fc541 addressing two issues found while reviewing this against real Claude CLI transcripts.

**1. The incremental-parse escalation effectively disabled incremental sync for Claude.**

Ordinary Claude Code CLI transcripts carry a top-level `"entrypoint":"cli"` field on most message lines (in a 200-transcript sample: 174 files, median ~80% of lines — assistant, user, attachment, system, and progress entries). `claudeSessionIdentityUpdate` treated any appended line with a non-empty `entrypoint` as an identity edit and returned `ErrClaudeIncrementalNeedsFullParse`, so nearly every append to every live Claude session escalated to a full re-parse of the whole file. The PR's tests didn't catch it because the fixtures don't carry `entrypoint` on ordinary lines.

The fix threads the stored `agent_label`/`entrypoint` from the session row through `IncrementalRequest` into the incremental parser. Identity is first-non-empty-wins across the file, so an append can only change a field whose stored value is still empty — the parser now escalates only in that case, which bounds escalation to at most once per field per session. Coverage:
- parser: append carrying already-stored identity stays incremental; a not-yet-stored field (`agentSetting` appearing for the first time) still escalates
- sync integration: uses the malformed-line count as the observable (the incremental reader skips appended malformed lines without recounting; a full parse re-derives the count), verified to fail when the gate is removed

**2. Every existing Claude session would have grown a "cli" badge.**

Because `entrypoint:"cli"` is near-universal, the data-version resync would have backfilled it everywhere and the sidebar/breadcrumb would render a redundant "cli" tag on nearly every session — noise, given the issue's goal is distinguishing SDK/CI sessions from the default. The new `entrypointBadge` helper suppresses the default "cli" value (and trims whitespace) at render time; the value is still parsed and stored, so filtering or surfacing it later stays possible. Non-default entrypoints like `sdk-cli` render as before.

PR #1136, opened 2026-07-14
Pin recorded models in resume commands

### LINKED ISSUE #1104: Include the session's recorded model in the generated resume command
The resume command agentsview generates doesn't pin the model the session was recorded with, so resuming drops you into whatever model the harness currently has active. It should carry the recorded model flag, e.g. `--model` for Claude Code and `-m` for Codex.

## Why

Neither Claude Code nor Codex re-selects a resumed session's original model on its own, so you silently continue under the wrong one. Codex at least warns:

```
⚠ This session was recorded with model `gpt-5.5` but is resuming with `gpt-5.6-luna`. Consider switching back
  to `gpt-5.5` as it may affect Codex performance.
```

Claude Code says nothing. Pinning the model avoids both the performance drift Codex warns about and a cost hit: prompt caches are per-model, so resuming under a different model re-sends the whole accumulated context at full uncached input rate instead of the cheap cache-read rate.

agentsview already knows the model (`sessions.model`, and per-event `usage_events.model`), so filling the template in is straightforward. When a session used more than one model, which one to pin is the maintainer's call.

## Where

The templates in `internal/server/resume.go` only interpolate the session ID:

```go
var resumeAgents = map[string]string{
    "claude":   "claude --resume %s",
    "codex":    "codex resume %s",
    ...
}
```

So the emitted command is `claude --resume <id>` / `codex resume <id>`, with no model pinned.

## The ask

Fill the recorded model into the command for agents whose CLI takes a model flag:

- `claude --resume <id> --model <model>`
- `codex resume <id> -m <model>`

Leave templates unchanged for agents whose CLI has no model flag.

rodboev:
Resume commands could ignore the model recorded in the session and fall back to the CLI default, which breaks prompt-cache reuse.

This change keeps the backend and frontend resume paths aligned on the same selected model, preserves shell quoting for model values, and matches JavaScript UTF-16 tie-break ordering when multiple models have the same usage count. Backend resume generation now uses compact per-model counts instead of hydrating full transcripts, and the model-aware command is reused for command-only responses, launcher paths, and clipboard fallbacks.

Frontend local fallback pinning now stays behind a store-owned complete-history predicate. The breadcrumb and keyboard fallbacks pin only when the requested session matches a successfully complete, current, idle history; partial pages, later-page failures, and in-flight or failed reloads stay unpinned, while display-only model badges remain unchanged.

Message-point forks and sessions without model data keep their existing behavior.

Closes #1104

### DIFF
--- a/frontend/src/lib/components/layout/SessionBreadcrumb.svelte
+++ b/frontend/src/lib/components/layout/SessionBreadcrumb.svelte
@@ -273,11 +273,7 @@
   );
 
   let resumeModel = $derived(
-    messagesStore.sessionId === session?.id
-      && !messagesStore.loading
-      && !messagesStore.hasOlder
-      ? messagesStore.mainModel
-      : "",
+    session ? messagesStore.resumeModelFor(session.id) : "",
   );
 
   const gradeStyle = $derived(
--- a/frontend/src/lib/components/layout/SessionBreadcrumb.svelte
+++ b/frontend/src/lib/components/layout/SessionBreadcrumb.svelte
@@ -272,6 +272,14 @@
       : "",
   );
 
+  let resumeModel = $derived(
+    messagesStore.sessionId === session?.id
+      && !messagesStore.loading
+      && !messagesStore.hasOlder
+      ? messagesStore.mainModel
+      : "",
+  );
+
   const gradeStyle = $derived(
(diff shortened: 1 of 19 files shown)


wesm:
thank you!

PR #1151, opened 2026-07-15
feat(i18n): add French locale

dpanbug:
Adds a complete French message catalog and wires `fr` through the locale plumbing: inlang project settings, `SUPPORTED_LOCALES`, browser-locale matching (`fr`, `fr-FR`, `fr-CA`, `fr-CH`), and the language picker. `settings_language_french` is added to every existing catalog so the key sets stay identical across locales.

## Terminology

The catalog follows what French-speaking developers actually say rather than the official francisations: `token`, `prompt`, `embedding`, `skill`, `chunk` and `worktree` stay in English, while the surrounding copy is translated. Rendering `token` as *jeton* or `prompt` as *invite* is defensible on paper but reads as machine-translated to the target audience.

## French plural rules — worth a reviewer's attention

French has `one`/`other` like English, but CLDR places **0 in the `one` category**. Zero therefore takes the singular form ("0 appel d'outil") where English says "0 tool calls". Most messages map across unchanged.

One message does not map cleanly, and it is the main thing to look at:

`session_vitals_calls_summary` relies on 0 landing in `other` to hide the running-count suffix. In French 0 lands in `one`, so a faithful copy of the English structure would render "· 0 en cours" whenever nothing is running. The French catalog therefore selects on `runningCount` directly (`0` vs `*`) instead of on a plural category, which keeps the suffix hidden when there is nothing to report. This is the only place where the French variant structure deviates from the base catalog.

While looking at that message: the `runningPlural=one` branch appears to be unreachable in `zh-CN`, `zh-TW` and `ko`, since those locales have no `one` category — so the running count seems to never surface there. I have not touched those catalogs beyond the new language name, and this may well be intentional.

## Limitations

`settings_language_french` required a word for "French" in the existing catalogs (`法语`, `法文`, `프랑스어`). These follow the pattern of the neighbouring language names, but a native reader should confirm them.


wesm:
Coucou et merci !

dpanbug:
You're welcome !

Leuconoe:
`프랑스어` is correct. alt : `불어`

PR #1154, opened 2026-07-15
fix(sync): bound daemon memory and watcher recovery

wesm:
## Why

On macOS, recursive kqueue watching kept state proportional to every file in a long-lived archive. Passive daemon memory therefore grew with archive size, while broad polling and recovery scans could keep CPU and I/O active.

## What changed

- Active recursive Darwin roots use native FSEvents, so passive watcher state scales with configured roots rather than stored files. Callback queues retain the existing path and byte limits.
- Discovery and authoritative reconciliation stream through disk-backed pages. Provider enumeration, duplicate selection, ordered aggregation, Trae workspace discovery, and Hermes profile/member discovery no longer require archive-sized Go maps or repeated value concatenation.
- Watcher overflow, directory lifecycle events, fallback coverage, and polling retain bounded retry scopes where correctness permits. Non-Darwin fsnotify coverage applies the same watch budget to newly created and moved-in subtrees.
- Missing-source deletion is conservative and recoverable. Only complete machine-local provider scopes establish deletion proof, including virtual members in multi-session containers, and scheduled reconciliation defers scopes whose physical roots are unavailable. Tombstones durably invalidate base and hash-qualified skip entries before hiding a session, and audit workers refresh the daemon cache so byte-identical restored files are parsed and revived across workers and restarts.
- Source revival replaces stale content before becoming visible, and provider-specific discovery preserves canonical identities across symlinks, overlapping roots, expanded Hermes profile archives, and mixed filesystem/SQLite layouts.

## Failure behavior

Root-local discovery failures may still yield valid sessions and do not block later configured roots. Those partial candidates can be stored, but the failed provider scope cannot add baselines or tombstone missing sessions; independent healthy providers remain authoritative. Callback, cancellation, temporary-store, and cleanup failures stop immediately, and the watcher retains the affected roots for retry.

True event loss or ownerless recovery can still require global authoritative reconciliation. Runtime watcher fallback and low-frequency reconciliation remain available for filesystems without reliable native events.

Focused macOS FSEvents CI runs only on trusted pushes to `main` through the self-hosted runner group.

### DIFF
--- a/.github/workflows/ci-macos-main.yml
+++ b/.github/workflows/ci-macos-main.yml
@@ -1,0 +1,37 @@
+name: CI (macOS)
+
+on:
+  # Persistent self-hosted runners must never be selectable from a workflow
+  # definition controlled by a pull request. The ci-runners group must allow
+  # this exact workflow path only from refs/heads/main.
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test-macos-fsevents:
+    name: Go Test (macOS FSEvents)
+    runs-on:
+      group: ci-runners
+      labels: [self-hosted, macOS, ARM64]
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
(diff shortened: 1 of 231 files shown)


wesm:
I upgraded locally and am going to monitor background memory use for a few hours. I noticed performance regressions on resyncing so looking at that now

wesm:
Working more on this since the memory use is still not where I'd like it

wesm:
This has been holding solid so I'm going to merge it once roborev approves

wesm:
This is close, but it still has some full sync performance regressions versus main

wesm:
Declined. Merging this when CI is green

PR #1156, opened 2026-07-16
feat(frontend): guided semantic-search setup in the command palette

mjacobs:
When a Semantic or Hybrid search fails in the command palette because vector search is not serving, the palette currently surfaces the raw backend error string ("semantic search not available: enable [vector] in config.toml and run 'agentsview embeddings build'"). This replaces that with a state-aware setup panel.

On a non-fulltext 501, the search store classifies the failure as a distinct `semantic-unavailable` kind and the palette mounts a new `SemanticSetupHelp` panel, which probes `GET /api/v1/embeddings/status` to choose guidance:

- **`[vector]` not configured** (status probe returns the generic 501): a short walkthrough with a copyable minimal `config.toml` snippet (mirroring the docs' Ollama example) and a copyable `agentsview embeddings build` command, plus a restart note.
- **Configured but no index built**: a "Build embeddings" button wired to the existing `POST /api/v1/embeddings/build`, with live progress polling (scanning phase, then percent embedded). When the build completes, the failed search re-runs automatically. A 409 from an already-running build (CLI or scheduler) is watched rather than surfaced as an error. When the search's 501 carries a more specific reason than the generic base message (stale index after an embedding-config change, vectors.db schema mismatch), that reason is shown verbatim alongside the button.
- **Vector serving disabled for this daemon run** (a specific 501 reason, e.g. `vectors.write.lock` held at startup): the daemon's reason is shown verbatim, since config guidance would be misleading.

A full config-from-the-UI flow was deliberately left out: enabling `[vector]` requires a model, dimension, and embeddings endpoint only the user knows, and the daemon wires vector serving at startup — so the panel guides the config half and automates the build half.

On the build route, `include_automated` is now tri-state: a request omitting the field builds with the daemon's configured `[vector].include_automated` scope (new `WithEmbeddingsIncludeAutomatedDefault` server option), matching how scheduler and CLI builds resolve it, instead of silently defaulting to false; an explicit value overrides the config for that build. The CLI's daemon client correspondingly always sends the field (its previous `omitempty` serialization would have dropped an explicit `--include-automated=false` override against a daemon configured true).

503 transient encode failures keep the existing generic/timeout presentation with retry; only 501 gets the setup panel. Full-text 501s stay generic.

New message keys are added to all five locales. Reviewers may want to look at the 501-reason discrimination in `SemanticSetupHelp.svelte` (string comparison against the server's generic "embeddings manager not available" default) — a machine-readable error code would be cleaner, but the current API shape only carries a message.


mjacobs:
looking at these issues now

wesm (approved):

PR #1157, opened 2026-07-17
feat(parser): add Omnigent session support

### LINKED ISSUE #1145: New agent support: Omnigent (open-source meta-harness, SQLite chat.db)
### Agent name and link

Omnigent (open-source meta-harness) — https://github.com/omnigent-ai/omnigent (Apache-2.0)

It orchestrates other coding agents (Claude Code, Codex, OpenCode, Pi, Cursor, Hermes, ...) under one harness.

### Where does it store sessions?

A single SQLite database at `{OMNIGENT_DATA_DIR|~/.omnigent}/chat.db` (WAL mode), same path on all platforms. The location can be overridden with `OMNIGENT_DATABASE_URI` (which may point at a non-SQLite backend — out of scope for a first parser).

### File format

Not per-session files — one SQLite DB, most similar to the existing Shelley/Kiro/ZCode providers:

- `conversations` = one row per session, with the sub-agent hierarchy (`parent_conversation_id`, `root_conversation_id`, `kind` default/sub_agent, `sub_agent_name`), `workspace`, `git_branch`, `model_override`, and `session_usage` (token/cost).
- `conversation_items` = one row per transcript event (`message`, `function_call`, `function_call_output`, `reasoning`, `compaction`, `error`, `slash_command`, `terminal_command`, ...), with plaintext-JSON `data` shaped by `type`, ordered by a monotonic `position`. FTS5 is already present.

Two things worth flagging for a parser:

1. **The schema has multiple generations.** Older builds use one `conversations` table with VARCHAR enum columns; newer builds encode `type`/`kind` as SMALLINT codes and split session metadata into `omnigent_conversation_metadata` (+ `agent_configuration`). The `data` JSON shape is identical across generations, so feature detection on the table/column shape handles both. A deployment whose metadata lives in a separate physical database can't be read from `chat.db` alone and should be skipped rather than parsed partially.
2. **Because it's a meta-harness, one parser covers every orchestrated sub-agent** — the transcript is normalized into the same `conversation_items` schema regardless of which underlying harness produced it, so new sub-agents need no parser changes.

### Sample session file

A real `chat.db` is a large single database of private transcripts, so rather than attach one I can provide a synthetic, self-contained fixture in the `kiro_sqlite` style (a schema-const DDL + one small JSON payload per item type), which is what a test would build against anyway. Happy to share a redacted excerpt if useful.

### Can you help?

- [x] I can share a sample session file (synthetic fixture)
- [x] I use this agent and can test a PR or build a parser

I already have a working branch — a multi-session container provider modeled on the Shelley parser (one `chat.db` fanned out into one session per conversation), with the dual-generation schema adapter above, tool-call folding, reasoning, sub-agent hierarchy with cwd/branch inheritance, and session usage/cost (via the existing `klauspost/compress` zstd dep). It's tested (unit + a cardinality-scaling regression) and verified end-to-end with `agentsview sync`.

Filing this first per the issue-first norm to confirm the approach before opening the PR — in particular whether the dual-generation feature-detection adapter and the "skip when metadata isn't co-located" behavior match how you'd want it handled. Will open the PR referencing this issue.

riordan:
Adds Omnigent session ingestion from `~/.omnigent/chat.db`, creating one AgentsView session per conversation. Closes #1145.

The parser supports the older single-table schema and newer co-located split-table schema. It preserves parent, continuation, and sub-agent relationships; workspace and git metadata; transcript events; tool calls; reasoning; compaction events; and session-level token and cost usage. Split-schema identity includes `workspace_id`, so reused conversation IDs cannot collide across workspaces.

SQLite watcher and periodic sync paths resolve conversations through a bounded `updated_at` fast lane, resumable row cursors, and paged stored-source hints. The tracker retains only scalar cursor state per container, caps each discovery batch at 32 sources, and performs cold-start, deletion, and failure recovery over resumable pages rather than materializing the archive. Whole-container cache entries include the parser data version, and SQLite shared-memory coordination changes do not trigger content probes.

Unsupported or non-co-located schemas are skipped non-destructively so archived sessions remain intact. Configuration details live in the Zensical documentation; the README supported-agents table lists only the default database path.

Non-SQLite backends and deployments whose metadata is stored in a separate physical database are not parsed.


wesm:
Looks like this needs a little work. I'll take the wheel from here

wesm:
Stopping the agent. May need a new approach here

riordan:
Oh that's way better. I feel like I should've seen the container framework was already there. Way better approach you're taking with #1167.

PR #1159, opened 2026-07-17
feat(recall): extraction generation store, client, and prompt profiles

wesm:
Second slice of the automatic recall extraction pipeline, after the #1158 segmenter: the persistence layer that makes long extraction runs safe to interrupt, and the model-facing client that turns one unit into entries.

**Generation registry** (`recall_extract_generations`) — an extraction run is keyed by a fingerprint digesting everything that changes its output: the model identity (name plus optional deployment label), segmentation strategy and parameters, resolved prompt contents, request shape including the output token budget, and a protocol version covering the response schema and recovery behavior baked into the binary. Same fingerprint → interchangeable corpora; any difference → a new generation, so reconfiguration never mixes entries produced under different settings. States are `building → active → retired`, with activation and retire performed as single atomic statements (no check-then-update races), and `EnsureExtractGeneration` idempotent so a restarting daemon re-registers safely.

**Resumable progress store** (`recall_extract_progress`) — one row per (session, generation) carrying a unit cursor, unit total, and content digest of the session's derived units. Because unit derivation and candidate ids are deterministic, a daemon killed mid-session resumes from the cursor and re-upserts at worst one unit's entries. If a session's content digest changes (new messages arrived), the upsert resets the cursor instead of silently extracting against stale unit boundaries; a session with zero units lands directly in done. Cursor advancement requires the caller's expected digest and enforces monotonic, bounded cursors; failure marking additionally requires the cursor the worker observed, and a completed row can never be demoted — a worker that lost any race gets a typed `ErrStaleExtractProgress` instead of corrupting the row. Every mutation holds the database write mutex so connection swaps during resync quiesce cleanly, and full resync carries both tables across via the existing attached-copy path, tolerating older archives that predate them.

**Extraction client** (`DistillWithRecovery`) — one unit per call through an OpenAI-compatible chat endpoint with strict JSON-schema decoding, enforced again client-side (exact keys, no nulls, typed fields, closed type enum) so a server that ignores `json_schema` cannot advance progress over silently lost entries. Failures are classified: network errors, timeouts, rate limits, and server errors retry with exponential backoff honoring `Retry-After`; permanent rejections fail fast; an HTTP 400 whose body names a context-length problem is a typed `ErrContextOverflow`; truncated output surfaces immediately as `ErrPersistentTruncation`. Both typed errors mean "split this unit" — splitting belongs to the caller, which owns unit identity, and preserves every entry. There is deliberately no retry that caps the entry count: a capped response looks complete while silently dropping entries. Token usage is summed across every attempt, successful or not.

**Prompt profiles** — three layers resolved most-specific-first: embedded base prompts per unit role (zero config works), built-in model-family profiles bundling prompt variants with request shape (the `qwen` profile disables hidden reasoning via template arguments — under constrained decoding it otherwise consumes the entire token budget before any content), and per-role override files (`intent.txt`/`action.txt`/`generic.txt`; an empty file is an error rather than a silently blanked prompt). Resolutions are deep copies, so caller mutation cannot corrupt the registry.

Nothing schedules extraction yet; the manager that wires segmenter → client → store together, plus CLI and configuration, is the next slice.

### DIFF
--- a/internal/db/recall.go
+++ b/internal/db/recall.go
@@ -234,6 +234,9 @@ func (db *DB) CopyRecallEntriesFrom(sourcePath string) error {
 	if err := copyRecallQueryEventsFromAttachedTx(ctx, tx); err != nil {
 		return err
 	}
+	if err := copyRecallExtractStateFromAttachedTx(ctx, tx); err != nil {
+		return err
+	}
 
 	res, err := tx.ExecContext(ctx, `
 		INSERT OR IGNORE INTO recall_entries (
--- a/internal/db/recall_extract.go
+++ b/internal/db/recall_extract.go
@@ -387,13 +387,24 @@ func (db *DB) AdvanceExtractCursor(
 	)
 }
 
+// ExtractFailure identifies the exact progress state a worker observed
+// when it failed, so a stale worker cannot demote fresher state.
+type ExtractFailure struct {
+	SessionID      string
+	Fingerprint    string
+	ExpectedDigest string
+	ExpectedCursor int
+	LastError      string
+}
+
 // MarkExtractProgressFailed records a failure without losing the resume
 // point: the cursor stays where it was so a retry continues, not restarts.
-// A row that already reached done is left untouched: the failure comes from
-// a worker that raced completion, not from the completed extraction.
+// The update applies only when the stored digest, cursor, and non-done
+// state all match what the failing worker observed; anything else means
+// another worker moved the row on, and the failure is reported as
+// ErrStaleExtractProgress instead of demoting newer progress.
 func (db *DB) MarkExtractProgressFailed(
-	ctx context.Context,
-	sessionID, fingerprint, expectedDigest, lastError string,
+	ctx context.Context, failure ExtractFailure,
 ) error {
 	if err := db.requireWritable(); err != nil {
 		return err
@@ -409,45 +420,56 @@ func (db *DB) MarkExtractProgressFailed(
 			updated_at = strftime('%Y-%m-%dT%H:%M:%fZ','now')
 		WHERE session_id = ? AND generation_fingerprint = ?
 		  AND content_digest = ?
+		  AND unit_cursor = ?
 		  AND state != ?`,
-		ExtractProgressFailed, lastError, sessionID, fingerprint,
-		expectedDigest, ExtractProgressDone,
+		ExtractProgressFailed, failure.LastError,
+		failure.SessionID, failure.Fingerprint,
+		failure.ExpectedDigest, failure.ExpectedCursor,
+		ExtractProgressDone,
 	)
 	if err != nil {
 		return fmt.Errorf(
 			"marking extract progress failed for session %s: %w",
-			sessionID, err,
+			failure.SessionID, err,
 		)
 	}
 	affected, err := result.RowsAffected()
 	if err != nil {
 		return fmt.Errorf(
 			"marking extract progress failed for session %s: %w",
-			sessionID, err,
+			failure.SessionID, err,
 		)
 	}
 	if affected > 0 {
 		return nil
 	}
-	progress, ok, err := db.ExtractProgress(ctx, sessionID, fingerprint)
+	progress, ok, err := db.ExtractProgress(
+		ctx, failure.SessionID, failure.Fingerprint,
+	)
 	if err != nil {
 		return err
 	}
 	if !ok {
 		return fmt.Errorf(
 			"no extract progress row for session %s under generation %s",
-			sessionID, fingerprint,
+			failure.SessionID, failure.Fingerprint,
 		)
 	}
 	if progress.State == ExtractProgressDone {
 		return fmt.Errorf(
 			"session %s is already done under generation %s: %w",
-			sessionID, fingerprint, ErrStaleExtractProgress,
+			failure.SessionID, failure.Fingerprint, ErrStaleExtractProgress,
+		)
+	}
+	if progress.ContentDigest != failure.ExpectedDigest {
+		return fmt.Errorf(
+			"failing session %s past digest change: %w",
+			failure.SessionID, ErrStaleExtractProgress,
 		)
 	}
 	return fmt.Errorf(
-		"failing session %s past digest change: %w",
-		sessionID, ErrStaleExtractProgress,
+		"failing session %s behind cursor %d: %w",
+		failure.SessionID, progress.UnitCursor, ErrStaleExtractProgress,
 	)
 }
 
--- a/internal/db/recall_extract.go
+++ b/internal/db/recall_extract.go
@@ -1,0 +1,480 @@
+package db
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// Extraction generation states. A generation is one distillation
+// configuration's corpus; at most one is active at a time.
+const (
+	ExtractGenerationBuilding = "building"
+	ExtractGenerationActive   = "active"
+	ExtractGenerationRetired  = "retired"
+)
+
+// Extraction progress states for one (session, generation) pair.
+const (
+	ExtractProgressPending = "pending"
+	ExtractProgressPartial = "partial"
+	ExtractProgressDone    = "done"
+	ExtractProgressFailed  = "failed"
+)
+
+// ErrStaleExtractProgress reports a cursor or failure update that no longer
+// matches the stored row: the session's content digest changed (the row was
+// reset for re-extraction) or the cursor would regress. Workers treat it as
+// "re-read the row and re-derive units", never as data loss.
+var ErrStaleExtractProgress = errors.New("extract progress is stale")
+
+// ExtractGeneration is one row of the extraction generation registry.
+type ExtractGeneration struct {
+	Fingerprint string `json:"fingerprint"`
+	State       string `json:"state"`
+	Model       string `json:"model"`
+	Segmenter   string `json:"segmenter"`
+	ParamsJSON  string `json:"params_json"`
+	CreatedAt   string `json:"created_at"`
+	UpdatedAt   string `json:"updated_at"`
+}
+
+// ExtractProgress is the resume state for one session under one generation.
+// UnitCursor counts completed units of the session's deterministic unit
+// list; a restart resumes at the cursor instead of re-extracting.
+type ExtractProgress struct {
+	SessionID             string `json:"session_id"`
+	GenerationFingerprint string `json:"generation_fingerprint"`
+	UnitCursor            int    `json:"unit_cursor"`
+	UnitsTotal            int    `json:"units_total"`
+	State                 string `json:"state"`
+	ContentDigest         string `json:"content_digest"`
+	LastError             string `json:"last_error,omitempty"`
+	UpdatedAt             string `json:"updated_at"`
+}
+
+// EnsureExtractGeneration registers a generation if its fingerprint is new
+// and returns the stored row. An existing fingerprint wins: the caller's
+// metadata is ignored so a re-registration can never mutate a corpus's
+// recorded identity.
+func (db *DB) EnsureExtractGeneration(
+	ctx context.Context, gen ExtractGeneration,
+) (ExtractGeneration, error) {
+	var zero ExtractGeneration
+	if err := db.requireWritable(); err != nil {
+		return zero, err
+	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	gen.Fingerprint = strings.TrimSpace(gen.Fingerprint)
+	if gen.Fingerprint == "" {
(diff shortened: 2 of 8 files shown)


wesm:
All three findings addressed in the follow-up commit:

1. **Resync loses extraction state** — `CopyRecallEntriesFrom` now carries `recall_extract_generations` and `recall_extract_progress` (progress filtered to sessions present in the rebuilt DB, same rule as entries), with a table-existence guard so archives from releases without the tables still resync. Regression tests cover both the carryover (active state, params, cursor, dropped-session filtering) and the older-archive case.

2. **Stale workers overwriting reset progress** — `AdvanceExtractCursor` and `MarkExtractProgressFailed` now take the expected content digest and update only when it matches, the cursor is monotonic, and it stays within `units_total`; a racing worker gets a typed `ErrStaleExtractProgress`. Tests cover stale digest, cursor regression, and out-of-bounds advancement.

3. **Retire/activate race** — `RetireExtractGeneration` is now a single conditional `UPDATE` (`WHERE fingerprint = ? AND (force OR state != 'active')`), so the check and the mutation are one atomic statement; the refusal message is derived after the fact for the zero-rows case.

wesm:
All five review findings on `af9fbc1` are addressed in `5c87bff`:

- **Fingerprint omissions** — `max_tokens` moved into `RequestShape` (so it feeds the fingerprint through the request field), and a new `extractionProtocolVersion` constant covering the response schema and recovery behavior is digested into the identity. Built-in profiles now carry a working `max_tokens` default, and the client refuses a non-positive budget.
- **400 misclassification** — `ErrContextOverflow` now requires the response body to name a context/input-length error; any other 400 is a permanent error that returns immediately without burning the retry budget or triggering unit splits.
- **Compact retry losing entries** — the compact retry is gated behind `CompactFloorChars`: a truncated unit large enough to split surfaces `ErrPersistentTruncation` immediately (splitting preserves every entry), and only below-floor units, where splitting cannot help, accept the entry-capped compact retry.
- **done→failed demotion** — `MarkExtractProgressFailed` skips rows already in `done` and classifies the refusal as `ErrStaleExtractProgress`; the completed cursor and empty `last_error` stay intact.
- **Missing `db.mu`** — all six mutation APIs now hold the write mutex, matching the other write paths `CloseConnections` relies on to quiesce writes; a test verifies each mutation blocks while the mutex is held.

wesm:
The three findings on `5c87bff` are addressed in `6ea96fb`:

- **URL join** — `BaseURL` is trimmed of trailing slashes before appending `/chat/completions`; the test server now rejects any other path, and a trailing-slash test covers the round trip.
- **Byte vs rune floor** — the compact floor and the unit-size diagnostics now count code points (`utf8.RuneCountInString`), matching how segmentation budgets measure text; a test with three-byte runes verifies a 40-rune/120-byte unit still gets the compact retry under a 64-rune floor.
- **Overflow classification** — `ErrContextOverflow` now requires a specific input-length phrasing (`context_length_exceeded`, "maximum context length", "prompt is too long", "too many tokens", ...); a 400 that merely mentions "context" in an unrelated message is a permanent error. A phrasing the list misses fails the unit with the server's message intact, which is the recoverable direction of that tradeoff.

wesm:
The three findings on `6ea96fb` are addressed in `0df6655`:

- **Compact floor outside the fingerprint** — `CompactFloorChars` moved from `Client` into `RequestShape`, the one struct the fingerprint digests, so changing it produces a new generation instead of mixing capped-compact and split corpora under the same fingerprint. `Client` now has no output-affecting field outside `Request`.
- **Zero-unit sessions stuck pending** — `UpsertExtractProgress` stores a zero-unit session as `done` (both on insert and on a digest reset to zero units) and refuses negative totals.
- **Discarded usage** — `DistillWithRecovery` sums usage across every attempt and returns the total on success and on terminal failure; `distill` reports the server's parsed usage alongside truncation, empty-content, and entry-parse errors. Tests assert the accumulated totals on both the compact-retry success path and the persistent-truncation failure path.

wesm:
Both findings on `0df6655` are addressed in `ad4fd87`:

- **Reserved `ExtraBody` keys** — belt and suspenders: `DistillWithRecovery` rejects extra-body entries naming a client-owned field (`model`, `messages`, `max_tokens`, `temperature`, `response_format`) with an error identifying the key, and the payload now merges extras *before* assigning the owned fields, so a shadow is structurally impossible even if validation were bypassed. Rejecting loudly (rather than silently ignoring the override) keeps a misconfigured profile visible.
- **Retry policy** — failures are now classified: network errors, unreadable responses, HTTP 408/429/5xx, and malformed 200 bodies are transient and retry with exponential backoff (seeded by a configurable `RetryBackoff`, honoring `Retry-After` in both seconds and HTTP-date form, capped at 30s, cancellable via context); every other status — 401/403/404/422 and the rest — fails fast on the first attempt, as do deterministic outcomes like empty content at temperature zero. Tests cover the 5xx→429→success retry chain, fail-fast on 401, no-retry on empty content, and `Retry-After` parsing.

wesm:
The finding on `ad4fd87` is addressed in `e956cdd`: `ResolveProfile` now returns a deep copy — `MatchPrefixes` and `Prompts` cloned, `ExtraBody` recursively duplicated through its nested JSON-shaped maps and slices — so merging configuration into a resolved profile can never write through to the built-in registry or race concurrent resolutions. A mutation-isolation test alters the prefixes and both extra-body levels of one resolution and verifies the next resolution is untouched.

wesm:
The finding on `e956cdd` is addressed in the latest commit: the client now enforces the same constraints it requests via `json_schema` — the `entries` array must be present, unknown fields are rejected (`DisallowUnknownFields`, including trailing data), and each entry needs a type from the shared enum, a non-blank title and body, and an entities array. An explicit `"entries": []` remains valid, distinct from a missing array. Violations fail fast with a message pointing at the server's schema enforcement, since at temperature zero they are deterministic. The type enum is now a single shared slice feeding both the request schema and the validator. Table-driven tests cover `{}`, `null`, a top-level array, unknown types, blank title/body, missing entities, and unknown fields.

wesm:
The finding on `5b97c15` is addressed in the latest commit. `parseEntries` no longer relies on struct decoding (whose case-insensitive key matching, null-to-zero-value conversion, and `decoder.More()` blind spot caused exactly the holes described); it walks raw messages instead:

- objects are checked against exact, case-sensitive key sets — `"Entries"` is now an unknown key, not a match;
- `null` is rejected explicitly at every level (top-level, entries array, entry, string field, entity element), since `json.Unmarshal` treats it as a no-op rather than a type mismatch;
- entity elements must be strings — `[1]` and `["a", null]` both fail;
- whole-value `json.Unmarshal` rejects trailing data, closing the `{"entries":[]}]` case `decoder.More()` missed.

The request schema now also carries `minLength: 1` on title and body so the constraint is expressed server-side too, and the extraction protocol version bumps to 2 for that schema change. The violation table grows to twelve cases including key casing, nulls at each level, non-string entities, and trailing delimiters.

wesm:
Both findings on `197b639` are addressed in `40cc97f`:

- **Stale progress after a shrinking reset** — `AdvanceExtractCursor` now checks the digest mismatch before the cursor bounds, so a stale worker whose cursor exceeds the shrunken unit total gets `ErrStaleExtractProgress` (re-read and re-derive) instead of a bounds error. The regression test resets 10 units down to 4 and advances a stale cursor of 8.
- **Deployment identity** — `Fingerprint` now takes a `ModelIdentity` (model name plus an optional deployment label) and digests both, with a test that a label change produces a new fingerprint. I deliberately did not hash `BaseURL` itself: an endpoint move (new address, port, or hostname) does not change the model's output, and URL-keyed fingerprints would orphan the entire corpus on every such move. The label lets configuration distinguish servers that expose different weights under the same model name — that wiring lands with the manager slice, which owns configuration.

wesm:
The finding on `40cc97f` is addressed in the latest commit: overflow classification now requires either a structured error code (`context_length_exceeded`, `exceed_context_size_error`) or a length subject (context, prompt, input, token) *paired with* an explicit overflow term (exceed, too long, too large, maximum). Messages that merely mention the subject — "context window must be an integer", "invalid input length parameter" — are permanent errors, not splits. A direct table test covers five overflow phrasings across server implementations and five non-overflow validation messages.

wesm:
The finding on `a2600d3` is addressed in the latest commit: bare "token" is no longer an overflow subject, so output-budget validation errors ("max_tokens exceeds the maximum allowed value", "max_new_tokens exceeds the model limit") classify as permanent instead of triggering input splits. Every genuine overflow phrasing across server implementations also names the prompt, input, or context, so those remain the subjects; the structured-code matches are unchanged. Both max-tokens messages are in the non-overflow regression table.

wesm:
All three findings on `b9f35c8` are addressed in `454c591`:

- **Compact retry losing entries** — resolved by removing the compact retry entirely rather than adding another guard. Truncation now always surfaces `ErrPersistentTruncation`; the caller splits (preserving every entry), and a unit already at the split floor fails with an error saying to raise `max_tokens`. No code path can return a capped-but-"complete" response anymore. `CompactFloorChars` leaves `RequestShape` and the protocol version bumps to 3 for the recovery-behavior change.
- **Stale worker demoting advanced progress** — `MarkExtractProgressFailed` now takes an `ExtractFailure` struct carrying the cursor the worker observed; the update predicate requires digest, cursor, and non-done state to all match, and the refusal is classified (`done`, digest change, or cursor advanced) as `ErrStaleExtractProgress`. Regression test: worker B advances to 2, worker A fails at observed cursor 1, row stays partial at 2.
- **Prefixed output-budget validation** — messages naming `max_tokens`/`max_new_tokens`/`max_completion_tokens` are only overflow when they also name the context length explicitly, so TGI-style "Input validation error: max_tokens exceeds the maximum allowed value" fails fast instead of triggering input splits. The regression table covers the prefixed form. The remaining heuristic is biased toward the safe direction: an unrecognized true overflow fails the unit with the server's message intact rather than looping on splits.

The PR body is updated to describe the simplified recovery ladder.

wesm:
Both findings on `454c591` are addressed in `5d39db1`:

- **Equal-cursor replay resurrecting failures** — `AdvanceExtractCursor` now requires a strictly advancing cursor in its predicate; an equal-cursor replay is classified as an accepted no-op (it completed nothing new), so a delayed duplicate can no longer flip a `failed` row back to `partial` or erase `last_error`. Regression test: advance to 2, mark failed, replay advance(2), assert the row is still failed with its error intact.
- **"input" + overflow term misclassification** — bare "input" is no longer an overflow subject, since "Input validation error:" prefixes arbitrary parameter rejections. Length-specific input phrasings ("input is too long/large") match on their own; otherwise only context/prompt pair with overflow terms. Regression cases: the TGI-prefixed temperature rejection and "invalid input: repetition_penalty exceeds maximum".

wesm:
Both findings on `5d39db1` are addressed in the latest commit:

- **Orphaned overflow phrasings** — "input length" and "input tokens" are now overflow subjects when paired with an overflow term, so "input length 5000 exceeds maximum 4096" classifies as recoverable again. The pairing requirement keeps "invalid input length parameter" (no overflow term) and "Input validation error: temperature ..." (no length-qualified subject) permanent. Both phrasings joined the overflow side of the regression table, which now runs seven overflow against fourteen non-overflow messages.
- **Non-stop finish reasons** — only `finish_reason: "stop"` is accepted as complete. `length` remains the typed truncation signal, and every other reason (`content_filter`, etc.) is a fail-fast error naming the reason — a filtered response can carry valid JSON with fewer entries, and advancing progress on it would silently lose facts. Test covers a content-filtered response with otherwise-valid entries.

wesm:
The finding on `940edc2` is addressed in the latest commit. The classifier is now an explicit evidence order:

1. structured codes;
2. input-side length evidence — standalone phrases ("input/prompt is too long/large") or length-qualified subjects ("input length", "input tokens", "prompt length", "prompt tokens") paired with an overflow term — classifies as overflow **before** any suppression, so "input tokens plus max_tokens exceed the model maximum" is recoverable;
3. only then does an output-budget parameter name suppress: "max_tokens (5000) exceeds the context window (4096)" is a configuration error splitting cannot fix, and is in the regression table as non-overflow;
4. bare "context" paired with an overflow term remains the fallback for the common server phrasings.

Bare "prompt" joined bare "input" as a non-subject, so "prompt_logprobs exceeds maximum allowed value" no longer triggers splits. Both examples from the finding are in the table, which now runs eight overflow against sixteen non-overflow messages.

wesm:
Both findings on `0975eef` are addressed in the latest commit:

- **Missing phrases** — "prompt is too large" and "prompt too large" joined the standalone input-length evidence, with both in the overflow side of the regression table (now ten overflow vs sixteen non-overflow).
- **Choiceless-response usage** — the body parsed, so its reported usage is real cost: it now rides along with the transient error, and the accounting sums it across the retry. Test scripts a choiceless 200 carrying usage followed by a success and asserts the combined total.

PR #1160, opened 2026-07-17
feat(recall): extraction client with recovery ladder and model prompt profiles

wesm:
Third piece of the automatic recall extraction pipeline, after #1158 (segmenter) and #1159 (registry + progress store). This adds the model-facing half: the client that turns one unit into entries, and the prompt/profile layer that makes model differences configuration instead of code.

**Client (`DistillWithRecovery`)** — one unit per call through an OpenAI-compatible chat endpoint with strict JSON-schema decoding, so malformed output becomes a server-side sampling constraint rather than a client-side parse error. The recovery ladder matches how these servers fail in practice:
- transient errors retry up to a bound;
- truncated output (`finish_reason: length`) gets exactly one compact retry — at temperature zero a same-input retry is deterministic, so the retry changes the input instead;
- a context overflow (HTTP 400) or truncation that survives the compact retry returns a typed error (`ErrContextOverflow` / `ErrPersistentTruncation`) meaning "split this unit" — splitting belongs to the caller, which owns unit identity. `SplitFloorChars` bounds that recursion;
- empty content with a normal finish reason is an explicit error pointing at the request shape (the token budget went somewhere invisible), never silently zero entries.

**Prompt profiles** — three layers resolved most-specific-first: embedded base prompts per unit role (intent/action/generic, so zero config works), built-in model profiles bundling prompt variants with request shape (the `qwen` profile, matched by model-name prefix, disables hidden reasoning via template arguments — under constrained decoding it otherwise consumes the entire token budget before any content), and per-role override files (`intent.txt`/`action.txt`/`generic.txt`; an empty file is an error rather than a silently blanked prompt).

**Fingerprint** — digests model, segmenter identity and parameters, resolved prompt contents, and request shape into the identity the generation registry (#1159) keys corpora by. Editing a prompt file or switching models yields a new fingerprint, so reconfiguration creates a new comparable generation instead of mixing entries produced under different settings.

Nothing calls the package yet; the extraction manager and configuration wiring are the next slice.

### DIFF
--- a/internal/recall/extract/client.go
+++ b/internal/recall/extract/client.go
@@ -1,0 +1,262 @@
+package extract
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"maps"
+	"net/http"
+	"time"
+)
+
+// ErrContextOverflow reports a prompt the server rejected as too large for
+// the model context. The unit must be split before retrying; retrying the
+// same text can never succeed.
+var ErrContextOverflow = errors.New("unit text overflows the model context")
+
+// ErrPersistentTruncation reports output that stayed truncated even after
+// the compact retry. Like an overflow, the only recovery is splitting the
+// unit into smaller pieces.
+var ErrPersistentTruncation = errors.New(
+	"model output stays truncated at the token budget",
+)
+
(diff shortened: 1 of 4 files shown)


wesm:
Folded into #1159 — one PR for the whole slice.

PR #1166, opened 2026-07-17
feat(data): first-class Data mode for project inventory and reclassification

wesm:
## Summary

Adds a top-level Data mode for reviewing project identity across an archive and managing folder-to-project mappings. It replaces the mapping editor in Settings with an inventory, a project workspace, and centralized rules that can be previewed and applied atomically.

## Behavior and correctness

- Keys inventory rows by opaque project identity, so sanitized-label collisions remain distinct and every deep link and reclassification targets the exact project.
- Presents observed folders in a compact workspace suited to projects with many source directories. Unknown identities appear as Unclassified, and filesystem roots are not suggested as automatic mappings.
- Binds reclassification previews to the normalized draft, collision state, mapping revision, and exact affected-session snapshot. Changed state returns a conflict instead of applying a stale preview.
- Reconciles identity evidence when sessions move between projects while preserving trustworthy historical Git evidence for missing sources and orphaned sessions.
- Emits project/session changes after mutations and refreshes Data, Sessions, and Activity metadata across clients. Background rule refreshes update the list without discarding unsaved edits.
- Keeps mapping mutations local to writable SQLite archives. PostgreSQL and DuckDB expose the same inventory and governance data as read-only views.

## Publication and storage

Filtered PostgreSQL and DuckDB publication keeps mappings, snapshots, provenance, and tombstones within each configured project scope. Incremental watch work is bounded by changed sessions, project moves publish in both directions, placeholder snapshots are excluded, and mirrored sessions retain their source machine, file path, and exact database generation.

The SQLite data version advances from 76 to 77 to reparse project-identity source snapshots while preserving live and orphaned archive data through non-destructive migration. PostgreSQL gains source-scoped publication metadata and session provenance fields. DuckDB schema v8 adds source archive and mapping provenance; existing disposable mirrors rebuild rather than migrate in place.


wesm:
I'm still working on this

PR #1169, opened 2026-07-17
fix(usage): use Copilot-reported billing as authoritative session cost

xkrogen:
Use Copilot CLI `session.shutdown.totalNanoAiu` as authoritative USD cost through the existing generic `cost_usd` and `cost_source` model only for sessions starting on or after June 1, 2026, when GitHub moved new sessions from premium-request billing to usage-based AI Credits. Older sessions, sessions without reported billing, and other Copilot-family agents retain catalog-priced estimates.

The final cumulative shutdown value wins, including zero. Per-model token analytics remain intact, while model-filtered cost stays estimated because a session-level charge cannot be allocated reliably among models. SQLite, PostgreSQL, and DuckDB preserve matching aggregation behavior, including cost-only shutdown carrier rows that do not surface as token data, models, or breakdown rows. Pricing provenance records `mixed` or `reported` when an authoritative session total is selected instead of incorrectly reporting `computed`.

Existing usage contracts remain unchanged: the Copilot AI Credits summary card, usage-daily `copilotAICredits` totals field, session `ai_credits` field, CLI `AI Credits` line, localization keys, and generated API types are retained. Credits derive from the final selected USD cost, whether reported or catalog-computed.

No database schema change, Copilot-specific cost column, migration, compatibility probe, dual-read path, or usage-daily schema-version change is introduced. The contract remains at `schema_version: 2`. Parser data version 69 re-parses existing sessions so eligible sessions created since June 1, 2026 receive reported `cost_usd` and `cost_source`; older sessions remain catalog-priced. Orphan archive-copy behavior remains unchanged from `main`.


xkrogen:
:robot: Addressed the schema compatibility finding: PostgreSQL compatibility checks now probe `usage_events.ai_credits`, with regression coverage for an older schema missing the column.

mariusvniekerk:
Adjusting this one a bit, adding special magic ai credits for a single provider is the wrong approach for a vendor agnostic system like agentsview.  The notion of a source makes sense, but we store pricing as normalized usd.

wesm:
looking

xkrogen:
Thanks @mariusvniekerk @wesm ! Please let me know if there's anything else I can do to contribute here 🙏🏻

PR #1170, opened 2026-07-17
feat(sync): add machine-labeled session sources

xkrogen:
Add filesystem-only `[[session_sources]]` entries with `agent`, `dir`, and
optional `machine`. Structured sources are additive to existing per-agent
directory arrays, defaults, and environment variables; when roots overlap, the
structured entry supplies the machine label. Configured paths keep the spelling
the user wrote them in, and a separate cleaned comparison key collapses
equivalent roots, so Windows-style entries survive a config round-trip.

A session's machine is resolved from the configured root that produced it and
is assigned once, when the session is first ingested. After that the label is
immutable: ordinary periodic, watcher-driven, incremental, provider-backed, and
single-session syncs all retain the stored value rather than adopting a newly
configured one. `agentsview sync --full` preserves stored labels while
rebuilding the archive; it is not a relabel operation. Changing a configured
label therefore affects only sessions discovered afterward, which makes editing
configuration safe rather than destructive. Filesystem labels do not namespace
native session IDs.

Because relabeling is no longer part of sync, this removes the machine-only
update for trashed sessions, the all-machines baseline scan, and the ownership
baseline index keyed without machine. Watch reconciliation still uses the
baseline for the machine that originally admitted a session, so per-event work
stays bounded by the changed batch.

An earlier revision of this branch chased reattribution through each new writer,
trash path, snapshot path, and reconciliation edge. That approach kept widening
the persistent-archive surface and reintroducing consistency bugs, so it was
replaced with the ingestion-time contract above. A one-shot relabel command is
deliberately deferred: it needs its own invariants and recovery tests, and it is
not required to ship labeled ingestion.

The filesystem sync guide documents Git, rsync/file copy, and NFS/shared-mount
topologies, including atomic publication, archive ownership, freshness,
native-ID deduplication, label-change behavior, and when PostgreSQL is a better
fit.

`session_sources` intentionally rejects `s3://` roots. Existing Claude and Codex
S3 configuration and behavior remain unchanged through their existing per-agent
settings.

Reviewers should start at `internal/sync/engine.go`, where machine resolution
and the preserve-on-upsert rule live, then `internal/config/config.go` for
parsing and path handling. `docs/superpowers/specs/` carries the design
rationale.

This branch contains no DuckDB changes. Mirror source-machine attribution
already landed in #1302, and this PR neither duplicates nor modifies it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,6 +480,29 @@ March 2026), which shares that same database. *Kilo (legacy)* is the legacy
 RooCode-derived VS Code extension that wrote per-task JSON under
 `kilocode.kilo-code/tasks/`.
 
+## Filesystem Session Sync
+
+One primary AgentsView instance can ingest native agent session directories
+copied or mounted from other machines without PostgreSQL:
+
+```toml
+[[session_sources]]
+agent = "copilot"
+dir = "/srv/session-archive/buildbox/copilot"
+machine = "buildbox"
+```
+
+Structured sources are additive to existing `copilot_dirs`,
+`claude_project_dirs`, and other per-agent settings. They label sessions by
+source machine without namespacing native session IDs. Transport source session
+files only -- never copy `sessions.db` or its WAL files. Machine labels are
+captured at first ingestion; ordinary sync and `agentsview sync --full` preserve
+the stored label. Changing attribution for existing sessions is not currently
+supported.
+
+See the [Filesystem Session Sync guide](https://agentsview.io/filesystem-sync/)
+for Git, rsync, shared-mount, freshness, and operational guidance.
+
 ## PostgreSQL Sync
 
 Push session data to a shared PostgreSQL instance for team dashboards:
@@ -633,6 +656,7 @@ Full docs at **[agentsview.io](https://agentsview.io)**:
 [Usage Guide](https://agentsview.io/usage/) --
 [CLI Reference](https://agentsview.io/commands/) --
 [Configuration](https://agentsview.io/configuration/) --
+[Filesystem Sync](https://agentsview.io/filesystem-sync/) --
 [Architecture](https://agentsview.io/architecture/)
 
 ______________________________________________________________________
--- a/cmd/agentsview/archive_query_backend.go
+++ b/cmd/agentsview/archive_query_backend.go
@@ -250,6 +250,7 @@ func (b localArchiveQueryBackend) SessionUsage(
 	if known && !b.skipFreshData {
 		engine := sync.NewEngine(b.database, sync.EngineConfig{
 			AgentDirs:               b.cfg.AgentDirs,
+			SourceMachines:          b.cfg.SourceMachines,
 			IncludeCwdPrefixes:      b.cfg.SyncIncludeCwdPrefixes,
 			Machine:                 b.cfg.LocalMachineName,
 			BlockedResultCategories: b.cfg.ResultContentBlockedCategories,
--- a/cmd/agentsview/archive_write_backend.go
+++ b/cmd/agentsview/archive_write_backend.go
@@ -936,6 +936,7 @@ func (b *localArchiveWriteBackend) DuckDBPushWatch(
 
 	engine := syncpkg.NewEngine(b.database, syncpkg.EngineConfig{
 		AgentDirs:               b.appCfg.AgentDirs,
+		SourceMachines:          b.appCfg.SourceMachines,
 		IncludeCwdPrefixes:      b.appCfg.SyncIncludeCwdPrefixes,
 		Machine:                 b.appCfg.LocalMachineName,
 		BlockedResultCategories: b.appCfg.ResultContentBlockedCategories,
--- a/cmd/agentsview/archive_write_backend.go
+++ b/cmd/agentsview/archive_write_backend.go
@@ -1054,6 +1054,7 @@ func (b *localArchiveWriteBackend) PGPushWatch(
 
 	engine := syncpkg.NewEngine(b.database, syncpkg.EngineConfig{
 		AgentDirs:               b.appCfg.AgentDirs,
+		SourceMachines:          b.appCfg.SourceMachines,
 		IncludeCwdPrefixes:      b.appCfg.SyncIncludeCwdPrefixes,
 		Machine:                 b.appCfg.LocalMachineName,
 		BlockedResultCategories: b.appCfg.ResultContentBlockedCategories,
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -296,6 +296,7 @@ func runServe(cfg config.Config, opts serveOptions) {
 		var onStartupReconciled func(sync.SyncStats, error)
 		engine = sync.NewEngine(database, sync.EngineConfig{
 			AgentDirs:               cfg.AgentDirs,
+			SourceMachines:          cfg.SourceMachines,
 			IncludeCwdPrefixes:      cfg.SyncIncludeCwdPrefixes,
 			Machine:                 cfg.LocalMachineName,
 			BlockedResultCategories: cfg.ResultContentBlockedCategories,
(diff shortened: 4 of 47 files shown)


xkrogen:
:robot: Addressed both review findings: Hermes attribution now retains the originating configured root, and stale-file machine lookups are agent-aware and covered by cardinality-scaling regression tests. Full lint also surfaced and prompted a nil-safe source-root initialization fix.

mjacobs (dismissed):
Thanks Erik! Would love to get this feature in, but have a few requests before
it's mergeble.

I independently verified the latest roborev findings against `d11b10e`: all five
reproduce, and Windows CI is red for a real reason (details below). The pattern
that worries me more than any single bug is that the two high-severity issues
live in code added during this PR's review iterations (`e0e1c56e`)- the S3
identity-transition machinery is accreting patches in the least forgiving part
of the codebase (the persistent archive, tombstones, PG ownership), and I don't
think another targeted-fix pass converges.

## Main ask: split the PR

**Part 1 — keep here:** `[[session_sources]]` config plus filesystem machine
attribution. With the fixes below, this is small and low-risk.

**Part 2 — separate PR:** the S3 label-_transition_ machinery
(`session_identity_aliases`, `session_source_identities`, supersession during
writes, tombstone propagation into PG push, rebuild integration). In part 1,
changing an S3 machine label should simply be unsupported for now- document it,
or hard-error when a transition is detected, rather than half-supported with
data-loss edges.

For part 2 I'd like a short design note before more code, stating the invariants
and the test that exercises each one directly:

- Ownership is never crossed without `sameSessionOwner` (needs a cross-owner
  isolation test on push).
- Trash is not permanent exclusion (needs a soft-delete + relabel + restore
  test).
- Rebuilds preserve curation through the alias table (needs a
  rebuild-first-observes-relabel test for both Claude and Codex, with
  pins/stars/names/recall asserted).
- Steady-state sync work is bounded by the changed batch (needs a query-count
  cardinality test, not a query-plan test).

## Verified findings

**High — PG ownership bypass.** `filterPGExcludedPushSessions`
(`internal/postgres/push.go:1929`) unions pusher-supplied alias endpoints into
purge groups with no `owner_marker` check, and it runs at `push.go:350`, before
`applyPGSessionIdentityAliases` can validate ownership. A pusher can delete and
permanently tombstone another machine's session by excluding one of its own
sessions and publishing an alias linking it to the victim ID. Every sibling
write path enforces `sameSessionOwner`; this new path is the sole exception, and
the added tests only ever use a single owner.

**High — rebuild loses curated data / creates duplicates.** When a
parser-version rebuild first observes a label change: the Codex orphan-copy
exclusion (`internal/db/orphaned.go:161-168`) drops the old-identity row while
metadata restoration matches by unchanged ID only — pins, stars, display names,
and recall evidence are permanently lost once the old DB is discarded. Claude
takes the other branch and is restored as an orphan, producing temporary old/new
duplicates. The copied alias table is never used to remap curated data, and
`TestCopySessionMetadataPreservesIdentityAliases` pre-applies supersession,
sidestepping exactly this path.

**Medium — trash becomes a permanent tombstone.** `PrepareSessionSourceIdentity`
(`internal/db/sessions.go:2251`) treats `deleted_at IS NOT NULL` the same as
permanent exclusion and inserts every identity — including the replacement ID —
into `excluded_sessions`, which nothing ever removes. Trash a session, relabel
the S3 machine, and it can never be restored nor its replacement imported. Tests
cover `DeleteSession` but never soft delete.

**Medium — UUID-backed pins can reattach to unrelated messages.** Both
`restorePinsTx` (`internal/db/messages.go:1536-1576`) and the PG realignment SQL
(`internal/postgres/push.go:2991-3001`) fall back to the saved ordinal when a
non-empty source UUID vanishes or becomes ambiguous. After a transcript rewrite
that ordinal can hold unrelated content, silently moving the pin and its note.
Ordinal fallback should be reserved for legacy pins with no UUID; a UUID-backed
pin whose target can't be identified unambiguously should be dropped or staged,
not guessed.

**Medium — steady-state sync scales with archive size.** `filterFilesByMtime`
(`internal/sync/engine.go:3232`) issues one `SessionMachinesByFilePathMatch`
query per unchanged file below the incremental cutoff on every periodic pass —
O(total archive) round-trips when nothing changed, which violates the
bounded-by-changed-batch rule in AGENTS.md. The regression test asserts the
query plan is index-stable, not that the query count is bounded, so it doesn't
guard the actual invariant. Batch the check, or key it off a source-attribution
config revision so it only runs when labels change.

**Windows CI (part 1 blocker).** `resolveSessionSources()` runs `filepath.Clean`
over all configured agent dirs, including legacy ones, which rewrites
`/sessions/x` to `\sessions\x` on Windows — that's why pre-existing tests
(`TestResolveDirs`, `TestEnvOverridesConfigFile`, the legacy-JSON test) broke
alongside the new ones. Normalization needs to be confined to the new source
entries, or done comparison-only so stored paths and user-visible dirs keep
their original form.

xkrogen:
Thanks @mjacobs — agreed. I removed the S3 label-transition work entirely rather than splitting it into a follow-up PR. I went down that path just for the sake of consistency, but it seems like it introduces a lot of unnecessary complexity that isn't really relevant to the core goal here. This PR is now limited to filesystem-only [[session_sources]] configuration and machine attribution. Existing S3 configuration and behavior remain unchanged, and session_sources rejects s3:// roots.

I also addressed the two Part 1 blockers: legacy directory paths retain their original spelling on Windows, and machine reattribution is now gated by a persisted source-attribution revision. Unchanged periodic syncs therefore perform zero per-file attribution queries, with a small-versus-large archive regression covering that behavior.

Removing the S3 transition machinery also removes the alias ownership, rebuild, trash/tombstone, pin-migration, and multi-session Claude issues identified in the reviews.

xkrogen:
Working on slimming down / simplifying this PR, but will have to put it on pause for a few days until my credits reset 😬

mjacobs:
Erik — your split on the 27th was right, and it worked: `a2b39fdf` dropped
3,400 lines and both High findings with it. Asking for one more of the same
shape.

**Stop fixing roborev findings here.** Everything from `035422f` onward is a
consequence of retroactive relabeling — editing `machine =` updating sessions
already in the archive. Each review finds one more write path, and there is
always another. None of them need a targeted fix.

You reached the answer at `3600f521`. Make it absolute: **a session's machine
label is immutable after ingestion.** That includes `sync --full`, which should
carry each session's stored machine forward rather than reapply configured
labels — a parser-version rebuild shouldn't quietly change attribution.

Then these can go: `UpdateTrashedSessionMachine` and
`UpdateTrashedSessionMachineByPath` in `internal/db/sessions.go` plus their call
sites, the snapshot and observation SQL in `internal/db/orphaned.go`, and
`ListActiveSessionSourceOwnershipScopesAllMachinesPage` with
`idx_local_source_baselines_source`.

Relabeling stays on the table, later, as an explicit one-shot command rather
than a config side effect. Far easier to build on immutable attribution than to
retrofit onto continuous reattribution.

**I'm taking the DuckDB changes.** They fix a `remote_hosts` bug that predates
your branch (`internal/remotesync/import.go:138`) and carry a schema bump
affecting every user, so they need their own review. Drop them here; I'll open
it separately.

**CI ran.** Two failures: the Windows `session_sources` config tests, because
`normalizeSessionSourceDir` still calls `filepath.Clean` — normalizing for
comparison only should close it, and you already fixed the legacy half. And
lint wants a `strings.Builder` at
`internal/sync/session_source_machine_test.go:79`.

No rush on any of it.

mjacobs:
Addressed both findings in `6f2ca15`. The missing machine attribution was reproduced and fixed. Baseline bookkeeping now also uses the persisted machine label, but I couldn’t reproduce the reported relabel → resync → delete failure, so that case still needs a second look.

This commit also addresses the five findings from `49b3f4a` and fixes the docs formatting failure.

PR #1173, opened 2026-07-17
fix(parser): track Hermes skill_view usage

serverless83:
## Summary
- centralize Hermes tool-call parsing across JSONL, JSON, and `state.db` sources
- preserve Hermes tool input from nested or top-level `arguments`/`input`, in string or object form
- populate `ParsedToolCall.SkillName` from `skill_view` arguments
- bump the parser data version so historical Hermes sessions are reparsed and backfilled
- add regression coverage for all three Hermes input paths and the data-version bump

## Problem
Hermes records skill loads as `skill_view` tool calls with a JSON-string `name` argument. The parser preserved the tool call and its input, but never populated `SkillName`, so skill analytics always reported no data for Hermes sessions.

Existing sessions also need a data-version resync; otherwise unchanged Hermes sessions already stored at the previous parser version would remain without skill metadata.

## Test plan
- `go test -tags fts5 ./internal/db ./internal/parser -run '^(TestCurrentDataVersionHermesSkillName|TestParseHermesSkillViewSetsSkillName)$' -count=1 -v`
- `go test -tags fts5 ./internal/db ./internal/parser -count=1`
- `make vet`
- `make test-short` (all packages passed except the unrelated, filesystem-sensitive `TestFileChangeTimeDetectsSameStatRewrite`, independently reproducible as intermittent)

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -308,7 +308,9 @@ const projectIdentityRemoteScrubCompletedKey = "project_identity_remote_scrub_v1
 // and entrypoint fields on existing Claude rows.)
 // (67: Antigravity CLI reader metadata. Re-parsing populates parent_session_id
 // and relationship_type from agyReader.parentCascadeId in trajectory sidecars.)
-const dataVersion = 67
+// (68: Hermes skill_view metadata. Re-parsing populates tool_calls.skill_name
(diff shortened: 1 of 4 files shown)


mjacobs:
LGTM, thanks for the contribution!

PR #1176, opened 2026-07-18
feat(usage): roll up subagent cost in session header

### LINKED ISSUE #1139: When viewing a session there should be a roll up cost for all sub agents work
### Problem / motivation

When viewing a session you can see the agents cost total as well if you click into each sub agent you can see there cost, but we should be able to see total cost across all sub agents at a glance on the session agent view.

### Proposed solution

Image

Add either a visible indicator here of roll up cost or ability to click the cost as a link to pop up the total roll up cost. Alternatively make it configurable to keep as is or show rollup cost in its place. 

### Alternatives considered

_No response_

rodboev:
Session headers currently price only the open transcript, so delegated work remains hidden unless each subagent is opened separately.

This adds an opt-in subagent rollup to the existing session-usage response and uses it for the header badge when the selected session has explicit subagent descendants. The rollup walks the stored child-session graph, traverses bridge nodes such as continuations and forks, and aggregates the root plus explicit subagents under one shared dedup scope per backend, so replayed usage rows contribute once across SQLite, PostgreSQL, and DuckDB and shared duplicates stay attributed deterministically.

Fork and continuation usage stays outside the displayed total, but traversal continues through those bridge nodes so nested explicit subagents still contribute. Untimed or unparseable usage rows now participate in the rollup wherever `GetSessionUsage` already includes them. Replayed rows, deleted sessions, duplicate links, and incomplete pricing also stay outside the total marker, while the existing own-session cost and per-step breakdown remain available when a complete rollup is unavailable.

The header marks a value as total only when at least one subagent contributes to a complete priced rollup. The requested at-a-glance placement came from [@travis-jorge's report and screenshot](https://github.com/kenn-io/agentsview/issues/1139).

Closes #1139

### DIFF
--- a/docs/session-api.md
+++ b/docs/session-api.md
@@ -626,6 +626,13 @@ only honored on a localhost-bound daemon.
 Per-session token usage and cost estimate. Output shape is
 stable for the fields shown below; new fields may be added.
 
+The REST endpoint is `GET /api/v1/sessions/{id}/usage`. Add
+`?rollup=true` there to include the selected session's explicit
+`subagent` descendants recursively. `breakdown=true` remains scoped to
+the selected session; descendant usage is loaded as totals only. The CLI
+subcommand keeps its existing own-session output and does not expose the
+rollup fields below.
+
 ```bash
 agentsview session usage <id> [--format json]
 ```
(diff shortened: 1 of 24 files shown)


rodboev:
I checked the size on this — tests and locale strings make up most of it. The feature is a small rollup service plus a header change. The rest is the same session-usage read written once per backend since SQLite, PostgreSQL, and DuckDB each own their query path and the badge has to agree under all three serve modes. The read dedupes raw rows across the whole session set so usage replayed between parent and subagent transcripts counts once.

wesm:
Makes sense, thank you!

PR #1177, opened 2026-07-18
parser(hermes): snapshot profiles and deprecate SSH remote sync

jahabdank:
AgentsView now treats Hermes named profiles as live archives rather than a startup-only list of transcript directories. Long-running servers discover profiles created after initialization and import sessions plus `state.db` metadata, including database-only sessions and WAL-only updates.

HTTP remote sync now publishes each Hermes `state.db` as a standalone SQLite online-backup snapshot whose manifest identity includes uncheckpointed changes. Full and delta transfers therefore cannot mix a main database with a different WAL generation, while target resolution remains confined to transcripts and database files instead of credential-bearing profile contents.

Hermes overrides retain replacement semantics, trailing-slash handling, profiles-container parity, and supported flat transcript roots across local and remote resolution.

SSH remote sync remains available for compatibility, but it is deprecated and receives only critical fixes. Remotes with Python 3 and SQLite backup support get the same coherent database snapshots; every database snapshot omitted because that support is unavailable or snapshotting fails emits a path-specific warning, while transcripts and unrelated agent data continue transferring. The CLI warns once per process, and the help and remote-sync documentation direct new configurations to HTTP.

### DIFF
--- a/cmd/agentsview/cli.go
+++ b/cmd/agentsview/cli.go
@@ -334,15 +334,15 @@ func newSyncCommand() *cobra.Command {
 	)
 	cmd.Flags().StringVar(
 		&cfg.Host, "host", "",
-		"SSH hostname for remote sync",
+		"SSH hostname for deprecated remote sync",
 	)
 	cmd.Flags().StringVar(
 		&cfg.User, "user", "",
-		"SSH user for remote sync",
+		"SSH user for deprecated remote sync",
 	)
 	cmd.Flags().IntVar(
 		&cfg.Port, "port", 0,
-		"SSH port for remote sync (default: 22)",
+		"SSH port for deprecated remote sync (default: 22)",
 	)
 	cmd.Flags().StringVar(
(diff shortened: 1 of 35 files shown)


jahabdank:
Thanks for taking this well beyond my original patch, @wesm — moving to SQLite online-backup snapshots with WAL-consistent sync is exactly the robustness the naive sessions/-only approach was missing, and deprecating the fragile SSH path in favor of it makes sense.

Happy to help validate against a real target: I run the exact setup this feature is for — a dedicated named orchestrator profile plus profiles synced from a second always-on box, with both database-only and WAL-only sessions in play. Glad to exercise the snapshot + remote-sync paths end-to-end on that once you've settled the shape, and report back anything that misbehaves. 

Should I wait until you merge the PR into main?

wesm:
Yeah, if you could test after I merge to main and make sure I didn't break anything in my hardening commits, that would be helpful!

PR #1180, opened 2026-07-18
feat(parser): add Trae local sessions (#1178)

### LINKED ISSUE #1178: add Trae / TRAE SOLO CN session and token usage support
### Problem / motivation

Hi! I use AgentsView to track local AI coding-agent usage on Windows. It already detects my Codex and WorkBuddy sessions, and I would also like to include Trae usage.

My installed variant is `TRAE SOLO CN`. Its local user-data directory is:

`%APPDATA%\TRAE SOLO CN`

I found VS Code-style storage files under:

`User\workspaceStorage\...\state.vscdb`
`User\globalStorage\state.vscdb`

Could AgentsView support Trae session discovery and token usage parsing?

Requested behavior:

- Add a `trae` agent source with a configurable `TRAE_DIR` override.
- Discover local sessions and retain timestamps, workspace/project, conversation messages, and model name when available.
- Parse reported input, output, and cache token fields when present.
- If Trae does not store token counters locally, show the session but mark token/cost data as unavailable instead of estimating it.
- Keep processing fully local, consistent with the existing AgentsView privacy model.

Would you be open to a PR? I can provide a sanitized database schema or a redacted sample fixture if that would help identify the relevant storage format.

### Proposed solution

_No response_

### Alternatives considered

_No response_

rodboev:
Adds local Trae session discovery for standard Trae installs and the Windows-specific Trae CN and TRAE SOLO CN builds. The provider reads VS Code style `workspaceStorage` and `globalStorage` SQLite stores, carries workspace labels, timestamps, ordered user and assistant messages, and model names into the existing session pipeline, and supports `TRAE_DIR` or `trae_dirs` overrides for custom install roots.

Trae sessions remain browsable when the local payload has no token counters. Those sessions use AgentsView's existing unavailable token and cost state instead of an estimate, so the change stops at transcript discovery rather than inventing pricing support. The parser stays intentionally narrow: it decodes only `memento/icube-ai-agent-storage`, preserves ordered messages and optional model metadata, and uses `agentTaskContent` only for empty assistant content. Prompt-history or auxiliary-key variants remain out of scope until stronger Trae format evidence is available.

This rework builds Trae on the shared multi-session container base instead of keeping a bespoke lifecycle. Each `state.vscdb` is discovered as one physical container, exposed under stable `<db>#<session-id>` virtual paths, and parsed into one result per session. Only an explicit `list: [...]` payload is authoritative for delete handling. A valid empty list is treated as a complete result set so tombstones flow, while malformed top-level storage still errors and preserves archived rows, and structurally unknown stores such as `{}` or `{"list":null}` stay incomplete so they preserve the archive instead of acting like confirmed deletion. `RequireFreshSource` now rechecks current membership and can relocate a stored virtual path by raw session ID.

Trae now matches the sibling shared-container providers' freshness model. I did not add Trae to the global hash-aware cache-key or DB freshness switches. The proof cases cover cached and fresh-engine same-size same-mtime rewrites without that broadening. Trae bypasses the generic mtime-only skip cache like the other shared-container providers, Trae DB and WAL watcher events stay on the provider fingerprint path instead of forcing a full rewrite, each Trae session now carries its own record hash plus workspace label, and the engine drops unchanged sibling sessions after a container reparse. Parse-diff also keys `state.vscdb#...` by agent, so Trae partial removals collapse back to the container for presence sweeps while Windsurf keeps per-session sampling. There is no Trae-specific trusted-container gate or cache layer in sync; member lookups and presence checks reread the store directly, matching the sibling providers' gateless and cacheless shape.

HTTP and SSH remote target resolution now exclude Trae entirely until it has Windsurf-style curated file targets limited to `state.vscdb`, `state.vscdb-wal`, and `workspace.json`, instead of archiving the full Trae `User` profile.

Closes #1178


rodboev:
I rebuilt Trae on the shared multi-session container base instead of keeping a bespoke source-set lifecycle. Discovery is now one physical `state.vscdb` per source, parsing fans out one result per session, stored hints participate in tombstones, and `RequireFreshSource` rechecks live membership before trusting a stored virtual path.

On the remote exposure point, I kept `FileBased: true` but excluded Trae from both HTTP and SSH target resolution. The follow-up path is curated Windsurf-style file targets only: `state.vscdb`, `state.vscdb-wal`, and `workspace.json`.

On the tombstone and fresh-lookup points, a valid empty `list` is now a complete no-session result set so deletes flow, while malformed top-level storage still errors and preserves the archive. I also fixed the warm-cache case: Trae member presence is reused only while the live SQLite container state is unchanged, so a later delete or relocation in the same provider instance does not keep reading as present. Coverage now includes warm-cache relocation fallback and warm-cache tombstone scheduling.

rodboev:
Quick catch-up on rounds 4 through 7: I fixed the WAL watcher and parse-diff seams, then tightened Trae's authoritative membership so malformed or contentless storage can't tombstone archived sessions. The next rounds moved the hot path onto the shared container-state gate, trimmed the snapshot cache down to a byte-bounded lightweight record cache, and this push fixes the last reachable correctness seam by forcing a reparse when workspace.json changes so project labels refresh immediately. I'm not taking the per-event container enumeration finding further because that work is Trae joining the same shared gate capture path the OpenCode-family providers already use, and I'm not adding cross-container duplicate-ID precedence because the sibling multi-container providers don't do that either and we don't have a real collision to resolve.

rodboev:
Taking the container freshness gate and the snapshot cache back out of Trae. Most of the recent review findings were defects in that machinery rather than in the feature, which turned out to be a sign it shouldn't exist. Windsurf and Shelley handle the same shape without either, so Trae will just do what they do.

mjacobs:
Thanks for the iteration on this one — stripping the bespoke freshness gate and snapshot cache in favor of the Windsurf/Shelley shape was the right call, and it reads as a clean member of the shared-container provider family now. The container fingerprint covering blob + `-wal` + `workspace.json`, the one-decode-per-container tombstone pass, and the remote-sync exclusion all look right.

I pushed two cosmetic fixes directly to the branch as df1fa60 rather than sending you round again for them — heads up that you will want to pull before your next push:

- `internal/parser/trae_provider.go`: dropped the `|| !ok` from the `traeParseRecord` error check in `traeParseMember`. The earlier `if !ok { return nil, nil }` already guarantees `ok` is true there, so the condition was dead.
- `internal/sync/engine.go`: added a comment on `providerVirtualSourceBackedByEvent` noting that the `workspace.json` branch keys on the `state.vscdb` basename shared by Windsurf and Trae, so it covers both. It was already correct and covered by tests, but the `WindsurfStateDBName` reference makes the Trae coverage look accidental on a first read.

Verified locally with CGO and the fts5 tag: `make vet` clean, `make test` green across 37 packages including parser, sync, remotesync, ssh, and cmd/agentsview.

On the cross-container keying: I am taking your sibling-parity reasoning, so this is not blocking. Filing it as #1191 rather than leaving it in the thread, since Trae discovering both `workspaceStorage` and `globalStorage` gives it a second namespace Windsurf does not have, and a collision would surface as a silently overwritten session rather than an error. If Trae never writes the same session to both stores, that issue closes as a non-issue.

Squash-merging.

PR #1184, opened 2026-07-19
refactor(duckdb): pare the mirror back to a rebuildable derived artifact

wesm:
## Summary

The DuckDB/Quack code had grown into a third system of record alongside SQLite and PostgreSQL, with its own in-place schema migrations, remote push path, orphan repair, and per-table incremental sync state. This branch pares it back to what it was conceived as: a disposable read mirror of the SQLite archive. AGENTS.md records the mirror rules as standing policy so the machinery cannot regrow.

### What the code does now

- **The mirror is self-describing and disposable.** All push state (last-push cutoff, deletion/identity revisions, scope, machine, schema and data versions, and the source archive's `database_id`) lives in the mirror's own `sync_metadata`. Deleting the mirror file loses nothing; the next push rebuilds it.
- **Rebuild instead of migrate.** `createSchema` is create-only at `SchemaVersion=3`; all in-place mirror migrations are deleted. Any schema-version mismatch (either direction), source data-version change, source archive change (`database_id` mismatch — a mirror can never silently blend sessions from two archives; resync mints a new id, so the first push after a resync is a full rebuild, paralleling PostgreSQL's generation-scoped cursors), scope change, machine-name change, shape damage, or a mirror deletion cursor ahead of the local archive triggers a full rebuild into a temp file that is validated, checkpointed, and atomically swapped in.
- **Serve holds the mirror read-only; pushes probe freely.** The serve-side Store, reopen aliases, and `quack serve` open with `access_mode=read_only`, and DuckDB permits concurrent cross-process read-only opens — so a probe can always inspect a served mirror. Overwrites fail closed on a simple rule: an existing destination is replaced only if it carries the `agentsview_schema_version` sentinel in `sync_metadata`; a probe-time lock conflict means a writer holds the file and the push fails with an actionable error. Incremental pushes need exclusive write access, so while readers hold the mirror an automatic push defers (logged, no cutoff advanced) and an explicit push falls back to rebuild-and-swap, which never write-opens the destination. Serving processes adopt swapped-in files via a replacement watcher. The per-mirror work directory (rebuild temps, reopen aliases) is created user-only-writable and fails closed if it is a symlink, foreign-owned, or group/other-writable, so nothing else can interfere with a rebuild's temp file before the atomic swap.
- **Generated artifacts are confined.** Rebuild temps and reopen hardlink aliases live in a per-mirror private work directory (`<mirror>.agentsview-work/`), created only when needed; cleanup sweeps operate exclusively inside it and can never touch sibling user files. `ProbeMirror` and `duckdb status` never create anything; status reports a missing mirror explicitly and opens existing mirrors read-only. The daemon push endpoint pins the mirror path to the server's own configuration; the CLI never sends a path.
- **Incremental is session-replace only, and both backends share the window.** A trigger-maintained `sync_marker` column on SQLite sessions gives an inclusive watermark; the DuckDB and PostgreSQL pushes both select candidates via the shared `ListSessionsForMirrorWindow` — `sync_marker >= cutoff`, no upper bound, so a clock-skewed future `file_mtime` cannot mask later real changes on either backend. Candidates are re-fingerprinted against mirror-resident fingerprints and changed sessions are deleted and re-inserted wholesale. The fingerprint covers every mirrored session column, and every mirrored-data rewrite advances the marker (signal recomputes, usage-event rewrites, subagent linking, tool-call subagent assignment all bump `local_modified_at` in the shared db layer). The marker computation drops unparseable `created_at` values, and trigger DDL is concurrency-safe across racing `Open` calls and installed in one write transaction (no window where a concurrent handle's write can slip past an absent trigger; the deletion/identity journal triggers in schema.sql get the same transactional install).
- **Deletions are journal-driven and candidate-bounded.** A trigger-driven `session_deletion_changes` journal with an O(1) revision counter delivers deletions incrementally; consumed cursors live in the mirror. Journal continuity across resync is deliberately not preserved — a resync changes the archive's `database_id`, which forces a full rebuild anyway (PostgreSQL never consumes this journal; its deletions are tombstone-based with generation-keyed cursors).
- **Scope filters reconcile transitions.** The incremental window is enumerated without project filtering and partitioned in Go: in-scope changes are pushed, sessions that moved out of a filtered mirror's scope are removed from it, and deletion tombstones apply regardless of the deleted session's final project.
- **Curation refresh is curation-bounded.** Star/pin updates load one scoped snapshot of the local curation rows, validate just those session IDs against the mirror, and rewrite the curation tables from that same snapshot — bounded by curation size, gated by a fingerprint computed from the snapshot so the stored fingerprint always describes exactly what the mirror holds.
- **Quack is read-only.** The remote push path is deleted; `duckdb push` writes only the local mirror file, and `ValidatePushTarget` rejects URLs with a pointer to `duckdb quack serve`. The read side (serve backend, parameterized-SQL rendering in `quack_sql.go`) is unchanged.
- **Automatic pushes stay bounded.** Watch-mode pushes carry `SyncOptions.Automatic`: they defer when write access is unavailable instead of rebuilding per changed batch, they skip archive-scale diagnostics like the scope-wide session count, and cursor admin usage rows sync through a high-water id in mirror metadata so each push loads only appended events, not the full history. Replacement detection primes each baseline FileInfo's identity eagerly (Windows loads it lazily and stamps mtimes at ~15.6ms clock-tick granularity) and compares identity plus mtime/size; store reads hold the handle lock across query start so a swap cannot close a handle under a starting query.

### Behavior changes

- Remote `duckdb push` to a URL is removed.
- Daemon (`POST /api/v1/push/duckdb`) pushes write only the server-configured mirror path; a request naming a different path is rejected.
- A scope (project-filter) change rebuilds the mirror, and sessions that move out of scope are removed from filtered mirrors instead of being preserved.
- A source-archive change (new `database_id`, including after any resync) forces a full rebuild.
- Serve processes open the mirror read-only. An explicit push while readers hold the mirror rebuilds and swaps; automatic watch-mode pushes defer and catch up. A serve process built before this change holds the mirror read-write and must be restarted once after upgrading; until then pushes against it fail closed.
- Replacing an existing file without the agentsview metadata sentinel fails with an error instead of overwriting it.
- `duckdb status` on a missing mirror reports it as missing instead of silently creating an empty file.
- Usage-event rewrites, subagent linking, and tool-call subagent assignment now bump the session's `local_modified_at`, so they propagate to push targets (previously invisible to both the DuckDB and PostgreSQL pushes).
- The PostgreSQL push selects candidates through the same unbounded-above window as DuckDB, fixing the same future-timestamp staleness there.
- Rebuild temps and reopen aliases moved from mirror-sibling files into the per-mirror `.agentsview-work` directory (safe to delete when no push or serve is running).

### Known limitations

- Sessions whose only timestamp signal is a malformed `created_at` are invisible to incremental windows (matching the PostgreSQL push); a `--full` push covers them.
- Existing mirrors rebuild or re-push their current window once after upgrading (fingerprint composition and recorded metadata changed).
- While readers hold the mirror continuously, watch-mode pushes defer indefinitely; the mirror catches up on the next moment without readers or via an explicit push (which rebuilds).

### Where to look

- `internal/duckdb/probe.go`, `rebuild.go`, `sync.go`, `push.go`, `connect.go` — probe, sentinel recognition, source-archive binding, rebuild-and-swap, session-replace push, read-only opens, deferred/bounded automatic pushes
- `internal/db/schema.sql`, `db.go`, `session_deletions.go`, `sessions.go`, `signals.go`, `usage_events.go` — sync marker, deletion journal, candidate window, mirrored-data marker bumps
- `internal/postgres/push.go` — shared candidate window
- `internal/server/huma_routes_push.go` — daemon mirror-path pinning and the automatic flag
- `internal/duckdb/store.go`, `mirror_watch.go`, `cmd/agentsview/duckdb.go` — read-only serve, replacement watching, status, CLI
- `AGENTS.md`, `docs/duckdb.md` — standing mirror rules and operational guidance


wesm:
Declined. merging

PR #1194, opened 2026-07-20
fix(frontend): stop persisting rolling-derived session date bounds

### LINKED ISSUE #1086: Date-range preset persisted as fixed dates never rolls forward — app silently stops showing new sessions
## Summary

(Issue posted with an agent🤖, I experienced it, troubleshooted with the agent(fable) and had it draft this commit) The macOS desktop app persists the dashboard/sidebar date range as **fixed calendar dates** (`"mode": "fixed"`) in the webview's localStorage. A range that originated as a rolling preset ("Last year") never advances past the day it was saved, so after a few days the app silently stops showing new sessions — it looks like sync is broken, when in fact the backend is ingesting everything in real time.

## Environment

- AgentsView v0.37.5, macOS desktop app (Apple Silicon, macOS 26)
- Backend healthy: `agentsview serve` syncing in real time, `/api/v1/sync/status` current

## Symptoms

- Desktop app sidebar showed 640 sessions, newest "3d ago"; dashboard activity charts ended days in the past.
- The same server opened in a regular browser at `http://127.0.0.1:8080` showed 763 sessions with live "just now" activity — the two views diverge because each webview has its own localStorage.
- The only visual hint is the small green dot on the sidebar filter funnel icon.

## Root cause

The desktop webview's localStorage contains a pinned range that matches "last year as of the day the app was first run" (installed Jul 7):

```json
// session-filters
{"dateFrom": "2025-07-07", "dateTo": "2026-07-06", ...}

// yoked-dates
{"version": 1, "range": {"from": "2025-07-07", "to": "2026-07-06", "mode": "fixed", "updatedAt": 1783717314532}}
```

`updatedAt` keeps refreshing, but the range itself never rolls forward. With `dateTo` stuck at 2026-07-06, every session after that date is filtered out of the sidebar and all analytics. Confirmed against the API: `/api/v1/sessions/sidebar-index?...&date_from=2025-07-07&date_to=2026-07-06` returns nothing newer than 2026-07-06.

I can't say for certain whether I explicitly picked a custom range at some point, but the stored dates exactly match "Last year" computed on the install date, which suggests a preset was serialized to fixed dates rather than kept relative.

## Expected behavior

- Presets like "Last year" / "Last 30 days" should be stored as relative ranges and re-resolved against the current date on load; `"mode": "fixed"` should be reserved for explicitly user-chosen custom ranges.
- When an active date filter excludes today, show an explicit indicator (e.g. a banner like "Showing sessions through Jul 6") — the green dot on the funnel icon is easy to miss, and the failure mode reads as "sync is broken".

## Workaround

Clear the date range via the sidebar funnel icon (or re-select a preset in the dashboard date picker).

bangddong:
Fixes #1086.

Date bounds materialized from a rolling window (`window_days`) were saved into the `session-filters` localStorage entry and restored verbatim on the next launch. Once yoked dates could no longer overwrite them — e.g. after the v1 storage migration disabled them — the range stayed pinned to the day it was saved and newer sessions silently disappeared from the sidebar and analytics.

The sessions store now tracks the provenance of its date bounds as `dateFiltersWindowDays`: bounds applied from a rolling panel state (`App.svelte`, `AnalyticsPage.svelte`) or a `window_days` deep link (`initFromParams`, validated with the route layer's shared `parseWindowDaysParam`) carry their rolling intent, while explicitly chosen fixed ranges carry `null`. Persistence stores the intent rather than the materialized dates: the saved entry keeps `windowDays` and blank bounds, and `loadSavedFilters` rematerializes the bounds against the current date, so a rolling filter survives restarts and keeps rolling forward instead of pinning or vanishing. `applyPanelDateFilters` persists immediately, so a provenance flip between fixed and rolling with identical materialized bounds — which does not register as a filter change and may never trigger a `load()` — still updates the saved entry. Wholesale filter resets clear the intent, and the store is now the sole writer of session date bounds (`clearSessionDateFilters` removed as dead code).

Already-poisoned installs are also repaired: saved entries are stamped with a storage version, and unversioned (pre-provenance) entries have their date bounds dropped once on load and are rewritten in the new format, mirroring the yoked-dates v1 migration. The check is deliberately `!==` rather than `<` so a downgraded newer format is also not trusted; dropping bounds is the fail-safe direction in both. Other persisted filter fields are preserved.

Reviewers should look at `saveFilters`/`loadSavedFilters`/`applyPanelDateFilters` in `sessions.svelte.ts` and the two call sites. Known cosmetic note: a persisted one-day rolling window rematerializes as `date_from`/`date_to` rather than the single `date` param; both filter identically.

### DIFF
--- a/frontend/src/App.svelte
+++ b/frontend/src/App.svelte
@@ -317,7 +317,10 @@
   ): Record<string, string> | null {
     const dateParams = panelDateToSessionFilterParams(state);
     if (Object.keys(dateParams).length === 0) return null;
-    sessions.applyPanelDateFilters(dateParams, state.mode === "rolling");
+    sessions.applyPanelDateFilters(
+      dateParams,
+      state.mode === "rolling" ? state.windowDays ?? null : null,
+    );
     const params = { ...routeParams };
     for (const key of [
       "date",
--- a/frontend/src/App.svelte
+++ b/frontend/src/App.svelte
@@ -311,6 +311,20 @@
     }
   }
 
+  // Session route params composed from filters must carry the rolling
+  // intent: a URL with concrete dates but no window_days re-enters
+  // initFromParams as an explicit fixed range, so the rolling
+  // persistence would be lost after one reload.
+  function sessionFilterRouteParams(): Record<string, string> {
+    const params = filtersToParams(sessions.filters);
+    if (sessions.dateFiltersWindowDays !== null) {
+      params[SESSION_ANALYTICS_WINDOW_PARAM] = String(
+        sessions.dateFiltersWindowDays,
+      );
+    }
+    return params;
+  }
+
   function applySessionDateState(
     state: PanelDateState,
     routeParams: Record<string, string>,
@@ -330,7 +344,7 @@
     ]) {
       delete params[key];
     }
-    Object.assign(params, filtersToParams(sessions.filters));
+    Object.assign(params, sessionFilterRouteParams());
     if (
       state.mode === "rolling" &&
       state.windowDays
@@ -445,7 +459,7 @@
   $effect(() => {
     const activeId = sessions.activeSessionId;
     const currentUrlSessionId = router.sessionId;
-    const filterParams = filtersToParams(sessions.filters);
+    const filterParams = sessionFilterRouteParams();
     const filterParamsSignature = JSON.stringify(filterParams);
     untrack(() => {
       if (router.route !== "sessions") {
@@ -504,7 +518,7 @@
   $effect(() => {
     const route = router.route;
     const newParams = sessionRouteParamsForFilters(
-      filtersToParams(sessions.filters),
+      sessionFilterRouteParams(),
       router.params,
     );
     untrack(() => {
--- a/frontend/src/App.svelte
+++ b/frontend/src/App.svelte
@@ -317,9 +317,7 @@
   ): Record<string, string> | null {
     const dateParams = panelDateToSessionFilterParams(state);
     if (Object.keys(dateParams).length === 0) return null;
-    sessions.filters.date = dateParams["date"] ?? "";
-    sessions.filters.dateFrom = dateParams["date_from"] ?? "";
-    sessions.filters.dateTo = dateParams["date_to"] ?? "";
+    sessions.applyPanelDateFilters(dateParams, state.mode === "rolling");
     const params = { ...routeParams };
     for (const key of [
       "date",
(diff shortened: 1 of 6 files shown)


mjacobs:
Thanks for the contribution! Looks good generally, a few requests:

**Rolling filters on the Sessions tab now vanish entirely on browser restart, for most users.** `analyticsPageDates` is in-memory only (`#retained`, no localStorage), so after a reload `restoreWithIntent("sessions")` returns `explicitDateIntent: false` and `sessionEntryDateParams` won't apply it. The only persisted fallback is `yokedDates` — which defaults to `enabled: false` (`yokedDates.svelte.ts:130,354`). So a user with "last 30 days" on Sessions comes back to a completely unfiltered list.

That's strictly safer than the frozen window it replaces, so it's not a blocker. But the more complete fix is to persist the *intent* rather than dropping it: store `windowDays` in the `session-filters` entry and rematerialize bounds on load. That fixes #1086 without losing the user's filter, and it's what makes the deferred "showing sessions through `<date>`" indicator unnecessary.

### Smaller items

- **`dateFiltersDerived` is not `$state`** (`sessions.svelte.ts:280`). Every sibling field in the store — `filters`, `total`, `loading` — is a rune. It's read only imperatively today so nothing is broken, but it's public, and a future template read (exactly the deferred indicator) would silently never update. `dateFiltersDerived = $state(false)` is a free fix.

- **`clearSessionDateFilters()` is now dead code** (`AnalyticsPage.svelte:145-148`, called at :172). `applyPanelDateFilters` already writes `dateParams["date"] ?? ""` for all three fields, and `panelDateToSessionFilterParams` returns `{}` for an invalid state — so the line below it clears them anyway. The `before` snapshot is captured before the clear, so removing it doesn't change the diff result. It's also the last raw write to `filters.date*` outside the store, and it only stays safe because `applyPanelDateFilters` runs on the very next line. Deleting it removes a fragile coupling and makes the store the sole owner of date provenance. (`sessionDateFiltersAreClear()` is separate and still used — keep that.)

- **`initFromParams` keys on presence, not validity, of `window_days`** (`sessions.svelte.ts:410-411`). The route layer uses `isValidWindowDaysParam` everywhere else (`sessionRouteParams.ts:54`, gating `shouldPreserveSessionWindowDays`). With `window_days=abc` the store marks bounds derived while the route layer drops the param on the next filter change, flipping it back. The PR body justifies this and both directions are safe, but reusing the shared helper removes the divergence for free.

- **`version !== SESSION_FILTERS_VERSION` also fires on downgrade** (`:138`). A v3 entry from a newer build gets its dates dropped and is rewritten as v2. Defensible as fail-safe, just noting it's `!==` and not `<` — if that's deliberate, it deserves a word in the comment.

- **Nit on the migration write:** `loadSavedFilters` calls `saveFilters` from the `filters` field initializer, so importing the store writes localStorage at module load. Both sides are try/catch-guarded and this is an SPA with no SSR, so it's fine — just flagging the import-time side effect.

mjacobs:
Thank you!

PR #1201, opened 2026-07-21
fix(frontend): keep the open session across a filtered sidebar reload

### LINKED ISSUE #1190: Project name missing from breadcrumb when opening a session from search results
### agentsview version

agentsview v0.38.1

### Install method

install.sh / install.ps1

### OS / platform

macOS 26.5.2 (arm64)

### Which agent and version

n/a

### Which model(s)

n/a

When I open a session found by searching for text within sessions, the session page breadcrumb does not show the project name.

I expected the breadcrumb to include the project name.

n/a — this appears to be a UI navigation issue.

### Steps to reproduce

1. Open AgentsView.
2. Search for a term that appears within a session. Do not select a recent session.
3. Open a result.
4. Check the breadcrumb on the session page.

**Actual result:** The project name is missing.

**Expected result:** The breadcrumb shows the project name.

### Checklist

bangddong:
Fixes #1190.

Opening a session from search (`navigateToSession`) appends it to the sidebar list even when it falls outside the active sidebar filters — search ignores those filters. The subsequent index reload (`load()`, triggered by route/filter re-evaluation) rebuilt `this.sessions` from the filtered index alone, so the out-of-filter session was dropped. `activeSession` then resolved to `undefined` and the `SessionBreadcrumb` lost its project name, session name, and badges, which is what the reporter saw when opening a search result rather than a recent session.

`load()` now preserves the previously resolved active-session row when the new index omits it, mirroring the append behavior `navigateToSession` already relies on for out-of-index sessions. The row is only re-added when it was a fully hydrated session (not an index-only stub) and is not already present in the new index, so filtered sidebar contents and ordering are otherwise unchanged.

Reviewers: the change is confined to `load()` in `frontend/src/lib/stores/sessions.svelte.ts`; the getter it feeds is `activeSession`, and the affected consumer is `SessionBreadcrumb.svelte` via `App.svelte`.

### DIFF
--- a/frontend/src/lib/stores/sessions.svelte.ts
+++ b/frontend/src/lib/stores/sessions.svelte.ts
@@ -296,6 +296,11 @@ class SessionsStore {
   agents: AgentInfo[] = $state([]);
   machines: string[] = $state([]);
   activeSessionId: string | null = $state(null);
+  // Resolved detail for the open session, held outside the sidebar list so a
+  // session that falls outside the current filters or page (e.g. opened from
+  // search) can back activeSession without polluting groupedSessions or being
+  // double-counted when loadMore appends a later page. Cleared on navigation.
+  activeSessionDetail: Session | null = $state(null);
   activeSessionUsageVersion: number = $state(0);
   childSessions: Map<string, Session> = $state(new Map());
   nextCursor: string | null = $state(null);
@@ -358,7 +363,13 @@ class SessionsStore {
 
   get activeSession(): Session | undefined {
     const session = this.sessions.find((s) => s.id === this.activeSessionId);
-    return session?.is_index_only ? undefined : session;
+    if (session && !session.is_index_only) return session;
+    // The open session may be absent from the filtered/paginated sidebar list
+    // (opened from search, filtered out, or on an unloaded page), or present
+    // only as an index-only stub. Fall back to the separately resolved detail.
+    const cached = this.activeSessionDetail;
+    if (cached && cached.id === this.activeSessionId) return cached;
+    return undefined;
   }
 
   get groupedSessions(): SessionGroup[] {
@@ -526,21 +537,6 @@ class SessionsStore {
       this.sessions = index.sessions.map((row) =>
         sidebarIndexRowToSession(row, existing.get(row.id))
       );
-      // Preserve the open session when the (filtered) index omits it, e.g. one
-      // opened from search that falls outside the current sidebar filters.
-      // navigateToSession appended it to the list; dropping it on reload leaves
-      // activeSession undefined and strips the breadcrumb's project, name, and
-      // badges. Keep the previously resolved row so the detail view stays intact.
-      const activeId = this.activeSessionId;
-      if (
-        activeId !== null &&
-        !index.sessions.some((row) => row.id === activeId)
-      ) {
-        const preserved = existing.get(activeId);
-        if (preserved && !preserved.is_index_only) {
-          this.sessions = [...this.sessions, preserved];
-        }
-      }
       this.nextCursor = index.next_cursor ?? null;
       this.total = index.total;
     } catch {
@@ -808,6 +804,7 @@ class SessionsStore {
     this.refreshRead.cancel();
     this.childSessionsRead.cancel();
     this.activeSessionId = id;
+    this.activeSessionDetail = null;
     this.activeSessionUsageVersion = 0;
     this.refreshVersion++;
     this.childSessionsVersion++;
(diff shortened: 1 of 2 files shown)


bangddong:
Hi, thanks in advance for taking a look when you get a chance — I'd really
appreciate a quick steer on scope before I push further.

This PR (for #1190) started as a small fix: when you open a session from search,
its breadcrumb (project name, session name, badges) was getting lost after a
filtered sidebar reload. The fix caches the active session's detail
(`activeSessionDetail` in `sessions.svelte.ts`) so it survives the sidebar
rebuild.

Since then, the automated review (roborev-ci) has flagged a series of freshness
/ race conditions around that cache, and I've been addressing each path
(navigation, hydration, refresh, rename) one round at a time — we're now seven
rounds in, all from the bot, so I wanted to pause and get a human read before
the scope drifts too far.

The most recent round is more of a design-level point than another isolated bug:

> Gate all active-detail mutations with a per-selection request generation or
> shared latest-read coordinator, and add regression tests covering older
> navigation/hydration responses resolving after a newer refresh.

I have a local (not yet pushed) implementation of exactly that: a monotonic
generation counter where each async detail load claims a generation when it's
issued and only commits if it's still the latest, with every write to
`activeSessionDetail` routed through a single guarded commit so a stale older
response can't overwrite a newer one. Tests pass locally.

Before I push it, though, I'd be grateful for your call on scope, since it turns
a small UX fix into a state-management refactor of the sessions store:

- **A.** Fold the centralized generation-token refactor into this PR.
- **B.** Keep this PR to the minimal fix for #1190 and track the shared
  coordinator as a follow-up issue.
- **C.** Something else you'd prefer.

I'm happy to go whichever way works best for you — I just didn't want to expand
the scope on my own without checking first. Thanks!

mjacobs:
Thanks for the patience here, and for driving this through six rounds of hardening — the hard part (getting the design right: a separate detail cache instead of injecting rows into the filtered sidebar list) was all you, and each round's fix landed with a matching regression test. Much appreciated.

On your scope question: your instinct in **A** was right, and rather than cost you another round-trip I took the liberty of pushing the last piece onto this branch (hope that's okay — "maintainer can modify" was enabled). Two commits:

- `f8e3490` routes every read that commits to `activeSessionDetail` through one shared `LatestRead` (`activeDetailRead`), replacing the independent `navigateRead`/`refreshRead` pair. That's the root cause the bot kept circling: two coordinators that never cancelled each other. `renameSession` now cancels it before applying the rename, and the coordinator subsumes `refreshVersion`. The per-issued-load generation counter you described locally was exactly the right shape — the only change is reusing the existing `LatestRead` utility (from #1146, already used three times in this store) instead of a bespoke counter. Regression tests cover both cross-coordinator orderings: a stale navigation resolving after a newer refresh, and a pre-rename refresh resolving after the rename.
- `9944a13` clears the cache when a watcher refresh 404s for the still-active session — with the cache, a session deleted on another machine otherwise kept a ghost breadcrumb forever (at base it self-cleared on reload). Transient failures deliberately keep the cache.

For perspective: the remaining race was real but self-healing (a momentary field flicker on the same session, never a blanked breadcrumb or wrong session — the id invariant held throughout), so your six commits had already fixed everything user-visible from #1190. With the shared coordinator in, I consider this converged; no further scope expansion needed. I'll watch CI on the new head and take it from here. Thanks again for a genuinely careful contribution.

mjacobs:
Still iterating on the remaining findings.

mjacobs:
Ah, #1190 was recently fixed, independently

mjacobs:
Update on where this landed, and the plan from here.

While this PR was in its final review rounds, #1204 merged and closed #1190 independently — it fixes the same breadcrumb loss with a route-first opening flow, but its store-side approach appends and keeps the open session's row inside the filtered sidebar collection. That is the same shape the very first review round here flagged (a session excluded by the active filters appears in the list and its groups), which is what led this PR to the separate `activeSessionDetail` cache design instead.

Rather than keep rebasing this branch through that collision, the plan is:

- Close this PR in favor of a follow-up branch that keeps #1204's route-first flow and its genuine improvements (self-healing deep-link hydration, pagination dedupe, in-flight navigation join) while replacing the row-injection with the detail-cache design from this PR, plus the race hardening from the later rounds here (shared read coordinator, freshness generation, rename/404/delete handling) and the accompanying regression tests.
- The reconciled stack already exists and preserves authorship — @bangddong's commits land as-is, so the design and its regression suite stay credited to them. Two small remaining review findings (syncing accepted index updates and signal detail into the cached active session) get folded in before it opens.

Thanks again @bangddong — the detail-cache design is the piece that survives all of this, and the follow-up is largely your work landing under a different PR number.

bangddong:
Thanks @mjacobs, and no worries about pushing to the branch — the reconciliation plan makes sense to me. Keeping #1204's route-first opening (which closes the drop window at the source) and swapping the row injection for the detail cache seems like the right split. Reusing the existing `LatestRead` coordinator instead of the bespoke counter I'd sketched is clearly the cleaner fit too — good to know that utility was already there for exactly this. I really appreciate you preserving the authorship and the careful reviews throughout; happy to help with anything on the follow-up if it's useful.

mjacobs:
Closing in favor of carrying remaining improvements into a new PR rebased on main, as noted in previous comment.

PR #1203, opened 2026-07-21
feat(cli): add AGE column to session search results

mjacobs:
Renders the per-match message timestamp in the human output of `agentsview
session search`, so results can be associated with a session by when it
happened rather than only by the snippet text.

The timestamp was already populated on every match in all search modes and
already emitted in `--output json` — the CLI change is presentation-only.

- Table format gains an AGE column immediately after MATCH (before the
  optional SCORE): relative under a week (`3h`, `5d`), `Jan 02` within the
  current year, `Jan 2025` for prior years, em dash when the timestamp is
  missing or unparseable. The year appears for prior years because search
  spans the whole multi-year archive, unlike the resume-oriented
  `session list`.
- The `--context` record format gains the same age token on the match line,
  between the ordinal/score markers and the project.
- The relative buckets are extracted from `humanizeSessionAge` into a shared
  `humanizeAgeRelative` core; `session list` output stays byte-identical,
  and JSON output is unchanged.
- The renderers now take an injected clock (captured once per render) so
  tests pin output against a fixed time.

Following review, the branch also normalizes search-content timestamps at
the backend boundary: the DuckDB quack store and the PostgreSQL read store
(substring/regex scans and the shared semantic/hybrid hit enrichment) were
emitting SQL-style `timestamp::text` / `CAST(... AS TEXT)` strings, which
the AGE parser — and any other RFC3339 consumer — rejects. Both backends
now scan the raw timestamp column and format RFC3339Nano UTC, matching the
SQLite backend and the stores' existing session-timestamp helpers
(`formatDBTime`, `FormatISO8601`), with regression tests asserting the
returned instant equals the inserted one.

Where to look: `cmd/agentsview/session_search.go` (new `humanizeMatchAge`
helper and the two renderer changes),
`cmd/agentsview/session_list_render.go` (the extraction),
`internal/duckdb/store.go` and `internal/postgres/search_content*.go` (the
timestamp normalization).


wesm (approved):
just removing the superpowers specs

PR #1205, opened 2026-07-21
fix(parser): namespace Trae session identities

### LINKED ISSUE #1191: trae: cross-namespace session ID collision silently overwrites a session
Follow-up from #1180 review.

`parseTraeSessionRecord` keys every Trae session as `trae:<sessionID>` (`internal/parser/trae_provider.go:537`) with no container discriminator, while discovery walks two independent stores: `workspaceStorage/*/state.vscdb` and `globalStorage/state.vscdb` (`internal/parser/trae_provider.go:49-61`). The two namespaces therefore share a single session ID space.

Trae session IDs are UUID-ish, so an accidental collision is unlikely, and cross-container dedup was deliberately declined during review to keep parity with the sibling shared-container providers. Windsurf only discovers `workspaceStorage`, though, so it has no second namespace and no equivalent exposure. This is an accepted risk specific to Trae rather than a shape the siblings already carry.

The reason to track it rather than close it: if the same ID does appear in both stores, the rows collide on the sessions primary key and the later parse overwrites the earlier one. There is no error and no warning, so the symptom is a session that silently disappears, which is hard to attribute back to this cause after the fact.

Options: qualify the ID with its container (breaks existing Trae session IDs), or detect the collision at parse time and log or skip instead of overwriting. Worth first confirming whether Trae ever writes the same session to both stores; if the same ID in both places is legitimately one session, overwrite is the correct behavior and this can be closed as a non-issue.

rodboev:
Trae session discovery currently maps workspace and global records with the same raw session ID to one database key, so a later upsert silently replaces the earlier transcript.

This qualifies Trae's internal session IDs as `trae:workspaceStorage:<id>` or `trae:globalStorage:<id>` before they enter the shared sync and database path. Qualified recovery keys now resolve back to the matching storage namespace and bare producer ID, so relocated workspace containers remain findable through the engine fallback. Shared bare-ID consumers, `session get`, `session usage`, and the deprecated `token-use`, keep the old convenience path when exactly one Trae namespace matches and now fail explicitly when workspace and global candidates collide, including host-qualified stored IDs. Raw source session IDs, virtual source paths, transcript parsing, freshness, and tombstone behavior retain their existing semantics; both physical records remain addressable when their raw IDs match. Unsupported container provenance returns an error instead of creating an ambiguous identity.

Trae support landed in #1180 after v0.38.1, so this corrects the identity representation before its first release without changing the database schema. The parser data version now advances so existing archives rebuild legacy `trae:<raw>` rows into namespaced identities and avoid reviving the superseded alias during orphan preservation. PostgreSQL push now also deletes owned legacy `trae:<raw>` rows as their namespaced replacements land, so remote archives converge instead of keeping stale duplicates beside the qualified sessions. The collision case came from @mjacobs' review of the provider.

Closes #1191

### DIFF
--- a/cmd/agentsview/archive_query_backend.go
+++ b/cmd/agentsview/archive_query_backend.go
@@ -243,9 +243,12 @@ func (b localArchiveQueryBackend) SessionUsage(
 	applyCustomPricing(b.database, b.cfg)
 	ensureUsagePricing(b.database, b.offline, b.cfg.CustomModelPricing)
 
-	resolvedID, known := resolveRawSessionID(
+	resolvedID, known, err := resolveRawSessionIDDetailed(
 		ctx, b.database, b.cfg.AgentDirs, sessionID,
 	)
+	if err != nil {
+		return nil, tokenUseExitErr, err
+	}
 
 	if known && !b.skipFreshData {
 		engine := sync.NewEngine(b.database, sync.EngineConfig{
--- a/cmd/agentsview/session_get.go
+++ b/cmd/agentsview/session_get.go
@@ -119,20 +119,25 @@ var traeSessionLookupNamespaces = []string{
 func resolveTraeRawSuffixAmbiguity(id string, matches []string) error {
 	workspaceID := string(parser.AgentTrae) + ":workspaceStorage:" + id
 	globalID := string(parser.AgentTrae) + ":globalStorage:" + id
-	hasWorkspace := false
-	hasGlobal := false
+	workspaceMatch := ""
+	globalMatch := ""
 	for _, match := range matches {
-		switch match {
+		_, stripped := parser.StripHostPrefix(match)
+		switch stripped {
 		case workspaceID:
-			hasWorkspace = true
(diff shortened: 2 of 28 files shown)


rodboev:
Closing this. Before pushing it further I went back and verified the premise on a real install, and it doesn't hold on current builds. On a fresh international Trae 3.5.78, chat transcripts never touch `state.vscdb`: the `memento/icube-ai-agent-storage` key holds one session stub per workspace with an empty `messages` array, `globalStorage/state.vscdb` never gets the key at all, and the actual conversation store is an encrypted database under `ModularData/ai-agent` with timestamp-prefixed session IDs from a single allocator. The cross-store collision this PR namespaces against can't arise in that layout, so the migration surface here protects a case that current Trae doesn't produce. Measurements are written up in #1191. If session-ID namespacing ever becomes worth doing, I think it starts as a design conversation about a generic ID migration path rather than a provider-specific retrofit of this size.

PR #1209, opened 2026-07-21
add Kilo legacy VSCode extension agent support

scross01:
This PR adds support for the original (legacy) Kilo Code agent VSCode extension. 

This agent parser covers the pre-OpenCode legacy extension (the RooCode-derived codebase). Kilo released a rebuilt VS Code extension on an OpenCode core in March 2026, which is already supported by the main `kilo` parser in AgentsView.


scross01:
All automated review feedback addressed. Ready for review.

PR #1215, opened 2026-07-22
fix(frontend): back the open session with a detail cache instead of a row injected into the filtered sidebar

mjacobs:
Supersedes #1201's store internals and replaces the approach #1204 introduced for #1190, while keeping #1204's route-first opening flow.

## Why

#1204 fixed the lost-breadcrumb bug (#1190) by injecting and keeping the open session's row inside the filtered sidebar collection. That leaks the open session into `groupedSessions`: a session excluded by the active filters (opened from search, or fallen off the loaded pages) visibly appears in the sidebar and shifts group membership and pagination accounting. This PR instead backs the open session with a separate `activeSessionDetail` cache in the sessions store: `activeSession` falls back to the cache when the session's row is absent or index-only, and the sidebar collection stays exactly what the filters say it is.

## What it does now

- Keeps #1204's route-first opening flow, deep-link self-healing, and loadMore dedupe; the same-session navigate join is ported onto the shared active-detail read coordinator.
- Coordinates every read that commits to active-detail state (navigation, watcher refresh, rename, sidebar hydration, signal detail) through a latest-read coordinator plus per-session commit generations: a session's generation advances only when fresher state actually commits (navigation/refresh resolved, a rename or hydration committed its snapshot, a 404 committed a deletion), so a read that merely began — and may fail — never invalidates anything, and interleaved responses cannot overwrite fresher detail with a stale snapshot. Generations are per session id, so activity on one session never invalidates an in-flight hydration of another — a selection may join that hydration and depend on its result. A watcher refresh likewise joins a pending same-session navigation (both issue the identical detail GET) instead of cancelling it, so a transiently failing refresh can never strand a deep link with an empty cache.
- Enforces the invariant that every update a sidebar row receives, the cache receives: hydration merges, renames, delete/404 removals, index commits in `loadSidebarPage`/`loadMore`, an index-only stub of a cache-only session re-entering the index, and signal-detail merges all sync the cache.
- After every index commit the active session's row and cache reconcile in whichever direction is fresher, gated on committed state: the merged row resyncs the cache when nothing newer committed mid-flight (a just-committed index is at least as fresh as any commit that resolved before the index request was issued); a rename, detail refresh, or hydration that committed while the index was in flight replaces the re-listed row; and a 404 deletion that committed mid-flight removes the re-listed row (tombstone). The same ordering applies in the other direction: a navigation/refresh/hydration response that resolves after an index commit re-published its row contributes only detail-owned fields, keeping the committed index's fields (display_name, counts) instead of reverting them (falling back to the cache when a later reload dropped the row). Ordering is by request issue: every commit records the index-request ordinal its underlying request began at (renames and 404s are write-derived and always win), and committed detail outranks an index commit only when its request is at least as new as the index's. A rename that lands on an empty cache also chases with a fresh detail read, since the rename endpoint's response lacks the derived detail-only fields.
- A definitive 404 for the active session drops both the cached detail and the sidebar row, so a session deleted on another machine empties the breadcrumb instead of lingering as a ghost.

## Tradeoffs and limitations

- Freshness generations are scoped to active-detail reads, which only exist for the open session: a rename of a session that is not open can still be transiently reverted in the sidebar by an index response already in flight, until the next refresh corrects it.
- A remote-delete 404 leaves `activeSessionId` set (only the detail and row are dropped) so the selection can recover if the session reappears; local delete still clears the selection. The asymmetry is deliberate.

## Where to look

Nearly all changes are in `frontend/src/lib/stores/sessions.svelte.ts` and its colocated test file; `frontend/src/lib/utils/latest-read.ts` holds the read coordinator.

## Credit

The detail-cache design and its original regression suite are @bangddong's work from #1201 (the first six commits here are theirs); this PR carries that design forward, rebased onto current main and reconciled with #1204's flow, with additional race hardening on top.

### DIFF
--- a/frontend/src/lib/stores/sessions.svelte.ts
+++ b/frontend/src/lib/stores/sessions.svelte.ts
@@ -551,18 +551,7 @@ class SessionsStore {
       );
       this.nextCursor = index.next_cursor ?? null;
       this.total = index.total;
-      // An included active row may have absorbed index refreshes (renames,
-      // counts) that the detail cache never saw; resync it so a later reload
-      // that excludes the row doesn't revert the breadcrumb to stale fields.
-      // Skip when a navigation/refresh/rename committed for the now-active
-      // session while this index was in flight — the cache is then fresher
-      // than the merged row.
-      if (
-        this.activeSessionId !== null &&
-        this.detailGenerationUnchanged(detailGenerations, this.activeSessionId)
-      ) {
-        this.cacheActiveSessionDetailFromList(this.activeSessionId);
-      }
+      this.syncActiveSessionAfterIndexCommit(detailGenerations);
     } catch {
       // Restore previous state so a transient failure
       // doesn't wipe the visible session list.
@@ -750,14 +739,7 @@ class SessionsStore {
       ];
       this.nextCursor = index.next_cursor ?? null;
       this.total = index.total;
-      // A re-listed active row can carry index refreshes the detail cache
-      // never saw; resync it (same reasoning as loadSidebarPage).
-      if (
-        this.activeSessionId !== null &&
-        this.detailGenerationUnchanged(detailGenerations, this.activeSessionId)
-      ) {
-        this.cacheActiveSessionDetailFromList(this.activeSessionId);
-      }
+      this.syncActiveSessionAfterIndexCommit(detailGenerations);
     } catch (error) {
       if (signal.aborted || isAbortError(error)) return;
       throw error;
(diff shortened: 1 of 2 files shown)


mjacobs:
stopping, this ran all night and not converging, I'll try to come back to analyze it at some point but it's not a top priority

mjacobs:
This attempt is in a bad state - it deserves a fresh implementation and should be a lot simpler.

Suggestions from Codex 5.6 Sol:
A fresh implementation should use much simpler ownership:
  - Sidebar results exclusively own membership, ordering, pagination, and totals.
  - Active-session detail exclusively owns the open session; the UI should not prefer a hydrated sidebar row over it.
  - Active detail reads should be coalesced per session, with one pending follow-up when an event arrives during a read.
  - Rename/delete/restore should invalidate and refresh the relevant source instead of locally reconciling stale index snapshots.
  - Brief sidebar staleness after a failed refresh should be acceptable; don’t build a client-side transaction protocol to keep every intermediate count exact.
  The tests are generally behavior-oriented rather than tautological. The problem is their scope: many late tests now enforce exact intermediate ordering under rare interleavings rather than the core observable contract. I would port only a compact set covering search/deep-link      
  breadcrumb persistence, exclusion from filtered results, pagination deduplication, rename, deletion, and refresh coalescing.
  If preserving a checkpoint is important, 59eee80c—the end of the original six commits—is the only sensible semantic boundary. But it already had lifecycle holes, so I would use it as design reference, not cherry-pick it wholesale.

PR #1217, opened 2026-07-22
trae: surface unsupported encrypted modern layout

### LINKED ISSUE #1210: trae: current international builds keep transcripts in an encrypted store, so the provider emits no sessions
Splitting this out of #1191, since it's independent of the collision question there.

On a fresh international Trae install (3.5.78 stable, Windows, VS Code base 1.107.1), the provider from #1180 discovers the stores but emits nothing. `memento/icube-ai-agent-storage` in `workspaceStorage/*/state.vscdb` holds one stub per workspace shaped like `{"list":[{"isCurrent":true,"sessionId":"6a5f87dc…","messages":[]}],"currentSessionId":…}`, and `globalStorage/state.vscdb` doesn't carry the key at all. I checked both while the app was running and after a full quit; quitting rewrites the files and still leaves `messages` empty. Discovery isn't the gap, `Trae/User` is already in the default roots (`internal/parser/types.go:327`); the sessions just parse to nothing because `parseTraeSessionRecord` drops zero-message entries (`internal/parser/trae_provider.go:530`).

The actual conversation data lives in `%APPDATA%\Trae\ModularData\ai-agent\database.db`, an encrypted SQLite (the header has no `SQLite format 3` magic), with per-session snapshot directories under `ModularData\ai-agent\snapshot\<sessionId>`. The only plaintext trace of a chat in any vscdb is the `icube-ai-agent-storage-input-history` key, which holds what was typed into the box rather than the transcript. So the inline-messages shape the parser decodes looks like the legacy layout from the original #1178 report (`TRAE SOLO CN`); current international builds have moved on.

Since the store is encrypted and there's no plaintext side-channel (the snapshot dirs are git copies of workspace files; input-history holds only typed prompts), parsing it is out per the same policy that keeps the Antigravity `.pb` decryptor frozen as a last resort. What fits instead is the pattern #953/#969 established for unrecognized Antigravity schemas: detect the layout (memento stubs with empty `messages`, or the key absent, alongside `ModularData/ai-agent/database.db`) and surface it through `doctor` and the sync anomaly counters. The one difference is that the `agy-schema:` marker rides on emitted sessions, and this layout emits none, so the store-level surfaces would carry it alone.

rodboev:
Current international Trae builds can still expose the `Trae/User` state stores that AgentsView already discovers, while moving the actual transcript rows into `ModularData/ai-agent/database.db`, an encrypted non-SQLite store. The visible `state.vscdb` files now carry empty session stubs or no session key, so the provider decodes zero usable inline sessions and silently returns a no-session skip. That makes a modern encrypted Trae install indistinguishable from an empty or broken one.

This change follows the Antigravity direction and surfaces the unsupported state instead of trying to read data AgentsView cannot read. A Trae-specific layout owner in `internal/parser` classifies the measured store shape, the provider maps that state to the existing `SkipUnsupportedSource` reason, sync records it in the existing per-run parser anomaly summary, and `agentsview doctor sync` reports the affected Trae roots directly from disk. The encrypted database stays untouched.

Legacy behavior stays intact. Inline-message Trae stores still parse exactly as before, valid explicit empty stores stay quiet, incomplete or malformed stores stay on the current preservation path, the existing Trae discovery roots remain unchanged, and the HTTP and SSH exclusions added with Trae support remain unchanged. The new signal only fires when the discovery-visible Trae store yields no usable inline sessions and the sibling `ModularData/ai-agent/database.db` matches the measured unsupported encrypted layout.

Closes #1210

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,7 @@ thread JSON files.
 | VSCode Copilot        | `~/Library/Application Support/Code/User/` (macOS)                                                                                                                      |
 | Visual Studio Copilot | `%LOCALAPPDATA%\\Temp\\VSGitHubCopilotLogs\\traces\\` (Windows), `~/Library/Caches/VSGitHubCopilotLogs/traces/` (macOS), `~/.cache/VSGitHubCopilotLogs/traces/` (Linux) |
 | Windsurf              | `~/Library/Application Support/Windsurf/User/` (macOS), `~/.config/Windsurf/User/` (Linux), `%APPDATA%\\Windsurf\\User\\` (Windows)                                     |
+| Trae                  | `%APPDATA%\\Trae\\User\\` (Windows), `~/Library/Application Support/Trae/User/` (macOS), `~/.config/Trae/User/` (Linux)                                                                                             |
 | Warp                  | `~/.warp/` (platform-dependent)                                                                                                                                         |
 | WorkBuddy             | `~/.workbuddy/projects/`                                                                                                                                                |
 | ZCode                 | `~/.zcode/cli/db/`, `~/.zcode/cli/`                                                                                                                                     |
--- a/cmd/agentsview/doctor.go
+++ b/cmd/agentsview/doctor.go
@@ -52,6 +52,7 @@ type doctorSyncReport struct {
 	doctorDBInspection
 	TempFiles           []string
 	AgentRoots          []doctorAgentRoot
+	TraeEncryptedRoots  []string
 	DebugLines          []string
 	DebugLogErr         error
 	HasResyncFailureLog bool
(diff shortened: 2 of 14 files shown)


rodboev:
Both findings are the same case: unsupported evidence beside a store that still parses. The current behavior is intentional. Detection is scoped to stores yielding zero sessions because that's the only captured layout, and an empty stub is also what a just-opened chat looks like on an inline-writing build; warning on stubs whenever the agent database exists would flag healthy installs every run.

The mixed root already surfaces through sync: stores classify independently, so the stub-only or key-less store reports `SkipUnsupportedSource` in the run summary. Doctor defers while part of the root still parses. If a capture ever shows readable records and encrypted transcripts in one install, extending the classifier is a small follow-up.

PR #1218, opened 2026-07-22
fix(sync): bound OpenCode watcher updates

### LINKED ISSUE #1208: fix(sync): bound OpenCode background work across watcher and polling modes
# OpenCode Bounded Background Sync Design

## Context

Issue #1208 and PR #1218 attempted to stop writes to a shared OpenCode SQLite
database from fanning out to every virtual session. The proposed fast path read
OpenCode's event journal after a database or WAL watcher event, deferred
streaming part updates, and parsed only sessions that reached a settlement
event.

That direction optimized only the native watcher callback. It did not cover the
generic fallback used when native watch coverage is unavailable.

A production-scale diagnosis exposed the gap:

- The process-wide recursive watch budget was exhausted at 8,192 directories.
- OpenCode received no recursive coverage and was transferred to the generic
  two-minute unwatched-root poller.
- Every poll called authoritative `ReconcileWatchRoots`, bypassing the journal
  classifier and enumerating the OpenCode archive.
- A poll took longer than two minutes, so one queued ticker wake started the
  next poll immediately. The daemon had no idle interval.
- The process used roughly one CPU core continuously and performed hundreds of
  gigabytes of logical reads in less than an hour.
- A captured stack was in `pollUnwatchedRootsOnce -> ReconcileWatchRoots ->
  parseOpenCodeDBSession -> encoding/json.Unmarshal`, decoding a 68.7 MB
  message from a 23.7 GB OpenCode database. The remaining parser workers were
  waiting behind the parse-retention budget.

Large and hybrid OpenCode installations are especially likely to enter this
fallback because the current provider declares the complete root recursive when
both legacy storage and `opencode.db` exist. They are therefore exactly the
installations for which the watcher-only optimization is least reliable.

This design supersedes the previous #1208 specification and the implementation
direction in PR #1218. The replacement starts from `main`; it does not preserve
or extend the current branch implementation.

## Governing Invariant

Every recurring in-process OpenCode background pass must scale with configured
roots and the newly changed batch, never with total archive cardinality.

In steady state:

```text
work = O(configured roots + admitted journal rows + changed session IDs)
```

Archive-scale OpenCode discovery is allowed only during an explicit
authoritative startup or repair audit. It must never run directly from a native
watcher callback, a degraded-coverage poll, a queued retry, or journal overflow.

## Goals

- Negligible CPU while OpenCode is idle, regardless of archive size.
- Native and degraded SQLite coverage use the same bounded logical change feed.
- A one-session update performs the same classification, metadata lookup, and
  parse count in a small archive and a production-scale archive.
- Streaming part updates do not repeatedly parse the active session's complete
  history.
- Recursive watch-budget exhaustion cannot prevent bounded SQLite observation.
- Polls are scheduled from completion and cannot become a continuous loop when
  one pass exceeds its nominal interval.
- Passive memory is bounded by configured roots and fixed changed-batch limits.
- Hybrid storage shadowing, deletion recovery, and persistent archive behavior
  remain correct.
- Startup and repair audits remain authoritative without coupling source
  journal state to the agentsview database transaction.

## Non-goals

- Persisting an OpenCode journal cursor or pending-session set.
- Adding a generic durable acknowledgement or retry protocol.
- Incrementally parsing individual messages or parts from an OpenCode session.
- Eliminating the cost of one necessary full parse of a changed session.
- Treating Kilo, MiMo Code, or Icodemate as journal-compatible without separate
  format evidence.
- Raising the recursive watch budget until the problem disappears.
- Changing SQLite archive schema, PostgreSQL, DuckDB, frontend, or server APIs.
- Running an authoritative archive scan every two minutes while coverage is
  degraded.

### Provider-owned coverage units

OpenCode must describe SQLite and file-backed storage as separate coverage
units instead of one recursive configured root.

The SQLite coverage unit shallow-watches the configured OpenCode root. One
directory watch observes direct changes to `opencode.db`, its WAL, atomic
database replacement, and storage-root lifecycle. It does not recurse through
logs, caches, binaries, or session trees.

Storage coverage uses recursive roots only for parser-owned storage subtrees.
The plan must preserve independently attributable SQLite and storage failures
so a failure transfers only the affected unit to degraded handling. The exact
number of storage coverage keys is an implementation choice.

`WatchRoot` gains a stable provider coverage key. Polling obligations retain the
provider, configured root, physical probe, and coverage key; they must not
collapse back to an untyped logical root.

All shallow roots are registered before recursive roots consume the shared
budget. Recursive allocation must not allow an earlier provider to starve a
later provider's shallow coverage. Missing storage subtrees are activated from
bounded ancestor lifecycle events when they appear.

Parser-visible symlinked storage directories require explicit coverage. If the
native backend cannot recurse through a directory symlink that discovery reads,
that path becomes a degraded storage obligation rather than being reported as
fully covered.

### Bounded change-feed capability

Add an optional provider capability for a bounded logical change feed. It is
unsupported by default and instantiated only for providers that declare it.
OpenCode is the first provider to opt in.

The feed accepts a coverage key and a reason:

- Native filesystem paths.
- A degraded-coverage poll wake.
- A lost-event or coverage-discontinuity notification.

It returns a bounded result containing logical session changes, whether more
journal work remains, and whether an authoritative audit is required. The
scheduler consumes this result; generic changed-path classification and generic
root reconciliation are not alternate OpenCode paths.

The provider capability is the scheduling contract. The daemon must not infer
journal behavior from an agent name or from a compatible-looking database.

### SQLite scheduling

Healthy shallow SQLite coverage is event-driven and has no idle timer.

When shallow coverage is unavailable, the shared coordinator performs a cheap
container probe on the degraded polling cadence. It drains the bounded journal
only when the container state or journal high-water mark changed. It never
calls `ReconcileWatchRoots` for that SQLite coverage unit.

The coordinator has at most one active pass per provider root. A wake received
while work is active coalesces into one pending wake. The next scheduled poll is
measured from completion, not from the previous ticker deadline. Transient
failures use bounded backoff. A pass that lasts longer than the nominal interval
therefore cannot restart immediately in a loop.

When the journal reports more work, continuation uses a short completion-based
cooldown so a backlog makes progress without monopolizing a core. There is no
per-session timer.

### Journal state and bounds

Maintain one disposable in-memory state entry per recognized OpenCode SQLite
container. It contains only:

- The last consumed journal row ID.
- A compact anchor identifying the consumed row.
- A bounded set of session IDs with intermediate events pending.
- A bounded set of ready session IDs awaiting targeted processing.

State contains no raw event payloads, transcript content, open descriptors, or
per-session timers. It is never persisted in agentsview's SQLite archive.

Each drain is bounded independently by:

- Journal row count.
- Total event payload bytes admitted for classification.
- Distinct pending and ready session IDs.
- Wall-clock or cancellation budget where required for shutdown responsiveness.

The first implementation must choose explicit constants and pin them in tests.
Row count alone is not a sufficient bound because an event payload can grow
with generated content.

Known event types that can be classified from row ID, aggregate ID, and type
must not read the event payload. Read journal metadata and payload sizes before
admitting payload bytes for the remaining types. A row that exceeds the byte
budget must not be materialized merely to discover that it is too large. The
implementation must bound source bytes read and SQLite JSON work, not only Go
heap allocation after SQLite has already scanned the payload.

Consume an admitted prefix and advance only through its last consumed row. If
more rows remain, return `More` and continue after the scheduler cooldown. Do
not jump to the captured high-water mark and discard an overflowed range.

Capture one fixed high-water boundary for a logical drain. Hold tentative ready
IDs until every bounded page through that boundary has been classified. This
lets an intermediate event on a later page cancel an earlier settlement before
the session is parsed. Events committed above the boundary belong to the next
drain.

If an individual row cannot be classified within the fixed bounds, request one
authoritative audit and stop retrying that row in a hot loop. The unsupported
cause remains latched until the container or schema state changes or the daily
audit runs; repeated equivalent rows cannot schedule repeated archive scans. A
successful audit rebaselines the disposable feed state.

### Settlement behavior

Recognized intermediate message and part events update only the bounded pending
set. They do not parse a transcript.

A session becomes ready after a positively verified settlement boundary:

- A user message.
- A completed assistant message.
- A failed assistant message.
- A session metadata event that affects archived output.
- A positively identified message or part removal, which triggers targeted
  replacement even when no later completion event follows.

Event ordering within a journal range is authoritative. A later intermediate
event cancels readiness established by an earlier settlement for the same
session until another settlement arrives.

No quiet-period timer is required for individual sessions. If OpenCode exits
without writing a settlement event, startup or the next authoritative audit
repairs the final state. This is a deliberate promptness tradeoff that avoids a
timer count proportional to active sessions.

### Targeted metadata and canonical source resolution

For each ready ID, query only that session's metadata and its matching project
metadata. Prefer one targeted join. Never list the complete `session` or
`project` table to resolve or parse a changed journal batch, and do not retain a
project map whose size follows the archive.

Batch-load existing canonical archive paths for the ready IDs:

- If the archive points to the matching SQLite virtual source, process that
  SQLite session.
- If the archive points to a storage source, preserve storage as canonical and
  process it only when the corresponding storage coverage reports a change.
- For a new hybrid session, use targeted metadata such as `project_id` to test
  its expected storage path directly. If canonicality cannot be proven with
  bounded lookups, defer that ID and request one audit.

The journal-selected source carries a narrow content-change hint. It bypasses
coarse pre-parse container, mtime, size, and skip-cache gates, but it does not
force replacement and does not bypass the post-parse session fingerprint.

Missing metadata is not proof of deletion. A source-row disappearance preserves
the archived session until an authoritative audit resolves ownership and any
hybrid replacement. Existing transcript payloads are never hard-deleted merely
because a live source vanished.

### Storage scheduling

Native storage events continue to identify changed files and sessions directly.
Their processing must remain bounded by the changed paths. A missing or
undecodable part must not search every message directory to recover its session
ID; use bounded stored ownership hints or defer the ambiguous deletion to the
coalesced audit path.

Legacy file-backed storage has no equivalent source journal. If recursive
storage coverage is lost, determining all missed changes requires enumeration.
The degraded poller must therefore not pretend to provide a cheap incremental
scan.

Instead, storage uncertainty records one coalesced audit request for the
provider root. The worker runs after a quiet cooldown and with failure backoff.
While coverage remains degraded, the regular daily audit is the recurring
backstop; a two-minute poll must not repeatedly scan storage.

Filesystem event overflow follows the same rule: SQLite coverage drains the
journal from its last verified anchor, while storage coverage records one audit
request. Neither case synchronously starts full discovery in the watcher.

### Authoritative audits

Startup reconciliation and explicit repair audits are the only archive-scale
paths. Repair audits run in the existing worker-process architecture so the
long-running daemon does not retain their temporary memory.

Audit requests are deduplicated by provider root and reason. Repeated watcher
errors, unknown journal events, unsupported schemas, or storage overflow cannot
queue repeated scans. Failure applies backoff rather than a tight retry.

An OpenCode database created after startup is audit-first. Capture its journal
anchor, run one authoritative audit, and install the pre-audit anchor only after
success so initial rows and events committed during initialization are not
silently baselined away.

Before an audit, capture the journal anchor. Install that pre-audit baseline
only after the audit succeeds so events committed during discovery remain above
the watermark and are consumed afterward. A failed or cancelled audit does not
clear pending change state.

The daily audit remains the safety net for abnormal termination without a
settlement event, ambiguous hybrid transitions, source deletion, and legacy
storage changes missed while coverage was degraded.

### Database replacement and discontinuity

The feed validates its saved row anchor before continuing. A lower high-water
mark, a missing anchor, or a different event at the same row ID marks the
database state untrusted without depending on platform-specific inode or device
identity.

A discontinuity requests one authoritative audit and suppresses hot retries.
After a successful audit, install a fresh baseline. A replacement that preserves
the complete observed prefix is equivalent to a continuation.

### Parse cost boundary

A settled OpenCode session is still parsed as a complete logical session. That
work is O(changed session history), and a session containing an unusually large
message can use a core while it is parsed.

This design guarantees that the parse occurs at most once for the admitted
settlement and is never multiplied by unchanged archive cardinality, streaming
part events, polling ticks, or watcher recovery. Incremental OpenCode transcript
parsing is separate future work.

## Failure Semantics

- No native SQLite coverage: bounded probe and journal drain, never root
  reconciliation.
- No native storage coverage: one coalesced repair audit, then the daily audit.
- Journal backlog: consume bounded prefixes with completion-based cooldown.
- Oversized or unknown journal row: request one audit and avoid a hot retry.
- Pending-ID capacity reached: stop admission and request one audit; do not grow
  memory or enumerate the archive.
- Repeated unsupported rows: keep one latched audit reason until source state
  changes; do not schedule one audit per event.
- Transient SQLite lock: retain the cursor and retry with bounded backoff.
- Targeted parse or write failure: retain a bounded ready ID or defer it to one
  audit; do not create an unbounded generic watcher retry loop.
- Database replacement: invalidate disposable state and request one audit.
- Source deletion: preserve archived content until authoritative ownership
  reconciliation.
- Shutdown: cancel active reads and workers; no per-session timer delays exit.
- Unsupported event schema: no journal targeting and no recurring in-process
  full scan.
- Empty WAL creation, SHM changes, and checkpoint-only notifications caused by
  read-only SQLite access: no-op without reopening a feedback loop.

### Increase the recursive watch budget

This is host-dependent, watches unrelated trees, and still fails under resource
exhaustion or runtime coverage loss. It does not fix generic fallback polling.

### Add only a shallow database watch

This fixes the observed registration starvation but leaves storage degradation,
watcher overflow, fully unavailable watcher backends, and generic polling
unbounded.

### Call the current #1218 hook from the poller

The current hook is changed-path-specific, bounds rows but not bytes, discards
overflowed ranges, and does not model coverage or scheduling. Reusing it would
preserve the design gap instead of establishing one bounded path.

### Poll container mtime and reconcile when it changes

Container state proves that something changed but cannot identify the session.
It still fans out across the shared database under active use.

### Poll storage directory mtimes

Child rewrites do not reliably update parent directory mtimes. Correctness still
requires per-file enumeration, which belongs in a cooled-down audit.

### Persist journal acknowledgement state

Durable cursors couple source progress to archive writes, schema migration,
database swaps, and retry transactions. Startup and authoritative audits already
provide the required correctness boundary.

### Audit immediately after every failure

This converts event storms into archive-scale work. Audit requests must be
coalesced, cooled down, and backed off.

## Implementation Boundaries

The replacement should be a clean implementation from `main`, not a
compatibility layer over PR #1218.

Expected areas of change:

- Provider capability and watch-root metadata.
- Watch registration ordering and provider-aware polling obligations.
- Shared bounded change-feed scheduling.
- OpenCode journal reading, settlement classification, and targeted metadata.
- OpenCode canonical hybrid-source lookup.
- Worker audit request coalescing and journal rebasing.
- Deterministic operation counters and regressions.
- OpenCode format provenance for the event schema and settlement semantics.

The design must preserve generic behavior for providers that do not advertise a
bounded change feed. New expensive scheduling inputs default to unsupported and
are computed only for providers that observably consume them.

## Acceptance Criteria

- With the recursive budget already exhausted, the shallow OpenCode SQLite root
  still registers and one journaled update produces one targeted parse.
- With the filesystem watcher entirely unavailable, a journaled update is found
  by bounded provider polling without invoking OpenCode discovery or
  `ReconcileWatchRoots`.
- Archives with 10 and 5,000 unchanged SQLite sessions perform identical journal
  row reads, session and project metadata row reads, canonical lookups, parses,
  retained project entries, and writes for one changed session.
- Journal work is capped by rows, bytes, and distinct IDs. A payload larger than
  the byte budget is not materialized by the daemon.
- A journal backlog spans multiple bounded pages without skipping rows or
  starting a tight continuation loop.
- A settlement at the end of one journal page followed by an intermediate event
  on the next page causes no premature parse.
- Thousands of intermediate part events cause zero transcript parses and one
  parse after the final settlement.
- A very large changed session is parsed once for one settlement, not once per
  poll or event batch.
- A degraded poll that exceeds its configured interval waits for a
  completion-based cooldown before another pass.
- SQLite watcher overflow recovers through the journal. Storage overflow records
  one audit request and performs no synchronous archive scan.
- Known SQLite-backed and storage-backed hybrid sessions preserve their
  canonical source. A new hybrid session uses bounded direct resolution or
  defers to one audit.
- Missing source rows preserve archived sessions and transcripts.
- Standalone message and part removals trigger one targeted replacement without
  scanning all message directories.
- A database first created after startup is audited without losing its initial
  rows or events committed during the audit.
- Database replacement and journal discontinuity request one audit, then process
  events committed during that audit after it succeeds.
- Unsupported schemas and transient locks cause no archive enumeration and no
  hot retry loop.
- Empty WAL, SHM, and checkpoint-only events cause no reader-induced callback
  loop.
- Parser-visible storage symlinks are either natively covered or represented by
  explicit degraded obligations.
- Passive feed memory scales with configured OpenCode roots and fixed active-ID
  limits, not stored session count.
- A production-scale isolated profile shows negligible idle CPU and no recurring
  stack under `pollUnwatchedRootsOnce -> ReconcileWatchRoots ->
  parseOpenCodeDBSession`.

Tests use deterministic operation counts and fake clocks rather than wall-clock
timings. Cardinality regressions compare small and large archives while holding
the changed batch constant. Production-scale profiling uses isolated source and
archive clones, never live agent data.

## Delivery

Close or supersede PR #1218 after this design is accepted. Build the replacement
from current `main` on a new feature branch. Prefer focused commits for coverage
planning, bounded scheduling, OpenCode feed semantics, and regressions rather
than carrying forward the existing merge history.

berenddeboer:
**Problem**

During active OpenCode use, agentsview repeatedly consumed roughly 244% to 937% CPU. Profiling showed 89.5% of CPU time under OpenCode SQLite parsing: each write to `opencode.db` or its WAL was expanded to every virtual session, while streamed part updates repeatedly reparsed the active session's full transcript. Watcher work therefore scaled with archive size and transcript history instead of the changed batch.

**Fix**

OpenCode watcher events now use a capped journal range read and targeted metadata lookup. Intermediate message and part events are held in a bounded, process-local pending set; user messages, completed or failed assistant messages, and session events settle the session and trigger one targeted parse. Journal-selected sources bypass coarse freshness checks while retaining post-parse content fingerprint filtering and hybrid storage shadowing.

This deliberately replaces #1189's durable cursor acknowledgement, quiet-period scheduler, and retry subsystem. Live watcher sync is a bounded best-effort optimization with no persisted state or timers; startup and periodic full sync remain the authoritative repair path for overflow, incompatible schemas, transient failures, and deferred hybrid sessions.

Closes #1208.

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -341,6 +341,11 @@ func runServe(cfg config.Config, opts serveOptions) {
 			},
 		)
 		defer stopWatcher()
+		// Establish the live engine's disposable journal watermark after
+		// collection begins and before startup reconciliation. An in-process
+		// full sync may replace it; worker-backed startup needs it so the first
+		// queued settlement is not consumed as a lazy baseline.
+		engine.InitializeOpenCodeWatchBaseline(ctx)
 		onStartupReconciled = newStartupReconciliationHandler(
 			ctx,
 			database.CheckpointWALTruncateWithRetry,
--- a/internal/parser/opencode.go
+++ b/internal/parser/opencode.go
@@ -27,6 +27,203 @@ type OpenCodeSessionMeta struct {
 	FileMtime   int64
 }
 
+// OpenCodeJournalEvent is the session-level interpretation of one recognized
+// OpenCode event row. Settlement is false for streaming updates that should be
+// coalesced until a turn boundary arrives.
+type OpenCodeJournalEvent struct {
+	SessionID  string
+	Settlement bool
+}
+
+// OpenCodeJournalBatch is a bounded, transactionally consistent event range.
+// Supported is false when the source does not expose the verified OpenCode
+// journal schema. Safe is false when targeted classification cannot be proven.
+type OpenCodeJournalBatch struct {
+	HighRowID int64
+	Events    []OpenCodeJournalEvent
+	Supported bool
+	Safe      bool
+	Overflow  bool
+}
+
+// OpenCodeJournalHighWater reads the current journal rowid without enumerating
+// sessions. The journal is an optimization; callers treat unsupported schemas
+// and read failures as reasons to wait for authoritative full sync.
+func OpenCodeJournalHighWater(
+	ctx context.Context, dbPath string,
+) (int64, bool, error) {
+	db, err := openOpenCodeDBContext(dbPath)
+	if err != nil {
+		return 0, false, err
+	}
+	defer db.Close()
+	supported, err := openCodeEventJournalSupported(ctx, db)
+	if err != nil || !supported {
+		return 0, supported, err
+	}
+	var high int64
+	if err := db.QueryRowContext(ctx,
+		"SELECT COALESCE(MAX(rowid), 0) FROM event",
+	).Scan(&high); err != nil {
+		return 0, true, fmt.Errorf("reading opencode journal high-water mark: %w", preferContextError(ctx, err))
+	}
+	return high, true, nil
+}
+
+// ReadOpenCodeJournal reads at most limit rows plus one overflow sentinel from
+// (afterRowID, currentHighRowID] in one read transaction.
+func ReadOpenCodeJournal(
+	ctx context.Context, dbPath string, afterRowID int64, limit int,
+) (OpenCodeJournalBatch, error) {
+	if err := ctx.Err(); err != nil {
+		return OpenCodeJournalBatch{}, err
+	}
+	if limit <= 0 {
+		return OpenCodeJournalBatch{}, fmt.Errorf("opencode journal limit must be positive")
+	}
+	db, err := openOpenCodeDBContext(dbPath)
+	if err != nil {
+		return OpenCodeJournalBatch{}, err
+	}
+	defer db.Close()
+	supported, err := openCodeEventJournalSupported(ctx, db)
+	if err != nil || !supported {
+		return OpenCodeJournalBatch{Supported: supported}, err
+	}
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		return OpenCodeJournalBatch{}, fmt.Errorf("starting opencode journal read: %w", preferContextError(ctx, err))
+	}
+	defer func() { _ = tx.Rollback() }()
+	batch := OpenCodeJournalBatch{Supported: true, Safe: true}
+	if err := tx.QueryRowContext(ctx,
+		"SELECT COALESCE(MAX(rowid), 0) FROM event",
+	).Scan(&batch.HighRowID); err != nil {
+		return OpenCodeJournalBatch{}, fmt.Errorf("reading opencode journal high-water mark: %w", preferContextError(ctx, err))
+	}
+	if batch.HighRowID < afterRowID {
+		batch.Safe = false
+		return batch, nil
+	}
+	rows, err := tx.QueryContext(ctx, `
+		SELECT aggregate_id, type, data
+		FROM event
+		WHERE rowid > ? AND rowid <= ?
+		ORDER BY rowid
+		LIMIT ?
+	`, afterRowID, batch.HighRowID, limit+1)
+	if err != nil {
+		return OpenCodeJournalBatch{}, fmt.Errorf("listing opencode journal events: %w", preferContextError(ctx, err))
+	}
+	defer rows.Close()
+	rowsRead := 0
+	for rows.Next() {
+		if err := ctx.Err(); err != nil {
+			return OpenCodeJournalBatch{}, err
+		}
+		if rowsRead == limit {
+			batch.Overflow = true
+			batch.Safe = false
+			break
+		}
+		rowsRead++
+		var aggregateID, eventType, data string
+		if err := rows.Scan(&aggregateID, &eventType, &data); err != nil {
+			return OpenCodeJournalBatch{}, fmt.Errorf("scanning opencode journal event: %w", err)
+		}
+		event, ok := interpretOpenCodeJournalEvent(aggregateID, eventType, data)
+		if !ok {
+			batch.Safe = false
+			continue
+		}
+		batch.Events = append(batch.Events, event)
+	}
+	if err := rows.Err(); err != nil {
+		return OpenCodeJournalBatch{}, fmt.Errorf("listing opencode journal events: %w", preferContextError(ctx, err))
+	}
+	if err := tx.Commit(); err != nil {
+		return OpenCodeJournalBatch{}, fmt.Errorf("finishing opencode journal read: %w", preferContextError(ctx, err))
+	}
+	return batch, nil
+}
+
+func interpretOpenCodeJournalEvent(
+	aggregateID, eventType, data string,
+) (OpenCodeJournalEvent, bool) {
+	if !strings.HasPrefix(aggregateID, "ses_") || aggregateID == "ses_" || !json.Valid([]byte(data)) {
+		return OpenCodeJournalEvent{}, false
+	}
+	event := OpenCodeJournalEvent{SessionID: aggregateID}
+	switch {
+	case strings.HasPrefix(eventType, "session."):
+		event.Settlement = true
+	case eventType == "message.updated.1":
+		role := gjson.Get(data, "info.role").String()
+		completed := gjson.Get(data, "info.time.completed").Exists()
+		failed := gjson.Get(data, "info.error").Exists()
+		if role != "user" && role != "assistant" && role != "" {
+			return OpenCodeJournalEvent{}, false
+		}
+		event.Settlement = role == "user" || completed || failed
+	case eventType == "message.removed.1",
+		eventType == "message.part.updated.1",
+		eventType == "message.part.removed.1":
+		// Intermediate transcript mutation; wait for a settlement event.
+	default:
+		return OpenCodeJournalEvent{}, false
+	}
+	return event, true
+}
+
+// ListOpenCodeSessionMetaByID loads only selected session rows. Missing rows
+// are omitted so watcher-time source deletion never becomes archive deletion.
+func ListOpenCodeSessionMetaByID(
+	ctx context.Context, dbPath string, sessionIDs []string,
+) ([]OpenCodeSessionMeta, error) {
+	if len(sessionIDs) == 0 {
+		return nil, nil
+	}
+	db, err := openOpenCodeDBContext(dbPath)
+	if err != nil {
+		return nil, err
+	}
+	defer db.Close()
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("starting opencode metadata read: %w", preferContextError(ctx, err))
+	}
+	defer func() { _ = tx.Rollback() }()
+	stmt, err := tx.PrepareContext(ctx,
+		"SELECT time_updated FROM session WHERE id = ?",
+	)
+	if err != nil {
+		return nil, fmt.Errorf("preparing opencode metadata read: %w", preferContextError(ctx, err))
+	}
+	defer stmt.Close()
+	metas := make([]OpenCodeSessionMeta, 0, len(sessionIDs))
+	for _, id := range sessionIDs {
+		if err := ctx.Err(); err != nil {
+			return nil, err
+		}
+		var updated int64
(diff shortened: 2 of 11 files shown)


berenddeboer:
Rejecting the requested changes. This PR was quite dificult to pull together ask roborev kept asking to fix more and more corner cases. I feel this is sufficient middleground for a solid implementation.

berenddeboer:
Also I noticed this still has a significant cpu usage: consistently about 130% on my system. Doing some more investigation.

PR #1219, opened 2026-07-22
feat(parser): parse OMP subagent transcripts as nested sessions

chikei:
#1202 was closed because being dumb when sync my repo :facepalm:


mariusvniekerk:
happens :D

PR #1225, opened 2026-07-22
feat(pricing): add OpenRouter as a second pricing source

godlockin:
Adds OpenRouter's public model catalog as a second pricing source under LiteLLM.

`pricing.FetchCatalog` fetches LiteLLM and OpenRouter together; if either fetch fails, the last stored catalog is kept. `pricingrefresh` drives every refresh through it, so the `usage` CLI, the statusline, and the server's existing daily refresh loop all pick up OpenRouter with no separate code path.

At store time, `Catalog.Reconcile` merges the snapshot over the stored table: an OpenRouter row is dropped when the LiteLLM catalog, or any stored row from another source (LiteLLM, embedded, supplemental), already lists the same canonical model name under any spelling or provider prefix. Adding OpenRouter therefore never changes a lookup that already resolved, and two spellings of one model never tie in the resolver. Rows are copied whole (a zero rate is a valid free-model price, so nothing is backfilled across sources), and no bare aliases are emitted because the resolver already matches a bare session name against a lone qualified key.

A single sentinel row, `_openrouter_models`, records which stored rows came from OpenRouter. A later refresh or fallback reseed retires a tracked row once another source covers its model under a different spelling, transfers ownership of exact patterns, and keeps rows OpenRouter merely delisted so old sessions stay priced. `db.ReconcileModelPricing` applies the deletes, upserts, and sentinel in one transaction.

`db.PlanModelPricingSync` gives the PostgreSQL and DuckDB pushes the same rules in both directions: a local OpenRouter row is withheld when a target row from another source already covers its model, a target-owned row is removed only when the local catalog covers its model under a different spelling, and target and local ownership are merged so no machine strips ownership another machine recorded on a shared target. PostgreSQL runs the read, plan, and write in one transaction under a schema-scoped advisory lock so concurrent pushes cannot overwrite each other's merged ownership. Sentinel rows compare by value in `FilterChangedModelPricing`; PostgreSQL writes them through a dedicated statement because their `updated_at` is not a timestamp, and DuckDB commits band deletes before parent deletes because it rejects both in one transaction.

The OpenRouter parser produces microdollar rates, keeps text-output models (including multimodal-input models), and skips negative or malformed entries per row so a single odd upstream entry cannot block the whole refresh.

Docs: pricing-source, `--offline`, and privacy sections describe the two catalogs and their precedence.

Reviewers: `internal/pricing/openrouter.go` (`Catalog.Reconcile`, `ShadowedPatterns`), `internal/pricingrefresh/refresh.go` (`storeCatalog`), `internal/db/pricing.go` and `pricing_sync.go`, and the push changes in `internal/postgres/pricing.go` and `internal/duckdb/push.go`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ agentsview usage statusline
 
 Features:
 
-- Automatic pricing via LiteLLM rates (with offline fallback)
+- Automatic pricing via LiteLLM and OpenRouter rates (with offline fallback)
 - Authoritative Copilot CLI billing totals when session logs provide them
 - Prompt-caching-aware cost calculation (cache creation / read tokens)
 - Per-model breakdown with `--breakdown`
--- a/cmd/agentsview/archive_query_backend.go
+++ b/cmd/agentsview/archive_query_backend.go
@@ -300,7 +300,7 @@ func (b localArchiveQueryBackend) SessionUsage(
 	}
 	if len(u.UnpricedModels) > 0 && !b.offline {
 		refreshed, refErr := pricingrefresh.RefreshIfStale(
-			b.database, pricing.FetchLiteLLMPricing,
+			b.database, pricing.FetchCatalog,
 			pricingrefresh.RefreshCooldown, time.Now(),
 		)
 		if refErr != nil {
--- a/cmd/agentsview/archive_query_backend.go
+++ b/cmd/agentsview/archive_query_backend.go
@@ -306,7 +306,10 @@ func (b localArchiveQueryBackend) SessionUsage(
 		if refErr != nil {
 			fmt.Fprintf(os.Stderr,
 				"warning: pricing refresh failed: %v\n", refErr)
-		} else if refreshed {
+		}
+		// A degraded refresh (see pricing.FetchCatalog) stores rows and
+		// reports an error at once, so re-read whenever rows changed.
+		if refreshed {
 			if u2, e := load(); e == nil && u2 != nil {
 				u = u2
 			}
--- a/cmd/agentsview/pricing_schedule_test.go
+++ b/cmd/agentsview/pricing_schedule_test.go
@@ -20,6 +20,8 @@ import (
 // swap, particularly on Windows where open handles prevent the rename.
 const pricingResyncTestTimeout = 10 * time.Second
 
+// pricingCatalogTransport answers the LiteLLM and OpenRouter catalog
+// requests a refresh makes and records their URLs.
 type pricingCatalogTransport struct {
 	requests chan *http.Request
 }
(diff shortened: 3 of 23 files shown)


mjacobs:
So, this is not converging cleanly enough to keep patching finding-by-finding:  new states are emerging faster than reviews/fixes are closing them.

Assessment from Codex:

 I would keep the fetch/parser work but reset the reconciliation layer:
  - Bound and validate both upstream responses.
  - Reject structurally invalid or implausibly empty snapshots.
  - Skip OpenRouter entries with unsupported overrides for now.
  - Replace alias plus transient-shadow metadata with one complete source-ownership manifest—or explicit source ownership per row—so every backend can derive deletions from its last observed snapshot without seeing every intermediate refresh.
  That is a contained architectural correction, not a wholesale PR rewrite. After that, one review should cover a much smaller state space. Continuing with another narrow tombstone patch is likely to produce another round.

wesm:
looking at this

wesm:
Reviewed this PR and pushed two follow-up commits.

**7d9c3a7 — CockroachDB-compatible pricing sync lock** (fixes the roborev finding). The pricing sync ran `pg_advisory_xact_lock(hashtext(...), hashtext(...))` on every push, but supported CockroachDB versions implement neither advisory locks nor `hashtext`, so `pg push` against CockroachDB failed before any session synchronization. The sync now serializes by locking a dedicated `sync_metadata` row (`model_pricing_sync_lock`) with `SELECT ... FOR UPDATE`, created idempotently with `ON CONFLICT DO NOTHING`. That behaves the same on PostgreSQL and CockroachDB, is naturally scoped to the target schema (no `hashtext(current_schema())` needed), and matches the `FOR UPDATE` locking already used elsewhere in the push path. Verified against a real PostgreSQL instance with the pgtest pricing suite, including the new `TestPushRetiresOpenRouterPricingRows`.

**beafd2a — keep LiteLLM rates fresh during an OpenRouter outage.** Before, a failure of either fetch discarded the whole snapshot, so OpenRouter being down also froze LiteLLM rate updates. `Reconcile` already handles an empty OpenRouter side safely (previously stored OpenRouter rows stay priced and tracked as delisted), so `FetchCatalog` now returns a LiteLLM-only snapshot together with the error; refreshes store it and callers surface the error as a warning. A LiteLLM failure still fails the snapshot and keeps the last stored rates. Two smaller changes rode along: a corrupt `_openrouter_models` sentinel on a shared target now produces an error that names the remediation (delete that `model_pricing` row) instead of silently blocking every machine's push, and `storeCatalog` documents the cross-process refresh race on SQLite and why it self-heals.

Two observations left as-is: `_fallback_version` / `_litellm_last_attempt` sentinel rows are no longer pushed to targets (nothing reads them remotely; stale rows from older versions just remain), and the legacy microdollar migration in `schema.go` uses the same advisory-lock construct but only runs on databases that predate CockroachDB support.

PR #1228, opened 2026-07-22
fix(grok): emit one usage event per turn instead of last-snapshot-per-session

### LINKED ISSUE #1227: Grok usage undercounted up to ~200x: turn_completed usage is per-turn, but only the last snapshot per session is ingested
## Summary

Follow-up to #1193 / #1206. The Grok usage ingestion added in #1206 treats `params.update.usage` objects in `updates.jsonl` as **cumulative session snapshots** and keeps only the **last** one per session (`parseGrokUsageEvents` overwrites `usage` on every candidate line, then emits one event per (session, model) with dedup key `session:<id>:<model>`).

Grok Build's `turn_completed` usage is actually **per-turn**, so this captures a single turn per session and drops the rest. On a real 44-session archive (2,976 `turn_completed` records), agentsview main (`75b9e8b`) reports:

| metric | agentsview | sum of per-turn records | undercount |
|---|---|---|---|
| input (net of cached) | 662,978 | 18,985,293 | 28.6x |
| output | 53,230 | 2,661,643 | 50x |
| cachedRead | 12,658,048 | 2,582,853,120 | **204x** |

**Exact-match confirmation:** independently summing only the *last* usage object per session over the same raw files reproduces agentsview's totals to the exact token (662,978 / 53,230 / 12,658,048) — so the delta is fully explained by last-snapshot-per-session semantics, not by parsing gaps.

## Evidence that usage is per-turn, not cumulative

- `cachedReadTokens` across consecutive `turn_completed` events in one session goes 11264 → 11264 → 6016 → 128 → 25472 — non-monotonic, impossible for a cumulative counter.
- `inputTokens` across those turns grows by ~50–70 per turn (25340 → 25411 → 25463…) while each turn actually resends a ~25k-token context — consistent with per-turn totals, not cumulative increments.
- `usage.modelCalls` resets per turn (e.g. 13 for a single multi-call turn).

## Secondary issue: date bucketing

`occurredAt` is set from the session's `endedAt`/`startedAt`, so a session's entire usage lands on one calendar day. Long-running sessions (days-long agent sessions are common) make the daily breakdown wrong even once totals are fixed — in the archive above, the single heaviest day (929M tokens) is missing entirely from `usage daily` output because its sessions ended on other days.

## Suggested fix

Emit one usage event **per `turn_completed` line** (per model from `usage.modelUsage`), with:
- dedup key `session:<sessionId>:<prompt_id>:<model>` (`prompt_id` is unique per turn — verified collision-free across the archive),
- `occurredAt` from the line's top-level `timestamp` (unix seconds),
- the existing `max(input - cachedRead, 0)` netting kept as-is (it's correct).

Happy to provide additional record samples or test against a branch.

delattre1:
Fixes #1227.

## Problem

`parseGrokUsageEvents` treats `params.update.usage` payloads in `updates.jsonl` as cumulative session snapshots and keeps only the last one per session. Grok Build's `turn_completed` usage is actually **per-turn** (`cachedReadTokens` is non-monotonic across a session's turns; `modelCalls` resets each turn), so multi-turn sessions are undercounted by orders of magnitude and all usage lands on the session's end date. Full evidence incl. an exact-token reproduction of the undercount is in #1227.

## Change

- Emit one usage event per usage payload and model (from `usage.modelUsage`, with the existing top-level fallback), instead of one per session.
- Dedup key gains a turn component: `session:<sessionID>:<prompt_id>:<model>` — `prompt_id` is unique per turn and stable across re-parses (verified collision-free on a 2,976-turn archive); a `turn-<n>` ordinal covers payloads without `prompt_id`.
- `OccurredAt` comes from each payload's own top-level `timestamp` (unix seconds), falling back to the session window when absent — a days-long session now buckets each turn on the day it happened.
- The existing `max(input − cachedRead, 0)` netting and `costUsdTicks` handling are unchanged.
- `dataVersion` 69 → 70 (same convention as 51, the Gemini cumulative-to-delta reparse) so existing Grok rows are re-parsed on upgrade.

## Before / after on a real archive

44 sessions / 2,976 `turn_completed` payloads, `usage daily --agent grok` totals:

| metric | before (main `75b9e8b`) | after | ground truth (independent re-parse of raw files) |
|---|---|---|---|
| input (net) | 662,978 | 19,023,100 | 19,023,100 |
| output | 53,230 | 2,674,197 | 2,674,197 |
| cachedRead | 12,658,048 | 2,591,768,960 | 2,591,768,960 |
| days present | 6 | 8 | 8 |

The heaviest day of the archive (929M tokens) was previously missing entirely from the daily output (its sessions ended on other dates); it now reports correctly.

## Testing

- `go test -short ./internal/...` passes; `gofmt`/`go vet` clean.
- Reworked `TestGrokProviderLatestUsageSnapshot` → `TestGrokProviderPerTurnUsageEvents` (two payloads → two events, non-monotonic cache preserved, session total output = sum).
- New: `TestGrokProviderPerTurnUsageDedupsByPromptID` (multi-model turn, prompt_id-keyed dedup) and `TestGrokProviderPerTurnUsageDatesByTurnTimestamp` (multi-day session buckets per turn; timestamp-less payload falls back to the session window).

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -313,7 +313,12 @@ const projectIdentityRemoteScrubCompletedKey = "project_identity_remote_scrub_v1
 // (69: Copilot shutdown events persist the authoritative AI-credit total as
 // reported cost. Re-parsing populates cost_usd and cost_source on existing
 // Copilot rows from session.shutdown totalNanoAiu values.)
-const dataVersion = 69
+// (70: Grok per-turn usage reparse. turn_completed usage payloads are
+// per-turn measurements, not cumulative snapshots — one event per turn
+// and model replaces the single last-payload event per session, with
+// occurred_at from each turn's timestamp. Existing Grok rows undercount
+// multi-turn sessions and need re-parsing.)
+const dataVersion = 70
 
 const tokenCoverageRepairStatsKey = "token_coverage_repair_v1"
 
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -930,9 +930,9 @@ func TestMigration_ToolResultEventsTable(t *testing.T) {
 		"expected tool_result_events table after reopen")
 }
 
-func TestCurrentDataVersionCopilotReportedCost(t *testing.T) {
-	assert.Equal(t, 69, CurrentDataVersion(),
-		"Copilot reported-cost parsing requires a data version bump")
+func TestCurrentDataVersionGrokPerTurnUsage(t *testing.T) {
+	assert.Equal(t, 70, CurrentDataVersion(),
+		"Grok per-turn usage parsing requires a data version bump")
 }
 
 func TestInsertMessages_PreservesToolResultEvents(t *testing.T) {
--- a/internal/db/usage_test.go
+++ b/internal/db/usage_test.go
@@ -324,6 +324,50 @@ func TestUsageEventsReplaceAndList(t *testing.T) {
 	require.Len(t, got, 0, "usage events after clear =")
 }
 
+func TestUsageEventsReplaceRejectsDuplicateDedupKeysAndRollsBack(t *testing.T) {
+	// A parser emitting two events with the same dedup key (e.g. Grok
+	// retry/replay lines sharing a prompt_id, before the parser-side
+	// last-wins dedupe) violates idx_usage_events_dedup and must fail
+	// the whole replace, leaving the prior events untouched. Parsers
+	// therefore dedupe in memory before handing events to the DB.
+	d := testDB(t)
+	ctx := context.Background()
+
+	insertSession(t, d, "grok:dup", "proj", func(s *Session) {
+		s.Agent = "grok"
+	})
+
+	prior := []UsageEvent{{
+		SessionID:   "grok:dup",
+		Source:      "session",
+		Model:       "grok-4.5",
+		InputTokens: 1,
+		DedupKey:    "session:grok:dup:p-1:grok-4.5",
+	}}
+	require.NoError(t, d.ReplaceSessionUsageEvents("grok:dup", prior))
+
+	dup := []UsageEvent{{
+		SessionID:   "grok:dup",
+		Source:      "session",
+		Model:       "grok-4.5",
+		InputTokens: 2,
+		DedupKey:    "session:grok:dup:p-2:grok-4.5",
+	}, {
+		SessionID:   "grok:dup",
+		Source:      "session",
+		Model:       "grok-4.5",
+		InputTokens: 3,
+		DedupKey:    "session:grok:dup:p-2:grok-4.5",
+	}}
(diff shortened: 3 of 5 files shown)


delattre1:
Addressed the duplicate-DedupKey review point in d6ef285 — thanks, it was a real gap.

**The gap:** `parseGrokUsageEvents` appended an event per usage payload, so retry/replay lines re-emitting a turn's payload under the same `prompt_id` produced duplicate DedupKeys, which would violate `idx_usage_events_dedup (session_id, source, dedup_key)` and roll back the entire `ReplaceSessionUsageEvents` transaction. And correct: the existing `DedupsByPromptID` test never actually sent a duplicate.

**The fix:** in-memory last-wins dedupe keyed on DedupKey — a re-emitted payload is a rewrite of that turn, so the last one replaces the earlier event in place (order preserved). Ordinal-keyed payloads (no `prompt_id`) are deliberately not deduped: each gets a distinct `turn-<n>` key, because a duplicate there is indistinguishable from a genuine new turn and collapsing would undercount.

**Tests added:**
- `TestGrokProviderPerTurnUsageLastWinsOnDuplicatePromptID` — two payloads, same `prompt_id`+model → one event carrying the second payload's tokens/timestamp; session total output reflects only the survivor.
- `TestUsageEventsReplaceRejectsDuplicateDedupKeysAndRollsBack` (db-level) — a batch with duplicate dedup keys fails the replace and the prior events survive the rollback, pinning exactly the failure mode the parser-side dedupe protects against.

Also caught in this pass: the `CurrentDataVersion` pin test still asserted 69 — updated to 70 (was missed in the first commit).

`gofmt`/`go vet` clean; `internal/parser` and `internal/db` suites pass fully. The `internal/sync` resync-contributor tests fail identically on unmodified main `75b9e8b` in my environment (macOS), so those appear environmental rather than related to this change.

mariusvniekerk:
Checked this against `xai-org/grok-build@7cfcb20`. Grok persists one complete `TurnCompleted.usage` snapshot per prompt and explicitly suppresses a second terminal in the cancel/completion race; replay does not append rewritten terminals. See [the terminal ownership guard](https://github.com/xai-org/grok-build/blob/7cfcb20d2b50b0d18801a6c0af2e401c0e060894/crates/codegen/xai-grok-shell/src/session/acp_session_impl/turn_end.rs#L258-L303) and [its double-emit regression test](https://github.com/xai-org/grok-build/blob/7cfcb20d2b50b0d18801a6c0af2e401c0e060894/crates/codegen/xai-grok-shell/src/session/acp_session_tests/turn_completion_emit_tests.rs#L527-L556).

The finding is internally correct only if duplicate `prompt_id` payloads are treated as whole-turn rewrites, but that is not a supported producer shape. I would avoid adding group-replacement hardening/tests for this hypothetical state and instead remove or simplify the duplicate-rewrite handling.

PR #1234, opened 2026-07-23
feat: unify usage metrics and add Kimi Work support

liaocaoxuezhe:
Unifies cost and token analytics in one Usage page with a URL-backed metric toggle, metric-aware visual formatting, and server-side rankings across SQLite, PostgreSQL, and DuckDB. Legacy token-usage links remain compatible.

Adds Kimi Work desktop discovery and parsing with documented format provenance, plus timestamp-aware Kimi alias pricing. Exact custom alias rates retain precedence, while schema-v4 pricing metadata preserves reported model names and every resolved priced model across cutoff-spanning reports.

Adds macOS production and development launchers that use private temporary log directories and open only process-authenticated, dynamically discovered loopback URLs.

K3 rates remain curated estimates until they are available from the upstream pricing catalog; consumers of the changed pricing provenance shape must use schema version 4.


mjacobs:
looking

mjacobs:
Thanks for the substantial contribution! This PR currently combines four independently reviewable changes: the Token Usage view, Kimi Work provider support, Kimi alias pricing, and macOS launchers. Please split the first three into focused PRs so each can be evaluated and merged independently.

Please drop the launcher scripts from this series; AgentsView already has a managed background-server lifecycle, and we do not want to introduce separate PID, logging, port-discovery, and shutdown behavior in repository shell scripts. A macOS end-user launcher would need a dedicated design discussion and should build on the existing lifecycle. The development launcher is better kept as a local helper.

The Kimi pricing PR should cover SQLite, PostgreSQL, and DuckDB consistently, including activity reports and custom pricing overrides. The Kimi Work provider PR should include the required format-provenance documentation.

wesm:
I think it's useful to see the tokens view, but it should simply be a toggle in the Usage tab, not a new page. I'm making these changes now

wesm:
I'm just going to do some manual QA on this before I merge it

wesm:
working on this

PR #1243, opened 2026-07-23
feat(recall): add corpus browser page

wesm:
Adds Recall as a top-level corpus browser. Accepted entries appear in a padded, single-row table for quick comparison; each row expands in place to show the full fact, trigger, uncertainty, extraction metadata, provenance state, and transcript evidence. Search and filters cover project, type, active extraction generation, and review state.

Browsing stays bounded with cursor pagination. Ranked searches preserve relevance semantics, reject stale continuations after corpus mutations, and disclose the existing 500-result cap. The page also reports extraction coverage and stays unavailable on read-only backends.


wesm:
I need to do some QA on this before merging

mjacobs:
I tested #1243 locally end-to-end this evening.

**test setup:** checked out the PR, pointed `[recall.extract]` at `gemma-4-31B-it-qat` behind llama.cpp (OpenAI-compatible endpoint) on a RTX 5090, and ran the full extract → activate → browse flow against:

### synthetic archive

a first test (testfixture-generated, ~14 sessions) to shake out the pipeline before committing a several GPU hours processing my real session data.

Looks good, no functional issues. extract doctor → clean probe on the first try, extraction ran with zero schema failures (Note: llama.cpp enforces the json_schema response format with grammar-constrained decoding, so a local gemma that was never tuned for this still produces valid entries every time; IME gemma/gemini wouldn't always do that naturally).

### my real session data

Same setup with real data completed in a few hours (I don't have specific timestamps off hand, but it was a few hours) processed 483 sessions with 5 sessions that couldn't complete the extraction.  Note that 488 sessions was a small subset of my overall session data, but those were the sessions that were eligible for extraction given the expected constraints of the extraction (tl;dr protecting potentially sensitive data).


## Recall UI

Recall page does what the PR description claims: status card with coverage/backlog, project/type/generation/review-state filters, lexical search flipping between keyset and ranked pagination, and the evidence links jump to the source transcript at the right ordinals.

Screenshot:
snippy_1784867769694
(I blurred several text blocks)

Some minor adjustments to the UI may be worthwhile aesthetic enhancements, but definitely not blocking IMO.

wesm:
Thank you. I'll work on making the UI less ugly! I'm not sure the 3-column layout is useful either

PR #1247, opened 2026-07-23
Fix nested subagent hierarchy: re-parent depth≥2 subagents to their spawner

### LINKED ISSUE #1246: Nested subagents (depth ≥ 2) get parent_session_id pinned to the main session
**agentsview version:** v0.38.1 (commit 1e173b95, built 2026-07-13T21:41:46Z)
**Install method:** install.sh
**OS / platform:** Linux (Ubuntu, x86_64), kernel 6.8

> Heads up: I'm not a deep-code person — I used Claude Code (AI) to investigate this
> properly and write the patch. It traced the root cause, produced a fix plus a
> regression test (fails before, passes after), and I verified the patch applies cleanly
> to current `main` (`34d326c`). PR incoming that fixes this. Happy to point the AI at any
> follow-up questions.

## Summary

In a Claude Code session where a subagent spawns its **own** subagents:

```
main → orchestrator → {codex, kimi}      (spawnDepth 1 and 2)
```

agentsview indexes all four sessions and captures every tool call correctly, but
`sessions.parent_session_id` for the depth-2 sessions points at the **main session**
instead of the orchestrator. The stored tree is flat:

| session | stored parent | true parent |
| --- | --- | --- |
| orchestrator | main | main ✅ (correct by luck) |
| codex | main | **orchestrator** ❌ |
| kimi | main | **orchestrator** ❌ |

`tool_calls.subagent_session_id` already has the correct nested edges
(main→orchestrator, orchestrator→codex, orchestrator→kimi) — the authoritative data is
in the DB; only the session hierarchy discards it.

Consequences: `GET /sessions/{orchestrator}/children` returns empty; per-session usage
rollups (BFS over `parent_session_id`) attribute the whole subtree to the main session.

## Root cause

Two parent sources fight, and the wrong one wins for depth ≥ 2:

1. **Path-derived parent (parse time):** `claudeCompanionParentSessionID`
   (`internal/parser/claude.go`) takes the path segment before `subagents/`. Claude
   Code stores **all** depths flat in `<main-session>/subagents/`, so every subagent
   gets parent = main and `relationship_type = 'subagent'`.
2. **Authoritative parent (post-parse):** `LinkSubagentSessions()`
   (`internal/db/sessions.go` ~L1552) re-points parents from
   `tool_calls.subagent_session_id` — but its `WHERE relationship_type != 'subagent'`
   guard skips every row mechanism (1) already tagged. The correction never runs for
   nested subagents.

Depth 1 is correct only by luck (its flat path-parent equals its true parent);
depth ≥ 2 stays wrong.

## Fix (one clause)

Drop the `relationship_type` guard; re-point whenever the authoritative parent differs
(null-safe, so it still covers the existing continuation→subagent upgrade):

```sql
UPDATE sessions
SET parent_session_id = (
      SELECT tc.session_id FROM tool_calls tc
      WHERE tc.subagent_session_id = sessions.id LIMIT 1),
    relationship_type = 'subagent',
    local_modified_at = strftime('%Y-%m-%dT%H:%M:%fZ','now')
WHERE EXISTS (SELECT 1 FROM tool_calls tc
              WHERE tc.subagent_session_id = sessions.id)
  AND parent_session_id IS NOT (
      SELECT tc.session_id FROM tool_calls tc
      WHERE tc.subagent_session_id = sessions.id LIMIT 1);
```

## Verification

- New regression test `TestLinkSubagentSessionsReParentsNestedGrandchild`
  (main → orchestrator → grandchild with the buggy stored state + correct tool_calls
  edges): **unpatched → FAIL** (grandchild parent `main`, want `orchestrator`);
  **patched → PASS**.
- All existing `Subagent|LinkSubagent` tests in `./internal/db` stay green, incl.
  `TestLinkSubagentSessionsOverridesContinuation`.
- No frontend change needed for reachability — transcript drill-down
  (`ToolBlock` → `SubagentInline`) is already driven by `tool_calls.subagent_session_id`
  at any depth. The fix makes `/children` and usage rollups correct.

## Optional follow-ups (not in the PR)

1. **Sidebar visual nesting:** the sidebar renders a root's descendants as one flat
   "Subagents (N)" list (`emitGroupItems`,
   `frontend/src/lib/components/sidebar/session-list-utils.ts`). Per-row
   `parent_session_id` is already in the payload; indenting grandchildren under their
   spawner is a client-only change.
2. **Claude sidecar metadata:** Claude Code writes `subagents/agent-<id>.meta.json` with
   explicit `parentAgentId` + `spawnDepth`; the Claude parser doesn't read it (Codex's
   `parent_thread_id` already gets this treatment in `internal/parser/codex.go`).
3. **Cross-agent lineage:** a `codex exec` run from inside a Claude subagent's Bash tool
   creates a standalone Codex session (indexed from `~/.codex/sessions`) with no link to
   the Claude session that spawned it — two unlinked views of the same event.

nmishaan:
Fixes #1246.

## Problem

Nested Claude Code subagents (`main -> orchestrator -> grandchild`, spawnDepth >= 2) all
live in the same flat `<main-session>/subagents/` directory, so path derivation assigns
`parent_session_id = main session` at every depth and tags each row
`relationship_type = 'subagent'`. `LinkSubagentSessions()` — which corrects the parent
from the authoritative `tool_calls.subagent_session_id` spawn edge — skipped those rows
because of its `WHERE relationship_type != 'subagent'` guard. The hierarchy stayed flat
past depth 1: `GET /sessions/{orchestrator}/children` was empty and per-session usage
rollups attributed the whole subtree to the main session.

## Change

All in `internal/db/sessions.go`, plus one call-site change in `internal/sync/engine.go`:

- **Re-parent from the spawn edge.** The linking predicate now updates a session when it
  is not yet tagged `subagent` (the pre-existing upgrade path) or when its stored parent
  differs from the spawn edge (null-safe `IS NOT`). The tool-call edge records the actual
  spawn, so it is authoritative over the path heuristic.
- **Deterministic conflict resolution** (`subagentSpawnerExpr`). Copied or forked history
  can leave several sessions claiming the same child. Resolution is a pure function of
  the stored edges — never of ingestion order — so a link written from a partial view
  self-corrects on a later sync: candidates are ordered by start time (a fork derives
  from its source, so it starts later), with `started_at` normalized via `strftime`
  because raw RFC3339Nano strings do not sort chronologically when fractional seconds
  are stripped. Unknown or malformed start times rank last, and the session id breaks
  remaining ties.
- **Cost bounded by edges, not archive.** The global statement is driven from the
  `idx_tool_calls_subagent` partial index (`s.id IN (SELECT tc.subagent_session_id ...)`)
  instead of a full `sessions` scan, so bulk-sync linking cost tracks the number of spawn
  edges rather than archive size.
- **Watcher syncs bounded by the batch.** `SyncSingleSessionContext` — called by the
  session watcher on every file change — now uses `LinkSubagentSessionsForSessions`,
  which restricts the pass to children reachable from the sessions that sync wrote
  (spawner-side and child-side branches, both index seeks). Bulk paths (full sync,
  reconciliation, resync) keep their coalesced global pass. Query-plan tests pin both
  cost shapes at two archive sizes.
- **All single-session branches link.** The incremental append path (which can
  introduce a spawn edge when the Task tool_use and its subagent mapping land in the
  same append) links scoped to the appended session, and children referenced by a
  batch's pre-write edges are captured (`SubagentChildSessionIDs`) before rewrites or
  parser-exclusion deletes cascade those edges away, so a former child still
  re-resolves to its remaining spawner in the same sync.

## Tradeoffs and limitations

- When conflicting edges exist and the real spawner has no usable `started_at`, a dated
  copy wins; protecting it would require reading the child's stored parent, which
  reintroduces ingestion-order dependence. The link self-corrects once the spawner's
  start time becomes known. Documented in `subagentSpawnerExpr`.
- Fork/copy provenance policy (self-edge guards, deeper lineage tracking) is scoped to
  #1250.
- No frontend change: transcript drill-down already follows `tool_calls` at any depth;
  this fix makes `/children` and usage rollups correct.

Reviewers should start at `subagentSpawnerExpr` and the two linking statements in
`internal/db/sessions.go`, then the `SyncSingleSessionContext` call site.

---

Disclosure: investigated and written with Claude Code (AI).

### DIFF
--- a/internal/db/link_subagent_nested_test.go
+++ b/internal/db/link_subagent_nested_test.go
@@ -1,0 +1,97 @@
+package db
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestLinkSubagentSessionsReParentsNestedGrandchild reproduces the
+// nested-subagent bug: when a subagent spawns its own subagent (depth >= 2),
+// the grandchild is parsed with a path-derived parent pointing at the MAIN
+// session (all subagents live in the same flat <main>/subagents/ dir) and is
+// tagged relationship_type='subagent'. LinkSubagentSessions must re-point it
+// to the intermediate subagent that actually spawned it, using the
+// authoritative tool_calls edge.
+//
+// Tree under test:  main -> orchestrator -> grandchild
+func TestLinkSubagentSessionsReParentsNestedGrandchild(t *testing.T) {
+	d := testDB(t)
+
+	mainID := "main"
+	orchestratorID := "orchestrator"
+	grandchildID := "grandchild"
+
+	// Main session (root).
+	insertSession(t, d, mainID, "p", func(s *Session) {
+		s.MessageCount = 1
+	})
+
+	// Orchestrator: a depth-1 subagent. Path derivation put its parent at
+	// the main session (correct here) and tagged it 'subagent'.
+	insertSession(t, d, orchestratorID, "p", func(s *Session) {
+		s.MessageCount = 1
+		parent := mainID
+		s.ParentSessionID = &parent
+		s.RelationshipType = "subagent"
+	})
+
+	// Grandchild: a depth-2 subagent. Path derivation ALSO put its parent at
+	// the main session (WRONG — it should be the orchestrator) and tagged it
+	// 'subagent'. This is the buggy stored state we expect linking to fix.
+	insertSession(t, d, grandchildID, "p", func(s *Session) {
+		s.MessageCount = 1
+		wrongParent := mainID
+		s.ParentSessionID = &wrongParent
+		s.RelationshipType = "subagent"
+	})
+
+	// The authoritative spawn edges, exactly as the parser records them in
+	// tool_calls from toolUseResult.agentId:
+	//   main         --Task--> orchestrator
+	//   orchestrator --Task--> grandchild
+	insertMessages(t,
+		d,
+		Message{
+			SessionID: mainID, Ordinal: 0, Role: "assistant",
+			Content: "spawn orchestrator", HasToolUse: true,
+			ToolCalls: []ToolCall{{
+				ToolName: "Agent", Category: "Task",
+				SubagentSessionID: orchestratorID,
+			}},
+		},
+		Message{
+			SessionID: orchestratorID, Ordinal: 0, Role: "assistant",
+			Content: "spawn grandchild", HasToolUse: true,
+			ToolCalls: []ToolCall{{
+				ToolName: "Agent", Category: "Task",
+				SubagentSessionID: grandchildID,
+			}},
+		},
+	)
+
+	require.NoError(t, d.LinkSubagentSessions(), "LinkSubagentSessions")
+
+	// Orchestrator stays under main.
+	orch, err := d.GetSession(context.Background(), orchestratorID)
+	requireNoError(t, err, "GetSession orchestrator")
+	if assert.NotNil(t, orch.ParentSessionID, "orchestrator parent") {
+		assert.Equal(t, mainID, *orch.ParentSessionID,
+			"orchestrator.parent_session_id")
+	}
+
+	// Grandchild must be re-parented to the orchestrator, NOT the main
+	// session. This is the assertion that fails on the current
+	// `WHERE relationship_type != 'subagent'` guard.
+	gc, err := d.GetSession(context.Background(), grandchildID)
+	requireNoError(t, err, "GetSession grandchild")
+	assert.Equal(t, "subagent", gc.RelationshipType,
+		"grandchild relationship_type")
+	if assert.NotNil(t, gc.ParentSessionID, "grandchild parent") {
+		assert.Equal(t, orchestratorID, *gc.ParentSessionID,
+			"grandchild.parent_session_id must be the orchestrator, "+
+				"not the flat main session")
+	}
+}
--- a/internal/db/link_subagent_nested_test.go
+++ b/internal/db/link_subagent_nested_test.go
@@ -95,3 +95,50 @@ func TestLinkSubagentSessionsReParentsNestedGrandchild(t *testing.T) {
 				"not the flat main session")
 	}
 }
+
+// TestLinkSubagentSessionsUpgradesTypeWhenParentAlreadyMatches guards the
+// regression flagged in review: LinkSubagentSessions sets BOTH parent_session_id
+// and relationship_type='subagent'. A session can already carry the correct
+// (authoritative) parent while still being misclassified as continuation / fork
+// / empty. The type upgrade must run even when the parent does not change, or
+// the session is grouped wrong.
+func TestLinkSubagentSessionsUpgradesTypeWhenParentAlreadyMatches(t *testing.T) {
+	d := testDB(t)
+
+	// Parent session with a tool call referencing the child.
+	insertSession(t, d, "parent", "p", func(s *Session) {
+		s.MessageCount = 1
+	})
+
+	// Child ALREADY has the correct parent (== the tool-call spawner) but is
+	// misclassified as a continuation (e.g. a header parentId that coincides
+	// with the spawner). parent_session_id won't change; relationship_type
+	// must still be upgraded to 'subagent'.
+	insertSession(t, d, "child", "p", func(s *Session) {
+		s.MessageCount = 1
+		parent := "parent"
+		s.ParentSessionID = &parent
+		s.RelationshipType = "continuation"
+	})
+
+	insertMessages(t, d, Message{
+		SessionID: "parent", Ordinal: 0, Role: "assistant",
+		Content: "spawn child", HasToolUse: true,
+		ToolCalls: []ToolCall{{
+			ToolName: "Agent", Category: "Task",
+			SubagentSessionID: "child",
+		}},
+	})
+
+	require.NoError(t, d.LinkSubagentSessions(), "LinkSubagentSessions")
+
+	child, err := d.GetSession(context.Background(), "child")
+	requireNoError(t, err, "GetSession child")
+	assert.Equal(t, "subagent", child.RelationshipType,
+		"relationship_type must upgrade to 'subagent' even when the parent "+
+			"already matches the tool-call spawner")
+	if assert.NotNil(t, child.ParentSessionID, "child parent") {
+		assert.Equal(t, "parent", *child.ParentSessionID,
+			"child.parent_session_id")
+	}
+}
--- a/internal/db/link_subagent_nested_test.go
+++ b/internal/db/link_subagent_nested_test.go
@@ -182,3 +182,60 @@ func TestLinkSubagentSessionsLinksNullParentSubagent(t *testing.T) {
 			"orphan.parent_session_id")
 	}
 }
+
+// TestLinkSubagentSessionsPreservesParentOnConflictingEdges guards against
+// arbitrary re-parenting when a child is referenced by spawn edges from more
+// than one distinct session (reachable via copied/forked history). The parent
+// must be resolved only from an unambiguous edge; with conflicting edges the
+// current parent is preserved rather than picked arbitrarily by a bare LIMIT 1.
+func TestLinkSubagentSessionsPreservesParentOnConflictingEdges(t *testing.T) {
+	d := testDB(t)
+
+	insertSession(t, d, "p1", "p", func(s *Session) {
+		s.MessageCount = 1
+	})
+	insertSession(t, d, "p2", "p", func(s *Session) {
+		s.MessageCount = 1
+	})
+	// "kid" is already a correctly parented subagent under p2.
+	insertSession(t, d, "kid", "p", func(s *Session) {
+		s.MessageCount = 1
+		parent := "p2"
+		s.ParentSessionID = &parent
+		s.RelationshipType = "subagent"
+	})
+
+	// Two DISTINCT sessions both carry a spawn edge to "kid". p1's edge is
+	// inserted first (lower rowid), so a bare `LIMIT 1` would pick p1 and
+	// wrongly re-parent kid away from its current p2. The unambiguous-edge
+	// resolution must instead preserve p2.
+	insertMessages(t, d,
+		Message{
+			SessionID: "p1", Ordinal: 0, Role: "assistant",
+			Content: "spawn kid (copy A)", HasToolUse: true,
+			ToolCalls: []ToolCall{{
+				ToolName: "Agent", Category: "Task",
+				SubagentSessionID: "kid",
+			}},
+		},
+		Message{
+			SessionID: "p2", Ordinal: 0, Role: "assistant",
+			Content: "spawn kid (copy B)", HasToolUse: true,
+			ToolCalls: []ToolCall{{
+				ToolName: "Agent", Category: "Task",
+				SubagentSessionID: "kid",
+			}},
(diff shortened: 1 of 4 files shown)


nmishaan:
Good catch  confirmed and fixed in 9d2e0b6.

You're right that `LinkSubagentSessions` sets **both** `parent_session_id` and `relationship_type = 'subagent'`, so gating only on "parent differs" skipped rows whose parent already matched the tool-call spawner but were still classified `continuation`/`fork`/empty the type upgrade never ran and they grouped wrong.

The predicate is now:

```sql
AND (
    relationship_type != 'subagent'
    OR parent_session_id IS NOT (SELECT tc.session_id FROM tool_calls tc
                                 WHERE tc.subagent_session_id = sessions.id LIMIT 1)
)
```

so the type upgrade fires even when the parent is unchanged, while already-correct subagents match neither branch (no `local_modified_at` churn). Added `TestLinkSubagentSessionsUpgradesTypeWhenParentAlreadyMatches` (existing parent already equals the tool-call parent, `relationship_type = 'continuation'` → upgraded to `subagent`); it fails on the previous predicate and passes now. Existing `Subagent|LinkSubagent` tests stay green.

(Disclosure: investigated/authored with Claude Code; changes verified locally.)

nmishaan:
Pushed a small polish commit (`9e58019`) after a second review pass:

- Updated the `LinkSubagentSessions` doc comment, which previously described only the type upgrade and not the nested re-parenting this PR adds.
- Added `TestLinkSubagentSessionsLinksNullParentSubagent` — a `subagent`-tagged session with a NULL parent + a spawn edge must be linked to its spawner. Mutation-verified: it fails if the null-safe `IS NOT` is swapped for `!=`. (Two reviewers independently suggested this case.)

All `internal/db` `Subagent|LinkSubagent` tests pass locally.

One **pre-existing** observation, intentionally left out of scope: the `SELECT tc.session_id … LIMIT 1` has no `ORDER BY`, so if a single subagent session were ever referenced by spawn edges from two *different* parents, the chosen parent would be arbitrary. This predates the PR (the `SET` subquery already used bare `LIMIT 1`) and shouldn't arise from valid Claude Code data (an `agentId` maps to one spawn). Flagging only in case you'd want a deterministic `ORDER BY` / dedupe as a separate follow-up.

(Disclosure: iterated with Claude Code and a second-opinion review pass; all changes verified locally.)

nmishaan:
Thanks — real catch. The temporal case is genuine but narrow: it needs a forked/copied edge ingested *before* its source's edge, and a fork derives from its source (so the source is normally parsed first, where the current `COALESCE`-preserve behavior is already correct). It's reachable mainly through reconciliation/copy orderings.

Fully resolving it needs stable provenance — e.g. resolving ambiguous edges to the earliest-started spawner, or tracking source-parent state to reverse a provisional link — a fork-lineage design decision I scoped out of this depth-2 parenting fix. The core change here is a strict improvement over the previous bare-`LIMIT 1` behavior, and its worst case for conflicting edges is bounded (parent-chain consumers use `UNION` recursive CTEs that dedupe/terminate — no loop or corruption).

Opened #1250 to track the provenance work (ingestion-order determinism + a self-edge guard), with your suggested before/after-conflicting-edge regression test noted. Happy to send a follow-up PR against whichever resolution approach you prefer.

mjacobs:
Thanks for the investigation and for opening #1250 — agreed the broader fork/copy provenance policy is separate design work. But the ingestion-order case needs to be resolved in this PR, because it's introduced here: on `main`, a correctly-parented `subagent` row can never be re-parented; with this change, a copied edge that is temporarily the only stored edge overwrites the correct parent, and once the real edge arrives the `COALESCE` preserve-on-ambiguity rule locks the wrong parent in — with no convergence path until follow-up work that has no timeline.

The existing conflicting-edge test inserts both edges before linking, so it doesn't exercise this sequence. Please add: (1) child correctly parented, (2) insert the copied edge, link, (3) insert the real edge, link again, (4) assert the child converges to and remains under the real spawner.

Concretely, resolving the edge to the **earliest-started spawner** (join `sessions`, `ORDER BY started_at`) instead of requiring a unique spawner would make step 4 pass: a fork always starts after its source, so this is deterministic across ingestion orders and self-corrects already-linked rows on the next sync. The self-edge guard and any deeper provenance tracking can stay in #1250.

nmishaan:
Pushed `00660f1` + `aeb3313`, covering @mjacobs's ingestion-order request and both outstanding roborev findings.

## 1. Ingestion-order convergence (@mjacobs) — done in `00660f1`

Took your suggested resolution: the spawner is now resolved to the **earliest-started** candidate (`LEFT JOIN sessions ps`, ordered by start time) instead of requiring a unique one. You were right that the previous rule was the problem rather than the mitigation — `COALESCE(unambiguous_edge, parent_session_id)` read the row's *own* parent, so its result depended on what earlier syncs had written, which is exactly how a provisional parent got locked in. Resolution is now a pure function of the stored edges, so a link written from a partial view self-corrects on the next sync.

`TestLinkSubagentSessionsConvergesAcrossIngestionOrder` implements your four steps literally:

1. child correctly parented under the real spawner,
2. insert the copied edge → `LinkSubagentSessions()`,
3. insert the real edge → `LinkSubagentSessions()` again,
4. assert the child converges to the real spawner **and stays there** on a third link (idempotent, not oscillating).

Plus the mirror order (real edge first, later-arriving copy must not steal the child), so neither order is privileged.

The self-edge guard and deeper provenance tracking remain in #1250 as you scoped.

## 2. Timestamp ordering (roborev on `00660f1`) — real bug, fixed in `aeb3313`

This one was a genuine defect in the fix above, and worse than a corner case. `started_at` is TEXT written by `timeutil.Format` → `time.RFC3339Nano`, which **strips trailing zeros** from the fractional second. So production values are not fixed width:

```
time.Date(2026, 1, 1, 0, 0, 0, 0, time.UTC)        -> "2026-01-01T00:00:00Z"
                       + 100 * time.Millisecond    -> "2026-01-01T00:00:00.1Z"
```

`.` (0x2E) sorts before `Z` (0x5A), so ordering the raw strings puts the **later** timestamp **first**. Any spawner whose start lands on a whole second stores no fractional part at all and sorts behind every fractional one — precisely undercutting the earliest-started rule.

Fixed by ordering on `strftime('%Y-%m-%dT%H:%M:%fZ', ps.started_at)`, which re-renders each value as fixed-width `...T00:00:00.000Z` where lexical order *is* chronological. That's the same normalization the `sync_marker` triggers in `db.go` already use. It also subsumes the old `NULLIF(...,'')` wrapper: `strftime` returns NULL for an `started_at` that is unset, empty **or malformed**, so the leading `IS NULL` key now sorts unparseable timestamps last too — which the previous form did not. Resolution is millisecond-granular (SQLite's parser truncates there); anything closer falls through to the existing `tc.session_id` key, so the order stays total.

`TestLinkSubagentSessionsOrdersStartTimesChronologically` builds both timestamps with `timeutil.Format` itself so it can't drift from what the parsers write, and asserts up front that the raw strings really do sort inverted — otherwise it would pass without the fix. Mutation-verified: restoring the raw ordering fails it with parent `copied-spawner`.

## 3. Archive scaling (roborev on `c544f11`) — fixed in `aeb3313`, and it's now faster than `main`

Worth stating plainly: the global `UPDATE` is not introduced by this PR. `main` plans the same way —

```
-- main today
QUERY PLAN
|--SCAN sessions                       <-- full scan of the archive
`--CORRELATED SCALAR SUBQUERY
   `--SEARCH tc USING INDEX idx_tool_calls_subagent (subagent_session_id=?)
```

`EXISTS(...)` reads like an index probe but plans as a scan of `sessions` with a correlated probe per row. Driving the row set from the edges instead — `s.id IN (SELECT tc.subagent_session_id ...)` — seeks the `idx_tool_calls_subagent` partial index and then looks each child up by primary key:

```
|--SEARCH s USING INDEX sqlite_autoindex_sessions_1 (id=?)
|--LIST SUBQUERY
|  `--SEARCH tc USING COVERING INDEX idx_tool_calls_subagent (subagent_session_id>?)
```

Measured on a seeded archive in the steady state a watcher-driven sync actually hits (every subagent already linked, `UPDATE` matches zero rows), holding spawn edges constant at 8k:

| variant | 40k sessions | 200k sessions | growth |
| --- | --- | --- | --- |
| `main` today | 17.1 ms | 102.5 ms | **6.0×** |
| this PR, `EXISTS` form | 35.9 ms | 109.4 ms | **3.0×** |
| this PR, `IN` form (shipped) | 25.1 ms | 30.9 ms | **1.2×** |

So this isn't just a recovered regression — at 200k sessions linking is now **3.3× faster than `main`**, because per-sync cost tracks the number of spawn edges rather than the size of the archive. (I also measured a single-pass `UPDATE .. FROM` with a `ROW_NUMBER()` window: 27.4 ms — it trades 8k tiny sorts for one large materialized one, so it's a bigger diff for no gain. Rejected.)

**On "limit reconciliation to child IDs affected by the current sync":** I deliberately didn't, because it would break the convergence in §1. A later-arriving spawn edge is written into the **spawner's** transcript, not the child's — re-checking a child that this sync didn't touch is exactly what lets a provisional link self-correct. Narrowing the scope to sessions written this sync reintroduces the lock-in @mjacobs asked to fix. Making the global statement index-driven gets the scaling without that trade.

`TestLinkSubagentSessionsPlanScalesWithSpawnEdges` asserts the **query plan** rather than wall-clock time — the plan is what decides the scaling and it's deterministic in CI (same approach as the existing `EXPLAIN QUERY PLAN` assertions in `source_path_hints_test.go`). It runs at two archive sizes so it can't pass merely for want of statistics. Mutation-verified: swapping the `IN` back to the equivalent-reading `EXISTS` re-introduces `SCAN s` and fails it **while every behavioural test stays green** — which is the silent regression it exists to catch.

## Verification

Go 1.26.5, `-tags fts5`:

- `./internal/db` — full package, green (all `LinkSubagent*` tests included)
- `./internal/duckdb` — full package, green, including `TestPushIncrementalMirrorsSubagentLinkBackfill`, which depends on this function bumping `local_modified_at` so the linked session re-enters the incremental mirror window
- `./internal/sync` — all `Subagent|Link|Parent|Hierarch` tests green

Both new tests are mutation-verified fail→pass as described above.

Two flaky-test notes, both pre-existing and outside `internal/db`:

- I could not get a clean **full**-package `internal/sync` run on either branch: `TestFileChangeTimeDetectsSameStatRewrite` fails for me on **unmodified `main`** (`9e5324c`) in 4 of 5 isolated runs. Looks environment-sensitive rather than related to any of this.
- `Go Test (windows-latest)` is red on `main` too, with a *different* test each run — [`TestPushWatchProductionOwnersRetryPartialAuthoritativeBatch`](https://github.com/kenn-io/agentsview/actions/runs/30299130630) and [`TestWatcherOverflowCollapsesPathAndByteLimitsToOneFullSync`](https://github.com/kenn-io/agentsview/actions/runs/30159669338) on `main`, `TestVerifiedSourceGateRechecksAfterStatAndWatcherInvalidation` on this branch. All three are in `internal/sync`; this PR touches only `internal/db`.

Happy to open a separate issue for the `internal/sync` flakes if that's useful — they're unrelated to this change but they do make the signal here harder to read.

(The CI workflow runs for both pushes are sitting at `action_required` pending maintainer approval, so the only check reporting on this PR right now is roborev.)

---

(Disclosure: investigated and authored with Claude Code; the timing/plan numbers above were measured locally on a seeded SQLite archive, and both new tests were mutation-verified fail→pass.)

wesm:
I can't push here anymore for some reason so opening a new PR to supersede this

PR #1252, opened 2026-07-24
fix(claude): keep ide context out of session titles

### LINKED ISSUE #1238: Strip out / parse <ide_opened_file>/<ide_selection> in Claude Code via VSCode extension
### agentsview version

agentsview v0.38.1 (commit 1e173b95, built 2026-07-13T21:41:46Z)

### Install method

Other / not sure

### OS / platform

Windows 11 + WSL2 Ubuntu 22.04.3

### Which agent and version

Claude Code for VSCode 2.1.218

### Which model(s)

claude-sonnet-5

Actual result: (see screenshot)

Image

Expected result: ide_selection / ide_opened_file hints do not show in session selection title and are parsed in User messages.

```text

```

### Steps to reproduce

1. Install Claude Code for VSCode and sign in with a claude.ai subscription
2. Open a file in the workspace, then communicate with claude via the IDE
3. Load the session in agentsview - notice the ide tags

### Checklist

tylergibbs1:
the claude code vscode extension writes file-open and selection hints as standalone user-role xml envelopes. agentsview currently treats them as operator prompts, so they can become sidebar titles, count as user turns, and render raw markup.

complete standalone `ide_opened_file` and `ide_selection` envelopes now promote to hidden system metadata with distinct subtypes. the match is intentionally strict, so malformed wrappers and messages with real prompt text remain ordinary user content.

dag fork selection applies the same user-text preprocessing before counting turns, so reminder-prefixed ide context cannot promote an obsolete branch. incremental sync keeps promptless ide and continuation appends bounded to the new bytes, falling back to a full parse only when the first title-eligible prompt arrives.

the source data version advances to 74 because main now uses 73 for opencode bash exit-failure detection. existing sqlite archives are reparsed on upgrade, and duckdb mirrors rebuild on their next push through the independent source data-version gate; the mirror schema is unchanged.

closes #1238

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -75,7 +75,10 @@ Grok section and remove the explicit registry exception in the coverage test.
   `attachment.type=queued_command` are written mid-stream, in file order
   between consecutive `assistant` records that share one `message.id`, so a
   queued command can fall inside a streaming run that straddles an incremental
-  sync boundary.
+  sync boundary. Reverified 2026-07-23 against the transcript shape reported in
+  [#1238](https://github.com/kenn-io/agentsview/issues/1238): Claude Code for
+  VS Code writes standalone `user` records wrapped in `ide_opened_file` or
+  `ide_selection` tags for editor context rather than operator prompts.
 
 ## OpenClaude (`openclaude`)
 
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -330,7 +330,10 @@ const projectIdentityRemoteScrubCompletedKey = "project_identity_remote_scrub_v1
 // status N" output text, so existing rows carry no failure signal. Re-parsing
 // attaches the errored event so tool-health failure counts cover historical
 // OpenCode sessions on every platform.)
-const dataVersion = 73
+// (74: Claude Code IDE context reparse. Standalone ide_opened_file and
+// ide_selection wrappers are promoted to system metadata so existing
+// VS Code sessions no longer use them as titles or user turns.)
+const dataVersion = 74
 
 const tokenCoverageRepairStatsKey = "token_coverage_repair_v1"
 
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -1007,9 +1007,9 @@ func TestMigration_ToolResultEventsTable(t *testing.T) {
 		"expected tool_result_events table after reopen")
 }
 
-func TestCurrentDataVersionOpenCodeBashExitFailure(t *testing.T) {
-	assert.Equal(t, 73, CurrentDataVersion(),
-		"OpenCode bash metadata.exit failure detection requires a data version bump")
+func TestCurrentDataVersionClaudeIDEContext(t *testing.T) {
+	assert.Equal(t, 74, CurrentDataVersion(),
+		"Claude IDE context parsing requires a data version bump")
 }
 
 func TestInsertMessages_PreservesToolResultEvents(t *testing.T) {
--- a/internal/parser/claude.go
+++ b/internal/parser/claude.go
@@ -2431,7 +2431,7 @@ func isCommandEnvelope(content string) bool {
 	return strings.TrimSpace(stripped) == ""
 }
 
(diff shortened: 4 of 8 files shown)


mjacobs (changes requested):
Thanks! The standalone-tag classification and the linear/incremental coverage
look good. The DAG branch-selection issue called out above still needs
addressing though: `countUserTurns` counts IDE-context `user` records before
they are classified as promoted system metadata, so enough of them can change
the fork heuristic. Please exclude those records from branch-turn counting and
add a UUID/DAG fork regression.

Two version-number requests as well:

**1. Please leave DuckDB `SchemaVersion` at 5.**

This PR does not change the mirror schema. A push already rebuilds when the
mirror's recorded `agentsview_source_data_version` differs from
`db.CurrentDataVersion()`. The reparse resync also creates a new SQLite
`database_id`, which independently forces a full rebuild on the next push.

Raising `SchemaVersion` has an additional effect: a new binary refuses to serve
existing v5 mirrors until their owner pushes again. Without the bump, the mirror
remains usable—with stale titles—until that push. For this metadata/title
correction, the latter is the better tradeoff and matches the mirror's normal
stale-until-next-push behavior.

`AGENTS.md` currently instructs authors to bump `SchemaVersion` for source
data-version changes, _so you followed the repository guidance correctly_. We
are going to narrow that policy so `SchemaVersion` tracks mirror-shape changes,
while source data-version changes use the existing push-time rebuild gate. I
will handle that documentation change separately; no `AGENTS.md` change is
needed in this PR.

**2. Use `dataVersion` 71 on the current base.**

Data versions are assigned in merge order rather than reserved for in-flight
PRs. If #1166 lands first, this PR rebases and moves to 72; if this PR lands
first, #1166 does that. A gap would be mechanically harmless because the
migration logic is monotonic, but I'd guide away from reserving numbers for
unmerged work.

tylergibbs1:
Addressed the requested changes in `11dcec65`:

- IDE/system-injected user records no longer count toward DAG fork selection.
- Added a UUID/DAG fork regression with four IDE-context records.
- Kept DuckDB `SchemaVersion` at 5.
- Rebased onto current `main` and resolved the data-version conflict. Since `main` now uses data version 71 for the OpenCode change, this Claude parser reparse is 72.

The focused parser and DB regressions pass, and `go vet -tags fts5 ./...` is clean.

tylergibbs1:
Addressed the incremental-sync edge case in c812905e. Claude sessions with an empty stored preview now fall back to a full parse even when `user_message_count` is 0, covering sessions that initially contain only injected IDE context. Added `TestIncrementalSync_ClaudeIDEContextOnlyRepairedOnAppend` to verify that appending the first real prompt populates `first_message` and increments the real user count.

Validation:
- focused clear-only + IDE-context incremental regressions pass
- `CGO_ENABLED=1 go vet -tags fts5 ./...` passes
- full `internal/sync` run reaches an unrelated existing macOS resync-close test failure (`TestResyncAbortsWhenReplacementCloseFails`); the failing test also fails in isolation and does not overlap this change.

tylergibbs1:
@mjacobs, this is rebased onto current `main` and ready for re-review. The upstream OpenCode migration remains data version 72; this Claude IDE-context reparse is now version 73. The prior DAG and incremental-title requests remain addressed, parser/database tests and the focused incremental-sync regressions pass, and `go vet -tags fts5 ./...` is clean.

mjacobs (approved):

tylergibbs1:
Both findings from the latest review are addressed on the branch: 2759a93b parses appended Claude deltas before falling back only for the first eligible prompt (with incremental-only continuation/IDE/command regressions), and 5308806f preprocesses reminder-prefixed user text before DAG turn classification. I validated the exact current head with `go fmt ./...`, `CGO_ENABLED=1 go test -tags fts5 ./internal/parser ./internal/db`, `CGO_ENABLED=1 go test -tags fts5 ./internal/sync`, and `CGO_ENABLED=1 go vet -tags fts5 ./...`; all pass.

mjacobs:
thanks! I'm just waiting on the CI tests that are still running.

PR #1253, opened 2026-07-24
parser: capture sessionKind and promptSource from Claude Code transcripts

TweedBeetle:
## What

Capture two Claude Code launch/prompt-provenance fields that the indexer currently reads past and discards:

- **`sessionKind`** — session-level (e.g. `"bg"`). Present on background/headless sessions, absent on interactive ones. Lands in a new `session_kind` column on `sessions`.
- **`promptSource`** — per user turn (e.g. `"typed"`, `"queued"`, `"system"`, `"sdk"`). Marks whether a turn was a human-typed prompt or a queued/system/SDK-injected one. Lands in a new `prompt_source` column on `messages`.

## Why

Both signals exist on essentially every current Claude Code transcript but are thrown away at parse time, so the archive can't answer two questions it otherwise has the data for: was a session interactive or headless, and was a given prompt typed by a human or injected. This mirrors the recently-added `entrypoint` capture — same shape of information, same reason to keep it.

## How

The change follows the existing `entrypoint` / `agent_label` and `source_type` / `is_sidechain` handling rather than inventing a new pattern:

- `sessionKind` is treated as a session-identity field alongside `entrypoint`: first-non-empty-wins across the file, wired into the incremental identity-escalation check, and blanked by the Cowork and Qoder parsers that reuse the Claude parse body (exactly as they blank `entrypoint`).
- `promptSource` is treated as a per-message source column alongside `source_type`/`is_sidechain`, including the retained-field whitelist in `compactClaudeEntry` and the message token fingerprints the PG/DuckDB push fast-paths compare against.

Both columns are `NOT NULL DEFAULT ''`, so transcripts that predate the fields and other agents that never emit them read empty and are otherwise untouched. A `dataVersion` bump (79) triggers the standard reparse so existing Claude rows backfill both fields. Capture is mirrored across the SQLite, Postgres, and DuckDB stores (mirror `SchemaVersion` 9) and the artifact wire format, where the manifest advances to v3 and message segments to v2: older artifacts still decode with the fields empty, while pre-provenance peers park new artifacts behind the existing future-version gate instead of silently dropping the fields. The fields also round-trip through artifact import, re-enqueue a published session's export when only `session_kind` changes, and count toward the export preflight's message-byte bound.

## Scope / limitations

- Pure field capture — no counting, aggregation, health-signal, or automation-classification logic is touched.
- `session_kind`/`prompt_source` backfill via the dataVersion-79 reparse; rows whose source files no longer exist stay empty.
- No new API/CLI surface or UI is added; the columns are available for consumers that read the archive directly.

## Where to look

- `internal/parser/claude.go` — the capture site (`sessionKind` next to `entrypoint`; `promptSource` next to `isSidechain`) and the `compactClaudeEntry` retained-field whitelist.
- `internal/parser/claude_provider.go`, `provider.go`, `internal/sync/engine.go` — incremental identity threading for `sessionKind`.
- `internal/db/{schema.sql,db.go,sessions.go,messages.go}` — SQLite columns, migrations, and read/write plumbing.
- `internal/postgres/*` and `internal/duckdb/*` — the mirrored columns and fingerprints.

The source of truth for both fields is the Claude Code transcript format, where `sessionKind` and `promptSource` appear as top-level keys on session JSONL entries.


mjacobs:
looking

PR #1257, opened 2026-07-24
feat(parser): detect opencode bash tool failures from metadata.exit

ajinkyajacob:
## Problem

OpenCode bash tool failures leave `metadata.exit > 0` in the tool state but the output text does NOT contain the `exit status N` pattern that AgentsView's `isBashFailure()` checks. This is not Windows-specific: on a Linux `opencode.db`, all 24 bash parts with a non-zero exit (1, 127, 128) had output text with no `exit status` marker, while all 81 successful parts recorded `exit=0`. Example outputs from real sessions:

- `'choco' is not recognized as an internal or external command` - exit=1, no "exit status" text
- (empty output) - exit=1, no text at all
- `rsync error: some files could not be transferred` - exit=1

The exit code is reliably recorded in `metadata.exit` by the agent, but the parser never reads it.

## Impact

On a Windows machine with 273 opencode sessions, 1233 bash tool failures are missed because they lack the `exit status` output pattern. Only 51 are currently detected (via the `tool:"invalid"` fix). The remaining ~1200 are invisible to signals, insights, and health scoring.

## Fix

In `extractOpenCodeToolCall()`, after parsing the tool state, check `metadata.exit`:

```go
if len(state.Metadata) > 0 {
    var m struct { Exit int `json:"exit"` }
    if err := json.Unmarshal(state.Metadata, &m); err == nil && m.Exit > 0 {
        isFailure = true
    }
}
```

This is opt-in per agent - only opencode records metadata.exit. Other agents are unaffected.

`dataVersion` moves to 73 so existing opencode rows are re-parsed and historical sessions backfill the failure events, and the format evidence is recorded in `docs/internal/session-format-sources.md`.

## Branch

https://github.com/ajinkyajacob/agentsview/tree/feat/opencode-metadata-exit-failure-detection

## Risk

- Non-zero exit doesn't always mean failure (e.g., `grep` returns 1 for no match). However, AgentsView's existing `exitStatusRe` already makes the same assumption via output text. This just adds parity for the metadata path.
- Only applies when `metadata` key exists in the tool state (opencode format), so other agents are isolated.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -244,8 +244,8 @@ Grok section and remove the explicit registry exception in the coverage test.
   failure signal. The tool's own output text carries no `exit status N`
   marker, and the shell is `COMSPEC`/`cmd.exe` on Windows, so text-pattern
   matching alone misses these failures on every platform. Agentsview treats a
-  non-zero `state.metadata.exit` on any tool part as a failure and attaches an
-  errored result event. Only `bash` parts record `exit`; other tools omit the
+  non-zero `state.metadata.exit` on a `bash` tool part as a failure and attaches
+  an errored result event. Only `bash` parts record `exit`; other tools omit the
   key. Verified 2026-07-24 against a live `opencode.db` where all 24 bash
   parts with `exit` in `{1, 127, 128}` had output text without an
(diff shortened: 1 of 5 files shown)


mjacobs:
found an edge case that should be handled, updating PR soon

PR #1261, opened 2026-07-24
Add Codebuff/Freebuff agent support

scross01:
Parses Codebuff and Freebuff sessions from `~/.config/manicode/projects/<project>/chats/<timestamp>/`. Both agents share the same on-disk layout; sessions are classified by the `agentType` field in `run-state.json`.

### Highlights

- Full transcript parsing: user/ai/error messages, tool calls with skill attribution, thinking blocks, subagent invocations
- Block types: text, tool, agent, mode-divider, plan, ask-user, image
- Credits tracking: extracts `creditsUsed` from `run-state.json` and maps to cost (1 credit = $0.01)
- Freebuff auto-classification: sessions with "free" in `agentType` appear as a separate filter

### Unique considerations

- Freebuff shares the Codebuff provider (single registration) to avoid double-discovery and skip-cache contention
- The `agentType` field is an agent template name (e.g. `base2-free-deepseek`), not the actual LLM model -- the real model is selected server-side
- `contextTokenCount` provides context window size but per-message token breakdown is not available in the on-disk format

### Limitations

- Mid-session model switches are invisible -- the on-disk format has no per-turn model field
- Freebuff has no credits (ad-supported); cost tracking only applies to Codebuff
- The `agent` field on `ChatMessage` is not populated in the persisted format, so per-turn agent attribution relies on subagent blocks only

### DIFF
--- a/cmd/agentsview/session_get.go
+++ b/cmd/agentsview/session_get.go
@@ -199,66 +199,66 @@ func resolveBareCodebuffID(
 		}
 	}
 
-	// Phase 2: Host-prefixed remote-synced sessions.
-	// This runs independently of the filesystem walk (even when
-	// len(locations)==0) so remote-only sessions — imported via
-	// pg push or remotesync with no local on-disk copy — can still
-	// be resolved. The filter requires a host prefix ("~"), a
-	// codebuff or freebuff agent marker, and the exact timestamp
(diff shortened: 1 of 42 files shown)


wesm:
rebasing

PR #1265, opened 2026-07-25
fix(claude): keep IDE-context envelopes out of first_message previews

### LINKED ISSUE #1238: Strip out / parse <ide_opened_file>/<ide_selection> in Claude Code via VSCode extension

adammw:
Fixes #1238.

Claude Code's VS Code extension sometimes prepends an IDE-context wrapper
(`<ide_opened_file>` or `<ide_selection>`) onto the user's real prompt inside a
single transcript entry. Those sessions showed raw markup as their first
message and in the transcript.

After this change:

- The parser splits such entries into a hidden metadata row plus the visible
  prompt, so `first_message` and the transcript show only what the user typed.
  Only well-formed leading wrappers with a non-empty remainder are split;
  malformed or mid-message markup is left alone.
- Data version 88 re-parses existing sessions so already-archived rows are
  cleaned up too.
- Pins survive that re-parse. The split inserts a row and shifts ordinals, so
  pin restoration now follows one rule, applied identically in SQLite and
  PostgreSQL: a unique source UUID follows its row; any other pin follows its
  message's role, content, and position among identical messages, provided
  that group kept its size. Unchanged messages keep their pins across inserted
  rows, ordinal shifts, and metadata reclassification, and a streamed partial
  response that completes in place keeps its pin as well.
- Uploads no longer discard parser metadata: hidden/system status, source
  identity, sidechain, and compact-boundary fields are preserved, so uploaded
  sessions render the same as locally synced ones.

Two limits are deliberate. Editing a pinned message that has no source UUID
drops the pin in both stores: the edit destroys the only identity the pin
could follow, and dropping consistently beats guessing. And while the one-time
re-parse push is rewriting a session, pin, unpin, and star on that session
wait for the push batch to commit.

Where to look: the split logic in `internal/parser/claude.go`, pin
restoration in `internal/db/messages.go` and `internal/postgres/push.go`,
resync copying in `internal/db/orphaned.go`, and the upload field mapping in
`internal/server/upload.go`.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -180,6 +180,13 @@ Grok section and remove the explicit registry exception in the coverage test.
   `promptSource` (per user turn, e.g. `"typed"`, `"queued"`, `"system"`,
   `"sdk"`). Neither key is documented upstream or covered by the codeburn
   notes; the evidence remains local observation under `no-public-source`.
+  Reverified 2026-07-31 against the transcript shape reported in
+  [#1265](https://github.com/kenn-io/agentsview/pull/1265): the extension can
+  also prepend one `ide_opened_file`/`ide_selection` wrapper directly onto a
+  real operator prompt inside a single `user` record (envelope first, prompt
+  text after the closing tag, one shared `uuid` for the whole record); the
+  parser splits these into a hidden system-metadata message plus the visible
+  prompt.
   Reverified 2026-08-09 against controlled `--resume <session> --fork-session`
   reproductions and inspection of the Claude Code 2.1.226 bundle
   ([#1370](https://github.com/kenn-io/agentsview/issues/1370)): the background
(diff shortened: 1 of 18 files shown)


mariusvniekerk:
The pin reattachment parts look a bit overwrought here?

mjacobs:
Agreed... this has grown beyond the original bug. I’m going to keep the prompt as one message and remove the leading IDE wrapper in place. That avoids shifting message ordinals, so the pin reattachment changes can come out of this PR.

wesm:
looking

PR #1268, opened 2026-07-25
Support API key env for recall extraction endpoints

binyangzhu000-sudo:
## Summary
- add `api_key_env` to `[recall.extract.servers.<name>]` so authenticated OpenAI-compatible recall extraction endpoints can use bearer tokens without embedding credentials in URLs
- wire the resolved server key into the extraction client and keep endpoint-provided diagnostics withheld when a bearer credential is configured
- document an Atlas Cloud HTTPS example using `ATLASCLOUD_API_KEY` and `deepseek-ai/deepseek-v4-pro`

## Testing
- `git diff --check`
- Atlas Cloud live model catalog: `GET https://api.atlascloud.ai/v1/models` returned HTTP 200 with 121 models; confirmed `deepseek-ai/deepseek-v4-pro` and `qwen/qwen3.5-flash`
- Not run: `gofmt` / `go test ./internal/config ./internal/recall/extract ./cmd/agentsview` because this machine has no Go toolchain in PATH (`go` and `gofmt` command not found); Docker is installed but the daemon is not running

### DIFF
--- a/cmd/agentsview/recall_extract.go
+++ b/cmd/agentsview/recall_extract.go
@@ -98,6 +98,7 @@ func resolveExtractDistillation(
 	return extractDistillation{
 		Client: &extract.Client{
 			BaseURL:    server.Endpoint,
+			APIKey:     server.APIKey(),
 			Model:      cfg.Model,
 			HTTPClient: httpClient,
 			Request:    request,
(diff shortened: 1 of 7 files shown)


wesm:
Makes sense

PR #1270, opened 2026-07-25
Route trusted Linux builds to managed runners

wesm:
Same-repository pull-request branches are controlled by repository collaborators and should use organization-managed Linux capacity. Fork pull requests remain arbitrary code—even when the author is an organization member—and must continue to use GitHub-hosted runners.

Each PR now triggers only `pull_request`, so required checks are associated with the current PR revision. The dispatcher calls a main-pinned reusable workflow, and that immutable workflow independently requires both the head and base repositories to equal `github.repository` before selecting the managed public fleet. The caller supplies no runner or trust input.

The organization runner policy restricts access to the exact main-branch reusable workflows. Main pushes use the managed Linux fleet; Windows, macOS, ARM Linux, tags, and fork PRs retain their hosted runners.

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -1,7 +1,8 @@
 name: Bench Gate
 
-# Fork PRs use GitHub-hosted runners. Same-repository PRs call the
-# main-branch workflow, whose Linux jobs can use the managed public fleet.
+# The PR dispatcher always calls the main-pinned workflow. That workflow
+# independently routes same-repository Linux jobs to the managed public fleet
+# and fork jobs to GitHub-hosted runners.
 on:
   pull_request:
     # Docs/frontend-only PRs cannot change the gated Go paths. If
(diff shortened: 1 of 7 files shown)


wesm:
This finding is already mitigated by the organization runner-group policy and does not require a workflow change here.

I verified the live `ci-runners-public` runner group:

- `restricted_to_workflows` is `true`.
- The allowlist contains the reusable implementations pinned to `refs/heads/main`, including `kenn-io/agentsview/.github/workflows/ci.yml@refs/heads/main`, `bench.yml@refs/heads/main`, and `desktop-artifacts.yml@refs/heads/main`.
- It does not contain `ci-pr.yml`, `bench-pr.yml`, or `desktop-artifacts-pr.yml`.

A fork can modify a `pull_request` dispatcher and request `kenn-linux-x64-public`, but that dispatcher workflow is not authorized to use the runner group, so GitHub will not assign one of these runners. Only the main-pinned reusable workflow can acquire the managed runner, and that workflow independently requires both the PR head and base repositories to equal `github.repository`; otherwise it selects `ubuntu-latest`.

This is the layered boundary intended by the organization runner policy: selected-workflow enforcement prevents direct dispatcher access, while the reusable workflow independently derives admission from immutable event context. The finding would become valid if `restricted_to_workflows` were disabled or a fork-controlled workflow reference were added to the allowlist.

PR #1275, opened 2026-07-27
fix(signals): treat raw terminal API errors as errored outcomes

### LINKED ISSUE #1269: ClassifyOutcome grades a session completed when it ends on a raw, unrecovered API error - giveUpPatterns doesn't match error-shaped final messages
**agentsview version:** agentsview v0.38.1 (commit 1e173b95, built 2026-07-13T21:41:46Z)

**Install method:** install.sh (`curl -fsSL https://agentsview.io/install.sh | bash`)

**OS / platform:** Arch Linux (rolling), x86_64

**Which agent and version:** Claude Code 2.1.205 (the affected session's own recorded `version` field)

**Which model(s):** claude-opus-4-8

`internal/signals/outcome.go`'s `ClassifyOutcome` grades a session `completed` (medium confidence) when
its transcript ends on a raw, never-recovered API/connection error.

Looking at the current source: `OutcomeInput` only carries `EndedWithRole`, `LastAssistantText`,
`FinalFailureStreak`, `MessageCount`, `IsAutomated`, `LastActivity` - nothing about whether the final
assistant message was a genuine answer vs. a transport-level error (`isApiErrorMessage` /
`"error":"server_error"` on the raw transcript entry never reaches this struct). For a session that ends
on `EndedWithRole == "assistant"`, the only check protecting against a bad final message is
`hasGiveUpPattern`, which substring-matches a fixed phrase list (`"i'm unable to"`, `"i can't proceed"`,
`"i don't have access"`, `"i cannot proceed"`, `"i am unable to"`). A raw error string like
`"API Error: Unable to connect to API (ConnectionRefused)"` doesn't match any of them, so the session
falls through to the generic `completed`/`medium` branch.

Expected: a terminal message that is a raw API/connection error (not the agent's own prose) shouldn't be
treated the same as a delivered answer - either a distinct `outcome` (you already have `errored` for the
`FinalFailureStreak >= 3` case) or at minimum a lowered `health_grade`/`outcome_confidence`.

## Repro

Confirmed on a real session, `sync --full`'d immediately beforehand so this isn't stale data (I found and
discarded a look-alike earlier that turned out to be pure sync lag - the source file was still growing;
this one's file size/mtime exactly match what's ingested, so it's genuinely final):

- Transcript's last line is a raw error entry, e.g.:
  ```json
  {
    "type": "assistant",
    "message": { "role": "assistant", "content": [
      { "type": "text", "text": "API Error: Unable to connect to API (ConnectionRefused)" }
    ]},
    "error": "server_error",
    "isApiErrorMessage": true
  }
  ```
  with no further lines ever appended after it (2 weeks old at time of filing, so not in-flight).
- Derived row: `outcome=completed`, `outcome_confidence=medium`, `health_grade=A`, `ended_with_role=assistant`.
- Corpus check (local `sessions.db`, after a full resync): 1 match out of 1110 `subagent`-relationship
  sessions whose ingested final message begins with the literal string `API Error:`. This is a narrow,
  literal-prefix detector on a small local corpus, not a rate - just what I could confirm by hand.

## Suggested direction (non-prescriptive)

Same shape as the existing `giveUpPatterns` check: add a check for a terminal message that's a raw
transport error (either pattern-match `"API Error:"`/similar prefixes the same way `hasGiveUpPattern`
does, or thread the raw `isApiErrorMessage`/`error` field through into `OutcomeInput` for a precise
check) and route it to `errored` or a lowered confidence, instead of falling into the generic
`completed`/`assistant` branch.

Happy to share the full anonymized transcript or run more queries against my local DB if useful.

rodboev:
Outcome classification currently marks a session completed when its final assistant message is a raw unrecovered API error, because the assistant-ended branch only lowers confidence for give-up prose.

This adds a dedicated terminal API-error discriminator before the generic assistant-completed branch and leaves the rest of the outcome ladder intact. User-ended sessions, repeated failure streaks, give-up phrases, and normal assistant answers keep their current classifications. The change stays in `internal/signals` with focused outcome regression coverage.

The terminal transcript entry and local corpus check came from @godilley's issue report.

Closes #1269

### DIFF
--- a/internal/db/sessions.go
+++ b/internal/db/sessions.go
@@ -182,7 +182,7 @@ func scanSessionRow(rs rowScanner) (Session, error) {
 	return s, err
 }
 
-const CurrentQualitySignalVersion = 2
+const CurrentQualitySignalVersion = 3
 
 // QualitySignals groups persisted deterministic quality-signal
 // columns for API callers while keeping the database representation
--- a/internal/signals/outcome.go
+++ b/internal/signals/outcome.go
@@ -42,6 +42,13 @@ func ClassifyOutcome(in OutcomeInput) OutcomeResult {
 		return OutcomeResult{"unknown", "low", false}
 	}
 
+	if in.EndedWithRole == "assistant" && hasTerminalAPIErrorText(in.LastAssistantText) {
+		if isRecent(in.LastActivity) {
+			return OutcomeResult{"unknown", "low", true}
+		}
+		return OutcomeResult{"errored", "medium", false}
+	}
+
 	if in.MessageCount == 2 && in.EndedWithRole == "assistant" {
 		return OutcomeResult{"completed", "medium", false}
 	}
@@ -92,4 +99,9 @@ func hasGiveUpPattern(text string) bool {
 		}
 	}
 	return false
+}
+
+func hasTerminalAPIErrorText(text string) bool {
+	lower := strings.ToLower(strings.TrimSpace(text))
+	return strings.HasPrefix(lower, "api error:")
 }
--- a/internal/signals/outcome_terminal_api_error_test.go
+++ b/internal/signals/outcome_terminal_api_error_test.go
@@ -1,0 +1,78 @@
+package signals
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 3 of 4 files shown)


wesm:
LGTM

PR #1276, opened 2026-07-27
fix(vector): classify whitespace-only embedding rejections as permanent

### LINKED ISSUE #1272: Vector build aborts fatally on empty/whitespace-only embedding input instead of skip-stamping
## What I was doing

Running an embeddings build over a large session archive.

## What happened

The build embedded thousands of documents successfully, then aborted fatally
on one document instead of skipping it:

```
Embedded 20501 documents (20532 chunks), skipped 0, stale 0
fatal: encode document <doc>: encode batch at 0: [vector.embeddings] status 400: {"error":{"message":"Invalid embedding input: Input strings must not be empty or whitespace-only","type":"embedding_error"}}
```

Some document apparently produced an empty or whitespace-only chunk, and the
embeddings endpoint rejected the batch with a 400.

`HTTPStatusError.Permanent()` in `internal/vector/encoder.go` is supposed to
classify input-specific 400/413/422 rejections as permanent so
`skipPermanentEncodeError` in `internal/vector/build.go` can skip-stamp just
that one document and let the build continue — the doc comment on
`skipPermanentEncodeError` explicitly lists "whitespace-only content some
servers refuse" as a case this is meant to cover:

```go
// skipPermanentEncodeError implements kitvec.FillOptions.OnEncodeError: a
// document the embeddings endpoint permanently rejects for input-specific
// reasons (e.g. a token-window overflow, whitespace-only content some servers
// refuse, or a content-policy rejection) is skipped ...
```

But the actual classifier, `hasDocumentSpecificEmbeddingError`, only
recognizes two message shapes: an overflow body (a size word like "too
long"/"length"/"limit"/"exceed"/"maximum" paired with "token"/"context"/
"input"/"text") or a content-policy body ("content" + "policy"). The message
"Input strings must not be empty or whitespace-only" matches neither pattern,
so `Permanent()` returns `false` and the whole build aborts instead of
skip-stamping the one poison document, contradicting the documented intent.

## Why it matters

A single document with empty/whitespace-only chunk content wedges the build
at that scan position: since pending documents are scanned in doc-key order,
later documents never get embedded, a first build never reaches `Missing==0`,
and auto-activation never fires — the exact "poison document" failure mode
`skipPermanentEncodeError`'s own comment warns about.

`hasDocumentSpecificEmbeddingError` in `internal/vector/encoder.go` needs a
case for empty/whitespace-only input rejections (e.g. "empty" or
"whitespace-only" paired with "input" or "string"), alongside the existing
overflow and content-policy cases.

rodboev:
Vector builds currently abort when the embeddings endpoint rejects one document as empty or whitespace-only, even though the same surface already skip-stamps other document-specific permanent rejections.

This extends the existing `hasDocumentSpecificEmbeddingError` classifier so whitespace-only input rejections route through the shipped permanent-error path instead of creating a second special case in the build loop. Token-window overflow and content-policy rejections keep skipping, while auth, routing, model, media-type, rate-limit, and server failures still abort the build. The change stays in `internal/vector` with focused classifier regression coverage.

The fatal error body and build symptom came from @jesserobbins's issue report.

Closes #1272

### DIFF
--- a/internal/vector/encoder.go
+++ b/internal/vector/encoder.go
@@ -133,6 +133,9 @@ func hasDocumentSpecificEmbeddingError(body string) bool {
 	if strings.Contains(body, "content") && strings.Contains(body, "policy") {
 		return true
 	}
+	if describesEmptyEmbeddingInput(body) {
+		return true
+	}
 	overLimit := strings.Contains(body, "too long") ||
 		strings.Contains(body, "too large") ||
 		strings.Contains(body, "too many") ||
@@ -148,6 +151,15 @@ func hasDocumentSpecificEmbeddingError(body string) bool {
 		strings.Contains(body, "context") ||
 		strings.Contains(body, "input") ||
 		strings.Contains(body, "text")
+}
+
+func describesEmptyEmbeddingInput(body string) bool {
+	hasWhitespaceRejection := strings.Contains(body, "whitespace-only") ||
+		strings.Contains(body, "whitespace only")
+	if !hasWhitespaceRejection || !strings.Contains(body, "empty") {
+		return false
+	}
+	return strings.Contains(body, "embedding input")
 }
 
 // embeddingsRequestBody is the OpenAI-compatible embeddings request.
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
pretty sure this fix should be done in kit and not here.

PR #1289, opened 2026-07-28
fix(copilot): use exact tool execution timestamps

### LINKED ISSUE #1288: Copilot CLI `task_complete` duration extends to the next user message in Session Analysis
### agentsview version

agentsview v0.38.1 (commit 1e173b95, built 2026-07-13T20:00:35Z)

### Install method

install.sh / install.ps1

### OS / platform

Windows 11 (10.0.26100.8893, x64)

### Which agent and version

GitHub Copilot CLI 1.0.76-0

### Which model(s)

gpt-5.6-sol

## What happened
In **Session Analysis**, a completed Copilot CLI `task_complete` tool call is shown as taking **12h 38m** and becomes the session's slowest call. The same interval is also attributed to the **Other** time-spent category.

The raw Copilot event stream shows that the tool actually completed in approximately **3.7 seconds**. The displayed 12h 38m duration instead matches the gap from the assistant message containing the tool request to the next user message after the session was resumed.

Observed timeline:

| Event | UTC timestamp |
|---|---|
| Assistant requests `task_complete` | `2026-07-24T11:56:24.189Z` |
| Tool execution starts | `2026-07-24T11:56:24.198Z` |
| Tool execution completes | `2026-07-24T11:56:27.923Z` |
| Next user message | `2026-07-25T00:34:49.483Z` |

Actual tool execution time:

`2026-07-24T11:56:27.923Z - 2026-07-24T11:56:24.198Z = 3.725 seconds`

AgentsView displays approximately:

`2026-07-25T00:34:49.483Z - 2026-07-24T11:56:24.189Z = 12h 38m`

## Expected behavior: 
when Copilot provides matching `tool.execution_start` and `tool.execution_complete` events, the per-tool duration should use those timestamps, paired by `toolCallId`. At minimum, idle time before the next user turn should not be attributed to the completed tool call.

This is separate from #83 and #869. PR #869 caps inter-message gaps for **Analytics → Top Sessions** active duration, but the incorrect **Session Analysis → slowest call / time spent** value remains in v0.38.1.

## Screenshot: 
attach the screenshot showing `task_complete — 12h 38m`, `slowest call task_complete`, and the following user message at the time of filing. But the actual tool execution time: `2026-07-24T11:56:27.923Z - 2026-07-24T11:56:24.198Z = 3.725 seconds`
Image

## Additional technical observation

Current `internal/db/timing.go` derives a tool-bearing message's duration from the next stored message timestamp:

```text
LEAD(m.timestamp) OVER (ORDER BY m.ordinal)
```

A solo non-subagent call then inherits that turn duration. This appears to explain the result when the next effective stored message after pairing is the later user turn, even though the Copilot parser receives a timestamped `tool.execution_complete` event.

This observation is included only as a possible cause; the reproducible behavior is the mismatch between the raw completion timestamp and the duration shown by Session Analysis.

```text
The following is a sanitized, minimal representation of the relevant events. Tool arguments, result content, user content, paths, and the original call ID have been removed or replaced.


{"timestamp":"2026-07-24T11:56:24.189Z","type":"assistant.message","data":{"toolRequests":[{"name":"task_complete","toolCallId":"call_example","arguments":{"summary":"[redacted]"}}]}}
{"timestamp":"2026-07-24T11:56:24.198Z","type":"tool.execution_start","data":{"toolCallId":"call_example"}}
{"timestamp":"2026-07-24T11:56:27.923Z","type":"tool.execution_complete","data":{"toolCallId":"call_example","result":"[redacted]"}}
{"timestamp":"2026-07-25T00:34:49.483Z","type":"user.message","data":{"content":"[redacted next request]"}}
```

### Steps to reproduce

1. Start a GitHub Copilot CLI session that has a `task_complete` tool available.
2. Let the agent invoke `task_complete` and verify that the Copilot event stream records matching `tool.execution_start` and `tool.execution_complete` events within a few seconds.
3. Leave the session idle for a significant period without entering another user message.
4. Resume the same session and enter another user message.
5. Sync or reload the session in AgentsView.
6. Open the session and select **Analysis**.
7. Observe that the final `task_complete` call is assigned approximately the entire idle interval up to the next user message and is reported as the slowest call.

### Checklist

EuanTop:
## Summary

- parse Copilot CLI `tool.execution_start` and `tool.execution_complete` events and attach them to the matching tool call
- preserve explicit Copilot tool failures from `success: false` and treat both successful and failed terminal events as exact timing endpoints
- use those event timestamps for per-call and per-turn timing in SQLite, PostgreSQL, and DuckDB/Quack
- fall back to the existing message-gap calculation when exact events are missing, incomplete, or invalid
- replace Copilot messages on later full parses so newly appended execution events update existing tool calls
- reparse existing Copilot sessions by bumping the data version
- preserve existing tool result content when timestamp-only events are present

## Behavior

Copilot tool calls previously inherited the interval between the assistant message and the next user message. For resumed sessions, this could attribute hours of idle time to a tool that completed in seconds.

When matching execution events are available, AgentsView now uses the actual execution interval. Parallel calls use the earliest valid start and latest valid terminal event, while each call retains its individual duration. Explicit `success: false` completion events remain failures for tool health signals without losing their exact end timestamp. Later syncs also replace previously stored Copilot messages so execution events appended after the initial tool call are persisted.

## Validation

- added parser coverage for successful, failed, and legacy completion events, plus unknown and reused IDs
- added SQLite coverage for exact, incomplete, invalid, parallel, and errored terminal intervals
- added PostgreSQL and DuckDB/Quack parity coverage for successful and errored terminal events
- added a two-sync regression where execution events arrive after the tool-call message
- added regression coverage for timestamp-only result events
- manually reproduced the issue and verified that idle time is no longer attributed to the completed tool call
- ran `go test -tags fts5 ./internal/parser ./internal/db ./internal/duckdb ./internal/sync`, `go test -tags fts5 ./internal/postgres`, the focused PostgreSQL `pgtest`, `go fmt ./...`, and `go vet ./...`

Fixes #1288

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -162,8 +162,12 @@ Grok section and remove the explicit registry exception in the coverage test.
   including reindex behavior, but not the event or database schema. The
   [configuration-directory reference](https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-config-dir-reference)
   further identifies `events.jsonl` and workspace artifacts. No
-  producer-side serializer is public. For independent legacy CLI and sibling
-  Copilot-store observations, clone
+  producer-side serializer is public. Reverified 2026-07-28 against the
+  sanitized session shape reported in
+  [#1288](https://github.com/kenn-io/agentsview/issues/1288): `events.jsonl`
+  emits matching `tool.execution_start` and `tool.execution_complete` records
+  keyed by `toolCallId`, whose timestamps provide the exact execution interval.
+  For independent legacy CLI and sibling Copilot-store observations, clone
   `https://github.com/getagentseal/codeburn.git` at
   `3472885629c41725b40c19c0780ecce148b067bf` and inspect its
   [Copilot format notes](https://github.com/getagentseal/codeburn/blob/3472885629c41725b40c19c0780ecce148b067bf/docs/providers/copilot.md)
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -167,6 +167,9 @@ Grok section and remove the explicit registry exception in the coverage test.
   [#1288](https://github.com/kenn-io/agentsview/issues/1288): `events.jsonl`
   emits matching `tool.execution_start` and `tool.execution_complete` records
   keyed by `toolCallId`, whose timestamps provide the exact execution interval.
+  Reverified 2026-07-28 against current local session artifacts that completion
+  records can include a boolean `success`; explicit `false` marks a failed
+  terminal execution, while older records may omit the field.
   For independent legacy CLI and sibling Copilot-store observations, clone
   `https://github.com/getagentseal/codeburn.git` at
   `3472885629c41725b40c19c0780ecce148b067bf` and inspect its
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -333,7 +333,10 @@ const projectIdentityRemoteScrubCompletedKey = "project_identity_remote_scrub_v1
 // (74: Claude Code IDE context reparse. Standalone ide_opened_file and
 // ide_selection wrappers are promoted to system metadata so existing
 // VS Code sessions no longer use them as titles or user turns.)
-const dataVersion = 74
+// (75: GitHub Copilot CLI tool execution events now provide exact tool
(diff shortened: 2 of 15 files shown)


wesm:
This was superseded by #1290

PR #1298, opened 2026-07-29
fix(duckdb): give local push-watch deferred scopes a polling owner

diazMelgarejo:
Local DuckDBPushWatch mirrored pg watch's pre-fix gap: no sync engine, no unwatched-root poller, and watcher options only wired OnCoverageDegraded. Interval pushes used runLocalSyncAuthoritative → SyncAll, which never tombstones missed deletions under deferred scopes.

Align local duckdb watch with local pg watch: shared watcher options, engine-backed batch sync, startup sync, and the probe-gated poller. Add TestLocalDuckDBPushWatchGivesDeferredScopesAPollingOwner parallel to the existing pg regression at archive_write_backend_test.go:821.

Daemon-delegated duckdb/pg watch remains unchanged (no local engine).

### DIFF
--- a/cmd/agentsview/archive_write_backend.go
+++ b/cmd/agentsview/archive_write_backend.go
@@ -74,7 +74,11 @@ type archivePushWatchHooks struct {
 	pgStartupSync func(
 		context.Context, *syncpkg.Engine, bool,
 	) (bool, error)
+	duckDBStartupSync func(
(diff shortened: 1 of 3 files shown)


wesm:
looking

wesm:
accepted, thank you!

PR #1306, opened 2026-07-30
fix(sync): bound OpenCode watcher fallback work

### LINKED ISSUE #1208: fix(sync): bound OpenCode background work across watcher and polling modes

rodboev:
OpenCode roots that miss recursive watcher coverage currently enter the generic two-minute unwatched-root poller. Each wake runs authoritative root reconciliation and lists the shared SQLite archive, and a pass longer than the interval can consume an already queued wake immediately after completion. Large installations can therefore sustain archive-scale reads and CPU while idle.

This carries provider-owned coverage units from watch planning into native and degraded scheduling. OpenCode's SQLite unit uses the event journal to select a bounded batch of settled session IDs, resolves each ID through targeted metadata, and feeds the existing provider parse path. The coordinator permits one active pass per coverage key, coalesces wakes, and schedules the next pass from completion. Other providers retain provider-scoped reconciliation.

Journal rows, payload bytes, pending IDs, ready IDs, and pass time have explicit caps. Checkpoints advance only after archive writes, retain work committed during an audit, and require another audit when the SQLite container is replaced. Unsupported, removed, or oversized journal state requests one provider-scoped authoritative audit and latches repeated journal work. Missing probes, recursive symlink gates, shared generic watch roots, storage canonicality, persistent archives, startup repair, and global overflow recovery keep their existing authority. The change adds no database schema, mirror, frontend, or server API surface.

The watcher-exhaustion trace and production CPU/read measurements came from @berenddeboer's issue report.

Closes #1208

### DIFF
--- a/cmd/agentsview/archive_write_backend.go
+++ b/cmd/agentsview/archive_write_backend.go
@@ -106,8 +106,9 @@ func newArchivePushUnwatchedPoller(
 		return hooks.newUnwatchedPoller(ctx, engine)
 	}
 	ticker := time.NewTicker(unwatchedPollInterval)
-	return newUnwatchedPollCoordinatorWithTicks(
+	return newUnwatchedPollCoordinatorWithSchedule(
 		ctx, engine, ticker.C, ticker.Stop, func(work func()) { work() }, nil,
+		unwatchedPollInterval,
 	)
 }
 
@@ -155,10 +156,16 @@ func archivePushWatchWatcherOptions(
 			return loop.NotifyCoverageDegraded(roots)
 		},
 		OnPollingRequired: func(obligation syncpkg.PollingObligation) error {
+			scopes := make([]pollingScope, 0, len(obligation.Scopes))
+			for _, scope := range obligation.Scopes {
+				scopes = append(scopes, pollingScope{
+					Agent: parser.AgentType(scope.Agent), Root: scope.SyncDir,
+					CoverageKey: scope.CoverageKey,
+				})
+			}
 			return poller.AddObligation(pollingObligation{
-				Key:   obligation.Key,
-				Roots: obligation.Roots,
-				Probe: obligation.Probe,
+				Key: obligation.Key, Roots: obligation.Roots, Probe: obligation.Probe,
+				NonBlockingProbe: obligation.NonBlockingProbe, Scopes: scopes,
 			})
 		},
 		OnPollingReleased: poller.RemoveObligation,
@@ -172,7 +179,7 @@ func archivePushWatchBatchCallback(
 ) syncpkg.WatchCallback {
 	return func(callbackCtx context.Context, batch syncpkg.WatchBatch) error {
 		scope := func() watchRecoveryScope {
-			return probeWatchRecoveryScope(appCfg)
+			return probeWatchRecoveryScopeWithOwner(appCfg, engine)
 		}
 		if err := syncWatchBatch(callbackCtx, engine, batch, scope); err != nil {
 			return err
--- a/cmd/agentsview/archive_write_backend_test.go
+++ b/cmd/agentsview/archive_write_backend_test.go
@@ -331,6 +331,19 @@ type pushWatchOwnerCase struct {
 	run  func(context.Context, *archivePushWatchHooks) error
 }
 
+func TestStartArchivePushWatcherWithoutLocalEngine(t *testing.T) {
+	root := t.TempDir()
+	cfg := config.Config{AgentDirs: map[parser.AgentType][]string{
+		parser.AgentOpenCode: {root},
+	}}
+	stop, open, _ := startArchivePushWatcher(
+		nil, cfg, nil, func(context.Context, syncpkg.WatchBatch) error { return nil },
+		syncpkg.WatcherOptions{},
+	)
+	open()
+	stop()
+}
+
 func pushWatchOwnerCases(t *testing.T) []pushWatchOwnerCase {
 	t.Helper()
 	// Keep SQLite setup outside the timed owner goroutines so channel deadlines
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -319,7 +319,7 @@ func runServe(cfg config.Config, opts serveOptions) {
 				// The serve ctx reaches watcher-driven syncs so SIGTERM can
 				// interrupt database reconciliation before Stop waits for it.
 				return syncWatchBatch(ctx, engine, batch, func() watchRecoveryScope {
-					return probeWatchRecoveryScope(cfg)
+					return probeWatchRecoveryScopeWithOwner(cfg, engine)
 				})
 			},
 			sync.WatcherOptions{
@@ -329,10 +329,17 @@ func runServe(cfg config.Config, opts serveOptions) {
 					})
 				},
 				OnPollingRequired: func(obligation sync.PollingObligation) error {
+					scopes := make([]pollingScope, 0, len(obligation.Scopes))
+					for _, scope := range obligation.Scopes {
+						scopes = append(scopes, pollingScope{
+							Agent: parser.AgentType(scope.Agent), Root: scope.SyncDir,
+							CoverageKey: scope.CoverageKey,
+						})
+					}
 					return unwatchedPoller.AddObligation(pollingObligation{
-						Key:   obligation.Key,
-						Roots: obligation.Roots,
-						Probe: obligation.Probe,
+						Key: obligation.Key, Roots: obligation.Roots,
+						Probe:            obligation.Probe,
+						NonBlockingProbe: obligation.NonBlockingProbe, Scopes: scopes,
 					})
 				},
 				OnPollingReleased: unwatchedPoller.RemoveObligation,
@@ -365,7 +372,7 @@ func runServe(cfg config.Config, opts serveOptions) {
 				// RecordStartupReconciled fires.
 				completeWorkerStartup = func() {
 					var gapErr error
-					if gapRoots := reconcileRootPaths(cfg); len(gapRoots) > 0 {
+					if gapRoots := reconcileRootPathsWithOwner(cfg, engine); len(gapRoots) > 0 {
 						gapErr = engine.ReconcileWatchRoots(ctx, gapRoots, false)
 					}
 					if gapErr != nil && ctx.Err() == nil {
@@ -770,6 +777,14 @@ func reconcileRootPaths(cfg config.Config) []string {
 	return probeWatchRecoveryScope(cfg).available
 }
 
+func reconcileRootPathsWithOwner(cfg config.Config, owner activeSourceProbe) []string {
+	return probeWatchRecoveryScopeWithOwner(cfg, owner).available
+}
+
+type activeSourceProbe interface {
+	HasActiveSessionSourceBelow(agent, path string) (bool, error)
+}
+
 // watchRecoveryScope is one probed availability snapshot of the configured
 // watch scope: the currently available reconciliation paths plus the
 // configured dirs whose physical scope is missing and therefore deferred to
@@ -799,22 +814,33 @@ func (s watchRecoveryScope) coversProviderRoot(root string) bool {
 // probeWatchRecoveryScope computes the probed reconciliation scope backing
 // reconcileRootPaths; see that function for the deferral semantics.
 func probeWatchRecoveryScope(cfg config.Config) watchRecoveryScope {
+	return probeWatchRecoveryScopeWithOwner(cfg, nil)
+}
+
+func probeWatchRecoveryScopeWithOwner(
+	cfg config.Config, owner activeSourceProbe,
+) watchRecoveryScope {
 	roots, unwatchedDirs, symlinkGatedDirs := collectWatchRoots(cfg)
 	deferred := make(map[string]struct{})
 	// A recursive symlink root never joins the watch roots, so its exact
 	// availability probe is the symlink target itself: os.Stat follows the
 	// link and fails while the target is gone, deferring the configured
 	// scope before an overlapping present path could expand into it.
-	for symRoot, dirs := range symlinkGatedDirs {
+	for symRoot, gate := range symlinkGatedDirs {
 		if _, err := os.Stat(symRoot); err == nil {
 			continue
 		}
-		for _, dir := range dirs {
-			deferred[filepath.Clean(dir)] = struct{}{}
+		root := watchRoot{path: symRoot, scopes: gate.scopes,
+			nonBlockingProbe: gate.nonBlockingProbe}
+		if gate.nonBlockingProbe && !watchRootHasArchivedSources(owner, root) {
+			continue
+		}
+		for _, scope := range gate.scopes {
+			deferred[filepath.Clean(scope.syncDir)] = struct{}{}
 		}
 	}
 	for _, r := range roots {
-		if r.exists {
+		if r.exists || (r.nonBlockingProbe && !watchRootHasArchivedSources(owner, r)) {
 			continue
 		}
 		for _, dir := range r.pendingPollingDirs {
@@ -859,6 +885,19 @@ func probeWatchRecoveryScope(cfg config.Config) watchRecoveryScope {
 		}
 	}
 	return watchRecoveryScope{available: paths, deferred: deferred}
+}
+
+func watchRootHasArchivedSources(owner activeSourceProbe, root watchRoot) bool {
+	if owner == nil {
+		return false
+	}
+	for _, scope := range root.scopes {
+		has, err := owner.HasActiveSessionSourceBelow(string(scope.agent), root.path)
+		if err != nil || has {
+			return true
+		}
+	}
+	return false
 }
 
 // overlapsDeferredScope reports whether the engine-side expansion of a
@@ -1805,6 +1844,17 @@ func startFileWatcher(
 ) {
 	t := time.Now()
 	roots, unwatchedDirs, symlinkGatedDirs := collectWatchRoots(cfg)
+	primeWatchCoverage(engine, roots)
+	authorityRoots := append([]watchRoot(nil), roots...)
+	for path, gate := range symlinkGatedDirs {
+		authorityRoots = append(authorityRoots, watchRoot{
+			path: path, scopes: gate.scopes,
+			nonBlockingProbe: gate.nonBlockingProbe,
+		})
+	}
+	onChange = annotateWatchCoverageWithAuthority(
+		onChange, roots, authorityRoots, activeSourceProbeForEngine(engine),
+	)
 	watcher, err := sync.NewWatcherWithCallback(
 		watcherBatchDelay,
 		watcherSyncMinInterval,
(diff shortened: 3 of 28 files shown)


berenddeboer:
Watching this with great interest @rodboev . My two cents is that I probably would have done a much purer queuing implementation. Watchers/pollers who dump work in the queue, and job workers who work on the queue, and probably have done this through a db queue, instead of in memory. If you are interested, you can see that implemented in [ready-for-agent](https://github.com/berenddeboer/ready-for-agent).

So I have [a job queue](https://github.com/berenddeboer/ready-for-agent/blob/main/packages/db-schema/src/schema.ts), implementing fairly standard queue semantics, and rely on db locking for queue semantics. [Queue implementation](https://github.com/berenddeboer/ready-for-agent/blob/main/packages/sqlite-queue-service/src/lib/sqlite-queue-service.ts) is an Effect TS service.

I suppose this isn't easy to read if you don't know TypeScript or use Effect TS (that's FP for TypeScript).

But I've found this pattern to be very stable, and survives restarts, which an in-memory implementation doesn't. For agentsview that's not needed, but still nice to have fast starts.

rodboev:
Agreed that the OpenCode journal feed should be a pure package. I am thinking of a split along these lines, listed in merge order; let me know what you think:

- D: Watch split. A non-recursive watch on the configured root for `opencode.db` and its WAL, exempt from the recursive budget, plus a recursive watch on `storage/` when that layout resolves. Keeps a budget-starved root natively watched rather than on the two-minute poller. Registration starvation only; overflow, unavailable backends, and poll cost stay with the later slices.
- A: Reconciliation-scoping half promised in [#1262 (comment)](https://github.com/kenn-io/agentsview/pull/1262/#issuecomment-5119886684): a reconcile pass walks only the roots it was asked about, with normalization done once through the provider contract.
- B: OpenCode journal feed as a pure package: journal rows in; ready and pending session identities out, plus continuation state and a latched audit reason. One contract change from the version here: the feed no longer resolves final sources. (Several findings here traced back to one provider trying to reconstruct archive ownership from the filesystem when the engine already owns that through shared stored paths and the existing FindSource path, so resolution moves engine-side.)
- C: Integration, stacked on #1307 and A + B: the engine resolves feed identities against archive ownership through the existing FindSource path, checkpoints commit only after the archive write and are serialized per coverage key, archive-scale reconciliation is gated behind the audit scheduler, and a keyed worker owns admission, wake coalescing, and one active pass per coverage unit.

B and C ship as one PR, since C is B's only consumer.

C is essentially the queue model you're suggesting: watchers and pollers enqueue keyed work, serialized workers apply it and ack after the archive write. In-process still makes more sense than db-backed, since OpenCode's journal is already the durable log and startup reconciliation the recovery path; persisting the queue would just add a second acknowledgement protocol.

The Effect TS service isn't importable from Go, but the semantics (unique keyed jobs, transactional claiming) are good reference points. Thanks for pointing that out.

rodboev:
It looks like #1291 takes out most of the reason for B and C; once it's merged, unchanged sessions will get skipped, and an idle pass over a large container becomes cheap.

So based on that:

- A new first slice: coverage units and a shallow SQLite watch, which I had left out. `WatchPlan` marks the whole configured root recursive (opencode_provider.go:558), so it competes for the recursive watch budget. Losing that competition is what drops OpenCode into the generic poller. A shallow root watches the db and its WAL without drawing on the budget.

- A is narrower than #1262 described: `ReconcileProviderRoots` already bypasses cross-provider expansion on main.

- B and C sit on top of #1291; their case is active-use cost rather than idle.

berenddeboer:
@rodboev looking at #1291 I feel it gets into the same issue I ran into with my third implementation. The code review never converges, and the code writing agent never solves all the problems you still have as its an architectural issue.

My hunch is that your first slice also won't converge, as you than run into the 2 minute poll / max time (I can't exactly describe the issue, but it feels this is what I still ran into).

So I would have pure work generators, which read the two opencode histories I know about (the old one and the new .db one), but do no work. They simply write session ids or session id ranges. These work generators keep scanning or polling. Then the job workers receive a range of session ids and read/parse them, and put them into agentsview. One can control the cpu/load of a session parser by the number of sessions it does at a time. Although for super large sessions this could still be substantial.

My first slice: write the work generators, queue and job worker purely for the old style opencode sessions, so that work gets removed from the normal path. If roborev is happy with this slice, the second step would be to move the modern opencode v1 db into this architecture.

The advantage of slice 1 is that this is old style, so you don't need to write the code to keep rescanning/polling this, we can be sure this is just importing old history. For slice 2 you then add the continual scanning for updates, but you have the queue architecture already in place.

Again I'm just sketching here, I don't know the agentsview code base nor the opencode architecture too well, this is just based on seeing my clanker attempting to solve this three times and running into one issue after another

berenddeboer:
@rodboev what I might do is just let the clanker write up a description of the two opencode architectures.

On the size mentioned in #1291 : my opencode is 27GB, so double.

berenddeboer:
@rodboev : I let the clanker write up a description of [opencode's storage format](https://github.com/berenddeboer/agentsview/blob/fix/1208-opencode-bounded-sync/docs/internal/opencode-storage.md). I think it's more useful detail than session-format-sources.md, but not sure you want this in your docs so haven't created a PR for this.

And the Xplain db format is extremely concise, and I can read this extremely fast and well, but it's very niche. But because it hides the details I find it great for humans.

I think this explains my hunch for slice 1 well: drop the json reading from the current path, do it via the new queue mechanism, and then you're done forever (except when you request a full scan).

Slice 2 is then fixing the issue that changing one session causes an entire repass. And hopefully this can fit nicely onto the new queue.

Please let me know if you want me to help out anywhere.

rodboev:
I think the 2 minute poll / max time is in `cmd/agentsview/main.go` (`unwatchedPollInterval = 2 * time.Minute`), and #1208 diagnoses it correctly: a poll took longer than two minutes, so the queued ticker wake started the next one immediately and the daemon had no idle interval.

I added that slice as D above. It lands ahead of A and keeps a budget-starved root off the poller; completion-based scheduling is in C.

The writeup is useful, thanks. Backfilling the JSON tree and treating the root as SQLite-only works makes sense for OpenCode, but I don't think it extends to the forks: my Kilo and MiMoCode ship the journal tables empty and still write `storage/` live, so they keep needing the recursive watch.

The backfill also flips hybrid shadowing, since the DB row would win where JSON does today. That plus the per-root state can be a follow-up slice; I'm trying to keep this reviewable.

berenddeboer:
Yes good point, db backfill shouldn't overwrite json. But implicitly I also wanted to improve the startup time of agentsview, so it doesn't start with a full scan, this simply gets scheduled when the parsing of all agents move to this architecture.

My opencode is 27GB now, and I'm getting better at this stuff, aiming for a consistent 150+ PRs a week, so startup time is going to be a real issue.

rodboev:
Yeah, I'm in the same boat with tens of GB of DBs spread across various harnesses. I would also like to see it queue things up in the background. The desktop app is much slower than having `agentsview serve` in the background on login and hit the web app on demand. There's still a quick reconciliation but much faster than desktop startup, so I avoid the app for now. Not sure which lag you're referring to, but try that if you haven't.

I've slowed down but [my rate](https://basedin.nyc/pr-stats/) was about the same at the 30-day mark with around 500 opened. I manually trim history to last 2 weeks to keep it manageable between agentsview and my hand-rolled history viewer. Sometimes this is the reason I use the hand-rolled one though — it caches on startup, but I also don't particularly care about the RAM hit.

berenddeboer:
I run it with agentsview serve. Just doesn't work. But my history goes all the way back to aider days. There's a lot. And I feel agentsview can potentially give me a lot of valuable data, and simply tracking improvements over time is already valuable.

PR #1307, opened 2026-07-30
fix(daemon): scope degraded-coverage polling to one provider per pass

rodboev:
When one provider loses native watch coverage, the unwatched-root poller
flattened every obligation into a single root list and called the unscoped
reconcile entry point, which expands each root across every configured agent.
One provider's coverage gap therefore dragged every overlapping provider
through a full authoritative discovery every two minutes, and an authoritative
pass for the wrong provider can tombstone sessions under a lifecycle-owned
missing root. Two smaller bugs compounded it: obligations for different
reasons over one physical root collapsed into one map entry (releasing one
silently released the other), and the poll interval was measured between
ticker deadlines, so a pass that ran longer than the interval was followed
immediately by the next one.

What this changes:

- `PollingObligation` now carries `(agent, root)` scopes instead of bare
  roots. Provider identity travels through registration, the watcher
  backends, and the polling handoff instead of being reconstructed
  downstream (`internal/sync/watcher.go`, `cmd/agentsview/main.go`).
- The coordinator groups available scopes by agent and reconciles each
  provider only over its own roots. Obligation keys are namespaced by reason
  and provider (`degraded:<agent>:<path>`, `nowatcher:<agent>:<path>`,
  `persistent:<dir>`), so overlapping obligations stay independently
  releasable (`cmd/agentsview/unwatched_poll.go`).
- Persistent-polling obligations carry scopes only for the providers that
  requested persistent polling for a directory, not every agent sharing it.
  `collectWatchRoots` records that provenance and `watchPollingObligations`
  consumes it (`cmd/agentsview/main.go`).
- A poll pass issues one grouped engine call,
  `Engine.ReconcileProviderRootsGrouped`, which runs each provider's bounded
  pass but shares a single epilogue — global subagent linking and skip-cache
  persistence — across the batch. Without this, that archive-sized work ran
  once per provider per tick. The sync lock is held across the groups and the
  epilogue so a concurrent pass cannot persist newer skip state that the
  shared epilogue would then overwrite; a canceled batch stops between groups
  and skips the epilogue rather than blocking shutdown on it; epilogue
  eligibility is recorded at each pass's own gate sites (after page writes
  commit, before tombstoning), so a tombstoning or cleanup failure cannot
  leave synced sessions unlinked; and session events are emitted once per
  batch outside the lock (`internal/sync/engine.go`).
- The pre-discovery SQLite container capture is scoped to the pass: a group
  outside the OpenCode container family probes no containers, and an
  in-family group probes only the containers overlapping its roots, so probe
  work per poll is bounded by the batch instead of group count or total
  configuration (`internal/sync/opencode_container_gate.go`).
- The worker waits a full interval measured from the previous pass's
  completion, not from the previous ticker deadline.

Deferral is conservative where scopes mix: an obligation without a provider
keeps the unscoped path, so a blocked provider-less root defers every provider
on it and vice versa. Probe gates, remote roots, watcher recovery, and the
daily audit are unchanged; the audit remains the backstop for anything a
scoped poll declines to prove.

Each provider's pass still performs a full discovery over its scope; making
the pass itself incremental needs provider-declared coverage units and is
deferred to #1318, which builds on the polling-scope API introduced here. The
15-minute scheduled sync (`runScheduledSyncPass`) still reconciles providers
individually and could adopt the grouped call in a follow-up.

Refs #1208.

### DIFF
--- a/cmd/agentsview/archive_write_backend.go
+++ b/cmd/agentsview/archive_write_backend.go
@@ -108,6 +108,7 @@ func newArchivePushUnwatchedPoller(
 	ticker := time.NewTicker(unwatchedPollInterval)
 	return newUnwatchedPollCoordinatorWithTicks(
 		ctx, engine, ticker.C, ticker.Stop, func(work func()) { work() }, nil,
+		time.Now, time.After,
 	)
 }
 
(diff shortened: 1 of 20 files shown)


wesm:
Looks good in principle, giving it some review and will merge

rodboev:
Sounds good. Fixed the lint issue, and opened the other two groundwork PRs for #1208: #1318 overlaps this one on the polling handoff and will rebase onto it once this lands, just to keep that as a smaller reviewable slice (versus adding ~2.5k lines); #1319 is based on #1291 so its diff shrinks when that merges.

The bounded change feed that actually closes the issue comes as a fourth PR on top of all three (this one, #1318, and #1319) — so the integration builds on code that's fully reviewed.

PR #1309, opened 2026-07-30
feat(vibe): read provider cache-hit tokens from session stats

verdie-g:
Mistral Vibe v2.23.2 (mistralai/mistral-vibe#969) started recording provider cache-hit (cached) tokens in the session stats written to `meta.json`. This teaches the Vibe parser to read them.

`stats.session_cached_tokens` is the provider cache-read count, reported as a subset of `session_prompt_tokens` (the OpenAI/Mistral `prompt_tokens` / `prompt_tokens_details.cached_tokens` wire shape). The parser now splits it out into the usage event's cache-read field and subtracts it from input tokens, so the cached prefix is priced at the discounted cache-read rate rather than the full input rate and is not double-counted. This mirrors the convention already used by the workbuddy and grok parsers.

Vibe exposes a single cached counter with no cache-creation breakdown, so `CacheCreationInputTokens` stays zero.

Reviewers should look at `vibeUsageEvents` in `internal/parser/vibe.go`. The format provenance entry in `docs/internal/session-format-sources.md` is updated in the same change.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -342,7 +342,12 @@ const projectIdentityRemoteScrubCompletedKey = "project_identity_remote_scrub_v1
 // (76: Copilot CLI tool execution boundaries. Re-parsing persists
 // tool.execution_start and tool.execution_complete timestamps as result events
 // so Session Analysis excludes resumed-session idle time from completed calls.)
-const dataVersion = 76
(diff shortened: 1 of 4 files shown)


EwenQuim (approved):

wesm:
will merge pending ci

PR #1311, opened 2026-07-30
fix: improve worktree path readability

mariusvniekerk:
Long worktree paths often share a low-signal prefix, while the final checkout or branch segments identify the context a developer actually needs. This keeps that useful tail visible in the compact session-vitals panel and preserves the actual order of mixed-direction paths through explicit LTR isolation.

The complete value is available through the shared tooltip for hover and keyboard focus. This worktree-specific tooltip can use up to half the viewport, expand left beyond the sidebar, and wrap long unbroken segments at the cap while remaining viewport-contained with its arrow anchored to the trigger; kit-ui’s generic 280-pixel tooltip default remains unchanged. Existing copy behavior and the empty-path state are preserved.

### DIFF
--- a/cmd/testfixture/main.go
+++ b/cmd/testfixture/main.go
@@ -449,6 +449,7 @@ func createDurationShowcaseFixture(
 		Project:          project,
 		Machine:          "test-machine",
 		Agent:            "claude",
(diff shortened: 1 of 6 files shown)


mariusvniekerk:
Screenshot showing the start-truncated worktree path and its full kit-ui tooltip.

PR #1312, opened 2026-07-30
fix(frontend): move sidebar toggle beside filters

mariusvniekerk:
Moves the desktop sidebar control out of the global header and into the session filter controls it affects. With the sidebar open, the filter is followed by the collapse action; when closed, the expand action precedes the relocated filter on both the analytics landing page and session detail breadcrumb.

The mobile hamburger and drawer behavior remain unchanged. A shared icon button provides state-specific accessible labels, with the new open-sidebar copy kept in sync across every locale.


wesm:
fixing the windows ci

PR #1314, opened 2026-07-30
feat(usage): report complete delegated session costs

mjacobs:
This PR makes session usage totals include work delegated to Claude Code subagents.

- Includes all descendant subagents in session usage by default; --own-only keeps the transcript-only view.
- Refreshes local and S3 subagent data before combined CLI reads, while keeping GET endpoints read-only.
- Requires daemon API v6 for remote combined totals so older daemons fail clearly instead of silently undercounting.
- Deduplicates repeated Claude streaming snapshots before calculating tokens and cost, while preserving the original session, project, agent, and machine attribution.
- Adds billed Claude web-search request fees and reports a combined cost only when every included token-bearing session has complete pricing coverage.
- Keeps daily usage, activity reports, reporting exports, SQLite, PostgreSQL/CockroachDB, and DuckDB consistent for filtering and snapshot attribution.
- Lets the embedding scheduler retry on a later periodic tick after one bounded retry cycle is exhausted.
- Known limitation: Claude Code does not store token usage for its separate web-search side call, so those unavailable tokens cannot be reported.


wesm:
rebasing and fixing

PR #1318, opened 2026-07-31
fix(opencode): split the watch plan into database and storage coverage units

rodboev:
The OpenCode provider declares one recursive watch root per configured directory, so registration walks the whole tree and draws on the process-wide recursive budget. On a large installation the budget runs out, the root gets no native coverage, and every poll becomes an authoritative reconciliation over the archive. The database and its write-ahead log are direct children of the configured root and never needed recursion.

This splits the plan into two coverage units per root: a non-recursive unit on the configured root covering only the database and its WAL, exempt from the recursive budget, and a recursive unit on the storage subtree, emitted only when the root resolves to file-backed storage. The units fail independently, so losing storage coverage no longer takes SQLite observation with it. A changed path is claimed by exactly one unit while the storage directory's existence is stable, and Kilo, MiMoCode, and ICodeMate inherit the same shape through the shared provider spec. The watch plan is cached at startup, so a storage tree appearing at runtime gets recursive coverage at the next startup, with the daily audit as backstop. A unit that loses coverage is still polled by its configured root at the same archive scale as before; carrying unit identity through the polling obligation belongs to the follow-up.

Measured on windows/amd64 with fsnotify v1.10.1: the non-recursive watch observes the database, its WAL, and storage-directory creation, and does not observe grandchildren. The watcher-exhaustion trace and the production measurements came from the issue report. It is one of three preparatory PRs for #1208, with #1307 and the reconciliation-scoping change (#1319), staged separately so roborev reviews each in isolation; the bounded change feed and coalesced audit that close the issue follow once they land. This branch overlaps #1307 on the polling obligation surface and rebases onto it after #1307 merges.

Refs #1208

### DIFF
--- a/cmd/agentsview/opencode_coverage_units_test.go
+++ b/cmd/agentsview/opencode_coverage_units_test.go
@@ -1,0 +1,293 @@
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"slices"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.kenn.io/agentsview/internal/config"
+	"go.kenn.io/agentsview/internal/parser"
+	agentsync "go.kenn.io/agentsview/internal/sync"
+)
+
+// newOpenCodeSQLiteWatchDir creates a configured OpenCode directory whose
+// only source is a SQLite database, the shape of the reported
+// budget-starvation regression.
+func newOpenCodeSQLiteWatchDir(t *testing.T) string {
+	t.Helper()
+	root := t.TempDir()
+	writeTestFile(t, filepath.Join(root, "opencode.db"), []byte("sqlite"))
+	return root
+}
+
+func registerCollectedWatchRoots(
+	t *testing.T, roots []watchRoot, budget int,
+) []agentsync.RecursiveWatchResult {
+	t.Helper()
+	watcher, err := agentsync.NewWatcherWithCallback(
+		time.Millisecond, 0,
+		func(context.Context, agentsync.WatchBatch) error { return nil },
+		nil, agentsync.WatcherOptions{},
+	)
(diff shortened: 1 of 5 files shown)


rodboev:
The storage tree narrowing is discussed in the PR body: the watch plan is cached at startup, so the shallow unit observes `storage/` creation but classifies the bare directory to no source, and files added at runtime wait for restart or the daily audit.

The reason an always-emitted storage unit doesn't work is documented on `watchUnits` itself:

https://github.com/kenn-io/agentsview/blob/bebdb8357e9ae4c967a4883c8e8fbedcfe68c528/internal/parser/opencode_provider.go#L560-L573

A unit emitted before the directory exists is a permanently missing probe on pure-SQLite roots. The poller defers every candidate overlapping a blocked probe, so it would silence the fallback polling that matters when watcher coverage degrades.

A pending watch that owns the ancestor lifecycle would avoid the probe. The Darwin backend has that shape, but the portable backend and the engine's cached changed-path plan don't, so dynamic activation belongs to the bounded change-feed follow-up on #1208.

The `unitScopeAllows` point is inert until then: with no recursive watch registered, no nested event is ever delivered, so classification never sees those paths.

rodboev:
Same finding as the last run; the commit in between only clarified the `watchUnits` comment that explains why. That rationale still holds: always emitting the unit leaves a permanently missing probe on pure-SQLite roots, which blocks the fallback poll, and activating it on creation needs ancestor lifecycle ownership in the portable backend plus invalidation of the engine's cached plan, so it goes with the change-feed follow-up on #1208.

rodboev:
This rebased cleanly on top of #1307 after that one merged.

rodboev:
Rebased onto current `main` and starting on the integration slice mentioned in https://github.com/kenn-io/agentsview/pull/1306#issuecomment-5137218124.

rodboev:
Closing as superseded by #1355.

PR #1322, opened 2026-08-01
fix(devin): treat session timestamps as epoch seconds

### LINKED ISSUE #1199: Devin CLI sessions sync but `started_at`/`ended_at` are never populated (all rows get NULL, same file_mtime)
### agentsview version

agentsview v0.38.1

### Install method

Other / not sure

### OS / platform

macOS

### Which agent and version

devin cli

### Which model(s)

sonnet-5

Since Devin CLI session support was added (#961), AgentsView correctly *discovers* and syncs all Devin CLI sessions from `~/.local/share/devin/cli/sessions.db`, but it never populates per-session timestamps. Every synced Devin session ends up with `started_at`/`ended_at` = NULL, and all rows share the exact same `file_mtime` (which turns out to be the mtime ofthe `sessions.db` file itself, not a per-session value). This breaks any date-based view (recent sessions list, activity heatmap, etc.) for Devin CLI — sessions appear to just stop updating past a certain date even though new sessions keep being created.

### Environment

- agentsview v0.38.1 (commit 1e173b95, built 2026-07-13T19:56:25Z)
- macOS
- Devin CLI, sessions stored in `~/.local/share/devin/cli/sessions.db`

### Steps to reproduce

1. Use Devin CLI regularly so `~/.local/share/devin/cli/sessions.db` accumulates sessions.
2. Run `uvx agentsview serve` and let it sync.
3. Query the local agentsview DB:
   ```sql
   sqlite3 ~/.agentsview/sessions.db \
     "select agent, count(*), max(started_at), max(file_mtime) from sessions group by agent;"

  4. Observe that the devin row count matches the actual number of sessions in Devin's own DB (in my case, 144 == 144), confirming discovery works — but max(started_at) is empty,
  and every Devin row shares one identical file_mtime value.

Root cause (as far as I can tell)

Devin CLI's own sessions.db has a sessions table with real per-session timestamps:

CREATE TABLE sessions (
  id TEXT PRIMARY KEY,
  working_directory TEXT NOT NULL,
  backend_type TEXT NOT NULL,
  model TEXT NOT NULL,
  agent_mode TEXT NOT NULL,
  created_at INTEGER NOT NULL,       -- unix seconds
  last_activity_at INTEGER NOT NULL, -- unix seconds
  title TEXT,
  ...
);

Example row:

id: 0575d664-04b7-4bab-8af0-41156cbfcef3
created_at: 1776328158        -- 2026-04-16 10:29:18
last_activity_at: 1776328400
title: "Check PR #873 for Addressed Comments"

But in AgentsView's own sessions table, every Devin row has:

  • started_at: empty/NULL
  • ended_at: empty/NULL
  • file_mtime: identical across all 144 rows — looks like it's using the mtime of sessions.db itself rather than mapping created_at/last_activity_at per session.

It looks like the Devin CLI parser (added in #961) isn't reading created_at / last_activity_at from the sessions table into the corresponding session fields.

Expected behavior

started_at should be populated from created_at, and ended_at/last-activity from last_activity_at, per Devin session — same as other agents (Claude, Codex, etc.) already do correctlyin my DB.

Impact

Any UI surface that relies on session dates (recent sessions, activity heatmap, usage/date filters) effectively treats all Devin CLI sessions as if they happened on the same day (ordrops/misorders them), making it look like Devin CLI sessions "stopped syncing" even though sync is actually working and picking up all sessions.

Suggested fix

In the Devin CLI parser, read created_at and last_activity_at (Unix epoch seconds) from each row of ~/.local/share/devin/cli/sessions.db's sessions table and map them to started_at/ended_at (and use per-row values instead of the DB file's mtime for file_mtime).

Happy to provide more DB samples/schema if useful.


```text

```

### Steps to reproduce

_No response_

### Checklist

JohnRiv:
The Devin CLI stores created_at and last_activity_at as Unix epoch seconds, but the parser treated them as milliseconds. This produced dates in 1970 that were discarded as invalid, leaving started_at and ended_at NULL for every Devin session.

Replace devinUnixMilli (time.UnixMilli) with devinUnixSec (time.Unix) and correct the FileMtime multiplier from 1e6 to 1e9. The same fix applies to message_nodes.created_at, which is also epoch seconds.

Rename variables and test helpers from the misleading MS/Millis suffixes to match the Devin DB column names. Update all test data from 13-digit millisecond values to 10-digit second values.

Add timestamp-unit evidence to the Devin provenance entry.

I tested this and confirmed it fixes #1199

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -820,7 +820,9 @@ Grok section and remove the explicit registry exception in the coverage test.
 ## Devin CLI (`devin`)
 
 - **Format:** `cli/sessions.db` for session metadata plus transcript JSON
-  artifacts.
+  artifacts. The `sessions.created_at`, `sessions.last_activity_at`, and
+  `message_nodes.created_at` columns are Unix epoch seconds (not
+  milliseconds). Verified against a live Devin CLI database 2026-07-31.
 - **Evidence:** `no-public-source`.
 - **Upstream:** Cognition's first-party
   [Devin documentation](https://docs.devin.ai/) and public repositories were
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -822,7 +822,12 @@ Grok section and remove the explicit registry exception in the coverage test.
 - **Format:** `cli/sessions.db` for session metadata plus transcript JSON
   artifacts. The `sessions.created_at`, `sessions.last_activity_at`, and
   `message_nodes.created_at` columns are Unix epoch seconds (not
-  milliseconds). Verified against a live Devin CLI database 2026-07-31.
+  milliseconds). Verified against a live Devin CLI database 2026-07-31, and
+  reverified independently against CLI 3000.3.22 the same day. Because the unit
+  is observed rather than documented, the parser rejects values outside the
+  nanosecond-representable epoch-second range instead of converting them, so a
+  future unit change surfaces as missing timestamps rather than as a silently
+  overflowed far-future mtime that would wedge resync.
 - **Evidence:** `no-public-source`.
 - **Upstream:** Cognition's first-party
   [Devin documentation](https://docs.devin.ai/) and public repositories were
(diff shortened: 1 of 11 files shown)


mjacobs:
thanks, @JohnRiv, looking now

PR #1325, opened 2026-08-01
test: stabilize background wait synchronization

mariusvniekerk:
Two background timing tests depended on receiver or scheduler ordering and could exercise the wrong state under differently scheduled CI runners.

The unwatched poll cooldown test now retains its one-shot completion notification so a completed reconcile pass cannot become a timeout. The daemon replacement startup-wait test now waits until ensureBackgroundServe enters the startup wait before publishing the incompatible runtime, ensuring it exercises replacement rather than a fresh launch.

Both changes are test-only; production polling and daemon lifecycle behavior are unchanged. The race/non-race job split already exists on main via #1323.


wesm:
looking

PR #1330, opened 2026-08-01
feat(import): support Google Takeout Gemini Apps activity exports

### LINKED ISSUE #1295: Feature request: Support importing Google takeout for Gemini
### Problem / motivation

The exports from Claude.ai as well as ChatGPT are supported. I used google's [gemini](https://gemini.google.com/) recently and google provides google takeout for users, https://support.google.com/gemini/answer/16920332?hl=en and thus it is possible that these exports can also be supported.

### Proposed solution

Investigate if this is possible and maybe implement it if it is useful for others/interesting.

### Alternatives considered

None

rodboev:
Adds `agentsview import --type gemini-apps` support for Google Takeout Gemini Apps activity exports. Whole-hour numeric zones could be interpreted as UTC when the timestamp matcher accepted only the `GMT` prefix after the numeric form failed, so the import could persist a valid-looking record with the wrong instant and stable session ID.

The importer admits Gemini Apps records by their document structure and compatibility grammar, skips non-prompt activity, and emits each compatible `Prompted` activity as one stable plain-text message. It keeps visible text from inline and preformatted code, including spaces, tabs, line breaks, and authored backticks, without treating HTML presentation as speaker roles or generating Markdown delimiters. The parser now consumes each complete timestamp zone token and validates it once at the existing Gemini Apps zone authority. It accepts the existing named zones plus complete `GMT±H`, `GMT±HH`, `GMT±H:MM`, and `GMT±HH:MM` forms, with omitted minutes treated as zero. Malformed and unsupported complete tokens fail before import callbacks, while canonical UTC timestamps continue to drive timestamp-local stable IDs.

Unsupported locale or timestamp grammar fails before callbacks, while unrelated Takeout products are ignored and valid Gemini records in the same archive remain importable. The import-only provider, shared persistence, reimport identity, FTS handling, CLI dispatch, and Gemini CLI behavior remain unchanged. Browser upload, HTTP, schema, sync, remote transfer, frontend import UI, and raw archive handling are out of scope.

Closes #1295


wesm:
thanks!

PR #1349, opened 2026-08-05
feat(parser): add TraeX (TRAE CLI 2.0) as a first-class agent

GhostFlying:
TRAE CLI 2.0 (`traex`) is a closed-source fork of codex-rs. It writes rollout
JSONL that is field-for-field the Codex shape into `~/.trae/cli/sessions`,
including the subagent `session_meta`, `token_count` records, and a sibling
`history.jsonl` at the same position relative to the sessions root. Until now
the only way to index it was to declare that root as a `codex` source, which
gives the sessions the wrong agent and an ID prefix whose one-click resume
command (`codex resume <id>`) cannot resume them.

This adds `traex` as its own agent, served by the Codex provider through a
relabel hook — the same shape `opencode_provider.go` already uses to serve
Kilo, MiMoCode, and Icodemate from one implementation. Parsing, discovery,
fingerprinting, and JSONL-tail incremental appends are shared; only the agent
label and the `traex:` ID prefix differ. `internal/parser/traex.go` is the
whole fork-specific surface and is the first place to look.

The relabel covers more than Kilo's does. Besides the session ID, parent, and
source-session ID, the Codex parser stamps `codex:` onto the
`SubagentSessionID` of tool calls and their result events, so a partial
relabel would leave a TraeX parent pointing at `codex:<uuid>` rows the
`traex:` namespace never stores. The provider's incremental path returns
appended messages with no session, so the hook takes `(*ParsedSession,
[]ParsedMessage)` and treats a nil session as that path.

Discovery keys are now agent-scoped (`CodexSourceKey` takes an agent). Both
agents name files by UUID, so a shared key would let one agent's session
resolve to the other's file.

In `internal/sync`, the branches that describe the *format* — same-UUID
duplicate resolution, dated-over-flat layout preference, reconciliation
identity, the incremental `file_mtime` basis, and the parse-diff raced guard —
move behind a new `isCodexFormatAgent` predicate. The branches that describe
Codex's *out-of-band files* stay keyed to `parser.AgentCodex` alone, because
TraeX writes neither: there is no `session_index.jsonl` sidecar (so no title
backfill, no index-driven cache invalidation, and no index watch or changed-path
fan-out), and no `s3://.../raw/codex/...` archive convention, so a `traex` root
that looks like an S3 URI is skipped rather than imported under Codex's
identity. That split is the main thing worth reviewing.

Two consequences of having no index file are worth stating. Titles come from
the rollout head only, so a TraeX session never picks up a renamed thread name.
And `~/.trae/cli` is not watched shallowly, unlike `~/.codex`, since the only
reason to watch it would be the index file and the directory otherwise holds
constantly-rewritten SQLite WALs.

Evidence for the format is observational and documented as `no-public-source`
in the format inventory: TRAE CLI ships only as a binary, and neither Trae's
product site nor the public `Trae-AI/Trae` repository publishes a session
schema. A de-identified rollout with subagent spawn/wait records is kept as a
fixture.

Switching an existing installation from the `codex`-source workaround to
`agent = "traex"` is not handled: the stored rows are keyed by file path, so
the old `codex:<uuid>` rows would satisfy the freshness skips and the new rows
would never be written. A full resync into a fresh database is the intended
path.


wesm:
looking

PR #1362, opened 2026-08-07
test(parser): shorten posit-assistant fixture paths for Windows

statik:
The posit-assistant testdata tree used full-length placeholder IDs: a 24-character workspace hash and 36-character UUIDs for the main conversation, its subagent, and the default-workspace conversation. The deepest fixture file sat 178 characters below the repository root. When the repository is checked out as a submodule under a deep CI workspace prefix on Windows, the absolute path exceeded the 260-character MAX_PATH limit and broke the clone.

The fixture directories now use truncated IDs (a1b2c3d4, 11111111, 22222222, 33333333), capping the deepest path at 106 characters. The matching ID references inside conversation.json, index.json, and the test constants are updated in step. The parser's IsValidSessionID only requires alphanumeric, dash, or underscore names, so discovery, find-by-ID, changed-path classification, and parent-linkage behavior are unchanged; a comment on the test constants records why the IDs are short so they are not restored to full UUIDs later.

Reviewers should look at the constants and comment at the top of internal/parser/posit_assistant_provider_test.go and the renamed tree under internal/parser/testdata/posit-assistant/workspaces/. The tradeoff is that fixture names are no longer format-realistic UUIDs; the parser does not validate UUID shape, so no covered behavior depends on it. The grok-build fixtures are now the longest paths in the repository at 144 characters relative and remain untouched here.


statik:
@wesm looks like the deep paths I added when working on posit assistant support cause problems on some windows builds. This tries to keep things under the path length limits.

PR #1365, opened 2026-08-08
feat(amp): record per-inference token usage and model

shrivatsas:
Amp thread exports include per-inference usage and timestamps, but AgentsView previously ignored them. Exported Amp sessions now show model-specific token use, cost, and accurate daily attribution.

- Records each inference’s model, timestamp, input/output tokens, and cache buckets.
- Preserves Anthropic-family semantics and treats OpenAI-family cache-creation tokens as uncached input to avoid inflated cost.
- Reprocesses existing Amp sessions with data version 84 and documents the format evidence.

This reads complete thread documents already on disk; it does not fetch server-side threads. Amp exports also omit subagent usage, so those tokens remain unavailable.

Review focus: `internal/parser/amp.go`, `internal/parser/amp_provider.go`, and the Amp fixtures in `internal/parser/amp_test.go`.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -665,11 +665,42 @@ Grok section and remove the explicit registry exception in the coverage test.
 
 - **Format:** One JSON thread document per session.
 - **Evidence:** `no-public-source`.
-- **Upstream:** The first-party [Amp manual](https://ampcode.com/manual) and
-  public Sourcegraph/Amp repositories were searched 2026-07-19; no
-  session-file producer or authoritative disk schema was found.
-- **Usage and cost:** The consumed thread documents do not expose token, cache,
-  reasoning, credit, or USD fields to Agentsview.
+- **Upstream:** The first-party [Amp manual](https://ampcode.com/manual), its
+  [appendix](https://ampcode.com/manual/appendix), the
+  [CLI guide](https://github.com/sourcegraph/amp-examples-and-guides/blob/main/guides/cli/README.md),
+  and public Sourcegraph/Amp repositories were searched 2026-07-19 and again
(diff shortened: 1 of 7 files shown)


shrivatsas:
Both findings were accurate. Fixed in 5e4471a9, one of them differently from the suggestion — reasoning below.

**Per-inference timestamps — fixed as described.** Confirmed: `amp.go` never set `ParsedMessage.Timestamp`, so `COALESCE(NULLIF(m.timestamp, ''), s.started_at, '')` fell back to the session start for every inference and a long-lived thread collapsed into one daily bucket. Amp does carry `usage.timestamp`, and it is now parsed. Across the threads I have, 96 of 105 model-bearing messages now carry their own inference time; the remaining 9 are an older thread that omits the field and keeps the previous fallback. Tests cover both a thread spanning two dates and a thread with no timestamp.

**Model-less rows — documentation corrected instead.** The finding is right that `usageMessageEligibility` requires `m.model != ''`, so these rows are absent from daily totals and breakdowns. My docs claimed they "contribute tokens but no cost," which overstated it; that wording is now corrected in both `docs/token-usage.md` and the provenance entry to say they are excluded from usage reporting while remaining stored and visible in the session's own usage view.

I did not implement the suggested remedy, for two reasons. A `<synthetic>`-style sentinel would not work — that value is explicitly excluded by the same predicates, so it changes nothing. And including token-bearing rows without a model as unpriced means changing `usageMessageEligibility` and its PostgreSQL and DuckDB counterparts, which alters totals for every agent that has ever written a model-less usage row, not just Amp. That is a cross-cutting reporting-semantics decision rather than an Amp parser change.

wesm:
looking at this

PR #1367, opened 2026-08-08
feat(recall): organize generated insights and quality

wesm:
- Recall now groups durable corpus entries and saved model-written reports under one URL-backed workspace, so users no longer have to infer how two kinds of reusable context relate.
- Generated insights expose their complete session scope before execution, preserve that scope across tab changes, and retain deep links plus export, publish, retry, and delete workflows.
- Deterministic recommendations and evidence move to Quality, making the model-written versus measured-data boundary explicit; the former `/insights` route is a clean break and falls through to Sessions.
- Read-only backends keep access to saved generated reports while hiding the unsupported corpus, and the public docs now describe both surfaces and their privacy boundary.


wesm:
UI captures from the screenshot fixture:

- Generated report content is synthetic.
- Quality shows aggregate metrics from a bounded, scrubbed fixture; real saved insights, machine identity, paths, and blocked/private terms are removed.

### Recall — Generated insights

Recall Generated insights

### Quality

Quality

PR #1368, opened 2026-08-09
fix(frontend): keep root sessions landing unfiltered

wesm:
A bare local visit should show the unfiltered Sessions landing page, but the remembered filter state was being applied immediately and could be written back during refresh or session-detail navigation. That made the root URL surprising and risked erasing the user’s saved Sessions preference.

This change treats the root path as an unfiltered landing state while preserving the saved preference. Explicit filter URLs remain authoritative, direct or provenance-preserving navigation to bare /sessions can restore the remembered view, and opening a session from the root does not overwrite the preference when the detail view refreshes or closes. The task-specific design artifacts have also been removed now that the behavior is implemented.

The full frontend suite passes in serialized single-worker mode. The default parallel runner is resource-sensitive in this environment and previously timed out in an unrelated Playwright-config test; that focused test passes independently.


wesm:
Just giving this a last review before merging

PR #1372, opened 2026-08-10
feat: add proactive cross-chat issue review

xboxmasters:
Adds a local-first Issue Review workflow for recurring failures, blockers, slow tools, repeated work, and remediation opportunities across scoped sessions. It includes deterministic evidence links, complete filters and pagination, cached and forced refresh, telemetry coverage, privacy redaction, and allocation reductions for production-scale archives.

Named browser-local filter views preserve complete Issue Review filter sets, support case-insensitive updates and deletion, reject invalid stored data, and cap retained views at 50.

Acknowledgement and timed or permanent suppression decisions now persist in SQLite and PostgreSQL, survive full archive rebuilds, and overlay the cached detector result on every request. DuckDB remains read-only and active-only; no automatic remediation or external repository, chat, skill, or GitHub mutation is introduced.


wesm:
Thanks for your contribution. I'll need a little time to dig into this

wesm:
Thanks for the work here — there is a lot of solid engineering in this PR (the detector is well tested, redaction and read-only handling look careful, and backend parity was clearly taken seriously). I did a close review and found several things that need to be addressed before it can move forward. Posting them up front to see if you'd like to pick them up.

**The PR renames the Insights page.** `nav_insights` becomes "Issue Review" in all locales, and existing section labels are rewritten ("Deterministic Recommendations" → "What to fix next", etc.), with test expectations edited to match. That's a product decision that shouldn't ride along in a feature PR. Please revert the renames — the new panel can live inside Insights as-is.

**The Codex parser change needs a data-version bump.** `internal/parser/codex.go` now stores commentary events as assistant messages, and the detector's blocker detection depends on them. But `dataVersion` in `internal/db/db.go` is still 80, so existing archives never reparse and old sessions silently miss those findings. It needs a bump to 81 with a history note, like every prior parser-output change.

**Two detector bugs:**

- Repeated-polling findings multiply the *first* call's duration by the run length instead of summing each call, and attach N identical copies of the same evidence entry (the run loop around `internal/db/issue_review.go:1095`). Reproduced: four polls of 40s/10s/10s/10s report 160s total and wasted instead of 70s.
- `dedupeIssueCalls` keys on tool + `tool_use_id` with no session scoping. RooCode, Kilo-legacy, and Poolside synthesize ordinal-based IDs (e.g. `roocode:execute_command:3`), so unrelated calls from different sessions collide and get merged or dropped. The fix needs care, since cross-session dedupe of imported copies looks intentional.

**Personal working files are committed.** `docs/issue-review-handover.md` is a personal release/ops log (machine-specific details, a local scheduled task), and `.claude/skills/run-postgres-integration-tests/` is a Windows-specific test flow that conflicts with the documented `make test-postgres` path. Both should come out, along with the `docs/insights.md` link to the handover doc.

**Some new frontend tests assert source text, not behavior.** The "detector contract" block in `InsightsPage.test.ts` greps the component source for strings instead of exercising it. `IssueReviewPanel.test.ts` in the same PR shows the right pattern — please rewrite or drop the grep block.

There are smaller items too (a SQLite-vs-Postgres join difference for orphaned tool calls, "At least 1 chats" pluralization, evidence links swallowing cmd/ctrl-click), but the above are the blockers. Happy to go into detail on any of these — let me know if you want to take them, or if you'd prefer I finish it off.

PR #1373, opened 2026-08-10
fix(parser): trim replayed prefix of backgrounded Claude sessions

### LINKED ISSUE #1370: Backgrounded Claude Code sessions are ingested as duplicate top-level sessions
### agentsview version

agentsview v0.40.0 (commit a2cebf55, built 2026-08-03T02:33:05Z)

### Install method

Other / not sure

### OS / platform

Ubuntu 26.04

### Which agent and version

Claude Code

### Which model(s)

claude-opus-5

Sending a running Claude Code session to the background copies the entire conversation into a new transcript under a new session id and continues there. The original file stops mid-conversation. Agentsview ingests both files as unrelated sessions, so one conversation appears twice in the sidebar and its usage is counted twice per-session.

```text

```

### Steps to reproduce

1. Start a Claude Code session in any project and send a few prompts.
2. Mid-turn, press the left-arrow key. Claude Code shows a session picker; select the same thread you were already in. The prompt then shows the thread's name in a highlighted box above the input — you are now in a new transcript continuing the old conversation. (`Ctrl+B` and `/background` are the other two triggers for the same handoff.)
3. Continue for another turn or two, then exit. The project's transcript directory now holds two `.jsonl` files.
4. If AgentsView is still watching the agents, it will show 2 sessions on the sidebar with pretty much the same content. One of the 2 has a few more turns due to step 3 above.

### Checklist

wesm:
Fixes #1370.

Backgrounding a Claude Code session (left-arrow picker, Ctrl+B, `/background`) copies the whole conversation into a new transcript with a new session id and no back-reference. agentsview ingested both files as unrelated sessions: the conversation showed up twice in the sidebar and its usage was counted twice.

Verified against controlled reproductions and the Claude Code 2.1.226 bundle: copied lines are byte-identical except a rewritten `sessionId` and an injected `sessionKind:"bg"`; only uuid-bearing chain entries are replayed; the marker is process-derived, so ordinary interactive forks carry none.

## Fix (`internal/parser/claude_lineage.go`)

- A bg-marked transcript finds the same-directory sibling that shares its root uuid and covers the longest prefix — its nearest ancestor — drops that replayed prefix, and links to it as a `continuation` (nested in the sidebar).
- Chains trim link by link. A candidate that strictly contains the fork is its descendant and never wins; exact-equal twins pick a deterministic direction by stem.
- A fork with no content of its own is excluded; a prompt queued at handoff counts as content.
- Everything ambiguous fails open (no trim, no link): unmarked manual forks, missing or divergent siblings, branches at the replay boundary. Worst case is the status-quo duplicate.
- Data version 83 re-parses existing archives to repair old duplicates.

## Limits

- Root-to-root matching only: sessions compacted before backgrounding stay duplicated for now.
- Lineage reads on-disk state at parse time; a sibling appearing after a fork was stored is picked up at the next resync.
- Only the local Claude provider participates; uploads, Cowork, Qoder, subagents, and s3 sources are untouched.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -98,18 +98,17 @@ Grok section and remove the explicit registry exception in the coverage test.
   persists verbatim inside the stored usage object. Verified 2026-07-30
   against two local web-search sessions: when the search is driven by the
   **CLI's** `WebSearch` tool, every assistant record carries
-  `server_tool_use: {"web_search_requests": 0, "web_fetch_requests": 0}` —
-  the search itself runs in an out-of-band side call that is **not written to
-  the transcript at all**. The only surviving evidence is the tool-result
-  record's `toolUseResult` object
-  (`{query, results, durationSeconds, searchCount}`), whose `searchCount`
-  matched the wire-billed `web_search_requests` (1 == 1) in both sessions.
-  Agentsview therefore credits the assistant message that issued the
-  `WebSearch` `tool_use` with its linked result's `searchCount`, and uses the
-  message's own counter instead whenever that counter is nonzero (which is
-  what sessions driving the API directly report), so a search is never
-  counted twice. **Known undercount:** the side call's own token usage — tens
-  of thousands of input tokens on `claude-haiku-4-5` per search — is not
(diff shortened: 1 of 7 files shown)


wesm:
declined. Merging this

PR #1380, opened 2026-08-11
feat: import Claude.ai conversations through desktop auth

pascalwhoop:
## Summary
- add a desktop-only authenticated Claude.ai connection and browser transport
- ingest browser-fetched Claude conversations locally with durable incremental sync, retry/cancel support, and optional while-running scheduling
- preserve non-text Claude blocks losslessly during parsing
- add explicit **Repair import**, which refetches conversations and restores only matching Claude.ai permanent-deletion markers

## Validation
- `go test ./internal/cloudsync/claudeai ./internal/db ./internal/server ./internal/parser`
- `npm --prefix frontend run check`
- `cargo check --manifest-path desktop/src-tauri/Cargo.toml`
- `go vet ./...`

## Security
Claude credentials remain in the isolated WKWebView/Keychain; the Go sidecar receives only browser response JSON and stores no credentials.


mariusvniekerk:
Can you adjust this to be more general?  Tying this to the Tauri app instead of the generic webapp feels like the wrong approach.

pascalwhoop:
I can pull as much as possible out of rust and into go but the actual routing of all the `fetch` requests through the webkit was on purpose. Because openai/anthropic have extensive blocking of non web based interfaces. so staying in the webkit engine just removes a whole lot of spoofing complexity. 

What I can aim for is

- Svelte: control UI
- Go: Main lifecycle, pagination etc
- Rust: Just perform requests that it's being told to execute from Go

mariusvniekerk:
Yeah the idea is we should be able to relatively easily sub in electron / some other browser thing to make this behave.

PR #1392, opened 2026-08-13
Tombstone stored Claude sessions a complete full parse no longer emits

wesm:
## Outcome

- Claude transcripts with legacy fork rows sync by append again. They no longer re-parse the whole file on each change.
- Legacy fork rows that the current parser no longer produces become source-missing tombstones. Nothing is hard-deleted. A later parse that emits the ID revives the row.
- Full rebuilds also retire these rows, and they do not slow down. The rebuild reads the archive's legacy fork rows once, not once per file.
- Clients refresh when a sync only tombstones rows.

## Why

Older parser versions split Claude transcripts into fork-branch rows that share the parent's file path. One real archive holds about 13,900 such rows. Today's parser never emits them, so no parse can refresh or remove them. While they exist, the incremental path refuses the transcript, and the path's minimum data version stays stale.

## What changed

- After a complete Claude parse, the sync compares the parsed sessions with the stored stale fork rows for that file. Rows that the parse did not produce become source-missing tombstones. This uses the existing baseline gate, so rows without deletion authority stay until a later pass proves it.
- The same check runs for zero-result transcripts, S3 sources, and rebuilds. A rebuild snapshots the archive's stale fork rows before it starts. The archive takes no writes during a rebuild, so the snapshot is exact.
- Stale forks that the CWD filter rejects stay in place. An unchanged source that holds only such rows is fresh, so it does not re-parse on every pass. A primary row tombstoned as source-missing counts as absent for this check.
- Deletion baselines can now be exact per session and path. A failed or canceled pass revokes only the rejected entries.
- `SyncStats.Tombstoned` is exported. Sync entry points emit a sessions event when it is non-zero.

## Limits

- Only fork rows with a data version below the current one are candidates. Primary rows and current forks are not touched.
- The rowless freshness marker is keyed on the CWD filter. A wider filter always re-parses.
- `engine.go` grows by about 1,000 lines. Some large functions gained branches instead of being split. A follow-up can extract the exact-baseline bookkeeping.

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -754,14 +754,18 @@ func deferStartupMaintenance(skipInitialSync, workerSyncDone bool) bool {
 func statsFromWorkerResult(r workerResult) sync.SyncStats {
 	if r.Stats != nil {
 		stats := *r.Stats
+		if r.Tombstoned > stats.Tombstoned {
+			stats.Tombstoned = r.Tombstoned
+		}
 		stats.Aborted = !r.DiscoveryComplete
 		return stats
 	}
 	return sync.SyncStats{
-		Synced:  r.Synced,
-		Skipped: r.Skipped,
-		Failed:  r.Failed,
-		Aborted: !r.DiscoveryComplete,
+		Synced:     r.Synced,
+		Skipped:    r.Skipped,
+		Failed:     r.Failed,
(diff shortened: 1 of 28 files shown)


mariusvniekerk (approved):

wesm:
fixing the CI

wesm:
Skipping. merging

Issue #1352, opened 2026-08-06
Roadmap: Hosted sync: upload raw agent artifacts and derive sessions and embeddings server-side


salmonumbrella:
## Summary

Make hosted sync raw-first. A laptop should watch agent source data and transport authenticated incremental raw artifacts to the hosted server. The server should own durable raw storage, parsing, PostgreSQL projection, and embeddings.

The authoritative pipeline is:

`laptop watcher → authenticated incremental raw upload → server object storage → server parsing → PostgreSQL → server embeddings`

This issue is the umbrella for that end-to-end change. The original request for server-side embedding remains required, but it is now the final derived stage rather than a standalone change to `pg serve`.

We are starting work on this architecture.

## Delivery plan

This is the shipping sequence. Update each item with its PR link or links as the work opens. One item may take more than one PR, but every PR should stay inside one of these boundaries.

1. [x] **Raw custody core** — canonical manifests, tenant-scoped immutable object and manifest storage, exact verification, PostgreSQL acceptance metadata, source-head fencing, idempotency, and parse-job creation. [PR #1396](https://github.com/kenn-io/agentsview/pull/1396)
2. [x] **Device enrollment and authenticated raw transport** — revocable device credentials, scoped upload tokens, missing-object negotiation, resumable checksummed uploads, manifest-last commit, and durable receipts. [PR #1459](https://github.com/kenn-io/agentsview/pull/1459), [PR #1473](https://github.com/kenn-io/agentsview/pull/1473), [PR #1488](https://github.com/kenn-io/agentsview/pull/1488)
3. [x] **Laptop watcher and uploader** — provider discovery, append-only fast paths, immutable-file and SQLite capture, bounded spool and backpressure, local checkpoints, and reconciliation. [PR #1498](https://github.com/kenn-io/agentsview/pull/1498), [PR #1532](https://github.com/kenn-io/agentsview/pull/1532), [PR #1544](https://github.com/kenn-io/agentsview/pull/1544)
4. [ ] **Server parsing and PostgreSQL projection** — isolated manifest materialization, reused provider parsers, leased processing, fenced transactional projection, and explicit source provenance. [PR #1607](https://github.com/kenn-io/agentsview/pull/1607) merged the parse-worker foundation: verified manifest materialization, provider-parser reuse, and generation-fenced job leases. Transactional PostgreSQL session projection and runtime activation are next; this item remains open.
[...]

PR #1396, opened 2026-08-13
feat(rawsync): add hosted raw custody core

salmonumbrella:
## Summary

Adds the storage layer for raw-first hosted sync (#1352): the server keeps the
original agent session files, not just the parsed rows. With the source files
on hand, the server can re-parse after a parser fix or rebuild lost data.

Nothing calls this yet. There is no upload API, no server-side parsing, and no
change to local sync or `pg push`. Those come in later PRs.

## How it works

A client uploads content-addressed objects, then a **manifest** that lists one
source (for example one Codex session file), its files, and the SHA-256 and
length of every piece.

- `internal/rawsync/manifest.go` validates the manifest and produces one
  canonical JSON document. Its SHA-256 is the manifest ID. Tenant and device
  IDs are embedded, so it cannot be replayed under another identity. Unknown
  providers and `RemoteSyncExcluded` providers (Omnigent, Trae) are rejected.
- `internal/rawsync/object_store*.go` stores objects and manifests immutably
  in the existing artifact store under a per-tenant prefix. Same content
  again is a no-op; different content under the same digest is a conflict.
- `internal/postgres/raw_ingest_*.go` records verified objects, accepted
  manifests, each source's current head, and a parse job per manifest, all in
  one transaction. Same capture ID returns the same receipt. A manifest must
  name the current head as its parent (compare-and-swap), and every object it
  references must already be verified. Accepted rows are append-only.
- `internal/rawsync/service.go` is the single entry point for later upload
  handlers. It rejects excluded providers and oversized objects before
  accepting bytes.

## Notes for reviewers

- Source keys and paths can be 4096 bytes, so unique keys use a SHA-256 of the
  text (see `rawIngestDDL`).
- On the schema-current `pg push` fast path, a role without `CREATE` skips
  the raw custody DDL with a log line instead of failing the push.
- End-to-end test: `internal/postgres/raw_ingest_custody_pgtest_test.go`.

Refs #1352

### DIFF
--- a/internal/postgres/raw_ingest_custody_pgtest_test.go
+++ b/internal/postgres/raw_ingest_custody_pgtest_test.go
@@ -1,0 +1,272 @@
+//go:build pgtest
+
+package postgres
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.kenn.io/agentsview/internal/artifact"
+	"go.kenn.io/agentsview/internal/parser"
+	"go.kenn.io/agentsview/internal/rawsync"
+)
+
+func TestRawCustodyEndToEnd(t *testing.T) {
+	pgURL := testPGURL(t)
+	cleanSchemaTestPG(t, pgURL)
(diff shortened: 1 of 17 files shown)


wesm:
looking at this

wesm:
looking

PR #1415, opened 2026-08-14
perf(codex): persist parse checkpoints and bound import memory

dqtz5vpvj9-create:
Codex sync persists parse checkpoints so unchanged rollouts can skip transcript reads and eligible appends resume without reparsing their history. Late results and incremental signals update with the content. Full writes seed signal state in the same transaction, without a follow-up revision read or commit. More than eight pending calls can still append without a checkpoint; the checkpoint becomes available when the pending set shrinks.

Imports of at least 128 MiB stream result payloads through scratch SQLite storage. Single-event summaries reuse staging metadata, and publication honors cancellation and disabled signal recomputation. Messages and metadata remain in memory. Raw event identity preserves deduplication after sanitization; old events missing this metadata require a source reparse before late updates.

Append validation checks the trailing 128 KiB and trusts the earlier prefix until an audit. The main review boundaries are checkpoint validation, transactional state, and staging ownership. Full-corpus performance remains a merge gate.

### DIFF
--- a/cmd/agentsview/live_activity_test.go
+++ b/cmd/agentsview/live_activity_test.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/assert"
@@ -149,91 +150,99 @@ func TestLiveActivityIndexedLookupSchedulesRowsWithoutCompleteStat(t *testing.T)
 }
 
 func TestStartLiveActivityRunTracksSyncAndWaitsForStop(t *testing.T) {
-	ctx, cancel := context.WithCancel(t.Context())
-	defer cancel()
-	base := t.TempDir()
-	sessions := filepath.Join(base, "sessions")
-	history := filepath.Join(base, "history.jsonl")
-	rollout := filepath.Join(base, "rollout.jsonl")
-	id := "019f0000-0000-7000-8000-000000000002"
-	now := time.Now()
-	require.NoError(t, os.WriteFile(history, []byte(
-		`{"session_id":"`+id+`","ts":`+
-			strconv.FormatInt(now.Unix(), 10)+
-			`,"text":"private prompt sentinel"}`+"\n",
-	), 0o644))
-	require.NoError(t, os.WriteFile(rollout, []byte("changed"), 0o644))
-	provider, ok := parser.NewProvider(parser.AgentCodex, parser.ProviderConfig{
-		Roots: []string{sessions},
-	})
-	require.True(t, ok)
-	hints, ok, err := parser.ResolveActivityHintProvider(provider)
-	require.NoError(t, err)
-	require.True(t, ok)
-	sources, err := hints.ActivityHintSources(t.Context())
-	require.NoError(t, err)
-
-	idled := make(chan struct{}, 1)
-	idle := server.NewIdleTracker(20*time.Millisecond, func() {
-		idled <- struct{}{}
-	})
-	go idle.Run(ctx)
-
-	entered := make(chan struct{})
-	release := make(chan struct{})
-	finished := make(chan struct{})
-	trackedSync := trackLiveActivitySync(idle,
-		func(context.Context, []string) error {
-			close(entered)
-			<-release
-			close(finished)
-			return nil
+	synctest.Test(t, func(t *testing.T) {
+		ctx, cancel := context.WithCancel(t.Context())
+		defer cancel()
+		base := t.TempDir()
+		sessions := filepath.Join(base, "sessions")
+		history := filepath.Join(base, "history.jsonl")
+		rollout := filepath.Join(base, "rollout.jsonl")
+		id := "019f0000-0000-7000-8000-000000000002"
+		now := time.Now()
+		require.NoError(t, os.WriteFile(history, []byte(
+			`{"session_id":"`+id+`","ts":`+
+				strconv.FormatInt(now.Unix(), 10)+
+				`,"text":"private prompt sentinel"}`+"\n",
+		), 0o644))
+		require.NoError(t, os.WriteFile(rollout, []byte("changed"), 0o644))
+		provider, ok := parser.NewProvider(parser.AgentCodex, parser.ProviderConfig{
+			Roots: []string{sessions},
 		})
-	runCtx, runCancel := context.WithCancel(ctx)
-	poller := agentsync.NewLiveActivityPoller(
-		[]agentsync.LiveActivityTarget{{
-			Provider: provider,
-			Hints:    hints,
-			Sources:  sources,
-		}},
-		func(context.Context, string) (agentsync.LiveActivitySource, bool, error) {
-			return agentsync.LiveActivitySource{Path: rollout}, true, nil
-		},
-		trackedSync,
-		nil,
-	)
-	stop := startLiveActivityRun(runCtx, runCancel, poller)
+		require.True(t, ok)
+		hints, ok, err := parser.ResolveActivityHintProvider(provider)
+		require.NoError(t, err)
+		require.True(t, ok)
+		sources, err := hints.ActivityHintSources(t.Context())
+		require.NoError(t, err)
 
-	select {
-	case <-entered:
-	case <-time.After(time.Second):
-		require.FailNow(t, "tracked sync did not start")
-	}
-	select {
-	case <-idled:
-		require.FailNow(t, "idle callback fired while sync work was active")
-	case <-time.After(3 * 20 * time.Millisecond):
-	}
+		idled := make(chan struct{}, 1)
+		idle := server.NewIdleTracker(20*time.Millisecond, func() {
+			idled <- struct{}{}
+		})
 
-	stopped := make(chan struct{})
-	go func() {
-		stop()
-		close(stopped)
-	}()
-	select {
-	case <-stopped:
-		require.FailNow(t, "stop returned before active sync work completed")
-	case <-time.After(20 * time.Millisecond):
-	}
-	close(release)
-	select {
-	case <-finished:
-	case <-time.After(time.Second):
-		require.FailNow(t, "tracked sync did not finish")
-	}
-	select {
-	case <-stopped:
-	case <-time.After(time.Second):
-		require.FailNow(t, "stop did not join the poller goroutine")
-	}
+		entered := make(chan struct{})
+		release := make(chan struct{})
+		finished := make(chan struct{})
+		trackedSync := trackLiveActivitySync(idle,
+			func(context.Context, []string) error {
+				close(entered)
+				<-release
+				close(finished)
+				return nil
+			})
+		runCtx, runCancel := context.WithCancel(ctx)
+		poller := agentsync.NewLiveActivityPoller(
+			[]agentsync.LiveActivityTarget{{
+				Provider: provider,
+				Hints:    hints,
+				Sources:  sources,
+			}},
+			func(context.Context, string) (agentsync.LiveActivitySource, bool, error) {
+				return agentsync.LiveActivitySource{Path: rollout}, true, nil
+			},
+			trackedSync,
+			nil,
+		)
+		stop := startLiveActivityRun(runCtx, runCancel, poller)
+		defer func() {
+			close(release)
+			stop()
+		}()
+
+		select {
+		case <-entered:
+		case <-time.After(time.Second):
+			require.FailNow(t, "tracked sync did not start")
+		}
+		// Exercise idle suppression only after the sync has acquired its work
+		// lease; slow file reads during startup are not part of this contract.
+		go idle.Run(ctx)
+		select {
+		case <-idled:
+			require.FailNow(t, "idle callback fired while sync work was active")
+		case <-time.After(3 * 20 * time.Millisecond):
+		}
+
+		stopped := make(chan struct{})
+		go func() {
+			stop()
+			close(stopped)
+		}()
+		select {
+		case <-stopped:
+			require.FailNow(t, "stop returned before active sync work completed")
+		case <-time.After(20 * time.Millisecond):
+		}
+		release <- struct{}{}
+		select {
+		case <-finished:
+		case <-time.After(time.Second):
+			require.FailNow(t, "tracked sync did not finish")
+		}
+		select {
+		case <-stopped:
+		case <-time.After(time.Second):
+			require.FailNow(t, "stop did not join the poller goroutine")
+		}
+	})
 }
--- a/cmd/agentsview/sync_worker.go
+++ b/cmd/agentsview/sync_worker.go
@@ -196,6 +196,12 @@ func runSyncWorkerStartup(
 		var stats sync.SyncStats
 		var tombstoned int
 		var auditErr error
+		// The audit is also the periodic content-verification pass for
+		// checkpointed sources: bypass the stat-trust gate so the provider's
+		// full-source fingerprint detects and repairs same-stat in-place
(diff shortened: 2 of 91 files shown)


wesm:
Thank you for working on this, the Codex path has been in need of a revamp and I just haven't had the time. Will review and get this merged soon

wesm:
*Review by Claude (Claude Code) on behalf of @wesm.*

Thanks for the work on this. Parse output is byte-identical to `main` (verified on ~37k real Codex sessions: rows, tokens, and computed costs match exactly), and cold-import peak memory does drop (~1.0 GB to ~0.8 GB on a large real archive). But the PR isn't mergeable yet. Findings in priority order:

**Must fix**

- **The "late tool result delta" path never runs on real Codex logs.** Codex writes every tool output followed immediately by a `token_count` event with new usage. That usage belongs to the assistant message that was already committed, so the incremental parser sets `unattachedTokenUsage` and falls back to a full re-parse. Tested: appending just the output line, incremental works; appending output + `token_count` (the real shape), full parse, 0/47 sessions. So the headline claim ("late tool results no longer trigger a full reparse", "13.6 s to 22 ms") doesn't hold, and the agent-state table + incremental signal fold + `signal_maintain.go` are effectively unused. Either add a usage-attach delta so a `token_count` for a committed call can be applied incrementally, or drop that machinery.
- **Bootstrap loop:** a stored session that can't earn a checkpoint (file doesn't end in `\n`, or fork replay gate still active at EOF) is force full-parsed and rewritten on **every** sync (`checkpoint.go` returns Bootstrap, `codexForceFullParse` bypasses all skip gates). Reproduced with the built binary; `main` skips the same file. Bootstrap must be one-shot.
- **First sync after upgrade re-parses the entire Codex archive.** On a real archive: 21 minutes for 37.5k sessions, and every session's `local_modified_at`/`sync_marker` is bumped so the next PG/DuckDB push re-evaluates all of them. The design doc says the opposite ("previous O(file) reads still apply until the next full parse"). Make bootstrap lazy (on first change) or document the one-time cost.
- **`golangci-lint` fails (18 issues, `main` has 0):** 3 `unused` (`codexParentResolution`, `hasEquivalentToolResultEvent`, `codexStagingAvailableBytes`), 1 ineffassign, 1 staticcheck, 13 modernize.
- **Audit mode isn't scoped to Codex.** `SetCheckpointAudit(true)` in the daily audit worker disables the stat/verified/incremental gates for every provider, so Claude/Gemini/etc. get content-hashed daily and grown Claude files take a full parse.
- **Incremental signal fold disagrees with the full recompute** for `runaway_tool_loop_count` (`ExactHistorical` never OR'd into output; 12-call windows only latch when 12+ calls arrive in one delta; missing `TotalCalls >= 12` guard). Health grade drifts (D vs F in the repro). Moot while the first bullet holds, but wrong as written.

**Should fix**

- Staged (>128 MB) path parity: duplicate `call_id` in one transcript assigns all events/summary to the last call; summaries differ when an event sanitizes to empty; anonymous events lose `subagent_session_id`; staged publish always deletes+reinserts and bumps `transcript_revision` even when unchanged.
- Staging scratch defaults to `os.TempDir()` (tmpfs on many Linux distros, which defeats the memory goal), fixed 256 MiB free-space check regardless of file size, and stale scratch files are never cleaned up after a crash.
- Windows: full-parse checkpoints store ctime in NTFS ticks but the gate compares Unix ns, so the stat-only skip never fires there.
- Warm sweep is slower than `main` (12.2 s vs 8.3 s on 37k files) because the checkpoint gate does ~7 DB queries per file before the free stat gates. Cold import is ~15% slower because bulk passes are now byte-budgeted for every provider (files of 16 MiB or more parse serially). Both should be called out or reduced.
- Checkpoint is upserted twice per full parse (in-tx and again in the flush loop); `tool_calls` has no `(session_id, tool_use_id)` index for the delta probes; blocked-category late results store a different `content_length` than a full parse.

**Context for the PR description**

- `main` already skips unchanged Codex files on stat alone since #1393, so the "warm sweep 1.9 s to 1 ms" before/after compares against a pre-#1393 build. The real remaining wins are bounded cold-import memory and O(delta) appends without late outputs.

dqtz5vpvj9-create:
@wesm Thank your attention to this PR and for your awesome tool. I will review the problems later, and I'm very happy to keep contributing and discussing!

dqtz5vpvj9-create:
@wesm , we’ve finished addressing the latest review findings. When you have a chance, could you please take another look and help trigger the CI checks?

dqtz5vpvj9-create:
> ## roborev: Combined Review (`6ac44e2`)
> Code looks sound overall, but existing Codex archives require invalidation to apply the parser fix.
> 
> ## Medium
> * `internal/parser/codex_cursor.go:278` — Reused call IDs now attach outputs by occurrence, but the parser data version remains 88. Unchanged archived sessions may pass freshness checks and retain previously misassociated outputs. Bump the parser data version or add Codex-specific invalidation so existing sessions are reparsed.
> 
> _Reviewers: 2 done | Synthesis: codex, 10s | Total: 20m7s_

The `6ac44e2` finding on reused call ids and the parser data version: checked rather than bumped.

Where the two attachment rules actually differ: main resolved a `function_call_output` through `callRefs[callID]` (the most recent call with that id); this branch resolves it through the pending-call list, FIFO (the earliest still-unresolved call with that id). Sequential reuse (call, output, call, output) yields identical results under both rules. They diverge only when two calls sharing one `call_id` are pending at the same moment an output arrives.

Measured on a real archive (860 Codex rollouts, 22.2 GiB, 769,933 `function_call`/`custom_tool_call`/`local_shell_call` items and 770,127 outputs): zero files contain a reused `call_id` at all, and therefore zero contain the simultaneous-pending pattern. Codex does not currently produce the input this code path is defensive against, so no stored session carries outputs that a reparse under the new rule would attach differently. The occurrence-qualified path exists to keep the incremental/late-result writer exact if that ever changes.

Given that, a parser data-version bump would buy nothing and cost a full reparse of every Codex archive on upgrade -- on that same machine, re-reading ~23 GiB -- which is precisely the startup cost this PR removes and why checkpoint adoption is lazy. Leaving the data version as is. If a future Codex release starts reusing call ids, any affected file is reparsed authoritatively on its next real change, and `parse-diff` reports drift between stored rows and a fresh parse for files that never change again.

dqtz5vpvj9-create:
Hi @wesm , thanks again for reviewing this PR. I've been using a build based on these changes for the past few days, and it has been working well for me.
When you have a chance, would you be open to taking another look at the latest version? Thanks!

wesm:
Yes, i will take a closer look, I'm setting up some benchmarking infrastructure from agentsview to allow me to consider changes like this with more confidence, and I've been traveling. Please bear with me

wesm:
now that 0.42.0 has released I will spend some time on this soon

dqtz5vpvj9-create:
Thanks, much appreciated. We are still actively working on this branch as well: it is up to date with current main and merges cleanly, and we are running it in production against a large archive, so anything you find in review we can turn around quickly.

wesm:
running benchmarks on this now

wesm:
Rebasing this and rerunning benchmarks. Preliminary test showed a performance regression

dqtz5vpvj9-create:
@wesm Thanks for fixing the Claude incremental-append performance regression and rerunning the benchmarks!

wesm:
Pushed the fixes in [`5a76f5f1c`](https://github.com/kenn-io/agentsview/pull/1415/commits/5a76f5f1ca52526883993840afbbc6035c2b6a49).

The reported full-import slowdown remains a merge blocker until the benchmark passes on the updated head. The contributor's profile points to the ordinary write path: files below 128 MiB never use staging, so staging improvements cannot explain away that result.

The fixes address these reproduced problems:

- Full writes now commit incremental signal state with the content, removing a second transaction and revision read. Successful and failed output checks are reused. Raw-event hashing runs in parse workers with bounded buffers. Agent summary state is created only when a call receives a late result, and other providers do not build Codex signal state.
- Short sessions no longer acquire a false runaway-loop signal. More than eight pending calls no longer permanently disable appends; checkpoints recover when the pending set shrinks, and full parses retain their file hash even without a checkpoint. Reused call IDs again attach to the latest call, matching main.
- Late outputs preserve raw event identity and the full parser's whitespace and blocked-content rules. Target lookups use message positions rather than scanning every call in the session. Existing events without raw metadata require one source reparse before late updates can proceed.
- Staged imports honor cancellation and disabled signal recomputation. Deduplication uses an index, and a sole-event summary uses metadata captured during staging instead of reading the payload again. Multi-event summaries still reread payloads; that part of the review remains open.
- Signal recomputation retries later after repeated snapshot conflicts, even if no further source change arrives.

The full test suite, vet, CI lint checks, and focused race tests pass. SQLite commit-hook tests confirm one content-and-seed commit, with both rolled back if seeding fails.

Performance is still unresolved. On a synthetic 384-file, 49,152-call full resync (all files below the staging threshold), five runs per build gave medians of 14.12 s for main and 14.73 s for this candidate (+4.3%). Ranges were 11.95–17.44 s and 12.12–16.54 s respectively. This timing variation prevents a reliable acceptance claim. A separate CPU profile measured less writer CPU (8.16 s → 7.78 s), and result-event bytes and omitted-summary bytes stayed identical. The actual-corpus Watchdog result must be rerun on the new head.

My broader assessment: this remains a large change combining checkpoints, late-result updates, incremental signals, parser refactoring, and scratch staging. The highest review risk is where these paths meet. A smaller series would be easier to review and roll back. Passing correctness checks is necessary, but the full-corpus wall-time comparison must also pass before merge.

wesm:
I'm still seeing an 8% performance regression on full syncs on my corpus after my fix (down from a 20% regression), so continuing to work on this

dqtz5vpvj9-create:
I reproduced the remaining full-sync regression locally and profiled where it goes, in case it saves you a round.

**Setup.** Isolated clone of a real local Codex corpus: 859 session files, 2.7 GB, every file under 128 MiB so nothing takes the staged path. Fresh data dir per run, `AGENTSVIEW_NO_DAEMON=1`, `sync --full` with the hidden `--cpuprofile`/`--memprofile`/`--trace` flags, plus `strace -c` for commit syscalls. `main` at 618f0fab versus this branch at 8552bf7, alternated for three rounds on an otherwise idle machine.

| | main | PR |
|---|---:|---:|
| wall, median of 3 | 107.6 s | 116.2 s (+8.0%) |
| user CPU | ~120 s | ~108 s |
| sys CPU | ~16.5 s | ~13.2 s |
| archive | 1033 MB | 1067 MB (+3.3%) |

The PR does less work overall: 10% less CPU and about 8.5 GB less allocation, almost all of it from dropping the fork-parent pre-scan, and `tool_result_events` is byte-identical at 849 MB on both sides, so the duplicate-storage fix holds. The extra 34 MB is `session_signal_state` (20 MB) and the new `idx_tool_calls_session_tool_use` (5.6 MB).

**Where the wall time goes.** The pipeline is bounded by the single collector goroutine: `writeBatchWithOutcomeContext` accounts for two thirds of wall time on both builds (63 s of CPU on main, 67 s on the PR), and the parse workers spend the difference blocked on the retention semaphore (512 goroutine-seconds on main, 693 on the PR in the runtime trace). Anything added inside a session's write path therefore lands one-for-one on wall time. Three things did:

1. A second commit per session. After `ReplaceSessionContentWithCheckpoint` commits, `seedSignalStateFromFull` opens its own writer transaction in `UpsertSessionSignalStateIfRevision`; the bulk path does the same per session after `WriteSessionBatchContext`. That transaction is cheap on its own (0.3 s of syscalls) but it re-dirties the `sessions` page and appends WAL frames after every content commit. `strace` shows 186k more `pwrite64` calls, 5–8% more `fsync` (autocheckpoints), and each fsync 18% slower (31.9 → 37.7 ms) because each checkpoint copies more pages. Roughly +7 s on the writer.
2. Writer CPU is up 2.5–4 s: `computeSignalsFromToolRows` + `SeedIncrementalState` about 2.5–2.9 s even after the reuse fix, and the per-session `TranscriptRevision` read about 1.3–1.5 s, immediately after the content transaction bumped that same revision.
3. New rows inside the content transaction: `tool_call_occurrence_agent_state` (one per call, 57k rows here), the checkpoint tables, and maintenance of the new index. Small in bytes, but all on the commit path.

Those three add up to the 8–10 s difference. I also ruled out the bulk admission change (the byte-cap flush only applies to daemon passes after the rebase), SQLite busy-handler sleeps (`nanosleep` counts are within 3%), and the per-byte SHA-256 (it replaces the separate fingerprint pass main already ran, net zero).

**Suggestion.** Fold the signal-state seed into the content transaction so a session commits once, and drop the `TranscriptRevision` re-read since the transaction already knows the new value; that should recover most of item 1 and all of the revision read. Deferring `idx_tool_calls_session_tool_use` until the end of a bulk import would take a smaller slice off item 3. Neither changes the checkpoint or revision contracts.

Happy to share the raw profiles if useful.

wesm:
Thanks for the investigation, I'll push fixes here shortly

wesm:
The Windows failure came from the live-activity test: its 20 ms idle timer could expire before startup finished reading the hint file. A 40 ms startup delay reproduced the same failure locally. The test now uses a controlled clock and starts idle checking after the sync acquires its work lease. It passed 50 repetitions and 20 repetitions of the focused race tests; removing work tracking or the shutdown join still makes it fail. Vet and commit lint checks passed. Windows CI must confirm the new commit, `d5dbba6d1`.

I could not reproduce the reported TraeX link bug through the provider API. Late wait outputs return `IncrementalNeedsFullParse` with `ForceReplace`, and the resulting full parse returns the correct `traex:` child link. The guard in `codexIncrementalNeedsFullParse` also rejects late subagent notifications and spawn results before the incremental relabel callback. The callback omission exists, but the reported bad-link path is not reachable under these guards, so I left production parsing unchanged.

The full-import performance gate from the earlier comment remains open.

wesm:
Confirmed the signal-invalidation finding and fixed it in `6eccf35c9`. Replacing staged content while signal recomputation is disabled now clears old secret findings, resets secret-scan freshness, and marks quality signals stale in the same transaction. An unchanged staged publish preserves its valid findings. The regression test failed before the fix and passes afterward.

The ordinal-mapping finding did not reproduce. `callKeyByPosition` is used during parsing, before `Finalize`. Publication rebuilds final coordinates from the normalized message model using the call ID and occurrence number; findings use that final mapping too. A real import with a claimed notification slot, an orphan notification, and three occurrences of the same call ID preserved all four events, their expected final ordinals, and their summaries on both collecting and staged paths. No mapping change was needed.

The staged-write race tests, vet, and commit lint checks passed. The earlier full-import performance gate remains open.

wesm:
Validated both findings and pushed 1f0b6414e.

The SQLite batch issue is valid. With the connection's parameter limit set to 999, a 350-event update failed. Backfilling summary state for 350 existing agents failed the same way. Both paths now split inserts into at most 166 rows (996 parameters), preserving event retention and first/latest summary ordering. These failures were reproduced through the database API; this does not establish their frequency in real session files.

The collector leak finding is not reachable with the current providers. Only Codex creates staged imports, and Codex does not supply the source-CWD state required to enter those error branches. A diagnostic using a real staged Codex parse and a rejecting directory filter confirmed that the reachable discard path removes scratch files and restores the GC setting. No collector change was needed.

The regression tests, focused database/sync race tests, go vet, and golangci-lint passed. The full-sync performance gate remains open.

wesm:
Benchmarking this again

wesm:
Benchmarks showed 3% higher sync time which is probably not statistically significant. I'm running it one more time to see and then will merge this

PR #1425, opened 2026-08-16
perf(signals): index duplicate prompt comparisons

naveenspark:
Duplicate-prompt scoring rebuilt a token set for every comparison against prior substantive prompts. On long sessions, signal recomputation therefore spent most of its time repeating token hashing and garbage collection even when the prompts were distinct.

This change preserves the existing exact and fuzzy scoring semantics while indexing accepted prompts by normalized text and token frequency. Each prompt now evaluates only prior representatives that share tokens. A deterministic reference test compares the indexed implementation with the original pairwise algorithm, and a large-session benchmark is added to the existing PR performance gate. On the same 800-prompt shared-vocabulary fixture, current main takes about 1.42 s and allocates 2.14 GB per operation; this branch takes about 13–19 ms and allocates 16.9 MB.

The postings index grows with retained prompt tokens. A pathological session where most distinct prompts share most of their vocabulary can still require many candidate comparisons, but the implementation no longer rebuilds or scans complete token structures for every prompt pair. The main review point is equivalence with the existing asymmetric score, where current tokens are treated as a set and previous tokens remain a multiset.


wesm:
LGTM

PR #1428, opened 2026-08-16
fix(db): ignore self-referential subagent edges

### LINKED ISSUE #1250: LinkSubagentSessions: provenance for conflicting fork/copy spawn edges (follow-up to #1247)
## Summary

`LinkSubagentSessions()` resolves a subagent's `parent_session_id` from the
authoritative `tool_calls.subagent_session_id` edge. When a child session is
referenced by spawn edges from **more than one distinct session** — reachable
via **copied/forked history** (e.g. after `CopyOrphanedDataFromExcluding`) — the
"real" parent is ambiguous, and `tool_calls` alone doesn't carry enough
provenance to disambiguate.

PR #1247 made the common case correct and safe (unambiguous edges resolve to the
one spawner; ambiguous edges fall back to preserving the current parent via
`COALESCE(<grouped subquery>, parent_session_id)` instead of an arbitrary
`LIMIT 1`). This issue tracks the remaining, deeper fork-reconciliation gaps that
are out of scope for that depth-2 parenting fix.

## Known gaps

1. **Ingestion-order dependence** (flagged by roborev on `c544f11`).
   If a forked/copied edge is stored *first* (initially the only edge), linking
   provisionally re-parents the child to it. When the real edge arrives later and
   the child becomes ambiguous, `COALESCE(..., parent_session_id)` permanently
   retains the (possibly wrong) provisional parent. Normal fork ordering is safe
   because a fork's source is parsed first, but reconciliation/copy paths don't
   guarantee that ordering.

2. **No self-edge guard.**
   Nothing prevents `tool_calls.subagent_session_id == session_id` (only
   reachable from a corrupt/crafted transcript). It would yield a self-parent.
   Consequences are bounded — every parent-chain consumer uses `UNION` recursive
   CTEs that dedupe and terminate on cycles, so no loop/crash; the row just drops
   from the sidebar. A one-line `AND tc.session_id <> sessions.id` on the
   subqueries + `EXISTS` closes it (and also lets a self-edge + a real edge
   resolve to the real spawner).

## Possible approaches

- Resolve ambiguous edges via **stable provenance** — e.g. the earliest-started
  spawner (`ORDER BY started_at`), which is stateless and self-correcting
  regardless of ingestion order.
- Or retain enough **source-parent state** to reverse a provisional linkage once
  candidates become ambiguous.
- Add the self-edge guard.

## Suggested regression tests

- Call `LinkSubagentSessions` **before and after** inserting the conflicting edge
  and assert the child converges to the correct spawner (roborev's suggestion).
- A self-edge case asserting no self-parent.

## Context

Split out of #1247 (nested subagent depth ≥ 2 re-parenting). The core fix there
is a strict improvement over the previous behavior; this is the follow-up design
work the maintainer may want to weigh in on.

rodboev:
SQLite subagent linking now treats a self-referential spawn edge (`tool_calls.session_id = tool_calls.subagent_session_id`, only reachable from a corrupt or crafted transcript) as non-evidence everywhere spawn edges are enumerated: parent resolution in `subagentSpawnerExpr`, the driver sets of the global and scoped linking statements, the surviving-edge check in `clearDanglingSubagentParentQuery`, and `SubagentChildSessionIDs`. A child with a self edge and a real edge resolves to the real spawner; a child with only a self edge is left alone, so a parser-derived parent and relationship type survive.

`SanitizeSession` rejects a `ParentSessionID` equal to the session's own id (and a self `ParserParentSessionID`), falling back to the parser-derived parent when it names another session and to no parent otherwise, so no write path — session batch, `UpsertSession`, or artifact import — can store a self-parent.

Rows that an earlier build linked to themselves are repaired the same way — `NULLIF(parser_parent_session_id, id)`, since the linker never writes that column — once per archive by a marker-gated pass at the start of `LinkSubagentSessions` (`parent_session_id IS id` cannot use `idx_sessions_parent`, so the scan is not repeated every sync). The archive-rebuild orphan copy applies the same clear to the rows it copies, since the fresh archive's one-time pass usually runs before orphans arrive.

Queue semantics, artifact-import behavior, and `CopySyncStateFrom` are unchanged from `main`.

Reviewers should look at `subagentSpawnerExpr` and `repairLegacySelfParentedSessions` in `internal/db/sessions.go`, `SanitizeSession` in `internal/db/validate.go`, and `clearCopiedSelfParents` in `internal/db/orphaned.go`.

Closes #1250

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -22,8 +22,9 @@ description: Release history for AgentsView
   `scan_protected_paths` to opt back in to Git remote, worktree, and branch
   detail there.
 - Ignore self-referential subagent spawn edges when linking session
-  hierarchy, drop a self-parent claim at ingest, and clear rows an earlier
-  build had linked to themselves once per archive.
+  hierarchy, and resolve a self-parent claim to the parser-derived parent (or
+  no parent) at ingest and once per archive for rows an earlier build had
+  linked to themselves.
 
 ---
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -21,6 +21,9 @@ description: Release history for AgentsView
   Sessions in those folders now keep path-only project identity; set
   `scan_protected_paths` to opt back in to Git remote, worktree, and branch
   detail there.
+- Ignore self-referential subagent spawn edges when linking session
+  hierarchy, drop a self-parent claim at ingest, and clear rows an earlier
+  build had linked to themselves once per archive.
 
 ---
 
--- a/internal/db/artifact_import_session.go
+++ b/internal/db/artifact_import_session.go
@@ -75,7 +75,12 @@ func (db *DB) applyArtifactImportedSession(
 	write = sanitizeSessionBatchWrite(write)
 
 	db.mu.Lock()
-	defer db.mu.Unlock()
+	locked := true
+	defer func() {
+		if locked {
+			db.mu.Unlock()
+		}
+	}()
 	tx, err := db.getWriter().BeginTx(ctx, nil)
 	if err != nil {
 		return result, fmt.Errorf("beginning artifact imported session: %w", err)
@@ -100,6 +105,14 @@ func (db *DB) applyArtifactImportedSession(
 		if err := tx.Commit(); err != nil {
 			return result, fmt.Errorf(
 				"committing suppressed artifact imported session: %w", err,
+			)
+		}
+		locked = false
+		db.mu.Unlock()
+		if err := db.RepairQueuedSubagentParents(); err != nil {
+			return result, fmt.Errorf(
+				"repairing queued subagent parents after artifact import: %w",
+				err,
 			)
 		}
 		result.Suppressed = true
@@ -136,6 +149,13 @@ func (db *DB) applyArtifactImportedSession(
 			fmt.Errorf("committing artifact imported session: %w", err)
 	}
 	pendingRecallRevocations.flush()
+	locked = false
+	db.mu.Unlock()
+	if err := db.RepairQueuedSubagentParents(); err != nil {
+		return result, fmt.Errorf(
+			"repairing queued subagent parents after artifact import: %w", err,
+		)
+	}
 	return result, nil
 }
 
--- a/internal/db/artifact_import_session.go
+++ b/internal/db/artifact_import_session.go
@@ -75,12 +75,7 @@ func (db *DB) applyArtifactImportedSession(
 	write = sanitizeSessionBatchWrite(write)
 
 	db.mu.Lock()
-	locked := true
-	defer func() {
-		if locked {
-			db.mu.Unlock()
-		}
-	}()
+	defer db.mu.Unlock()
 	tx, err := db.getWriter().BeginTx(ctx, nil)
 	if err != nil {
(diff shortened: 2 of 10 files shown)


wesm:
Thanks for this. The core change in `3306d8c9e` is right and is exactly what #1250 asked for: the `tc.session_id IS NOT s.id` guard in the spawner resolution, the driver `IN` sets, `clearDanglingSubagentParentQuery`, and `SubagentChildSessionIDs`. The thirteen follow-up commits added through the roborev cycles are where the problems are, so I am going to trim the branch back down myself and push to this PR rather than round-trip again. Here is what is wrong with the current state, most severe first:

**1. Pre-upgrade cleanup intent is discarded (behavior regression).** `upgradeSubagentParentRepairStateTx` moves every `subagent_parent_cleanup_queue` row into the relink-only queue. A legitimately queued former child whose spawner was parser-excluded or deleted therefore keeps a dangling `parent_session_id` forever. The PR's own test shows it: `TestRepairQueuedSubagentParentsMigratesLegacyCleanupIntent` was renamed to `...DemotesLegacyCleanupIntent` and inverted so `kid` now keeps parent `"spawner"`, a row that no longer exists. It also makes the cleanup insert in `migrateLegacySubagentParentRepairQueueTx` dead work and its comment ("carries cleanup intent") false. This trades a rare-but-real dangling parent for protection against a self-edge session whose parser parent has not been ingested yet, which needs a crafted transcript to reach.

**2. Artifact import consumes queue rows it did not create, and now runs the linker.** `removeSubagentParentRepairQueueForSessionsTx` deletes cleanup and repair rows for the imported id. Artifact sync does not hold `syncMu`, while the local sync's capture -> queue cleanup -> delete/rewrite -> deferred repair sequence in `internal/sync/engine.go` assumes the queue row survives that window. An import of the same session landing inside the window evaluates cleanup early (parent still present, no-op), deletes the row, and the deferred repair finds nothing. Separately, `repairSubagentParentsForSessionsTx` runs the full scoped linker for the imported session, which changes artifact-import semantics in a way unrelated to self-edges.

**3. The added `(subagentSpawnerExpr) IS NOT NULL` guards are dead.** They were added in `c565e205d` when the driver-set self-edge guard was removed, and `ff3a59829` restored the driver guard without removing them. `tool_calls.session_id` is `NOT NULL` and the driver `IN` uses the same predicate, so the spawner subquery is never NULL for a candidate row. Reverting both `WHERE` clauses to the original two-branch form passes the entire `internal/db` suite, including the continuation/fork preservation cases. As written they cost up to two extra correlated subquery evaluations per candidate row in the statement whose comments are about edge-scaled cost.

**4. There is a much smaller design.** Every write path funnels through `ValidateAndSanitize` (session batch, `UpsertSession`, and artifact import via `sanitizeSessionBatchWrite`). Nulling `ParentSessionID` when it equals `ID` there, plus the linker guard, means a self-parent can never be stored again, which removes the need for `queueSelfParentRepairTx`, the artifact-import drain, and the copied-orphan queueing. What remains for legacy rows is a marker-gated one-time `UPDATE ... WHERE parent_session_id IS id`.

I will keep the first commit's guards, add the sanitize guard, keep a one-time marker-gated repair for legacy rows, restore main's cleanup-queue and artifact-import behavior, and adjust the tests to match.

Issue #1418, opened 2026-08-14
sync_include_cwd_prefixes silently drops every session of agents that don't record cwd (Cursor, Kimi) — Kimi's cwd is in the source but never extracted


Helix-lyh:
### Summary

`sync_include_cwd_prefixes` silently drops **every** session from agents that do not populate the `sessions.cwd` column — including Cursor (the source file has no cwd field) and Kimi (the source *has* cwd, but the parser never extracts it). A whole primary agent can silently vanish from the archive with no indication that anything was skipped.

### agentsview version

v0.40.1

### Install method

Homebrew cask

### OS / platform

macOS 25.5.0 (darwin)

### Repro

config.toml:

```toml
sync_include_cwd_prefixes = ["/Users/helix/Code"]
```

1. **Cursor** — sessions under `~/.cursor/projects/<project>/...`. The jsonl lines contain only `role`/`message` keys; no cwd. The project path *is* encoded in the enclosing folder name (`Users-helix-Code-work-area`), but it is not recorded as cwd. → 0 Cursor sessions ingested.
2. **Kimi** — sessions under `~/.kimi-code/sessions/wd_<project>_<id>/.../agents/<id>/wire.jsonl` contain records like `{"type":"config.update","cwd":"/Users/helix/Code/mcp-hub", ...}`, yet the `sessions.cwd` column is empty for **all** Kimi sessions in the archive (verified directly in sqlite). → 0 Kimi sessions ingested, even though every one of them is ~/Code work.

### Expected behavior

- Kimi parser: extract cwd from `config.update` records (and/or the `wd_<project>` session-dir prefix).
- Cursor: derive cwd from the project folder name when no in-file cwd exists.
- At minimum: surface a count / log line of sessions skipped by the prefix filter, so a silently-emptied agent is not mistaken for a sync failure.

### Related

#942 (worktree mappings don't apply to empty-cwd sessions), #1248 (recovering OpenCode cwd).

PR #1432, opened 2026-08-16
fix(parser): resolve Cursor workspace cwd from the filesystem

rodboev:
This makes the Cursor parser resolve workspace identity through one filesystem-backed authority, with passive discovery and explicit resume using separate policies.

Remote imports never search the local filesystem, discovery reuses a resolution only within one operation, and generic sync refreshes stored Cwd when the resolved workspace changes while preserving it when passive probing has no authority.

Refs #1418


wesm (approved):

wesm:
Stopping here, will merge once CI is green

wesm:
Residual risk accepted

Issue #1333, opened 2026-08-02
Session upload derives session identity from the uploaded filename, letting a basename collision silently replace an archived session's messages


lcatlett:
**Observed on:** `agentsview v0.39.0` (commit `627a8afa`, built 2026-07-27), daemon on `127.0.0.1:8080`.

---

## Summary

`POST /api/v1/sessions/upload` takes the session's identity from the **uploaded
filename**, ignoring the `sessionId` recorded inside the transcript. Because the
resulting basename is the archive's primary key, uploading an unrelated file whose
basename collides with an existing session **replaces that session's messages**, with
an `HTTP 200` and no warning.

In one measured case a single POST reduced an archived session from 96 messages to 24.

## Reproduction

A 326-line Claude Code transcript was copied and its internal session UUID rewritten to
a fresh value, so no pre-existing record was involved. The same content was then
uploaded under different filenames.

| Step | Uploaded basename | `sessionId` inside file | Response `session_id` | Response `messages` | Archived `message_count` for `344e9899…` |
| --- | --- | --- | --- | --- | --- |
| 1 | `344e9899….jsonl` (326 lines) | `344e9899…` | `344e9899…` | 96 | 96 |
| 2 | `344e9899….jsonl` (first 100 lines) | `344e9899…` | `344e9899…` | 24 | **24 — 72 messages gone** |
| 3 | `344e9899….jsonl` (full again) | `344e9899…` | `344e9899…` | 96 | 96 |
| 4 | `renamed-truncated-copy.jsonl` (100 lines) | `344e9899…` | **`renamed-truncated-copy`** | 24 | 96 (untouched) |
| 5 | `zzz-unrelated-name.jsonl` (326 lines) | `344e9899…` | **`zzz-unrelated-name`** | 96 | 96 (untouched) |

Steps 4 and 5 isolate the variable: identical bytes produced three distinct archive
records, each keyed to its filename. Step 2 is the destructive case.

Requests were sent with `-H "Origin: http://127.0.0.1:8080"`, which is required — the
example curl in `docs/usage.md:1176-1179` returns `403` as written when remote access
is disabled. That is a separate, smaller documentation issue.

## Why this is a problem rather than a quirk

The failure needs no malice and no unusual input:

- **Recovered or archived transcripts frequently have altered basenames** —
  `session-copy-2.jsonl`, `foo.jsonl.bak.jsonl`, timestamped variants. Uploading those
  as-is mints archive records under identifiers that match nothing, dedupe against
[...]

PR #1435, opened 2026-08-16
fix(parser): use transcript identity for rooted uploads

rodboev:
Renamed Claude root uploads can collide with an existing archive because the multipart filename supplies session identity. Uploads whose transcript carries one stable archive-safe ID with explicit root markers now select that ID during the existing Claude parse, and the staged file follows it. Sidechain, markerless, mixed-ID, and local discovery parses keep their current identity rules.

Replacement semantics, existing filename-keyed rows, case-equivalent names, and concurrent-writer handling are unchanged.

Refs #1333

### DIFF
--- a/internal/parser/claude.go
+++ b/internal/parser/claude.go
@@ -64,12 +64,9 @@ type claudeQueuedCommand struct {
 // and also returns session IDs intentionally excluded from the
 // archive, such as content-free /usage probes. Sync uses those IDs
 // during full resync so orphan preservation does not restore rows the
-// current parser deliberately dropped. This is the provider-owned
-// parse body shared by the Claude provider (both its discovered-session
-// Parse path and its ParseUploadedTranscript entry) and the Cowork
-// parser (which reuses the Claude transcript format); it carries no
-// legacy entrypoint naming so the provider can call it without shimming
-// a Parse* free function.
+// current parser deliberately dropped. This is the zero-option parse body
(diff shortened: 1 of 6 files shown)


rodboev:
The underlying fix for #1333 genuinely crosses three authorities:

- internal/parser owns transcript identity and lineage.
- Upload handling owns the destination path, conflict response, and replacement policy.
- internal/db owns race-free persistence.

After the repeated bounces trying to work all this into one PR I reworked it to just the identity-collision authority. This establishes the canonical identity input:

- Parser selects a trustworthy explicit-root ID.
- Upload handling uses that ID as the destination.
- Existing DB persistence is reused.

This is a better design and turned out smaller. Once this foundation merges, the next PR would cover the destructive behavior described in the original issue without reopening identity or filesystem design. Once both merge, the reported data-loss bug should be fixed such that we can close out the original issue. There are other possible follow-ups in:

- existing archive audit/repair (mentioned in the original issue)
- ambiguous identity/namespace policy (surfaced during review of the larger fix I tried at first)

But their value needs to be evaluated once the two-part fix here is in, since there may not be a need for them.

rodboev:
The second of this two-part fix is tracked as PR #1443. If that merges, this can be closed; if fixes are needed here, that one gets rebased. The pair closes #1333.

wesm:
rebasing

wesm:
superseded by #1443, now merged

PR #1450, opened 2026-08-18
fix(activity): restore Grok agent-minutes

cpcloud:
Restores Grok activity timing by enriching chat-history messages from the timestamped authoritative update stream. Matching follows Grok turn boundaries, keeps parallel tool-call starts separate, uses populated tool identifiers as authoritative, and falls back to exact content only when identifiers are unavailable.

Grok tool starts and terminal updates now use the existing tool-result event model. Activity treats terminal completions as sparse timeline anchors: they extend activity after the final transcript message and reset the gap cap before later messages without adding hidden or visible transcript rows, including for backend tools. Same-session overlaps are unioned, and post-range completions remain eligible successors so report-boundary clipping stays correct.

The report keeps transcript pairing on its indexed query path and merges sparse completion candidates afterward, avoiding full-transcript scans for ordinary sessions. The parser data version is bumped to reparse existing archives, and the public Grok session-format evidence is documented. Regression coverage uses only synthetic, provider-shaped fixtures.


cpcloud:
Merging on green.

PR #1462, opened 2026-08-19
feat: add daemon-backed terminal interface

randomradio:
Adds `agentsview tui`, a full-screen Bubble Tea interface backed by the existing daemon HTTP API and SSE stream rather than direct archive access. It covers session browsing and transcripts, dashboards, usage, activity, trends, insights, pins, trash, recent edits, settings, mutations, exports and publishing, persistent filters, localized labels, and terminal-safe rendering. Shared report services also expand the MCP surface for analytics, activity, trends, pins, insights, and recent edits.

Startup serves the client before the initial archive sync completes. Report panels stream independently, session details load concurrently, and transcript and filter-editor work stays bounded by the viewport and render caches. Concurrent sync is normal active status, empty trends remain an input state, and sidebar filters open an explicit keyboard editor with cursor and paste handling.

The implementation deliberately reuses daemon and service contracts instead of adding a second SQLite path. Fixed keybindings and colon commands remain the terminal-specific tradeoff; sidebar text filters are free-form rather than metadata pickers. The main review surfaces are `internal/tui`, `cmd/agentsview/tui.go`, the shared report and message-window changes, and `docs/tui.md`.


mariusvniekerk:
I've tested this out locally on my large agentsview corpus.  As it stands the performance of this TUI is not of a sufficient bar to be included in agentsview.

PR #1469, opened 2026-08-19
feat(sync): generalize S3 session sources beyond Claude and Codex

### LINKED ISSUE #1447: S3 session sources: generalize beyond Claude and Codex
S3 session roots work for Claude and Codex but nobody else. Adding Cursor (or any other single-file provider) means touching ~5 files and adding another case to 9 switch statements that already branch on agent type. That doesn't scale.

Today the S3 path is hardcoded: `discoverClaudeS3` / `discoverCodexS3` in the parser, agent-type switches in `processS3Session`, `s3DiscoveredSessionID`, `statS3SourceObject`, `isS3AgentRootSegment`, and the engine dispatch. Each new agent repeats the same boilerplate.

The building blocks for a generic approach are already there. `s3PrefixScan` is reusable, `parseMaterializedS3Source` is agent-agnostic, and most single-file providers (Cursor, Copilot, Gemini, Hermes, Codex forks, ...) would need zero sidecar handling.

Proposal:

1. Add an optional S3 discovery interface to providers (scanner config, session ID derivation, stat, post-fetch hydrate). Provide a default implementation that covers the common single-file case.
2. Replace the per-agent switches in `s3.go` / `s3_source.go` / `engine.go` with interface dispatch.
3. Claude and Codex keep custom implementations behind the interface (sidecars, fork reconciliation, index resolution).
4. Any `FileBased` provider with a single-file format can opt in by setting a capability flag, no new S3 code needed per agent.

~30 of the ~60 providers have single-file formats and could work with S3 out of the box. SQLite-backed and multi-file providers (Windsurf, Forge, RooCode, etc.) stay filesystem-only.

This would also unblock Cursor S3 support without adding yet another one-off scanner.

Happy to put up a PR for this.

durandom:
## Summary

Implements the approach outlined in #1447: replaces the hardcoded Claude/Codex
S3 switches with an `S3Provider` interface so any single-file provider can opt
into S3 discovery by setting a capability flag.

- Adds `S3Provider` interface (scanner config, session ID, temp path, stat,
  post-fetch hydrate) with a `DefaultS3Provider` struct that covers the common
  single-file layout.
- Claude and Codex implement the interface in `claude_s3.go` and `codex_s3.go`
  so sidecar, fork, parent, and index behavior stay with those agents.
  `s3source.go` is generic transport plus `s3PrefixScan`.
- Provider discovery uses that same scanner path. There is no separate
  `discoverClaudeS3` / `discoverCodexS3` wrapper.
- Cursor uses a dedicated scanner in `cursor_s3.go`: keep the documented
  harvest layout and local `agent-transcripts` layouts, decode encoded project
  dirs only for the local layout, reject other objects, and prefer `.jsonl`
  then nested over flat for the same stem.
- Adds `docs/agents/s3-providers.md` for enabling S3 on later agents.
- Adds `SourceCapabilities.S3Discovery` flag and migration validation that
  enforces the interface contract at startup.
- Replaces per-agent switches in `s3.go`, `s3_source.go`, and `engine.go` with
  interface dispatch.
- Enables Cursor S3 ingestion as the first consumer of `DefaultS3Provider`
  (`s3://.../<machine>/raw/cursor/<project>/<uuid>.jsonl`).
- Caches `S3ProviderFor` lookups to avoid per-object provider allocations.

This has been running in a deployed container ingesting Cursor sessions from S3
without issues.

Reviewers should start at `internal/parser/s3_provider.go`, then
`claude_s3.go` / `codex_s3.go` / `cursor_s3.go`, then the remaining dispatch
points in `internal/sync/s3.go` and `internal/sync/s3_source.go`.

Closes #1447


durandom:
The same-stem overwrite is real. Cursor S3 discovery was keeping every `.jsonl`/`.txt` object, so `proj/abc.jsonl` and `proj/abc.txt` shared `cursor:abc`.

Landed in `cursor_s3.go`: keep the documented harvest layout (`<project>/<id>.{jsonl,txt}`) and the local `agent-transcripts` layouts, reject other objects, and prefer `.jsonl` per project+stem.

We did not require the local Cursor directory tree for every object. The harvest layout in `docs/configuration.md` is `s3://.../<machine>/raw/cursor/<project>/<uuid>.jsonl`, and that is what the deployed ingest uses.

durandom:
Both findings were real.

1. Mirrored `agent-transcripts` roots now run `DecodeCursorProjectDir` on the encoded project directory. The documented harvest layout (`<project>/<id>.jsonl`) still keeps the literal project name, so `my-cool-project` is not rewritten.

2. Same-stem selection now matches local Cursor order: `.jsonl` over `.txt`, then nested `<id>/<id>.ext` over flat `<id>.ext`, then lexical path. Grouping stays on the encoded directory so two dirs that decode to the same name do not collapse.

Landed in `cursor_s3.go`.

durandom:
👏  

thanks for considering this pr

PR #1482, opened 2026-08-21
fix(quality): show refresh progress for filter changes

mariusvniekerk:
Quality now shows the same refresh feedback as the other analytics pages when filters trigger a background query. The page keeps existing results visible while the shared refresh control spins and an animated progress bar marks the content busy.

Successful Quality reads also update the refresh timestamp. The main review area is `QualityPage.svelte`, where the page now follows the shared refresh scheduler and query-state pattern.

### DIFF
--- a/frontend/src/lib/components/quality/QualityPage.svelte
+++ b/frontend/src/lib/components/quality/QualityPage.svelte
@@ -613,7 +613,7 @@
     </div>
 
     <RefreshControl
-      lastUpdatedAt={analytics.lastUpdatedAt}
+      lastUpdatedAt={analytics.qualityLastUpdatedAt}
       busy={querying}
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
The `fbb9016` refresh-timer finding is invalid. Quality now uses the shared `RefreshControl`, whose app wrapper forwards `onRefresh` to the pinned kit-ui control. That control creates a scheduler with `DEFAULT_REFRESH_INTERVAL_MS`, calls `scheduleNext()` on mount, and calls `scheduler.stop()` during teardown. Adding another scheduler to `QualityPage` would create a second timer and duplicate refresh requests.

Finding: https://github.com/kenn-io/agentsview/pull/1482#issuecomment-5376494522

PR #1488, opened 2026-08-22
feat(rawsync): add resumable object uploads

salmonumbrella:
Adds authenticated resumable raw-object uploads to hosted PostgreSQL raw sync. Scoped upload tokens can start or resume an upload, read its exact offset, and append bounded chunks. Finalization independently verifies length and SHA-256 before tenant-scoped immutable custody; retries are idempotent, and offset or checksum conflicts return the authoritative resume offset.

Upload sessions are durable in PostgreSQL and staged in a private restart-safe spool with exact-offset fencing, crash-tail truncation, and bounded cleanup of expired or orphaned state. The runtime privilege probe now includes the upload table's exact DML requirements.

Raw custody remains lazily opened under its dedicated `raw-sync` repository, preserving ordinary `pg serve` startup and repository ownership behavior from #1473. Device enrollment UX, laptop watcher behavior, quotas, S3 hosting, and downstream processing remain outside this slice.

Refs #1352

### DIFF
--- a/cmd/agentsview/pg_raw_sync.go
+++ b/cmd/agentsview/pg_raw_sync.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"io"
 	"path/filepath"
 	"sync"
 	"time"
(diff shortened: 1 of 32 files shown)


salmonumbrella:
@wesm ready for you or perhaps... @mariusvniekerk

wesm:
looking

PR #1490, opened 2026-08-23
fix(sync): advance reconciliation past deferred sources

### LINKED ISSUE #1476: DataVersionNeedsRetry blocks reconciliation of all later sessions
### agentsview version

v0.41.1 (commit a902515a)

### Install method

install.sh / install.ps1

### OS / platform

Debian GNU/Linux 12 (bookworm), Linux x86_64

### Which agent and version

Codex CLI 0.118.0 (blocking source); Pi (starved source)

### Which model(s)

Gemini 3.7 Flash (Pi target); bug is model-independent

## What happened

A Codex fork whose parent transcript is missing returns `DataVersionNeedsRetry`. The retry is counted in `pageStats.Failed`, so `reconcileWatchRootsStreamedLocked` aborts the page before updating the cursor. The same first page is retried indefinitely, preventing later pages and providers from being reconciled.

Observed on v0.41.1:

- The deferred source was item 9 on the first 256-item page.
- The spool contained approximately 70,000 candidates.
- The same failure was logged 1,936 times after daemon startup.
- New Pi sessions remained unindexed.
- Direct path sync of a valid Pi transcript succeeded immediately.

```text
watcher callback: watch root reconciliation failed processing page: 1 failures
```

## Expected behavior

A retryable source should remain deferred and receive no freshness or deletion proof, while healthy sources commit, the cursor advances, and later pages and providers are processed. The pass may remain incomplete, and the deferred source should retry with backoff.

Current `main` still appears to contain the same abort-before-cursor-update behavior.

```text
{"type":"session_meta","payload":{"id":"<child-id>","forked_from_id":"<missing-parent-id>"}}
```

### Steps to reproduce

1. Configure Codex and Pi roots and index more than 256 candidates.
2. Include a Codex fork on page 1 whose `forked_from_id` parent transcript is absent.
3. Start daemon/watch-root reconciliation.
4. Add a valid Pi transcript sorted after page 1.
5. Observe the repeated page failure and missing Pi session.
6. Run `agentsview session sync /path/to/pi-session.jsonl`; observe that the Pi session indexes successfully.

### Checklist

rodboev:
Sync runs can commit healthy streamed results while deferring a Codex-format fork whose parent is unavailable, but the deferred state was private to watcher paths. Other completion gates could acknowledge the pass and retire remote journals, and TraeX's shared Codex parser still reported the same retryable condition as a hard failure.

This exposes deferred work on `SyncStats` and centralizes processing completion in the sync package. Changed-path, watcher, startup, worker, CLI and HTTP remote coordination, mirror push, SSH import, and remote import/rebuild consumers use that contract before acknowledging the pass or retiring durable retry state. Deferred runs also leave the finished-state timestamp unchanged. Codex and TraeX share the existing Codex-format classifier; unrelated providers keep their current failure semantics. Typed retry paths and roots outrank the origin's full-sync flag, preserving lost-event recovery and falling back to full recovery only when no bounded scope exists.

Healthy same-page and later-page sessions continue to commit, deferred sessions remain stale without freshness or deletion proof, and watcher retry/backoff remains bounded. The reproduction came from @mulyoved's issue report.

Closes #1476


wesm:
looking

PR #1491, opened 2026-08-23
feat(search): add optional Chinese FTS tokenizer

dqtz5vpvj9-create:
Adds optional Chinese word and single-character search to SQLite through the pinned simple/cppjieba sidecar. HTTP, CLI, and MCP message searches segment CJK queries, preserve explicit quoted expressions, and retain English stemming for ASCII-only queries. PostgreSQL/CockroachDB and DuckDB retain their existing matching behavior and now prepare raw queries at the store boundary to preserve literal embedded quotes.

The parallel Chinese index is derived data. Its first backfill, a changed library or dictionary fingerprint, or pending writes requires a full rebuild before writable startup completes; the session-only freshness ledger cannot repair old token entries individually. Startup waits and stale-index fallback now emit logs. TEMP writer triggers maintain the index without requiring older writers to load the extension.

Removing the sidecar drops the optional index while retaining the small freshness ledger and persistent session triggers; reinstalling rebuilds the index and clears pending rows. Session-result snippets highlight segmented matches, while highlighting inside an opened transcript still uses the original query. Configuration documentation describes these limits, and the sidecar workflow covers database, service, and MCP tests.

### DIFF
--- a/.github/workflows/chinese-fts.yml
+++ b/.github/workflows/chinese-fts.yml
@@ -1,0 +1,54 @@
+name: Chinese FTS
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/chinese-fts.yml"
+      - "Makefile"
+      - "internal/db/**"
+      - "internal/service/**"
+      - "internal/mcp/**"
+      - "scripts/build-simple-fts.sh"
+  push:
+    branches: [main]
+    paths:
+      - ".github/workflows/chinese-fts.yml"
+      - "Makefile"
+      - "internal/db/**"
+      - "internal/service/**"
+      - "internal/mcp/**"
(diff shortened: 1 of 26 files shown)


dqtz5vpvj9-create:
Addressed the archive-reopen issue in `9085f2ef`: `reopenLocked` now uses `sqliteArchiveDriverName`, with behavioral coverage that reopens the full DB, writes a new Chinese message, and retrieves it through the Chinese FTS path.

I also added a regression test for the sidecar-removal concern. It creates the simple-tokenized table through the extension-enabled driver, closes those connections, then opens the file through plain `sqlite3` and successfully drops `messages_chinese_fts` without loading the extension. This confirms the documented cleanup path does not instantiate the unavailable tokenizer.

Issue #1451, opened 2026-08-18
Optimize PostgreSQL usage queries with normalized facts


wesm:
SQLite usage reads are moving to a normalized facts cache so aggregation happens over narrow indexed rows and only grouped results cross the driver boundary. PostgreSQL currently retains its live usage queries. This preserves observable behavior but leaves a documented performance divergence.

Choose and implement a PostgreSQL-native optimization after the SQLite facts shape is stable. A direct server-side aggregate, materialized view, or facts table are all viable; do not assume the SQLite sibling-cache lifecycle belongs in PostgreSQL.

Acceptance criteria:

- preserve exact snapshot-survivor selection and per-survivor microdollar rounding;
- keep model-rate resolution authoritative in Go unless an equally maintainable shared representation replaces it;
- keep all five public usage methods behaviorally aligned;
- pass the literal SQLite-facts versus PostgreSQL-live parity fixture;
- add PostgreSQL-specific read and refresh benchmarks before changing the query shape.

PR #1493, opened 2026-08-23
test(postgres): add usage performance benchmarks

rodboev:
PostgreSQL still serves usage analytics from live message and usage-event queries, while SQLite has a normalized facts path. The repository has no PostgreSQL-specific benchmark for the five public usage methods or their refresh cost, so the query-shape choice in issue #1451 lacks a reproducible baseline.

This adds an opt-in PostgreSQL 16 benchmark substrate covering varied read windows, all five usage methods, breakdowns, complete SQLite/PostgreSQL result checks before timing, an initial full push into an initialized empty schema, a fixed-cardinality one-message delta, catalog probes, eligible usage input counts, token-usage bytes, allocations, and the existing parity fixture. It records the constraints for a later optimization while leaving production queries, schemas, pricing, and public results unchanged. The benchmark remains outside the normal CI bench gate because it requires Docker.

Refs #1451

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -359,7 +359,7 @@ postgres-up:
 	docker compose -f docker-compose.test.yml up -d --wait
 
 # Run opt-in PostgreSQL usage benchmarks against the pinned PG16 service.
-bench-pg-usage:
+bench-pg-usage: pricing-snapshot
 	docker compose -f docker-compose.test.yml up -d --wait postgres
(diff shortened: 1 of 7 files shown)


wesm:
I'm working on some more systematic benchmarking of the postgres support, so this will help, but I may make some PRs related to this in the near future

Issue #1375, opened 2026-08-10
Feature request: official compaction/compression story for large sessions.db (tool results dominate)


dqtz5vpvj9-create:
### Problem / motivation

`sessions.db` grows linearly with tool-result content and there is currently no non-destructive official way to reclaim space:

- On a normal Codex-heavy machine: 888 sessions / 534k messages / ~396k tool calls produce a ~7.0 GiB `sessions.db`.
- The largest consumers are `tool_calls.result_content` (~3.07 GB) and `tool_result_events.content` (~3.04 GB). These store the same output twice for the same call.
- Bash output alone is ~2.4 GB; `view_image` results add ~0.58 GB of base64 image payloads.
- `result_content_blocked_categories` only controls future ingestion; it does not shrink an existing archive.
- `agentsview prune` deletes the original JSONL session files as well as DB rows, and it does not `VACUUM` the SQLite file, so the on-disk DB may stay large even after pruning. (There is an internal `db.Vacuum()` helper but no exposed command.)
- SQLite files do not shrink by themselves after DELETEs.

### Proposed solution

A first-class storage-maintenance story that does not require deleting source transcripts:

1. `agentsview db compact` (or similar) that VACUUMs / rebuilds the archive and reports before/after sizes.
2. A non-destructive rebuild path that re-parses source JSONL and applies `result_content_blocked_categories` to existing sessions (`agentsview sync --full` already rebuilds; it would be helpful for that path to be the documented way to reclaim space).
3. Consider storing tool results once instead of duplicating the same content in `tool_calls.result_content` and `tool_result_events.content`.
4. Optionally support transparent compression for the large text columns (for example zstd), or at least leave a documented extension point for ZIPVFS/sqlite-zstd-style VFS/row compression.

### Alternatives considered

- `result_content_blocked_categories = ["Bash"]` plus `agentsview sync --full`: works for new/rebuildable data, but requires understanding that a full resync is the only retroactive path.
- `agentsview prune --before ...` + manual `VACUUM`: works, but deletes source JSONL files and is not discoverable as a size-reclaim flow.
- Using ZIPVFS (commercial) or sqlite-zstd externally: not currently integrated and would need official support to be safe with the daemon/writer model.

wesm:
Agreed that there is need for some work here. My sessions.db is 34GB so I have this problem too

PR #1495, opened 2026-08-23
feat(db): add staged SQLite archive compaction

dqtz5vpvj9-create:
Adds a staged, crash-safe compaction workflow for the local SQLite archive. `agentsview db compact` estimates reclaimable space (`--dry-run`), builds the compacted archive with `VACUUM INTO` in a caller-selected staging directory, verifies it (integrity, schema, row counts, SHA-256), and installs it with an atomic rename. An optional original backup (`--keep-backup`) and a recovery manifest cover interrupted replacements. A writable daemon serves the same operation via `POST /api/v1/data/compact` (localhost only), sharing the archive maintenance barrier with sync and resync. A production archive went from 11.85 GiB to 7.06 GiB with this workflow, reclaiming 4.79 GiB; that is a free-page/WAL result specific to that archive.

The command does not compress or deduplicate live tool-result payloads and does not change future growth from them; transparent payload compression remains a separate follow-up to #1375. This partially addresses #1375.

## Lifecycle and safety model

The lifecycle is built around an explicit commit point with two durable manifest phases:

- **prepared** — the verified candidate and backup are durable and writes are barred, so recovery may roll the replacement forward or back without losing anything. Recovery also accepts a content-verified archive whose hash moved only because the WAL-mode reopen rewrote the journal header.
- **committed** — written after post-install verification succeeds. From then on the installed archive is authoritative: recovery only cleans up staged artifacts and never restores the backup, so writes that land after a compaction can never be discarded by a later startup.

During the build, the archive stays readable while writes fail fast with the same retryable maintenance error a worker pass uses (`ErrWriterClosed`, mapped to 503 + Retry-After); connections pause only for the final swap. Compactions are serialized per archive (a concurrent attempt returns 409), and the barrier is enforced in the writer handle itself, so writer paths that do not take the database mutex are barred too, from the start of the build until the committed record is durable. Verification and commit run on an operation-owned context, so a client disconnect after the swap cannot roll back a completed compaction. The candidate is converted to WAL mode before hashing so its bytes stay stable across the install, and rollback prefers keeping or renaming an intact original over a second full-archive copy.

Manifest recovery runs only for the canonical archive under the write-owner lock (not in the generic `db.Open`), so derived databases such as a resync temp in the same directory never interpret the archive's manifest, and a relocated data directory recovers instead of refusing to start. Startup also sweeps staging directories abandoned before their manifest became durable. Failure paths restart the WAL checkpoint loop and the usage-cache backfill.

## Where risk concentrates

`internal/db/compact.go` carries the state machine: `Compact` (barrier, build, swap, commit), the swap failure paths, and `RecoverCompactManifest` with its prepared/committed recovery rights. The regression tests in `internal/db/compact_test.go` encode the guarantees: committed recovery preserves post-commit rows, hash-moved-but-verified archives are kept, cancel-after-install still succeeds, writes are barred mid-build and restored after, backfill restarts, recall FTS survives compaction, and derived opens ignore the manifest.

## Scope

Excludes zstd row compression, payload deduplication, automatic VACUUM after sync, and compact semantics for PostgreSQL or DuckDB mirrors.


wesm:
looking at this

wesm:
Hi would it be ok if I take over from here? Please don't merge anymore

dqtz5vpvj9-create:
> Hi would it be ok if I take over from here? Please don't merge anymore

Awesome, thanks! Glad to see it merged!

PR #1496, opened 2026-08-23
feat(parser): add t3code session support

arandomhooman:
# feat(parser): add t3code session support

t3code (`pingdotgg/t3code`, MIT) is an event-sourced desktop agent. It keeps
every thread in one SQLite database under its userdata root rather than a
session tree, so this registers it as a multi-session container provider on the
existing shared base: discovery surfaces `state.sqlite` once and Parse fans it
out into one session per thread, addressed by `<db>#<threadID>`. Change
detection is therefore per thread, so one live thread's writes never invalidate
its neighbours.

The append-only event log is folded into `projection_*` read models, and those
projections are the only tables a reader can interpret without replaying the
log. The visible conversation comes from `projection_threads` joined to
`projection_thread_messages`, with `projection_projects` supplying the workspace
root and `projection_thread_sessions` the backing provider. Threads are
soft-deleted via `deleted_at` and are skipped.

**`RemoteSyncExcluded` is the load-bearing part of this change.** t3code stores
its transcripts and its credentials in the same directory: `clerk-tokens.json`,
`cloud-auth-token.json`, and a `secrets/` directory of encrypted blobs are
siblings of `state.sqlite`, and the database itself carries `auth_sessions` and
`auth_pairing_links`. No part of the root can be copied off the machine safely,
so the agent opts out of every remote sync artifact.

## Tradeoffs and limits

Change detection is per thread and uses two signals, like Shelley. The mtime
token is the latest of the thread's, its messages', and its project's
created/updated timestamps -- message `updated_at` so an in-place edit advances
it, the project stamps because `workspace_root` determines the session's cwd
and project. A digest over every parser-observed thread, project, and message
field is stored in `file_hash` and compared under `UnchangedResultMTimeAndHash`,
because t3 is event-sourced: refolding the event log after an app update
rewrites projection rows whose timestamps derive from the events and do not
move, and the accepted legacy generations lack message and project `updated_at`
entirely -- the digest is the only signal that sees either. Discovery and parse
compute both signals from the same fields in the same order, so a
fingerprint-triggered reparse is never discarded as unchanged. The digest also
gates freshness in both directions: `FingerprintHashRequiredForFreshness` keeps
a timestamp-blind rewrite (size and mtime still matching the stored row) from
being skipped pre-parse, and t3 joins Hermes in
`providerFingerprintHashEstablishesFreshness` because members inherit the
shared container's stat -- another thread's growth moves every member's size,
and the matching digest is what keeps reconciliation from reparsing the whole
archive after every write anywhere in the database. The live watcher's token
folds the digest into the timestamp's unused sub-millisecond bits, the same
way Shelley's does, so a watched session also reacts to timestamp-preserving
rewrites instead of computing the digest and discarding it.
Session activity ends at the latest of the thread's and its messages' stamps,
so a message recorded past the thread's `updated_at` -- the defensive case the
token covers -- also moves `EndedAt` rather than leaving ordering and date
windows stale. Project stamps stay out of `EndedAt`: a workspace rename
advances the change token but is not conversation activity. Tombstone passes
resolve membership for all stored threads with one query via the
batch-presence seam. The digest is computed from the rows the parser
reads anyway; hashing the multi-megabyte `state.sqlite` file itself on every
fingerprint would add nothing on top of it, so `CompositeFingerprint` stays
declared unsupported rather than declared and not delivered.

A database event no longer fans out the whole container. The multi-session
base gains an opt-in `WithChangedContainerWatermarkListing` option: when the
caller allows watermark-merged sources and supplies stored freshness -- the
same generic `StoredMemberFreshnessPager` seam the OpenCode format consumes --
the base emits only the members whose watermark advanced, reusing the existing
paging cursor. t3 answers it with a session-row watermark listing -- one
indexed thread row plus its project row per live thread, never the messages
table -- so a no-op WAL touch emits nothing and one thread's write emits one
member. A message stamped without the thread row moving is deferred the same
way timestamp-blind rewrites are: the full meta scan catches it at scheduled
reconciliation. The persisted fingerprint embeds the session-row watermark
behind a versioned prefix (decoded on the stored side the same way the
OpenCode child digest's metadata watermark is), so the merge compares row
watermark against row watermark -- a stored file_mtime dominated by a late
message cannot mask a later thread or project update. Hashes from before the
format simply fall back to file_mtime and re-emit once. Without stored authority, or on a listing failure, the container
answer is preserved unchanged, and no other agent's behavior moves unless it
opts in. Like the OpenCode watermark listing, the merge cannot see a
timestamp-blind rewrite; the digest catches those during scheduled
reconciliation and full parses. Zed, Shelley, Trae, and Kiro still fan out
their containers; the new option is available to them but adopting it is their
own change.

t3code migrates its projection columns in place: `model_selection_json`
replaced a bare `model` column, and `attachments_json` and
`provider_instance_id` were added after the projections were introduced.
Optional columns are probed with `PRAGMA table_info`, so an older database
still parses with its model read from the legacy column. Required columns are
validated and a database without them fails closed.

Content coverage is deliberately narrow. Thinking, tool calls, tool results,
and per-message token usage are not projected into the message table; they live
in `projection_thread_activities` and the event log. Attachments render as
`[Image: name]` placeholder lines -- the shape the Codebuff and Claude-web
parsers use -- appended after any text the message carries, so an image-only
turn is not blank and a text-plus-image turn does not silently lose the
attachment. The capability declaration
says so rather than silently returning empty. t3code delegates execution to
another agent, so usage stays in that provider's own transcripts, which
agentsview already indexes separately.

`zedDBCompositeMtime` and `shelleyDBCompositeMtime` were byte-identical. They
are replaced by one `sqliteContainerCompositeMtime` rather than adding a third
copy for t3.

## Where to look

`internal/parser/t3.go` holds the reader and the schema probe;
`internal/parser/t3_provider.go` is the options wiring onto the multi-session
container base. The `internal/sync/engine.go` touchpoints follow the Shelley,
Zed, and Hermes patterns: deleted-container recognition, container composite
mtime, skip-cache exemption, the hash-establishes-freshness listing, and the
watcher's per-member source mtime, which routes to the digest-bearing token
before the generic provider-fingerprint path because t3 is not file-based.
`docs/internal/session-format-sources.md` records the format with pinned
upstream evidence.


mariusvniekerk:
So why do we actually need this, since the underlying sessions that T three code makes are already parsed correctly by agents for you. What does this give us?  3.3K lines is quite a lot for functionality that basically already exists?  @arandomhooman

arandomhooman:
ur right with the defualt case, where a t3 instance with no `homePath` leaves `CLAUDE_CONFIG_DIR` alone and its sessions land in `~.claude` for agentsview to parse but instances that do set `homePath` write to a private home no agentsview coverage, so those threads go unindexed unless the user adds the path by hand, and the database is the only place thread titles, `workspace_root` project grouping, and soft-delete state exist. Tests are 1,523 of the 3,242 lines and the parser is 1,396.

 I can split out the generic base option if that helps.

mariusvniekerk:
yeah i think making it easier to confugure alternate `CODEX_HOME`/ `CLAUDE_HOME` is a better path here., losing the nice session names is unfortunate, but t3code *SHOULD* be using the appropriate `/name` methods from codex / claude natively for this ideally which we do pick up.

wesm:
The PR description is also a wall of text. Ask you agent to rewrite it in plain language and be clear and concise, leading with outcomes

PR #1498, opened 2026-08-24
feat(rawsync): add raw upload transport client and checkpoint state

salmonumbrella:
## Summary

Adds the laptop-side half of the hosted raw-ingest transport, as one feature
commit on top of the merged server surface (#1459, #1473, #1488).

`internal/rawclient` is the HTTP client for the authenticated raw-sync
surface:

- scoped-token exchange with TTL-aware caching and single-flight refresh,
  plus one forced re-exchange retry on unauthorized responses;
- missing-object negotiation (`POST /objects/missing`);
- resumable chunked uploads (`POST`/`PATCH` on `/uploads`) with
  authoritative offset recovery on `upload_offset_conflict`, terminal
  `checksum_mismatch`, and a final empty PATCH when a resumed session
  already sits at the full offset but was not finalized (server crash
  window between append commit and finalization);
- manifest-last commit returning the durable receipt, with typed
  `head_conflict` surfacing the current receipt and generation.

`internal/rawcheckpoint` is the small dedicated local SQLite database for
transport state: one-row device identity and per-source head
receipts/generations keyed by `(provider, configured_root_id, source_key)`,
using the same dual mattn/modernc driver split as `internal/vector`. Head
advancement refuses anything without a durable receipt, so the checkpoint
can only trail server acknowledgment. The device credential is injected
through `Config` and never persisted or logged.

Tenant identity is never sent by the client; it stays derived from the
authenticated server context (spec invariants 3 and 4).

Proof comes from three layers: wire-contract unit tests against scripted
handlers, a full-cycle end-to-end test against the real huma server stack
with real rawsync services and in-memory stores, and a `pgtest` end-to-end
against the durable PostgreSQL stores that also covers interrupted-upload
resume (durable offset row plus spool prefix observed before and after a
fresh client re-uploads) and the stale-parent head conflict.

Known limits, deliberately deferred to the capture-agent slice: negotiation
calls are unbatched against the server's 1 MiB control-plane cap (roughly
9-10k object refs per call against the 16384-object manifest ceiling), and
there is no status scope because no server route exists yet.

Reviewers should start at `internal/rawclient/upload.go` (offset recovery
and finalization), `internal/rawclient/tokens.go` (refresh discipline), and
`internal/rawcheckpoint/store.go` (receipt-gated advancement).

### DIFF
--- a/internal/postgres/raw_client_e2e_pgtest_test.go
+++ b/internal/postgres/raw_client_e2e_pgtest_test.go
@@ -1,0 +1,527 @@
+//go:build pgtest
+
+// Package postgres_test carries the raw-client end-to-end proof as an
+// external test package: internal/server imports internal/postgres, so an
+// internal test file cannot import the server package without an import
+// cycle. Everything the exercise needs from the store layer is exported;
+// only the tiny pgtest conveniences (schema wipe, digest helper, table
+// counts, spool paths) are restated here.
+package postgres_test
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"database/sql"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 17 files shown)


salmonumbrella:
@mariusvniekerk read for you

wesm:
looking

PR #1504, opened 2026-08-24
feat(cli): query retained soft-deleted sessions (#1471)

### LINKED ISSUE #1471: Allow querying soft-deleted sessions via CLI and/or optional config
### Problem / motivation

Harnesses like Claude Code regularly prune or clean up older transcript files on disk. When agentsview sync reconciles provider roots and finds a file missing, it sets deletion_cause = 'source_missing' and populates deleted_at.

While AgentsView's local-first storage design intentionally preserves all messages, tool calls, and tokens in sessions.db, all standard CLI commands (session search, session list, session get, session messages) filter out records where deleted_at IS NOT NULL. This makes historical conversations invisible (from CLI and Web) even though the data remains intact in the SQLite archive.

### Proposed solution

1. CLI Flag (--include-deleted):
    Add --include-deleted to agentsview session search and agentsview session list so CLI queries can optionally include tombstoned/soft-deleted sessions.
2. Config Entry (preserve_missing_sources):
    Add an optional setting in config.toml:
    ```toml
      preserve_missing_sources = true
    ```
    When enabled, reconciliation leaves missing files active instead of tagging them with source_missing tombstones.

### Alternatives considered

_No response_

rodboev:
Agentsview's CLI list and content-search commands now accept an explicit
`--include-deleted` option for retained soft-deleted sessions. The option works
through direct and daemon-backed requests and includes both root and child
sessions.

Semantic and hybrid search reuse compatible vectors created before a session
was soft-deleted. Vector builds filter already-deleted transcripts before any
encoding call, so deletion never causes a new embedding request. If a
compatible vector does not already exist, semantic and hybrid search omit that
deleted session even when `--include-deleted` is set.

Default queries still hide soft-deleted sessions. Permanently excluded sessions
remain hidden, hard deletion still removes vectors, and unrelated sidebar,
usage, analytics, and synchronization paths keep their existing policy.

This change adds no table, corpus version, vector recreation, or migration.

Closes #1471

### DIFF
--- a/cmd/agentsview/session_list.go
+++ b/cmd/agentsview/session_list.go
@@ -27,6 +27,7 @@ func newSessionListCommand() *cobra.Command {
 		minUserMessages                         int
 		includeOneShot                          bool
 		includeAutomated, includeChildren       bool
+		includeDeleted                          bool
 		outcome, healthGrade                    string
 		minToolFailures                         int
 		hasSecret                               bool
@@ -69,6 +70,7 @@ func newSessionListCommand() *cobra.Command {
 				IncludeOneShot:   includeOneShot,
 				IncludeAutomated: includeAutomated,
 				IncludeChildren:  includeChildren,
+				IncludeDeleted:   includeDeleted,
 				Outcome:          outcome,
 				HealthGrade:      healthGrade,
 				HasSecret:        hasSecret,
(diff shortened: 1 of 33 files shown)


mariusvniekerk:
Closing as obsolete. #1505 fixed #1471's underlying issue by keeping source-missing sessions browsable without treating them as user-deleted. This PR now only adds opt-in queries over deliberately deleted sessions, which is a separate feature and is not needed for #1471.

PR #1505, opened 2026-08-24
fix(sync): keep sessions browsable when sources disappear (#1471)

rodboev:
Missing source files no longer hide or remove archived sessions. The session,
messages, and usage remain available. Sync records source absence separately and
reparses the source if it returns.

PR #1154 represented a missing source by setting `deleted_at` with
`deletion_cause = 'source_missing'`. Sync could distinguish that state from user
deletion, but ordinary reads only check `deleted_at`. They therefore hid intact
history when a source file disappeared.

The first version of this PR added `preserve_missing_sources` to avoid writing
that state. That was the inverse of the behavior we need: it kept sessions
visible by preventing sync from remembering that their source was gone.

The corrected model keeps the two facts independent:

- `deleted_at` records direct user intent. It controls trash, restore, and
  permanent removal.
- `source_missing_at` records local source availability. It disables freshness
  and incremental shortcuts until a successful reparse clears it.

A remaining provider path still treated a missing one-file-per-session source
as a parser exclusion and hard-deleted the archived row. Proven file absence now
uses `source_missing_at` there too. A present source that the parser deliberately
excludes still follows the existing parser-exclusion behavior.

The marker stays set until the returned session row, transcript, usage, and data
version have been written successfully. A failed refresh therefore keeps the
old archive visible and eligible for a later retry. This preserves the existing
eventual-consistency contract.

Existing SQLite rows with `deletion_cause = 'source_missing'` are migrated out
of `deleted_at`, making them visible again. PostgreSQL rows written with the old
state are repaired. The new source-availability marker stays SQLite-local
because it describes the filesystem seen by the local sync process.

The config setting is removed because preserving archived data is unconditional.

Windows SQLite test jobs now retain structured Go test events on failure and
show active, slowest, and last completed tests in the job summary. This does not
raise the timeout; it makes future overruns diagnosable. Fork pull requests run
the workflow pinned to `main`, so these diagnostics start applying after merge.

Refs #1471


mariusvniekerk:
@wesm This PR now separates source absence from user deletion instead of adding a config escape hatch. Since #1154 introduced the source-missing deletion model, could you review the corrected state model, especially the `source_missing_at` migration and the freshness/reparse boundaries? Missing-source sessions remain browsable; only direct user intent sets `deleted_at`.

mariusvniekerk:
On the atomicity finding: the multi-step write ordering predates this pull request. The contract here is eventual consistency. A session whose local source disappears remains browsable from archived data. If the source reappears, sync reparses it. The source-missing marker remains set until the session row, transcript, usage, and data-version writes succeed. A failed or interrupted refresh therefore remains eligible for retry and does not discard the retained transcript. There can be a temporary metadata/transcript mismatch during a failed refresh, but the next successful sync repairs it. Widening the existing transaction boundary is not needed for this pull request.

PR #1508, opened 2026-08-24
fix(remotesync): curate editor remote targets

### LINKED ISSUE #1492: Narrow remote sync targets for Cursor, VS Code Copilot and Zed
## Problem

HTTP remote sync recursively mirrors every regular file beneath most configured provider roots. For Cursor, VS Code Copilot and Zed, those roots contain much more than session data.

On one machine, `~/.agentsview/remote-mirrors` reached 11 GB:

```text
7.4G  Library/Application Support/Code/User
5.5G  Code/User/workspaceStorage
1.4G  Code/User/globalStorage
3.0G  Library/Application Support/Zed
1.6G  Zed/external_agents/registry
```

The mirror included unrelated extension data, external-agent packages, language servers, `node_modules`, Cursor canvases, MCP metadata and files named `mcp_auth.json`.

These files consume disk, increase initial transfer and manifest-walk costs, and may copy application state including secrets that the session parsers never read.

## Current behaviour

For ordinary providers, `ResolveTargets` adds the configured root to `TargetSet.Dirs`. `BuildManifest` and `WriteArchive` then recursively walk that directory.

Relevant code:

- [`internal/remotesync/resolve.go`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/remotesync/resolve.go#L15-L109)
- [`internal/remotesync/manifest.go`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/remotesync/manifest.go#L77-L107)
- [`internal/remotesync/archive.go`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/remotesync/archive.go#L42-L77)

The original [incremental HTTP remote sync design](https://github.com/kenn-io/agentsview/blob/b820558/docs/superpowers/specs/2026-07-07-incremental-remote-sync-design.md) chose a complete mirror so parsers could read sibling files without adding provider file layouts to the transport protocol.

The collector still needs every raw file a parser may read. It does not need unrelated files beneath a broad application-data root.

## Existing precedent

`TargetSet.Files` already supports curated provider exports.

RooCode exports only discovered task metadata and message files. Its resolver explicitly excludes `mcp_settings.json`, which may contain MCP environment variables, API keys and authentication headers, along with caches and checkpoints.

Kilo Legacy similarly exports only:

```text
task_metadata.json
ui_messages.json
api_conversation_history.json
```

Windsurf creates a sanitized export of selected state-database content. Poolside narrows its application-data root to `trajectories/`.

Relevant code:

- [`resolveRooCodeTarget`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/remotesync/resolve.go#L341-L379)
- [`resolveKiloLegacyTarget`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/remotesync/resolve.go#L381-L423)
- [`resolvePoolsideTarget`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/remotesync/resolve.go#L425-L446)

## Proposed change

Resolve Cursor, VS Code Copilot and Zed to the smallest complete set of raw parser inputs, including required companion files.

### Cursor

Use a verbatim file-scoped target containing the transcript layouts already recognised by the parser:

```text
<project>/agent-transcripts/<session>.jsonl
<project>/agent-transcripts/<session>.txt
<project>/agent-transcripts/<session>/<session>.jsonl
<project>/agent-transcripts/<session>/<session>.txt
```

See [`cursor_provider.go`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/parser/cursor_provider.go#L378-L487).

This should exclude canvases, `node_modules`, MCP state and unrelated project metadata.

### VS Code Copilot

Export only:

```text
workspaceStorage/<hash>/chatSessions/*.{json,jsonl}
workspaceStorage/<hash>/workspace.json
globalStorage/emptyWindowChatSessions/*.{json,jsonl}
globalStorage/transferredChatSessions/*.{json,jsonl}
```

`workspace.json` must remain included because it supplies project attribution.

See [`vscode_copilot_provider.go`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/parser/vscode_copilot_provider.go#L161-L301).

This should exclude unrelated extension storage, tooling databases, user history and caches.

### Zed

Zed parses the shared database:

```text
threads/threads.db
```

Its watcher also recognises:

```text
threads/threads.db-wal
threads/threads.db-shm
```

See [`zed_provider.go`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/internal/parser/zed_provider.go#L10-L85).

At minimum, narrow the target to the `threads/` directory. Prefer a consistent SQLite snapshot, similar to Hermes, to avoid copying the database and WAL while Zed is writing them.

This should exclude Zed’s external-agent registry, language servers, Node installations and other application state.

## Design compatibility

This change is compatible with the incremental HTTP remote-sync design. It preserves:

- collector-owned parsing
- raw-source transfer
- persistent per-host mirrors
- manifest-based delta transfer
- local reparsing
- recovery after mirror loss
- existing session identity and database semantics

Provider-specific knowledge remains in target resolution, not in the transport protocol. The protocol continues to exchange the existing `TargetSet`, manifest and archive structures. Windsurf, RooCode, Kilo Legacy and Poolside already establish this boundary.

“Complete mirror” should mean a complete copy of each provider’s resolved parse inputs, including required companions. It should not require recursively copying an application-data root containing unrelated files.

The change is also compatible with:

- [`docs/remote-access.md`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/docs/remote-access.md#L155-L246), including target allowlisting, mirror recovery and independent transfer and import breadth
- [`docs/internal/background-sync-efficiency.md`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/docs/internal/background-sync-efficiency.md#L23-L40), by reducing unrelated mirror cardinality
- [`Scoped Push and Session Provider Controls Design`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/docs/superpowers/specs/2026-08-09-scoped-push-and-provider-controls-design.md), because providers remain eligible for remote export and import
- [`SECURITY.md`](https://github.com/kenn-io/agentsview/blob/c276c9c00146cc7442df9b2032e98b04216bb7f3/SECURITY.md#L57-L76), because the collector continues parsing raw remote inputs as untrusted data while transferring less unrelated state

## Acceptance criteria

- Cursor sync transfers only parser-discovered transcripts and required companions.
- VS Code Copilot sync transfers only chat sessions and required workspace metadata.
- Zed sync transfers only data required to parse `threads.db`.
- A fresh curated sync produces the same parsed sessions as the existing full-root sync.
- Incremental sync transfers only changed selected files.
- Additions, updates, deletions and interrupted sync recovery continue to work.
- Existing symlink, path-confinement and forbidden-root protections remain in place.
- Missing files caused by target-resolution races do not fail the whole sync.
- Existing mirrors remove files absent from the new curated manifest.
- Tests prove unrelated files beneath each configured root do not appear in manifests or archives.
- Remote-access documentation describes curated provider exports and their disk requirements.

## Non-goals

- Sending parsed sessions between AgentsView instances.
- Changing the collector-owned parsing model.
- Changing session identity or database schemas.
- Narrowing every provider in the same change.

rodboev:
Remote sync currently recurses through broad Cursor, VS Code Copilot, and Zed roots, copying unrelated editor data into persistent mirrors. This bounds those providers to parser-required Cursor transcripts, VS Code chat sessions plus workspace metadata, and standalone Zed `threads.db` snapshots through the existing `TargetSet.Files`, manifest, delta, mirror, and import paths.

Empty Cursor and VS Code roots remain explicitly file-scoped when no sessions are discovered, so credential, settings, cache, and extension files cannot fall back to recursive export. When the last VS Code chat disappears, its workspace metadata is omitted with the stale chat and the persistent mirror evicts both entries.

Zed exports use SQLite online backup so committed WAL data becomes one consistent database without transferring WAL or SHM sidecars. Authenticated target selection preserves root and symlink confinement, unchanged mirrors produce no repeat fetch, and deletion races evict stale files without widening the selected set.

Closes #1492.

### DIFF
--- a/docs/remote-access.md
+++ b/docs/remote-access.md
@@ -231,6 +231,39 @@ each sync while the rest of the host remains eligible for delta transfer. The
 Windsurf content in the mirror is the sanitized export, not a byte-for-byte copy
 of the remote state database.
 
+### Curated Provider Exports
+
+Most providers export their configured directories as-is. Providers whose
+configured root is a broad application-data directory export a curated file
+list instead, so caches, extension state, and credential files stored near the
+sessions never enter the manifest, the archive, or the mirror:
+
+- **Cursor** exports only the transcript files its parser discovers under each
+  project's `agent-transcripts/` directory.
+- **VS Code Copilot** exports only chat session files under
+  `workspaceStorage/<hash>/chatSessions/` and
+  `globalStorage/{emptyWindowChatSessions,transferredChatSessions}/`, plus each
+  exporting workspace's `workspace.json` for project attribution.
+- **Zed** exports `threads/threads.db` as a consistent SQLite snapshot taken
+  with the online backup API, so committed write-ahead-log data arrives as one
+  standalone database without `-wal` or `-shm` sidecars. Hermes state databases
+  use the same snapshot mechanism.
+- **RooCode** and **Kilo Legacy** export only their discovered per-task session
+  files, **Poolside** narrows its root to `trajectories/`, and **Windsurf**
+  exports the sanitized copy described above.
+
+Budget mirror disk for the curated session files only, not for the full
+application-data directories those editors keep around them. When the last
+session under a curated Cursor or VS Code Copilot root disappears, the root
+stays advertised with an empty file list so the next sync empties the mirror
+instead of failing.
+
+An unreadable snapshot database is treated as missing: the sync continues, the
+mirror drops its cached copy, and the file transfers again once the remote
+application repairs it. Sessions already imported into the AgentsView database
+remain available throughout. Any other read failure during target resolution
+fails that host's sync without touching the mirror.
+
 ### When AgentsView Downloads A Full Archive
 
 A full HTTP transfer occurs in these cases:
--- a/internal/parser/cursor_provider.go
+++ b/internal/parser/cursor_provider.go
@@ -329,11 +329,16 @@ func (s cursorSourceSet) streamingSourceRefWithCache(
 ) (SourceRef, bool, error) {
 	root = filepath.Clean(root)
 	path = filepath.Clean(path)
-	info, err := os.Stat(path)
+	// A transcript listed by the walk can vanish before this stat — a
+	// routine deletion race, not a discovery failure; the next
+	// enumeration simply no longer lists it. Genuine read errors and
+	// dangling symlinks still propagate, matching
+	// cursorFindSourceFileInProjectStrict.
+	regular, err := streamingRegularFileCandidate(path)
 	if err != nil {
 		return SourceRef{}, false, fmt.Errorf("stat cursor transcript %s: %w", path, err)
 	}
-	if !info.Mode().IsRegular() {
+	if !regular {
 		return SourceRef{}, false, nil
 	}
 	rawID, ok := cursorRawSessionIDFromPath(root, path)
--- a/internal/parser/cursor_provider_test.go
+++ b/internal/parser/cursor_provider_test.go
@@ -254,6 +254,47 @@ func TestCursorStreamingDiscoveryPropagatesAuthoritativeResolutionErrors(t *test
 	})
 }
 
+func TestCursorStreamingDiscoverySkipsTranscriptVanishedBeforeStat(t *testing.T) {
+	root := t.TempDir()
+	transcriptsDir := filepath.Join(root, "Users-demo", "agent-transcripts")
+	healthy := cursorProviderWriteJSONLTranscript(
+		t, transcriptsDir, "healthy.jsonl", "still on disk",
+	)
+	ctx := withStreamingDirectoryReader(t.Context(), func(
+		_ context.Context, dir string, yield func(os.DirEntry) error,
+	) error {
+		if samePath(dir, transcriptsDir) {
+			if err := yield(failingInfoDirEntry{
+				name: "a-vanished.jsonl", err: os.ErrNotExist,
+			}); err != nil {
+				return err
+			}
+		}
+		entries, err := os.ReadDir(dir)
+		if err != nil {
+			return err
+		}
+		for _, entry := range entries {
+			if err := yield(entry); err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+	provider, ok := NewProvider(AgentCursor, ProviderConfig{Roots: []string{root}})
+	require.True(t, ok)
+
+	var streamed []string
+	err := provider.(StreamingDiscoverer).DiscoverEach(ctx, func(source SourceRef) error {
+		streamed = append(streamed, source.DisplayPath)
+		return nil
+	})
+
+	require.NoError(t, err,
+		"a transcript deleted between enumeration and stat must not fail discovery")
+	assert.Equal(t, []string{healthy}, streamed)
+}
+
 func TestCursorProviderResolvesDuplicateStemsWithinProject(t *testing.T) {
 	root := t.TempDir()
 	firstProject := "Users-fiona-Documents-first"
--- a/internal/parser/kilo_legacy_provider.go
+++ b/internal/parser/kilo_legacy_provider.go
@@ -1,6 +1,8 @@
 package parser
 
 import (
+	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -21,9 +23,7 @@ func newKiloLegacyProviderFactory(def AgentDef) ProviderFactory {
 			return NewSingleFileSourceSet(
 				def.Type,
 				cfg.Roots,
-				WithFileDiscovery(func(root string) []singleFileMatch {
-					return kiloLegacyDiscoverFiles(root)
-				}),
+				WithStreamingFileDiscovery(kiloLegacyDiscoverEach),
 				WithFileWatchRoots(func(roots []string) []WatchRoot {
 					return kiloLegacyWatchRoots(roots)
 				}),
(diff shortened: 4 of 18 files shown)


wesm:
looking at this, fixing a few things.

wesm:
I pushed two commits to this branch that change how it handles unreadable files.

**A corrupt Zed database no longer blocks the whole sync.** Before, if `threads.db` could not be opened, the manifest and archive requests failed, so one bad file stopped every agent's sync for that host. Now an unreadable database is treated like a missing one, the same way Hermes state databases already work. Nothing is lost: sessions already imported stay in the archive database, and the file transfers again once Zed repairs it.

**Read errors can no longer silently shrink the target set.** The curated resolvers swallowed discovery errors, so a transient permission failure made the server advertise nothing for that editor. The collector treats "not advertised" as "uninstalled" and would delete that agent's entire mirror, then re-download everything after the error cleared. Resolution errors now fail the request instead, which matches how the directory-scoped manifest walk already behaves. The resolvers also use the streaming discovery path, whose error handling the local sync engine already depends on; the batch path still converts traversal failures into an empty result.

Smaller cleanups in the same commits: the Cursor/VS Code "keep the root advertised with an empty file list" behavior is a named trait instead of repeated agent-type checks; the stale-file check returns one bool, making it explicit that a stale file is only ever omitted, never streamed; the Hermes-specific snapshot helpers now have shared sqlite-snapshot names; and `docs/remote-access.md` documents what each curated provider exports and how much mirror disk to budget.

New tests cover the corrupt-database and unreadable-root cases.

wesm:
Response to the roborev review of `33da0e4`: two of the three findings were real and are fixed in `def783a`; the Kilo Legacy finding does not reproduce.

**Kilo Legacy streaming discovery (refuted).** The single-file source set that Kilo Legacy is built on always implements `DiscoverEach` — when no streaming discovery function is configured, it iterates the batch `WithFileDiscovery` results internally instead of returning `ErrUnsupportedProviderFeature` (`internal/parser/single_file_source_set.go`, `singleFileSourceSet.DiscoverEach`). I verified with a real Kilo root: `ResolveTargets` returns the curated task files with no error. The unsupported-feature error only occurs for source sets that lack a `DiscoverEach` method entirely, and none of the resolvers routed through `discoverProviderSources` use one.

**RooCode and Kilo root stat errors (fixed).** Both resolvers now treat only a missing or non-directory root as absent; any other stat failure fails the resolution, matching the editor and Zed resolvers. Covered by a new unreadable-root test case.

**Explicitly empty curated file request (fixed).** `SelectAllowedTargets` now rejects a curated directory request whose file list is empty while the fresh resolution has curated files. The acceptance is limited to the case it exists for: a Cursor or VS Code scope that is itself empty. Covered by a new rejection case for every curated agent, Zed included.

wesm:
Response to the roborev review of `def783a`: the finding is real, but at a different line than cited — fixed in `028697d`.

The cited helper, `streamingRegularFileCandidate`, already handles its own race: it returns `false, nil` for a genuinely missing file and only propagates `os.ErrNotExist` for a dangling symlink. The actual leak was one stat later, in `streamingSourceRefWithCache` (`internal/parser/cursor_provider.go`), which propagated every stat error — including `ErrNotExist` for a transcript deleted between directory enumeration and the stat. With target resolution now failing closed, that routine deletion race became an HTTP 500 for the whole sync request.

The fix is in the parser rather than the resolver, so the local sync engine benefits too (it previously treated the same race as a failed discovery scope). The stat now goes through `streamingRegularFileCandidate`, matching the provider's own conventions in `cursorFindSourceFileInProjectStrict`: genuine absence skips the entry — the next enumeration simply no longer lists it and the mirror evicts it then — while dangling symlinks and real read errors still propagate. RooCode's streaming discovery already tolerates this race (`IsRegularFile`), and VS Code's swallows per-file errors, so Cursor was the only affected provider.

A regression test simulates the race with an injected directory reader that lists a transcript which no longer exists; discovery must yield the surviving transcript with no error. I verified the test fails against the previous code.

wesm:
Fixed in `6ade0dc`. The curated resolvers no longer swallow per-file stat failures: a missing file is still an omitted target, but any other stat error (an unreadable session file, workspace manifest, or task sibling) now fails the resolution instead of silently shrinking the curated set — for RooCode and Kilo Legacy, an all-files failure could previously drop the whole root. All the sites route through one helper, `statRegularRemoteSyncFile`, which carries the same contract the root and discovery paths already follow. A test covers the propagation via an unreadable parent directory.

On the Windows CI failure: `TestCaptureDistinguishesAnObservedSourceThatDisappears` in `internal/capture` is unrelated to this PR (it exercises the Claude capture flow, which this branch does not touch) and the identical failure appears on `main` — see the Windows job of run 33086658060 (commit `b83abc2`, Aug 27). It is a pre-existing timing-sensitive test on windows-latest; re-running the job should clear it here.

wesm:
Fixed in `01699a6`. The finding is valid: a database could pass the snapshot identity probe and then vanish or turn unreadable before the online backup, failing the whole archive request while the mirror kept its stale copy until the next cycle.

Rather than classifying backup error strings, the fix re-probes the source on backup failure. An unusable source becomes an omitted archive entry — exactly as if the identity probe had failed first — so the next manifest evicts the mirror's stale copy. A still-usable source means the failure was local (temp directory, destination write) and still fails the archive loudly. The existing snapshot-failure test pins the loud branch; a new test deletes the source mid-backup and requires the archive to succeed with the entry omitted.

wesm:
Valid — fixed in `8db977e`, at the parser layer rather than the cited resolver lines. The resolver's own per-file stats were already fixed in `6ade0dc`; the remaining swallow was inside discovery, where the resolver cannot compensate because a skipped task never reaches it.

Two distinct holes:

- **RooCode**: the task walk's `IsRegularFile` check silently skipped a task whose directory could not be searched — contradicting the contract its own unreadable-tasks-directory test states. The per-file stat now treats only genuine absence as "not a session" and propagates other failures.
- **Kilo Legacy**: worse — its batch discovery (`func(root) []singleFileMatch`) had no error channel at all, so even a fully unreadable `tasks/` directory enumerated as an authoritative empty scope. It now uses the same streaming discovery shape RooCode already uses, with the same error contract.

Tests: unreadable task-directory cases for both providers (both verified to fail against the old code — Kilo's old signature could not report an error at all), plus an end-to-end resolver case proving an unreadable Kilo tasks directory fails target resolution instead of resolving to an empty target set.

One correction to an earlier comment of mine: the `TestResync*` failures I described as pre-existing on this branch were an artifact of running `go test` without the `fts5` build tag — with `CGO_ENABLED=1 go test -tags fts5`, the full `internal/sync` suite passes on this branch. Disregard that part; the Windows capture flake remains real and pre-existing on `main`.

PR #1514, opened 2026-08-25
feat(data): add a project-first mapping workspace

mariusvniekerk:
Adds an opt-in, project-first Data workspace for correcting project classification. The existing rules view remains the default when `VITE_PROJECT_MAPPING_WORKSPACE` is unset.

Shift-click selects projects for bulk correction. The destination picker and Save stay above an independently scrolling folder list. Suggestions come from observed working directories, collapsing known worktree layouts and immediate shared parents. They remain grouped by project, then machine, with larger groups first within each machine.

An optional date picker narrows project counts, folder suggestions, and paged session previews. Automated previews are hidden by default. Folder rules still match across all dates; bulk previews count each session once and require confirmation when rules reach unselected projects. Save shows progress as individual corrections complete.

Individual sessions can be reassigned from previews or the Sessions list. Overrides survive resync and take precedence until cleared. Read-only mirrors remain inspection-only; writes belong to the Source Agents View.

Review candidate selection in `internal/db/worktree_candidates.go`, the Data workspace components, and override persistence in `internal/db/session_project_assignments.go`. Chunked saves, retry improvements, and correction history/Undo remain follow-up work.


rodboev:
The folder suggestions are useful, but their order currently reads as alphabetical. Could candidates be ranked or annotated using repository identity, drive, common parent path, or path distance? For example, a `D:/Repos/...` worktree should be easier to compare with another `D:/Repos/...` checkout than a path on `C:`. The destination should still require an explicit choice.

"Map whole project" and "Correct one folder" still leave the scope unclear. A shorter pair such as "One folder" and "All folders" may be easier to understand, with the bulk panel explaining that all listed folders will receive the selected destination.

The bulk view needs to make the operation clear before the folder list. Put the destination picker and Save action together near the top, with Save beside the picker, and keep that row visible while the list scrolls. A short instruction there could be: "Select folders to correct, then choose their destination project." Right now the destination appears near the bottom, so the source folders and the action feel disconnected.

In the example here it's not clear what the flow is since the dropdown and Save button are cut off. If they were pinned to the bottom of the container, so that they are always visible, it would be easier to connect that the user is taking the folders at the top and assigning them to a folder they select:

2F7F8FB3-AFAF-4DC7-BC55-5177A2FDEDB8

PR #1523, opened 2026-08-25
ci(desktop): run the full desktop unit suite in the Windows lane

prateek:
The Desktop Unit Tests (Windows) lane was added in #488 to cover the Windows update stop/install ordering, and its cargo test invocation filtered to `install_downloaded_update`. That filter means the rest of the desktop crate's unit tests run nowhere in CI: this is the only lane that executes desktop Rust tests, and the macOS desktop workflows only build artifacts.

This drops the filter so the lane runs the whole `--lib` suite. The full test module already compiles in this lane today (cargo's test filtering happens at runtime), so the added cost is only test execution, which is sub-second. Platform-gated tests are behind `#[cfg(unix)]` and the remaining tests use tempdirs and missing-binary spawns that behave the same on Windows.

Reviewers: the only change is the test step in `.github/workflows/ci.yml` (name + dropped filter).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -370,8 +370,8 @@ jobs:
             Start-Sleep -Seconds (10 * $attempt)
           }
 
-      - name: Run Windows desktop update tests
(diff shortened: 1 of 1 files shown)


wesm:
looking

PR #1525, opened 2026-08-25
feat(desktop): open sessions through agentsview deep links

### LINKED ISSUE #1524: Open desktop sessions through agentsview:// deep links
### Problem / motivation
External tools can resolve session IDs but cannot open them directly in the desktop app. Browser URLs require a known server port, and links to sessions that have not synced yet render a blank message pane.

### Proposed solution
Register `agentsview://sessions/<session-id>` for the desktop app, support an optional `?msg=<ordinal|last>` anchor, and handle links both at launch and while the app is running. Defer navigation until the backend is ready, focus the existing window for later links, and show a retryable not-found state that recovers when the session appears.

### Alternatives considered
Continue opening `http://<host>:<port>/sessions/<session-id>` in a browser. This remains a useful fallback when the desktop scheme is unavailable, but callers must discover the active server address and cannot target the desktop app directly.

prateek:
Registers the `agentsview://` scheme so external tools can open `agentsview://sessions/<session-id>` in the desktop app, optionally forwarding `?msg=<ordinal|last>` to the frontend message anchor. Unsupported paths, invalid anchors, and unrelated query parameters are discarded or ignored.

Deep-link dispatch now focuses the existing window, defers launch-time links until the sidecar is ready, and re-enters the deferred state before the backend republishes its port so restart redirects cannot overwrite the requested route. A link that arrives while the startup redirect itself is navigating queues behind it and replays once it lands, so the newest link always wins. Windows and Linux use single-instance forwarding for links delivered through a second launch, and diagnostics are mirrored to `agentsview-desktop.log` for packaged builds without usable stderr, including deferred routes that are dropped or fail to navigate.

The frontend distinguishes a missing session from an empty one across every selection path: detail fetches, sidebar-row hydration, and the message loader's metadata probe all report 404s, so a session deleted behind a cached sidebar row shows a retryable not-found state instead of a blank pane. The not-found flag stays set until a fetch for the session succeeds, and recovery re-runs the per-session loaders in place rather than bouncing the selection, so retries and late-arriving sessions keep the URL, history, and `?msg=` anchor intact. Stale hydration responses are guarded by the sidebar's version and epoch checks, stale message-load failures are ignored unless they belong to the current load, and other request failures are displayed as before.

The scheme is available only with the desktop app. macOS development builds do not receive scheme events because registration comes from the bundled app, unsupported links are logged without an in-app error, and Linux single-instance forwarding depends on a session D-Bus. Review the `DeepLinkDispatch` transitions and URL validation in `desktop/src-tauri/src/lib.rs`, plus the not-found and retry flow in the sessions store and message list.

The branch also carries a small CI repair: the usage guide's Session API link now points at the docs path, and the Codex capture integration test no longer overrides the finalization wait, which had made it time out.

Closes #1524


prateek:
@mariusvniekerk thanks for pushing this over the line!

Issue #1484, opened 2026-08-22
Antigravity CLI does not group sessions from different worktrees under one project


mgcnrx11:
### Summary

When the Sessions sidebar is grouped by Project, Antigravity CLI sessions from different worktrees are displayed as separate project groups. The same worktree layout is grouped correctly for Codex sessions.

### Steps to reproduce

1. Create multiple Git worktrees for the same repository.
2. Run Antigravity CLI sessions from at least two of those worktrees.
3. Sync the sessions into AgentsView.
4. In the Sessions sidebar, choose Group by Project.

### Expected behavior

Sessions from different worktrees of the same repository should appear under one project group, consistent with Codex.

### Actual behavior

Each Antigravity CLI workspace path is treated as a separate project, so the worktrees appear as separate project groups.

### Technical context

The Antigravity CLI provider reads the workspace value from history.jsonl and passes it unchanged as ProjectHint. The parser then uses that non-empty hint directly as ParsedSession.Project. It does not run the shared project/worktree normalization used by Codex (ExtractProjectFromCwdWithBranch), and the parsed Antigravity CLI session does not populate ParsedSession.Cwd.

As a result:

- the stored sessions.project value is the raw workspace path;
- the sync-stage worktree resolver is skipped because it requires a non-empty cwd;
- the frontend groups by the raw session.project value.

A likely fix is to reuse the shared project extraction logic for Antigravity CLI workspaces, while preserving the workspace path as Cwd where appropriate. Please add a regression test proving that two worktree paths for the same repository produce one project group.

### Related discussions

- #416 — Grouping worktrees into parent project
- #125 — Resolve project name for deleted nested worktrees
- #1294 — Attribute Git worktrees to repositories
- #579 — Infer Antigravity CLI project when history rows omit conversationId
- #942 — Worktree mappings do not apply to sessions whose cwd column is empty

### Checklist

- [x] I searched existing issues for related discussions
- [x] No private session files or absolute personal paths are included

PR #1527, opened 2026-08-25
fix(parser): group Antigravity CLI sessions across git worktrees

anhermon:
Antigravity CLI sessions from different git worktrees of the same repo now group under one project in the Sessions sidebar, matching how Codex and Claude sessions already behave. Previously the raw workspace filesystem path from `history.jsonl` was stored as `ParsedSession.Project` with no normalization, so each worktree showed up as its own project group (#1484).

The fix threads `context.Context` through `parseSessionWithStatus` and routes the workspace path through the shared `ExtractProjectFromCwdWithBranchContext` normalizer, which resolves a worktree path back to its parent repository name. Because of this, stored Antigravity project names change shape: raw paths like `/home/user/my-cli-project` become `my_cli_project`. The parser data version is bumped to 92 so archives written before the change reparse: their sources are byte-identical, so nothing fingerprint-side could otherwise trigger the renormalization, and the same repo would stay split between a path-named group and the normalized one. A sync integration test pins that boundary by stamping a raw-path row with version 91 and asserting an ordinary incremental sync renormalizes it.

Remote (SSH-imported) parses keep the lexical normalization but skip local git-root discovery, gated on `Config.PathRewriter` — the same remote signal the Cursor provider uses. Without the gate, a remote workspace path that also exists on the importing host would be attributed to the importer's local repository.

The engine's `DisableFilesystemProjectDiscovery` option now holds on every entry point: previously only `SyncPaths` put the policy on the parse context, so a discovery-disabled engine running a full sync, streamed reconciliation, single-session refresh, source lookup, or parse diff would still walk the local filesystem for git roots. A `parsePolicyContext` helper is applied at each funnel, and an integration test drives a git-worktree workspace through `SyncAll` on a discovery-disabled engine to prove the stored project stays lexical.

Review focus: the normalizer call and remote gate in `internal/parser/antigravity_cli.go`, the data-version bump in `internal/db/db.go`, and the `parsePolicyContext` call sites in `internal/sync/engine.go`; the rest is context plumbing and test fixtures (history.jsonl lines are built with `json.Marshal` so Windows workspace paths with backslashes survive JSON decoding).

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -428,7 +428,11 @@ CREATE INDEX IF NOT EXISTS idx_provider_freshness_updated_at
 // (91: Posit Assistant inferred cache-write normalization. Existing
 // non-Anthropic auto-cached model rows need re-parsing so their persisted
 // uncached prompt remainder is priced as input rather than cache creation.)
-const dataVersion = 91
+// (92: Antigravity CLI workspace project normalization. Existing rows store
+// the raw workspace path from history.jsonl as the project; re-parsing routes
+// it through the shared cwd normalizer so sessions from different git
+// worktrees of the same repo group under one project.)
+const dataVersion = 92
 
 const tokenCoverageRepairStatsKey = "token_coverage_repair_v1"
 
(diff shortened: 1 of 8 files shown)


wesm:
looking

PR #1528, opened 2026-08-25
feat(parser): add Cursor IDE (GUI) session provider

anhermon:
Cursor IDE, the GUI editor, stores chats in a VS Code-style global-state SQLite database (`state.vscdb`, table `cursorDiskKV`) rather than the `agent-transcripts` files the existing `cursor` (Cursor Agent CLI) provider reads, so GUI sessions were never discovered. This adds a distinct `cursor-ide` provider on the shared `multiSessionContainerSourceSet` framework (the same shape as Zed and Omnigent): `composerData:<uuid>` rows decode into sessions, and each session's `fullConversationHeadersOnly` array orders lookups into sibling `bubbleId:<composerId>:<bubbleUuid>` rows for message content, tool calls, and timestamps. The existing `cursor` provider is untouched.

Freshness and deletion semantics, where reviewers should look:

- Fingerprinting never hashes the full database (86MB+ observed locally, 500MB+ reported in the wild). The container fingerprint uses size, composite mtime, and a SQLite transaction-state hash over just the database and WAL headers (change counter, schema cookie, WAL size and salts), declared with `FingerprintHashInCacheKey` and `FingerprintHashRequiredForFreshness`, so a rewrite that lands on the same size and mtime still misses the skip cache. The per-composer fingerprint digests every parse input — the raw `composerData` document plus that composer's bubble rows — so an equal-length bubble rewrite, a header reorder, or a rename that leaves `lastUpdatedAt` untouched still reparses under `UnchangedResultMTimeAndHash`.
- A chat deleted inside Cursor IDE is marked source-missing and stays browsable, never hard-deleted. Stored-source hints fan a `state.vscdb` change event out into per-member tombstones using a batched single-connection membership check, and `internal/sync/engine.go` includes cursor-ide in complete-container ownership reconciliation, in the emptied-container source-missing seam, and in the whole-container skip-cache exemptions generalized from Omnigent (no stored row of its own, parser data version in the cache key, cache-after-write). Removing `state.vscdb` itself preserves every archived session.
- `state.vscdb` co-locates Cursor's live auth tokens (`ItemTable`) and per-composer sync encryption keys, so the provider is `RemoteSyncExcluded`, matching the Omnigent `chat.db` precedent.

Deletion, reconciliation, emptied-container, stale-edit, and same-size/same-mtime rewrite behavior is pinned by `internal/sync/cursor_ide_integration_test.go`; parsing and fingerprinting by `internal/parser/cursor_ide_test.go`. The `docs/internal/session-format-sources.md` entry records the schema evidence per the provenance convention.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -838,15 +838,68 @@ add an archived or maintained mirror without replacing the original identity.
   confirms local chat persistence and the separate SQLite history index.
   Cursor support on the official forum documents the
   [`~/.cursor/projects/<project>/agent-transcripts` layout](https://forum.cursor.com/t/chat-history-gone-after-pc-restart-agent-transcripts-files-emptied-how-to-recover/158251/5)
-  and identifies `state.vscdb` as metadata. Cursor's public GitHub
-  organization was also searched 2026-07-19; no transcript schema or producer
-  source was found.
+  and identifies `state.vscdb` as metadata **for this CLI producer**. That
+  characterization does not extend to Cursor IDE (the GUI editor, see below):
+  for the GUI, `state.vscdb` is the only transcript store, not metadata beside
+  one. Cursor's public GitHub organization was also searched 2026-07-19; no
+  transcript schema or producer source was found.
 - **Usage and cost:** The consumed text/JSONL transcripts have no reliable
   per-message token, cache, reasoning, credit, or monetary-cost fields.
 - **Agentsview:** `internal/parser/cursor.go`,
   `internal/parser/cursor_paths.go`, and `internal/parser/cursor_provider.go`;
   workspace identity uses a filesystem-backed unique-match resolver, while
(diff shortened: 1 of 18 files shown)


anhermon:
Updated (47b38434): ran an independent adversarial review against this PR (findings below) and applied its one actionable suggestion -- `cursorIDEClassifyPath` now passes `rejectShmSiblingEvents: true`, matching Omnigent's identical classifier. No behavior change, just drops a redundant `-shm` watcher event earlier.

Review verdict: **safe to merge**. Checked auth-token isolation (ItemTable is never touched), RemoteSyncExcluded enforcement across every sync/transfer path, SQLite read-only/WAL concurrency handling, and edge-case resilience (missing rows, zero timestamps). One other informational finding (zero-time UnixNano edge case) was confirmed already unreachable in practice -- no change needed there.

wesm:
looking

PR #1529, opened 2026-08-25
fix(vector): cap embedding batches by token budget

anhermon:
Embedding builds now accept a global model-context limit and a per-server aggregate input-token cap, then conservatively bound request width by floor(max_batch_tokens / model_context_tokens). This prevents deterministic oversized batches from being replayed forever after provider-side truncation.

The bound intentionally charges every input at the full model context instead of guessing model-specific tokenization. That can create smaller requests for short documents, but it guarantees the configured provider cap for ordinary and repair builds. Existing configurations keep document-count batching until both limits are declared.

The main review points are config validation and server wiring in internal/config and cmd/agentsview, plus the shared batch sizing in internal/vector/build.go.


anhermon:
Superseded by #1530 (same fix, cleaner PR body/tests). Closing the duplicate.

Issue #1356, opened 2026-08-07
embeddings build can construct a batch that exceeds the provider's per-request token cap, then retries it unchanged forever


salmonumbrella:
Batch composition counts documents, but the provider caps a request by *tokens*. Because each input is truncated to the model context, the worst case for a batch is `batch_size × model_context`, which can exceed the per-request cap. When it does, the same batch is rebuilt identically on every retry, so the build cannot advance past it.

Distinct from #1353: that one is about a 429 (a rate limit that clears with time). This is a 400 that never clears, because nothing about the request changes between attempts.

## What happened

Backfilling ~142k documents with `voyage-4-large` @ 1024. Voyage caps a single request at 120,000 tokens and truncates each input to the model context (32,000 tokens).

Three runs, three different `batch_size` values, same failure:

```
# batch_size unset (default)
status 400: The max allowed tokens per submitted batch is 120000.
Your batch has 147910 tokens after truncation.  TOO_MANY_TOKENS_IN_BATCH

# batch_size = 16
Your batch has 121625 tokens after truncation.

# batch_size = 4
Your batch has 128000 tokens after truncation.
```

The third is the tell: **128,000 is exactly 4 × 32,000.** A batch of four inputs, each truncated to the model context, lands precisely on the theoretical maximum — and 8,000 tokens over the cap.

So the safe bound is arithmetic, not tuning: the largest `batch_size` that *cannot* exceed 120,000 with a 32,000-token context is **3** (96,000). Anything above that can fail depending on the documents that happen to land together, which makes it look intermittent and content-dependent while actually being deterministic per batch.

## Why it strands the build

Batch composition is deterministic, so a rejected batch is rebuilt identically next time. The scheduler burns its retries on a request that cannot succeed and stops:

```
embed scheduler: build failed: encode document ...: status 400 ... TOO_MANY_TOKENS_IN_BATCH
embed scheduler: retry limit reached; deferring until new work
```

`deferring until new work` is the part that hurts on a backfill: there is no new work coming, so progress simply halts with a large `missing` count and no indication that it is permanent rather than slow. Combined with #1353, a long build has two independent ways to stop early and look merely slow.

## Suggested fix

Bound batch construction by an estimated token budget rather than a document count — cap at `min(batch_size, floor(request_cap / model_context))`, or accumulate estimated tokens and cut the batch when the next input would cross the cap.

[...]

PR #1531, opened 2026-08-25
fix(vector): keep embedding builds alive through rate limits

anhermon:
## Summary

- Document embeddings builds now retry an HTTP 429 (honoring `Retry-After` when present, capped exponential backoff otherwise) until it succeeds or the build is canceled, instead of spending the normal `max_retries` budget on it.
- Query encoders are unchanged: latency-sensitive, so a 429 still counts against `max_retries` and fails fast, same as before.
- Daemon shutdown now cancels a detached API-started build instead of waiting for it to finish. With unbounded 429 retries, waiting could block shutdown forever when a provider's quota never recovers. Build progress is durable, so a canceled build resumes incrementally on the next start; scheduler-driven builds already stopped with the daemon context, and API-started builds now behave the same way.

## Root cause

`embeddings build` treated a 429 as just another retryable error under the shared `max_retries` budget. A 429 is the one error class a provider tells you exactly how to handle (it clears with time, the request itself isn't broken) -- treating it like a permanent failure meant a large first build died partway through and needed an external retry loop wrapped around the command by hand (repro in the issue: 28,085 of 142,161 documents embedded, then a 429 aborted the whole run).

## Scope

`internal/vector/encoder.go`'s retry loop, a `RetryRateLimits` field threaded through the existing document-vs-query encoder constructors in `cmd/agentsview/embeddings.go` (only document builds enable it), and `internal/vector/manager.go`, where detached builds now run on a manager-owned cancellable context with a `Shutdown` method that `vectorServing.Close` calls before waiting. Reviewers should look at the reworked retry loop in `encode` and at the `Shutdown`/`Wait` ordering in `Manager` and `vectorServing.Close`.

**Note:** #1356's fix (separate PR, already up) touches the same `embeddings.go`/`config.go`/`docs/semantic-search.md` files for an unrelated reason (per-batch token-cap). Whichever of the two merges second will need a small conflict resolution against the other -- I hit that conflict myself while preparing this PR and reconciled it manually rather than force-resolving; happy to help resolve it again here once one lands.

### DIFF
--- a/cmd/agentsview/embed_scheduler.go
+++ b/cmd/agentsview/embed_scheduler.go
@@ -722,8 +722,11 @@ func setupVectorServing(
 		RecallScheduler:      recallScheduler,
 		RecallMutationNotify: recallMutationNotify,
 		Close: func() error {
-			mgr.Wait()
-			recallMgr.Wait()
+			// Shutdown, not Wait: a detached API build may be waiting out
+			// provider rate limits indefinitely and must be canceled for
+			// daemon shutdown to complete.
+			mgr.Shutdown()
+			recallMgr.Shutdown()
 			ixErr := ix.Close()
 			recallErr := recallIX.Close()
 			lockErr := lock.Close()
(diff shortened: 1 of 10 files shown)


wesm:
looking

PR #1532, opened 2026-08-25
feat(rawcapture): add durable capture outbox

salmonumbrella:
## What changed

- Add provider-owned raw capture plans for Claude and Codex sources.
- Capture stable full snapshots and verified append suffixes without parsing raw content.
- Add a bounded, crash-safe outbox with ordered delivery, fenced acknowledgements, recovery, and compact terminal state.

## Why

This delivers the durable local capture layer needed before hosted raw sync can run continuously. Provider code remains responsible for physical source membership, while generic capture and checkpoint code enforce containment, stability, capacity, and server-chain ordering.

## Usage

There is no new runtime command in this slice. The new package APIs are ready for the uploader and scheduling work in the next stacked change.

Refs #1352

Stacked on #1498.

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -1,3 +1,7 @@
+review_agent = "codex"
+review_model = "gpt-5.6-terra"
+review_reasoning = "high"
+
 review_guidelines = """
 agentsview is a single-user developer tool. Default mode binds to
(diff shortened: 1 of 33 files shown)


wesm:
rebasing

PR #1535, opened 2026-08-26
fix(pg): preserve legacy offset timestamps

### LINKED ISSUE #1534: fix(pg): preserve legacy offset timestamps during push
## Summary

`agentsview pg push` silently converts some valid legacy SQLite timestamp strings into PostgreSQL zero/NULL timestamps. A single affected root session can then make the Usage page's All-time range return HTTP 500.

## Reproduction

A preserved archive row contains timestamps in a space-separated, fractional, numeric-offset form:

```text
created_at = 2026-04-14 04:09:28.922+00
started_at = 2026-03-09 22:48:29.937+00
ended_at   = 2026-03-14 00:32:16.577+00
```

The row and its message timestamps are valid in SQLite. After a full PostgreSQL push:

- `created_at` becomes `0001-01-01T00:00:00Z`
- `started_at`, `ended_at`, and message timestamps become `NULL`

`ParseSQLiteTimestamp` accepts RFC3339 and a space-separated seconds-only form, but not fractional seconds plus a numeric offset. `pushSession` also discards the parse result for required `created_at`:

```go
createdAt, _ := ParseSQLiteTimestamp(sess.CreatedAt)
```

That passes Go's zero `time.Time` to PostgreSQL. Nullable timestamp helpers silently return `NULL` on the same parse failure.

When `/api/v1/stats` reports the zero date as `earliest_session`, the frontend's All-time Usage range requests `from=0001-01-01`. PostgreSQL usage filtering pads the lower bound by 14 hours, producing a year-0 timestamp (`0000-12-31T10:00:00Z`) and a 500 (`SQLSTATE 22008`).

## Expected behavior

- PostgreSQL push accepts valid legacy/round-tripped timestamp representations, including fractional seconds with numeric offsets.
- A non-empty timestamp that cannot be parsed must not silently become year 1 or `NULL`; push should return a contextual error instead of corrupting the mirror.
- Usage date-bound handling should not turn a boundary underflow into an internal server error.

## Suggested coverage

- Unit cases for space-separated fractional timestamps with `+00`, `+00:00`, and non-UTC offsets.
- PostgreSQL push regression proving `created_at`, `started_at`, `ended_at`, and message timestamps survive the mirror.
- A usage-bound regression for `from=0001-01-01` (either safe clamping or a client error rather than HTTP 500).

tpn:
Closes #1534.

Legacy and imported SQLite archives can retain valid space-separated timestamps with fractional seconds and numeric UTC offsets. PostgreSQL push previously failed to parse those values, silently mirrored nullable timestamps as `NULL`, and used Go's year-1 zero time for required `created_at` values.

The PostgreSQL timestamp parser now accepts both short and colon numeric offsets. Session, message, usage-event, and tool-result push paths reject any remaining non-empty unparseable timestamps with contextual errors before committing corrupt mirror data.

A target- and filter-scoped migration marker forces one full republish after upgrade, repairing timestamps that an earlier incremental push may already have mirrored as `NULL` or year 1. The marker is recorded only after the session push completes without errors, so interrupted or partially failed repairs retry automatically. Regression coverage exercises both the original legacy-format mirror and this upgrade repair against a real PostgreSQL database.

Usage candidate bounds now clamp timezone padding at year 1 in both SQLite and PostgreSQL, preventing an anomalous earliest-session value from underflowing into an invalid year-0 PostgreSQL timestamp. Normal date-range behavior is unchanged.


tpn:
Addressed the migration concern in cdbf9807. A scoped one-time marker now forces an error-free full session republish before recording completion, so legacy timestamps previously mirrored as NULL or year 1 are repaired automatically. The PostgreSQL regression test now simulates those stale rows and verifies that an ordinary incremental push repairs them.

PR #1537, opened 2026-08-26
feat(parser): ingest Posit Assistant usage-events sidecar

machow:
**tl;dr:** Posit Assistant tracks its cache warming mechanism in `usage-events.jsonl`. This PR ingests that file so warm-cache spend is part of cost calculation. I also quickly checked again some existing PA sessions I have on disk with cache warm events.

Posit Assistant records auxiliary per-request usage that never appears in the `lm-messages.jsonl` transcript — cache-keepalive pings and classifier calls — in an append-only `usage-events.jsonl` sidecar next to each conversation (subagents carry their own). That spend was previously invisible: idle sessions receive repeated keepalive pings that never reached the usage tables.

Each sidecar line becomes a request-scoped usage event with `Source` = `"posit-assistant-" + kind`, the same model-family cache-write normalization as message usage, a `MessageOrdinal` when its anchor node resolves, and a synthesized dedup key. The sidecar is supplementary to per-message usage — events are emitted only from sidecar lines and session token totals stay message-derived — so the additive messages-plus-usage-events cost queries never double count. Posit Assistant may add new kinds (e.g. `posit-assistant-greeting`) so `strings.HasPrefix` is used.

The sidecar also joins the composite source fingerprint and changed-path classification, so a keepalive append on an otherwise idle session triggers resync.

The request-scoped source policy previously lived in two hand-synced copies (`db.UsageSourceIsRequestScoped` and `usagefacts.sourceIsRequestScoped`); it is now consolidated in `usagefacts.SourceIsRequestScoped` with `db` delegating. The new sources are registered there by prefix (`posit-assistant-`) rather than as exact literals: one-request-per-line is a property of the sidecar format, not of specific kinds, so a future PA event kind stays correctly request-scoped (and gets per-request pricing, including long-context bands) without a second list to maintain.

Because existing archives predate sidecar events, the data version moves 91 → 92 so those sessions re-parse through the normal non-destructive resync path.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -1523,7 +1523,8 @@ add an archived or maintained mirror without replacing the original identity.
 ## Posit Assistant (`posit-assistant`)
 
 - **Format:** Workspace conversation directories containing `conversation.json`,
-  `lm-messages.jsonl`, and `ui-messages.jsonl`.
+  `lm-messages.jsonl`, `ui-messages.jsonl`, and an optional
+  `usage-events.jsonl` sidecar of auxiliary per-request usage.
 - **Evidence:** `no-public-source`.
 - **Upstream:** Posit's product documentation and the
   [posit-dev GitHub organization](https://github.com/posit-dev) were searched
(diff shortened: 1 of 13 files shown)


wesm:
rebasing

PR #1541, opened 2026-08-27
feat(pricing): apply Posit Assistant billing premium

### LINKED ISSUE #1538: Support provider specific billing calculations?
### Problem / motivation

Providers like Posit AI charge a model rate, plus a documented [10% premium on top of model rates](https://docs.posit.co/posit-ai/user/faq/#how-are-credits-included-with-the-pro-plan). Currently agentsview calculates only the underlying model rates, so does not include the 10% premium.

### Proposed solution

I would probably just lean towards only calculating the model rates for now (which is fairly well represented by the LiteLLM catalogue). But I'm happy to do any research / work on what provider-specific rates would look like.

### Alternatives considered

* Leave as is (I think this is reasonable).
* Add support for Provider dependent cost calculations (e.g. for Posit Assistant), but use model calculations by default.
* Do a sweep across Providers to evaluate how useful calculating Provider cost would be.

rodboev:
Posit Assistant usage previously inferred billing from the session agent, so managed Posit AI and bring-your-own-provider rows in the same session could not be distinguished. Computed usage therefore applied one billing rule to both cases.

This rework carries the provider identity from Posit Assistant message metadata and usage-event sidecars through usage facts, SQLite, PostgreSQL, DuckDB, cache rollups, activity reports, and session views. The exact `positai` provider receives the 11/10 managed-service adjustment after model, historical timestamp, request-band, and cache-tier selection. `anthropic`, empty, and other provider IDs remain at base rates, and custom pricing remains as configured. Explicit reported costs stay authoritative, cache savings for those rows use the resolved billed rates, and fixed web-search fees stay outside the adjustment.

Usage data version 94, usage-cache format 6, DuckDB schema 13, and export schema versions 6, 7, and 6 invalidate stale persisted results and document the changed cost semantics. No billing inference is made from model names, provider prefixes, or the session agent when row-level provider metadata is available.

Closes #1538


wesm:
cc @machow, do you want to have a look at this?

machow:
Yeah, looking now (I might have to pick this back up Monday though)

machow:
Ok, thanks so much for working on this @rodboev! One big piece is that Posit Assistant only charges a premium when the provider is Posit AI. In other words, if you do the BYO key approach with anthropic, no premium is charged. It's only when you use the Posit AI credit service. (Example sessions attached below)

A smaller piece is whether overriding the custom pricing rates should apply a premium on top of the custom rate. I suspect that only the custom rate should be used (no premium), since currently that's the only override available. (This might be a decision point though).

In review, I saw whether https://github.com/kenn-io/agentsview/pull/1541#issuecomment-5440950910 is still required flagged, but have not dug further into it.

## Example sessions: Posit AI billing vs bring-your-own key

[pa-provider-demo-sessions.tar.gz](https://github.com/user-attachments/files/31576126/pa-provider-demo-sessions.tar.gz)


To make the provider-identity question concrete, here are three minimal Posit Assistant
sessions (attached as `pa-provider-demo-sessions.tar.gz`, layout
`{workspace}/{conversationId}/…` as read by the PA parser). All three bill the **same
model** (`claude-sonnet-4-5-20250929`) with the **same token usage**
(100 input / 50 output / 500 cache-read / 1000 cache-write). The only distinguishing
signal is `providerId` inside the `providerMetadata.positai` envelope — the same object
the parser already reads `modelId` and `usage` from:

```json
{"id":1,"message":{"role":"assistant","content":[{"type":"text","text":"Same response"}],
  "providerOptions":{"providerMetadata":{"positai":{
    "timestamp":1768507210000,
    "modelId":"claude-sonnet-4-5-20250929","modelName":"Claude Sonnet 4.5",
    "providerId":"anthropic",   // ← "positai" in the Posit AI session, else the BYO provider
    "usage":{"inputTokens":100,"outputTokens":50,"totalTokens":1650,
             "cacheReadTokens":500,"cacheWriteTokens":1000}}}}}}
```

| Session | `providerId` | Who bills the user | PA's own accounting | agentsview today | agentsview under this PR |
|---|---|---|---|---|---|
| `aaaaaaaa-…` | `anthropic` | Anthropic directly (BYO key, no Posit premium) | $0.004950 | $0.004950 ✓ | $0.005445 (+10%, over-billed) |
| `bbbbbbbb-…` | `positai` | Posit AI credits (10% premium applies) | $0.005445 | $0.004950 (−9.1%) | $0.005445 ✓ |
| `cccccccc-…` | `positai` + sidecar | Posit AI credits, incl. auxiliary spend | $0.006633 | $0.004950 | $0.005445 |

(Sonnet 4.5 list rates: $3/$15 per MTok, $0.30 cache-read, $3.75 cache-write →
$0.004950 at list, × 1.1 = $0.005445 through Posit AI. Verified by syncing the pair
into a clean archive: current main prices both at $0.004950 — it has no signal to
tell them apart, since the parser doesn't read `providerId` and the usage facts
carry no provider dimension.)

Both routings are common in real PA usage, and a single session store mixes them —
Posit AI serves the same Anthropic model IDs a BYO key does, so the model name can't
substitute. Keying the premium on the session's agent identity therefore gets exactly
one of the first two rows right whichever way it goes; only per-request `providerId`
resolves both.

The third session adds PA's `usage-events.jsonl` sidecar (keepalive + classifier
spend, currently unparsed — tracked separately). Each sidecar event carries its own
`providerId`:

```json
{"type":"usage","kind":"keepalive","timestamp":1768507220000,"anchorMessageId":"node-1",
 "providerId":"positai","modelId":"claude-sonnet-4-5-20250929",
 "inputTokens":100,"outputTokens":2,"totalTokens":642,
 "cacheReadTokens":500,"cacheWriteTokens":40,"cacheWrite5mTokens":40}
{"type":"usage","kind":"classifier","timestamp":1768507230000,"anchorMessageId":"node-1",
 "providerId":"positai","modelId":"claude-haiku-4-5-20251001",
 "inputTokens":400,"outputTokens":10,"totalTokens":410,
 "cacheReadTokens":0,"cacheWriteTokens":0}
```

So the provider signal is pervasive across PA's usage surfaces (message metadata and
sidecar events), and plumbing `providerId` through the parser is the shared enabler
for both this premium and the sidecar ingestion.

rodboev:
Reworked this around the row-level provider ID. `positai` gets the 11/10 adjustment for billed model rates; `anthropic`, unknown or empty providers, and custom rates stay at base. Reported amounts remain authoritative, while cache savings for reported rows use the resolved billed rates. Provider identity now survives parsing, sidecar events, SQLite, PostgreSQL, DuckDB, rollups, activity, and session views. I also bumped the usage, activity, and session export schema versions and refreshed the docs and goldens.

wesm:
taking a look

wesm:
Thanks — the mechanism here looks sound. The 11/10 policy matches Posit's published FAQ ("Credit usage is calculated at a 10% premium over model provider rates"), the premium is scoped correctly in all three stores (reported costs stay authoritative, custom pricing rows are exempt, web-search fees stay at face value), and the invalidation story (data version 94, usage-cache format 6, DuckDB schema 13, export schema bumps with renamed goldens) is coherent. The focused billing, provider-scoping, and parser tests pass locally.

Three findings. The first two should be resolved before merge; the third is cleanup and not a correctness blocker.

**1. `docs/token-usage.md` describes the earlier agent-based design**

The "Posit Assistant Billing" section (`docs/token-usage.md:412`) says the premium applies "for sessions whose stored agent is exactly `posit-assistant`" and that BYOP usage cannot be distinguished. The implementation applies the policy only when the row-level provider ID is exactly `positai` (`internal/pricing/billing.go:26`), and the mixed-provider test confirms `anthropic` rows in the same session stay at base rates (`internal/db/usage_cache_daily_test.go:109`). Suggested replacement for the section body:

> For usage rows whose stored provider ID is exactly `positai`, AgentsView estimates computed token costs and cache savings at 110% of the selected catalog rates. Model, timestamp, and request-band selection happen before the adjustment. Rows with empty or other provider IDs — including observed `anthropic` BYOP rows — use base rates.
>
> Custom pricing overrides are not adjusted. Explicit reported costs remain authoritative, and fixed web-search fees remain at their published face value.

**2. Expand the Posit Assistant provenance entry**

This change alters Posit Assistant usage/cost accounting, so the `docs/internal/session-format-sources.md` entry needs updating in the same change (see the Provider Format Provenance rule in `CLAUDE.md`). The sidecar example there already shows `providerId`; what's missing:

- `providerOptions.providerMetadata.positai.providerId` in `lm-messages.jsonl` (consumed at `internal/parser/posit_assistant_provider.go:1121`).
- The observed values (`positai`, `anthropic`) and how AgentsView interprets them (managed vs. bring-your-own-provider).
- The exact-match billing behavior and the 10%-premium source (https://docs.posit.co/posit-ai/user/faq/).
- Data version 94 and its provider-identity reparse (`internal/db/db.go:440`).

Please keep the evidence classes distinct: the FAQ substantiates the 10% credit premium, while the serialized provider values are established by observed session artifacts — the FAQ does not itself define the meaning of `providerId`.

**3. Remove the abandoned agent plumbing (non-blocking)**

- `usagePriceInput.Agent` (`internal/db/usage_rollup_build.go:19`) is populated at three call sites but never read by `priceUsageFact`.
- `duckSessionUsageRow.agent` (`internal/duckdb/analytics_usage.go:3785`) is selected and scanned but never used by the breakdown pricing or rendering functions.

Both look like remnants of the earlier agent-keyed iteration. Removing just those fields and the columns/arguments that feed them is enough — the other agent fields in the rollup and DuckDB aggregation paths are still needed for attribution, filtering, and grouping. No deletion-specific test needed; compilation plus the existing behavior tests cover it.

PR #1544, opened 2026-08-27
Add hosted raw watch daemon

salmonumbrella:
## What changed

- Adds `raw-sync watch` and `raw-sync status` for continuous laptop-side raw
  capture and upload.
- Captures SQLite-backed providers through bounded online backups and file
  providers through ordered snapshots, append generations, and tombstones.
- Persists device-fenced checkpoints, durable retries, receipts, coverage
  state, and bounded outbox usage without storing parsed chat data.
- Reconciles missed watcher events with rotating audits while keeping upload
  and capture work serialized.

## Why

This advances the laptop watcher and uploader stage in #1352. Hosted raw mode
now has a durable local daemon path that can survive offline periods, watcher
gaps, source replacement, transport conflicts, and process restarts without
advancing an unsafe checkpoint.

## Limits

Remote transport requires HTTPS. Loopback HTTP is available only through the
explicit development flag. S3 roots stay out of this laptop filesystem watcher
and continue through their existing ingestion path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
         with:
@@ -195,6 +195,12 @@ jobs:
           )
           $packages = $packages | Where-Object { $_ -notin $isolatedPackages }
           go test -tags "fts5" $packages -count=1 -timeout=20m
+        env:
+          CGO_ENABLED: "1"
+
+      - name: Run raw capture tests (macOS)
+        if: runner.os == 'macOS'
+        run: go test -tags "fts5" ./internal/rawcapture -count=1 -timeout=10m
         env:
           CGO_ENABLED: "1"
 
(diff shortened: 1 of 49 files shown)


salmonumbrella:
@wesm ready for you

wesm:
Thanks, taking a look

wesm:
Thanks for building this. The direction is strong, but the current revision is not ready to merge. The checkpoint and durability design should stay.

Before merge, this PR needs to:

- Fix SQLite capture when the configured path contains symlinked parent directories. This currently fails on macOS and can cause repeated failed audits.
- Restore the normal Codex sync behavior. A changed-file event must not scan the full archive or silently switch to a different duplicate. Any new duplicate-selection policy belongs in a separate PR.
- Fix the upload queue so a transient failure followed by a server head conflict can recover without blocking every newer upload.
- Give SQLite backups one bounded BUSY/LOCKED policy on all supported drivers, with a time limit and repeated size checks.
- Use one raw-discovery implementation for collected and streaming results, with consistent S3 exclusion and bounded local/UTC Codex lookup.
- Keep raw-audit pacing out of the shared discovery contract.
- Remove the unrelated .roborev.toml change, unsupported non-cgo variants, dead compatibility exports, and inert audit code.
- Add focused regression tests, macOS capture coverage, and update the raw-sync documentation.

The private SQLite handle check can remain for now if its supported build boundary is explicit and tested. After these changes and the full Go checks pass, the PR should be ready for final review.

wesm:
invalid. Merging

PR #1547, opened 2026-08-28
fix(sync): report full resync finalization work

wesm:
Large full resyncs could display a completed session counter for minutes while database writes, source-state persistence, relationship repair, memory release, database-backed provider checks, final linking, and skip-cache persistence were still active.

This adds a bulk-only finalization phase with named steps and zero counters. Normal syncs keep their existing counted progress, while database-backed work keeps its accounting for final totals without replacing the finalization status. The command line, desktop startup status, status bar, and resync modal now render the named work; the modal hides its counter bar for uncounted phases.

Operation order, transactions, cancellation behavior, memory policy, and the database swap remain unchanged.


wesm:
tested this and it looks good

PR #1548, opened 2026-08-28
feat(pricing): add OrcaRouter as an online pricing source

clementguarino06510-glitch:
Add OrcaRouter's model catalog as a first-class pricing source, mirroring the existing OpenRouter integration. AgentsView already merges LiteLLM, OpenRouter, and the embedded fallback into the `model_pricing` table; OrcaRouter joins that set so sessions that route through OrcaRouter get accurate per-token costs. OrcaRouter publishes the same OpenRouter-shaped catalog envelope, so the fetch and parse paths are shared, and its rows are reconciled into the same marketplace ownership sentinel to keep the resolver free of duplicate spellings.

Like OpenRouter, OrcaRouter is an OpenAI-compatible AI gateway exposing a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance on the same endpoint, plus gateway-level zero-trust security that screens every prompt/response and governs tool calls on a default-deny basis with no application changes. [OrcaRouter](https://www.orcarouter.ai)

Verified: `go test ./internal/pricing/... ./internal/pricingrefresh/... ./internal/db/... ./internal/export/...` all pass, and a live fetch of `https://api.orcarouter.ai/v1/models` returns 167 priced models through the shared parser.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/docs/internal/pricing-source-survey-2026-08-25.md
+++ b/docs/internal/pricing-source-survey-2026-08-25.md
@@ -229,6 +229,18 @@ without a redistribution decision. Neither the API reference nor the official
 to the catalog. That is not proof that redistribution is forbidden; it means
 this survey found no affirmative open license comparable with MIT or CC BY 4.0.
 
+## OrcaRouter
+
+The unauthenticated live
+[`GET /v1/models`](https://api.orcarouter.ai/v1/models) returned 204 models on
+2026-08-28. It uses the same envelope and price fields as OpenRouter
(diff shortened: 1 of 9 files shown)


mariusvniekerk:
closing as advertising.

PR #1554, opened 2026-08-29
Add DSH-style session timing overview

iiiusky:
Adds a full-width session timing overview above every open transcript, modeled on DSH's Trajectory view. Input, model, and tool spans share an interactive timeline with duration/sequence projections, range focus, zoom and pan, transcript jumps, local search, and progressive history loading.

Tool spans prefer recorded execution start/completion events. Missing per-call timing remains visibly approximate rather than fabricated, while model spans use the timestamps available across supported session formats.

The toolbar also connects Turns to focused transcript mode and Calls to tool-detail visibility. The old right-column activity lane and its now-unused frontend fetch/store path are removed; Session Vital Signs remains available for aggregate and per-call analysis.

Primary review points are `TimingOverview.svelte`, `timingOverview.ts`, and the `MessageList.svelte` integration.


mariusvniekerk:
Thanks for the submission.

@cpcloud This seems similar to some of the work you did in the sidebar.

I'm not entirely sure if we want this one.

PR #1560, opened 2026-08-31
perf(sync): skip unchanged OpenCode sessions within a changed container

### LINKED ISSUE #1557: perf(sync): skip unchanged OpenCode sessions within a changed container
Follows up on #1208. The scheduling fixes in #1307, #1319, and #1355 resolved the continuous 100% CPU loop from poll starvation (idle CPU now reaches 0%). Residual 200-700% CPU spikes during active agent streaming remain (#1208, berendeboer Aug 19 comment).

### Root cause

The SQLite container gate (`opencode_container_gate.go`) is binary: container changed or not. During active streaming the WAL genuinely grows, so the gate correctly identifies the container as changed. When it does, the engine fans the event to every session via `sqliteSources` → `ListOpenCodeSessionMeta` → `SELECT id, time_updated FROM session` (all rows), and each session proceeds through the full parse path: `parseOpenCodeDBSession` opens the DB, queries session + message + part tables, builds the model, computes a content fingerprint. `dropUnchangedSharedSQLiteResults` then filters unchanged sessions post-parse by fingerprint comparison, but the parse work has already been done.

### Measured impact

On a live 1.2 GB OpenCode database with 123 sessions, sampling `time_updated` at 5-second intervals (matching the dispatch floor) across three consecutive cycles:

```
Interval 1: 1 changed, 122 unchanged (99%)
Interval 2: 1 changed, 122 unchanged (99%)
Interval 3: 1 changed, 122 unchanged (99%)
```

Different session IDs changed across intervals (2 distinct active sessions), confirming the signal isn't aliased. Every cycle, 122 sessions are fully parsed and discarded.

### Proposal: session-level time_updated pre-filter

After the container gate says "changed," compare per-session `time_updated` values from `ListOpenCodeSessionMeta` (already a lightweight index-only query) against a cached map. Only send sessions with a changed `time_updated` to the parse path.

The container gate comment at `opencode_container_gate.go:14` notes that a message or part row can change without bumping `time_updated`. This is real at the schema level (the test seeder confirms it). A periodic full-verification escape handles it: every Nth cycle (e.g. every 5th, roughly every 25 seconds at the 5s dispatch floor), bypass the pre-filter and verify all sessions by content as today. This bounds the staleness window while keeping the common path fast.

The active session's `time_updated` is being bumped during streaming (new messages arriving), so the fast path covers the dominant case.

rodboev:
Active OpenCode streaming advances the shared SQLite container's write markers on each dispatch interval. Discovery then runs the full `ForEachOpenCodeSessionMeta` query, which aggregates every `message` and `part` row to build per-session child identities, even when one session changed and 122 did not. The measured 1.2 GB production container repeats that archive-sized scan every five-second cycle.

This widens the container listing predicate so a recently digest-verified container uses `ForEachOpenCodeSessionWatermarkMeta` (session and project rows only, no child aggregation) while its verification age remains within a five-minute interval. Changed members resolve their composite fingerprint through the indexed per-session path. The reconciliation spool carries a compact discovery-state payload so watermark fidelity survives source rehydration across pages. Full digest verification runs on process start, after the interval expires, and when a container's file identity changes; it is stamped only after a complete successful pass.

Force parsing, resync, cancellation, failed passes, container replacement, legacy schemas, and non-OpenCode providers retain existing behavior. No database schema, archive, or frontend change.

Closes #1557

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -710,20 +710,21 @@ add an archived or maintained mirror without replacing the original identity.
   session or project row that advances past its own stored metadata watermark
   is always a candidate, wherever other sessions' watermarks or its own child
   timestamps sit. Periodic full passes and streamed reconciliation passes over
-  a container whose captured state still matches the last fully verified pass
-  also list the watermark form (`SQLiteContainerUnchangedSinceTrust`): every
+  a container whose last complete digest verification is less than five minutes
+  old may list the watermark form (`SQLiteContainerListsWatermarkOnly`): every
   member gate-skips before fingerprinting, so the child identity scan would be
-  archive-sized work nothing reads; any write breaks that trust and the next
-  pass carries the complete digest again. Watermark-only skips additionally
+  archive-sized work nothing reads. A changed container may continue using
+  this form during that bounded interval; after the interval, the next pass
+  carries the complete digest again. Watermark-only skips additionally
   require the pass's container capture to still be valid
   (`sqliteContainerPassCaptureValid`) — a container that changes between
   listing and the recapture check resolves full per-session digests instead,
   so a concurrent child-only write cannot hide beneath an unchanged metadata
   watermark. The trade is explicit: any child-only write that leaves the
   session and project rows untouched — wherever its timestamps land relative
-  to the stored composite — is invisible to a watcher pass and is reconciled
-  by the next full-discovery pass over the now-untrusted container, whose
-  digest still catches it; on the production container above, 96% of sessions
+  to the stored composite — is invisible to watermark-only discovery and is
+  reconciled by the next full digest pass, at most five minutes after the last
+  successful digest verification; on the production container above, 96% of sessions
   carry a session/project timestamp at or above every child, and actively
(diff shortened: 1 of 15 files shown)


rodboev:
The scenario the last review constructs is real only in the safe direction, so holding here.

Mid-pass, verification can only lapse, never renew (the stamp is written at pass finish and passes are serialized), so in any mixed pass the full digest listing runs last and re-lists the entire container, covering everything the watermark listing saw.

The stamp also requires every discovered source to complete, and finalization re-checks each digest-listed container against its pre-discovery capture, so a container that changed during the pass loses the stamp. The refresh never claims more than a completed full listing actually verified.

Both suggested fixes would add machinery without changing observable behavior, and the uniform-fidelity gate would actually regress, since it denies an earned stamp and forces a repeat scan.

wesm:
Thanks @rodboev! Merging this

rodboev:
If https://github.com/anomalyco/opencode/issues/36893 gets fixed, there won't be a need for full-digest runs at all (currently every 5 min).

PR #1564, opened 2026-09-01
Let config.toml choose how much of each session the archive keeps

naveenspark:
Adds `archive_content` to `config.toml` with three storage modes:

- `full` keeps the complete archive and remains the default.
- `transcripts` keeps conversations and tool metadata while dropping tool inputs and outputs.
- `usage` keeps token and cost accounting, model attribution, and automation classifications without conversation text.

`AGENTSVIEW_ARCHIVE_CONTENT` can set the same policy for a separate reporting archive. To apply a stricter policy to existing data, restart the daemon and run `agentsview sync --full`. Removed content requires the original source files to recover.

Writes, archive copies, and signal recomputation apply the policy consistently. Transcript-only copies conservatively remove text when older parser rows cannot separate tool output from prose. This may remove adjacent prose; it is a storage policy, not a guarantee that ordinary conversation text contains no secrets. Restricted Codex archives omit parser checkpoints whose resumable hash state can contain source bytes.

Usage-only mode also disables insights, recall, and vector operations. Opening the writable archive clears existing local vector indexes; PostgreSQL pushes remove indexed content for owned sessions while preserving other sessions. PostgreSQL automation audits preserve classifications only when a persisted marker says the prompt evidence was discarded. Tool-result rows remain part of the transcript totals but do not count as user prompts or prompt-quality evidence across SQLite, PostgreSQL, and DuckDB.


wesm:
looking

wesm:
Declining. Merging

PR #1569, opened 2026-09-01
fix(search): exclude live sessions and bake remote skill flags

### LINKED ISSUE #1511: SKILL.md's search recipe cannot find identifiers: --fts excludes tool_input, and prescribes it exactly for identifier hunts
The generated `agentsview-finding-history` SKILL.md teaches a search recipe that silently cannot find the thing agents most often look for: **who did this, and when**. Identifiers — issue/bead ids, file paths, commands, error strings — live in tool calls, and the recipe the skill teaches excludes tool calls.

Observed on `agentsview v0.40.1 (commit 9ef48912)`; `v0.41.1` release notes touch nothing in this area.

## 1. `--fts` and `--in` are mutually exclusive, and SKILL.md never says so

```
$ agentsview session search "<id>" --fts --in tool_input
fatal: --fts searches messages only; drop --in or --fts
```

This is client-side argument validation, so it applies to every install, local or `--server`.

SKILL.md's Core Workflow says to use `--hybrid` for concept questions and **`--fts` for "exact identifiers (paths, error strings, tool names)"** — precisely the case where `--fts` is the wrong mode, because those strings are usually in `tool_input`/`tool_result`, which `--fts` cannot reach. Plain search (neither `--fts` nor `--hybrid`) does reach them and is never mentioned as a mode at all.

The Rationalization Table reinforces it: *"Tool output matches are just as good." → "Start in messages for concepts"*. For an identifier hunt that advice inverts the correct order.

### Measured

Hunting the session that created a tracker id, same daemon, same query:

| Mode | Hits in the session that actually created it |
| --- | --- |
| `--fts` (what SKILL.md prescribes) | **0** — only the *currently running* session's chatter came back |
| plain + `--in tool_input` | **19** (plus 3 in a sibling session) |

The causer session is invisible on the prescribed path. Nothing errors; you get a smaller answer that reads like a finding. That is the part worth fixing — a wrong mode here does not fail loudly.

**Suggested fix:** give SKILL.md an explicit mode table, e.g.

| Hunting | Mode |
| --- | --- |
| prose, reasoning, "why did we…" | `--fts` (messages only) |
| identifiers: ids, paths, commands, error strings | plain search + `--in tool_input,tool_result` |

and drop `--fts` from the "exact identifiers" line in Core Workflow.

## 2. No way to exclude the running session

The archive ingests the live session, so for anything the current conversation just discussed, its own echoes are the freshest matches and outrank real history. With the skill's suggested `--limit 8` they can fill the page.

SKILL.md already tells the agent to *"down-rank active-session echoes"*, but there is no mechanism: `session search` has no `--exclude-session`. The workaround is `--limit 60` plus filtering the session id in post-processing, which the skill does not mention.

**Request:** `--exclude-session <id>` (repeatable), or `--exclude-current-session`.

## 3. Remote-daemon installs: every generated example is wrong

Where the archive is a shared remote daemon, every SKILL.md example is missing `--server` / `--server-token-file` and therefore silently queries the local SQLite index instead — under-reporting without any signal. There is no way to make the configured server the default, so the only fix is a hand-maintained sidecar file that says "override every example above", plus a post-`skills install` hook to inject a pointer to it. That works, but it means `skills install --force` regenerates guidance that is wrong for the deployment on every upgrade.

**Request:** let `skills install` bake a configured server (or `AGENTSVIEW_SERVER` / a config key) into the generated examples, or emit them server-agnostic.

---

Happy to send a PR for (1) if you point me at where the skill text is generated. (2) and (3) are your call on shape.

durandom:
## Summary

Content search can now drop specific sessions before `--limit`, and `skills install` can bake `--server` / `--server-token-file` into the finding-history examples.

Identifier hunts no longer have to lose the real history to the live conversation, and remote-daemon installs no longer teach commands that silently query local SQLite.

`--exclude-session` is repeatable on the CLI and `exclude_session` on HTTP; SQLite, PostgreSQL, and DuckDB apply it in SQL. For substring, regex, and FTS the excluded rows are gone before `LIMIT`, so they cannot consume the page. Semantic and hybrid still filter after the vector overfetch, so an excluded session can occupy top-k slots there.

`skills install --server` records the remote in an `# install-remote:` comment so a flagless reinstall stays current. Precedence is explicit flags, then the installed file, then `AGENTSVIEW_SKILLS_SERVER` / `AGENTSVIEW_SKILLS_SERVER_TOKEN_FILE` — an installed file decides even when it bakes no remote, so an exported variable cannot report existing skills as stale. `--server ""` un-bakes a remote. The variables are skills-only and named apart from `AGENTSVIEW_SERVER_TOKEN` so they cannot be read as a default read path for `session` commands, which have none.

Reviewers should look at two places. The `exclude_session` query tag in `internal/server/huma_routes_search.go`: Huma disables query-param explode by default and then splits a single value on commas, while the HTTP backend sends one param per ID, so without the `explode` tag every ID after the first was silently dropped on remote searches. And `skillRemoteSource.resolve` in `cmd/agentsview/skills.go`, which carries the precedence rule described above.

Baked values are shell-quoted, since the examples are meant to be run verbatim, and control characters are rejected because they would break the single-line `install-remote` comment.

Closes #1511


wesm:
looking

PR #1573, opened 2026-09-01
Retire inactive usage-cache generations safely

### LINKED ISSUE #1572: Retire inactive usage-cache generations safely
Usage-cache format or archive database-ID changes leave obsolete sibling databases on disk indefinitely because SQLite transaction locks cannot prove that another process has released an idle handle.

Introduce a format-gated cross-process retirement lease. Lease-aware openers hold a shared lease for the lifetime of the SQLite pool. Retirement takes the exclusive lease without waiting, revalidates application ID, cache kind, retirement protocol, format, source database ID, and exact generation path, then removes the database and its WAL/SHM sidecars. Preserve the current generation, active generations, pre-protocol generations, and any mismatched or unrecognized file. Keep the lease inode so racing openers remain coordinated.

naveenspark:
## What changed

- Adds a version 8 lease that every open usage-cache connection holds for its full lifetime.
- Retires an old version 8 cache only after an exclusive lease succeeds and the cache identity matches the expected source database, format, protocol, and filename.
- Removes the retired database and its SQLite sidecars. The current cache, active caches, older pre-protocol caches, caches written by a newer format, symlinks, and unknown files remain untouched. Skipping newer formats keeps a downgraded or older side-by-side binary from forcing the newer cache to rebuild.
- Closes every lease handle after success, contention, and errors.
- Removes the unused cache manager Reset method so no untested path can delete persistent cache files.
- Lets the first cache build finish while active session files keep changing. It publishes the stable snapshot and leaves changing sessions to the existing incremental updater.

## Why

Old cache generations could consume hundreds of megabytes after upgrades. Live testing with 40,951 sessions found a second failure: a continuously changing session could invalidate every publication attempt, so the 28-day usage query timed out.

Closes #1572


wesm:
looking

PR #1578, opened 2026-09-02
fix(codex): resolve forks of turnless parents instead of retrying forever

### LINKED ISSUE #1582: Codex fork of a turnless parent is marked NeedsRetry forever
### agentsview version

v0.41.1 (commit 55646d64); still present on `main` at d0395b50.

### Install method

Built from source

### OS / platform

Linux x86_64 (Debian-based)

### Which agent and version

Codex Desktop 0.147.0 (source `vscode`)

### Which model(s)

gpt-5.6-sol; the bug is model-independent

## What happened

Codex Desktop writes a rollout the moment a thread opens. Forking before the first prompt therefore produces a parent rollout that holds only `session_meta` and a `thread_settings_applied` event, with no `turn_context` at all. A chain of such forks (six sessions, three of them empty, one 221 MB) is sitting in a local archive.

`codexParentResolution` returns `resolved && len(turnIDs) > 0`, so a parent that exists and is fully readable but has no turn ids is treated exactly like a parent that cannot be found: the child is stamped `DataVersionNeedsRetry` with reason "codex parent turns unresolved" on every parse, and never reaches the current data version. Each pass re-parses the 221 MB rollout in full.

Before #1490 the retry also counted as a hard failure in `collectAndBatch`, which aborted every reconciliation page behind it; the archive logged `watch root reconciliation failed processing page: 6 failures` 7,977 times in one day. #1490 stops the abort from spreading, but the six forks stay permanently deferred and their fork lineage (`parent_session_id`) is never recorded, because `linkSubagentSessions` is gated on a clean pass.

This is distinct from #1476, whose parent transcript was genuinely missing. Here every parent is present.

## Expected behavior

A parent that was read but never opened a turn has nothing to replay, so the fork should parse as current unless the child itself carries the copied parent `session_meta` (the positive sign that replayed history exists and the parent file is simply behind). Only an unreadable parent should defer the child.

Parent rollout (complete file, two lines):

```text
{"type":"session_meta","payload":{"id":"<parent-id>","forked_from_id":null,"cli_version":"0.147.0","originator":"Codex Desktop", ...}}
{"type":"event_msg","payload":{"type":"thread_settings_applied", ...}}
```

Child rollout head:

```text
{"type":"session_meta","payload":{"id":"<child-id>","forked_from_id":"<parent-id>", ...}}
{"type":"event_msg","payload":{"type":"thread_settings_applied", ...}}
{"type":"turn_context","payload":{"turn_id":"<child-turn>", ...}}
```

### Steps to reproduce

1. In Codex Desktop, open a new thread and fork it before sending a prompt; send prompts in the fork.
2. Sync the Codex root.
3. Observe the fork stays below the current data version with `RetryReason` "codex parent turns unresolved" on every pass, while `doctor sync` shows the parent present.

Fix in #1578.

### Checklist

dqtz5vpvj9-create:
Codex Desktop writes a rollout the moment a thread opens, so a fork taken before the first prompt names a parent that holds only `session_meta` and settings events. The fork gate treated a parent with no turn ids the same way as a parent it could not find: the child was stamped `DataVersionNeedsRetry`, re-parsed on every pass, and never reached the current data version. Because a retry counts as a failure in `collectAndBatch`, one such fork on a reconciliation page aborted that page and every page behind it. A local archive carried six of these forks in one chain, every parent present and readable, and logged the identical abort every fifteen seconds for a day; the deferred-source work in #1490 stops the abort from spreading but leaves those forks permanently deferred.

`codexParentResolution` now treats any parent that can be read as resolved, whether or not it holds turn ids. A turnless parent never opened a turn, so the fork replayed nothing and the gate has nothing to drop; with an empty parent turn set the gate suppresses nothing either way, so parsing the child as current is the only change. A parent that cannot be read still defers the child, exactly as before.

An earlier revision of this PR kept a turnless parent blocking when the child carried a copied parent `session_meta`, on the theory that a replay prefix always begins with that copied meta. A survey of a large local Codex archive disproved it: about half of the forks there, including many written by Codex 0.147.0, replay parent turns with no copied parent meta at all. That gate and its test are gone, which is why the net diff is small.

Reviewers should look at `codexParentResolution` in `internal/parser/codex.go` and `TestCodexProviderTurnlessParentResolvesWithoutRetry`, which parses a fork of a meta-only parent and expects a current result with only the child's messages.

Closes #1582

https://claude.ai/code/session_0158DH4fPMj7ogzQdPmyRqUZ

### DIFF
--- a/internal/parser/codex.go
+++ b/internal/parser/codex.go
@@ -212,11 +212,6 @@ func (b *codexSessionBuilder) armForkGate(payload gjson.Result) {
 		b.parentTurnIDs, resolved = b.resolveParentTurns(parentID)
 	}
 	b.forkGate.armFromMeta(payload, resolved)
-	if resolved && len(b.parentTurnIDs) == 0 {
-		// The parent exists but holds no turns, so the fork replayed no
-		// history: every record in this rollout is the child's own.
-		b.forkGate.active = false
-	}
 }
 
 func (b *codexSessionBuilder) suppresses(
@@ -1670,16 +1665,21 @@ func (p *codexProvider) parentTurnResolver(
 func (p *codexProvider) codexParentResolution(
 	ctx context.Context, childPath string,
 ) (string, bool) {
-	parentID, resolutionNeeded := codexReplayParentIDContext(ctx, childPath)
-	if parentID == "" || !resolutionNeeded {
+	scan := codexReplayParentScan(ctx, childPath)
+	if scan.parentID == "" || !scan.resolutionNeeded {
 		return "", false
 	}
-	// A parent that was read but never opened a turn (Codex Desktop writes
-	// a rollout on thread open, and a fork taken before the first prompt
-	// copies nothing) is resolved: there is no replayed history to drop,
-	// and nothing a retry could learn later.
-	_, resolved := p.parentTurnResolver(ctx, childPath)(parentID)
-	return parentID, resolved
+	turnIDs, resolved := p.parentTurnResolver(ctx, childPath)(scan.parentID)
+	if !resolved {
+		return scan.parentID, false
+	}
+	// A readable parent with no turn ids is settled unless the child holds
+	// the copied parent session_meta. Codex Desktop writes a rollout when a
+	// thread opens, so a fork taken before the first prompt names such a
+	// parent and replayed nothing; the child is current. The copied meta is
+	// the one case where the parent wrote turns the fork copied but has not
+	// flushed to its own file yet, and only then does the child wait.
+	return scan.parentID, len(turnIDs) > 0 || !scan.copiedParentMeta
 }
 
 // CodexReplayParentID returns the explicit parent only when the rollout has a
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
The medium finding on `7680c46` (`internal/parser/codex.go:1677`, readable parent with no turn ids may be a stale snapshot) is **invalid**. Its failure path needs a fork to replay parent turns that the parent's rollout file does not yet contain, and Codex forks read their history from that file.

From upstream `openai/codex` `main`:

- `fork_thread` in `codex-rs/core/src/thread_manager.rs` loads history through `initial_history_from_rollout_path`, which reads the thread by rollout path with `include_history: true`.
- `attach_history_if_requested` in `codex-rs/thread-store/src/local/read_thread.rs` fills that history with `load_history_items(&path)` from the parent rollout file on disk.
- The subagent fork path calls `ensure_rollout_materialized` and `flush_rollout` on the parent first ("Persist queued rollout updates before reading the fork snapshot").

So the replayed prefix is a copy of what was already on disk, and rollout files are append-only. A parent with zero `turn_context` lines at fork time produces a child with zero replayed turns; nothing can later be duplicated.

Two more points:

- A survey of a local archive of 12,343 Codex rollouts (versions 0.133 through 0.152) found 73 turnless rollouts, none with a user message, and 303 forks, none with a turnless parent. No observed instance of the described state.
- The prior `len(turnIDs) > 0` check was never a completeness check. A parent with one flushed turn out of five was already "resolved" and only that turn was dropped. The zero-turn case is not special, and deferring it again would only restore the forever-retry this PR removes.

Caveat: the source read is current `main`; older tags were not checked. The archive survey covers those versions and shows no counterexample.

PR #1594, opened 2026-09-02
fix(parser): preserve Antigravity CLI effort for experimental serving variants

### LINKED ISSUE #1593: Antigravity CLI parser splits Gemini 3.7 Flash usage between gemini-3.7-flash-high and gemini-3.7-flash-exp-b
### agentsview version
`agentsview v0.42.0 (commit ff8fb4e8, built 2026-09-01T19:37:06Z)`

### Install method
Homebrew cask

### OS / platform
macOS 27.0 (arm64)

### Which agent and version
Antigravity CLI (agy) 1.1.24

### Which model(s)
`gemini-3.7-flash-high` / `gemini-3.7-flash-exp-b`

When using Antigravity CLI (`agy`) with `Gemini 3.7 Flash (High)`, AgentsView usage analytics splits the usage between two separate model rows:
1. `gemini-3.7-flash-high`
2. `gemini-3.7-flash-exp-b`

Expected behavior:
All usage events and step models should be unified under `gemini-3.7-flash-high`.

### Root Cause
In `internal/parser/antigravity.go` (added in #1461):
```go
func resolveAntigravityGenerationModel(data []byte, executorModel string) string {
    generationModel, hasDisplayLabel := extractAntigravityGenerationModel(data)
    if hasDisplayLabel || executorModel == "" {
        return generationModel
    }
    if antigravityBaseModel(executorModel) == generationModel {
        return executorModel
    }
    return generationModel
}

func antigravityBaseModel(model string) string {
    for _, suffix := range []string{"-low", "-medium", "-high"} {
        if base, ok := strings.CutSuffix(model, suffix); ok {
            return base
        }
    }
    return model
}
```
- In recent Antigravity CLI versions, Google's Cloud Code backend returns `gemini-3.7-flash-exp-b` in `gen_metadata` field 19 (an internal experimental serving variant identifier), while `executor_metadata` field 28 contains `gemini-3.7-flash-high`.
- `antigravityBaseModel("gemini-3.7-flash-high")` yields `"gemini-3.7-flash"`.
- Because `"gemini-3.7-flash"` != `"gemini-3.7-flash-exp-b"`, `resolveAntigravityGenerationModel` fails to match the base model slug. Consequently, the executor model is not applied, and the raw wire identifier `gemini-3.7-flash-exp-b` leaks into the usage events.

### Suggested Fix
Update `antigravityBaseModel` or normalize `generationModel` to handle `-exp` / `-exp-b` variant suffixes, e.g.:
```go
func antigravityBaseModel(model string) string {
    for _, suffix := range []string{"-low", "-medium", "-high", "-exp-b", "-exp"} {
        if base, ok := strings.CutSuffix(model, suffix); ok {
            return base
        }
    }
    return model
}
```
And check if `antigravityBaseModel(executorModel) == antigravityBaseModel(generationModel)`.

### Steps to reproduce
1. Use Antigravity CLI (`agy`) with model `gemini-3.7-flash-high`.
2. Run a few prompts and open `agentsview`.
3. Check the Models / Usage dashboard: observe both `gemini-3.7-flash-high` and `gemini-3.7-flash-exp-b` listed.

### Checklist

BUKOWSKIREAL:
Antigravity CLI 1.1.24 can store an experimental serving identifier such as `gemini-3.7-flash-exp-b` in generation metadata while the covering executor metadata records the effort-qualified product model, such as `gemini-3.7-flash-high`. The parser previously required the generation value to equal the executor's base slug, so the serving identifier leaked into usage events and split one model's dashboard accounting.

This change applies the executor range to both SQLite `gen_metadata` and preferred `.trajectory.json` sidecar generations. It recognizes only the observed `-exp-b` serving suffix when matching generation metadata to a covering executor range. Complete display labels and explicitly effort-qualified generation models remain authoritative, unmatched or missing executors preserve the raw generation model, and steps outside every declared executor range are not guessed. Existing archived rows are reparsed through data version 98 so historical message and usage records receive the correction.

The Antigravity format inventory now pins the official CLI 1.1.24 release and records the observed field 19 variant. Generic `-exp` and hypothetical serving suffixes remain distinct until evidence supports normalizing them.

The shared model-resolution and executor-range behavior is in `internal/parser/antigravity.go`; sidecar attribution is in `internal/parser/antigravity_cli.go`; regression coverage is in `internal/parser/antigravity_test.go`. The archive reparse boundary is in `internal/db/db.go`, and provenance is documented in `docs/internal/session-format-sources.md`.

Closes #1593.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -1801,10 +1801,10 @@ add an archived or maintained mirror without replacing the original identity.
   specification, or authoritative protobuf schema was found. The independent
   CodeBurn evidence pinned in the `antigravity` entry also covers CLI
   discovery, live RPC metadata, and the shorter capture window, but not the
-  encrypted producer format. Reverified 2026-08-20 against Google's official
-  [Antigravity CLI 1.1.16 macOS ARM64 release](https://github.com/google-antigravity/antigravity-cli/releases/tag/1.1.16),
+  encrypted producer format. Reverified 2026-09-02 against Google's official
+  [Antigravity CLI 1.1.24 macOS ARM64 release](https://github.com/google-antigravity/antigravity-cli/releases/tag/1.1.24),
   asset SHA-256
-  `a902e8b24fdc53504e7daf658e18f7ba47ebb9cfcf058aa6c01e37c5e610d389`. Its Go
+  `189af288ed9527f567ab3a53b35a6da2fc0c3812c6245f266c75a2a3604bdec3`. Its Go
   binary embeds `FileDescriptorProto` records for
   `third_party/jetski/cortex_pb/cortex.proto` and
   `third_party/jetski/codeium_common_pb/codeium_common.proto`. These compiled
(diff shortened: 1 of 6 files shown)


BUKOWSKIREAL:
Rebased onto latest `main` (`126e07a`). Conflict resolution: #1609 (Pi skill attribution) landed first and took `dataVersion = 98`, so the Antigravity reparse boundary in this PR was renumbered **98 → 99** (both version comments kept in `internal/db/db.go`, both boundary tests kept in `db_test.go`). No code changes otherwise; build and all parser/db tests pass locally.

PR #1597, opened 2026-09-02
Use stepped area chart for Cost Over Time

### LINKED ISSUE #1567: Use stepped area chart for Cost Over Time
### Problem / motivation

The "Cost Over Time" chart under the Usage view currently uses a stacked area chart with linear interpolation between dates. The sloped lines that connect data points presentan inaccurate view of the total - the area under the line isn't mathematically proportionate to spend. 

### Proposed solution

A stepped area chart would accurately reflect usage:

https://developers.google.com/chart/interactive/docs/gallery/steppedareachart

### Alternatives considered

A stacked bar chart would also be sufficient, but might not present continuity as well as a stepped area chart.

mslynch:
Resolves #1567.

The Usage Cost Over Time visualization now renders stacked daily amounts as centered steps, rather than diagonally interpolating between dates.

It also fills dates omitted by the daily usage response with zero-valued entries across the selected range, so zero-usage days appear at the baseline and the timeline remains continuous.

`d3-shape` is now a direct frontend dependency for LayerChart's step curve.

## Screenshot
Screenshot 2026-09-02 at 10 27 41

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13,6 +13,7 @@
         "@tanstack/virtual-core": "3.17.8",
         "d3-hierarchy": "3.1.2",
         "d3-scale": "4.0.2",
+        "d3-shape": "3.2.0",
         "d3-time": "3.1.0",
         "dompurify": "3.4.14",
         "layerchart": "2.3.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,6 +23,7 @@
     "@tanstack/virtual-core": "3.17.8",
     "d3-hierarchy": "3.1.2",
     "d3-scale": "4.0.2",
+    "d3-shape": "3.2.0",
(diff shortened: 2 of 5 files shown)


mslynch:
Pushed a commit to commit to address the failing test.

Regarding the review:

> `frontend/src/lib/components/usage/CostTimeSeriesChart.svelte:97` — `summary.daily.length` and `summary.daily.map(...)` assume `daily` is always an array, but the generated API model permits `daily: null`, which can crash the usage page. Guard or normalize the response, e.g. `const daily = summary.daily ?? []`.

It doesn't look like it's possible for this to be null in practice - it's more a mark of imprecision in the generated types. I think there's a broader issue surrounding that, which I've filed here: https://github.com/kenn-io/agentsview/issues/1610

PR #1607, opened 2026-09-03
feat(rawderive): add hosted parse worker foundation

salmonumbrella:
Adds a worker library for deriving sessions from retained raw captures. Workers verify manifest and object digests, reconstruct a read-only source tree, reuse provider parsing, and hand results to an atomic projection sink. PostgreSQL leases fence job completion to the current source generation.

Replay preserves provider inputs: SQLite snapshots use immutable reads, Codex retains alias-home title precedence and replay parents across configured directories, and Claude and Evener retain their lineage companions. Growing transcripts reuse captured prefixes. Hosted skill names come from recorded paths without reading worker-local frontmatter or cached local names.

Changing sources no longer stop an audit; full audits report skipped captures as incomplete so the watcher retries. Unchanged Claude sniff checks reuse filesystem evidence, and idle job cleanup excludes future retries.

Runtime activation and PostgreSQL session projection remain pending, so hosted browsing still requires `pg push`. The implemented capture-to-parser boundary has been exercised with Claude and WAL-mode ZCode inputs using PostgreSQL job leases and a recording projection sink.

Existing least-privilege raw-sync roles need `SELECT` and `UPDATE` on `raw_ingest_jobs`. Startup names missing requirements; grants are documented in the hosted raw-sync and PostgreSQL sync guides.

Rollout to existing raw-sync deployments still needs a compatibility decision: stricter validation rejects stored case-colliding paths, and an upload committed remotely but not bound locally before upgrading can retry indefinitely with a different manifest ID after modification times are added.

Refs #1352

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -32,6 +32,23 @@ description: Release history for AgentsView
   sync to apply the new category. A machine that only runs `agentsview usage`
   needs one `agentsview sync` to pick up subagent transcripts that already
   exist.
+
+- Hosted raw-sync operators now get a startup message identifying missing
+  database privileges when raw-sync routes are disabled. Existing least-privilege
+  runtime roles need `SELECT` and `UPDATE` on `raw_ingest_jobs`, in addition to
+  their existing grants. The internal hosted parse-worker foundation adds job
+  leases, source reconstruction, parsing, and retries; server startup and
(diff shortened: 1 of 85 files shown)


salmonumbrella:
@wesm ready

salmonumbrella:
@wesm ready

wesm:
looking

PR #1615, opened 2026-09-04
fix: blank invalid token_usage instead of panicking the messages API on every backend

### LINKED ISSUE #1614: Malformed token_usage panics GET /api/v1/sessions/{id}/messages: blank transcript in UI while HTML export renders
## Environment

| Field | Value |
| --- | --- |
| agentsview version | v0.42.0 |
| OS / platform | Windows 10.0.22631.7517 (Windows 11 Enterprise, Build 22631) |
| Install method | not recorded |
| Agent | believed Claude Code — unconfirmed; the unguarded raw-usage write is in the Claude parser (see root cause) |
| Model | not recorded |

Selecting a session in the web UI (via `duckdb serve`) shows the session in the list, but the
transcript pane stays **blank**. Exporting that same session to HTML renders all
of its content correctly, so the messages are present and correctly linked in
the database.

The messages request fails at the network level — Chrome reports
`ERR_EMPTY_RESPONSE`, not an empty result set. The server log shows the handler
panicking while **encoding the response**:

```
2026/09/03 18:55:33 GET /api/v1/sessions/<redacted>/messages
2026/09/03 18:55:33 http: panic serving 127.0.0.1:61982: error marshaling response for GET /api/v1/sessions/{id}/messages 200: json: cannot marshal from Go jsontext.Value: unexpected EOF within "/messages/0/token_usage" after offset 396
goroutine 166 [running]:
net/http.(*conn).serve.func1()
    net/http/server.go:1939 +0xbb
panic({0x7ff71ae718b0?, 0x3c125c1822c0?})
    runtime/panic.go:859 +0x125
net/http.(*timeoutHandler).ServeHTTP(0x3c125c5561c0, ...)
    net/http/server.go:4061 +0x76c
go.kenn.io/agentsview/internal/server.(*Server).humaTimeout.func1.1(...)
    go.kenn.io/agentsview/internal/server/huma_routes.go:426 +0x2aa
github.com/danielgtaylor/huma/v2/wrap.func1(...)
    github.com/danielgtaylor/huma/v2@v2.39.1/chain.go:16 +0x25
go.kenn.io/agentsview/internal/server.humaRequestInfoMiddleware(...)
    go.kenn.io/agentsview/internal/server/huma_routes.go:176 +0x108
...
```

Expected: a session whose stored usage metadata is malformed should still render
its transcript. A single unusable metadata field should not take down the
endpoint.

## Root cause

One message row has a **truncated `token_usage` blob**. `jsontext.Value` defers
parsing until marshal time, so the bad value is stored without complaint and
only fails when the API encodes the response. Four things combine:

1. **`internal/parser/claude.go:2671`** stores gjson's raw slice with no validity
   check:
   ```go
   msg.TokenUsage = jsontext.Value(usageResult.Raw)
   ```
   A torn read of a JSONL line still being written yields a truncated `.Raw`.
   `IsValid()` is already used for exactly this purpose in
   `internal/parser/hermes.go:899` and `internal/parser/piebald.go:713`, but not
   here.

2. **`internal/db/validate.go`** — the centralized sanitize pass every write flows
   through explicitly excludes the field: *"Message.TokenUsage (jsontext.Value)
   and transient ToolResults.ContentRaw are intentionally not run through this
   pass."*

3. **`internal/db/messages.go:2143` and `:2748`** — the read path only checks for
   non-empty, never for validity:
   ```go
   if tokenUsage != "" { m.TokenUsage = jsontext.Value(tokenUsage) }
   ```

4. **`internal/server`** contains no `recover()`, so the marshal error becomes an
   unrecovered handler panic and `net/http` drops the connection.

HTML export escapes all of this because `humaExportSession` →
`db.GetAllMessages` → `generateExportHTML` renders server-side and never
marshals `token_usage`. That asymmetry is what makes the symptom look like a
frontend rendering bug.

The Windows correlation appears incidental to *how the row got truncated* (a
torn read during concurrent writes); the crash itself is platform-independent
and reproduces anywhere with the same data.

## Steps to reproduce

Store a truncated usage object on any message, then request that session's
messages:

```sql
-- against %USERPROFILE%\.agentsview\sessions.db
UPDATE messages
SET token_usage = '{"input_tokens":4,"cache_read_input_tokens":123'
WHERE session_id = '<some-session>' AND ordinal = 0;
```

```
GET /api/v1/sessions/<some-session>/messages   -> connection dropped, panic in log
GET /api/v1/sessions/<some-session>/export     -> renders fine
```

A Go-level reproduction, which fails on `main`:

```go
_, err := json.Marshal(struct {
    Messages []Message `json:"messages"`
}{Messages: []Message{{
    TokenUsage: jsontext.Value(`{"input_tokens":4,"cache_read_input_tokens":123`),
}}})
// json: cannot marshal from Go jsontext.Value: unexpected EOF within "/messages/0/token_usage"
```

## Workaround for affected users

```sql
-- back up sessions.db first
SELECT session_id, ordinal FROM messages
WHERE token_usage != '' AND json_valid(token_usage) = 0;

UPDATE messages SET token_usage = ''
WHERE token_usage != '' AND json_valid(token_usage) = 0;
```

Non-destructive: it blanks only unparseable metadata and touches no transcript
content.

## Suggested fix

Guarding one end is not enough. A write-side fix leaves already-corrupted
databases broken until a resync; a read-side fix alone leaves invalid blobs
being persisted and pushed to PG. Both:

- **Write seam** — have `ValidateAndSanitize` blank invalid `TokenUsage`,
  following the file's stated sanitize-don't-drop policy. Covers every agent and
  every write path (sync, upload, import) rather than only the Claude extraction
  site.
- **Read path** — drop an invalid stored value in `scanMessages` and
  `GetMessageByOrdinal`, so existing databases render without a resync.

Worth reporting the blanked count alongside the other sanitize counters
(`SanitizeStats`) so the drop is visible in the sync summary instead of silent.

One deliberate behavior change falls out of this: invalid usage no longer
rejects a session from artifact export, because the value can no longer reach
the segment encoder (`internal/db/artifact_export_load.go:118` loads through
`scanMessages`). Previously a single truncated blob cost the whole session its
export — `TestExportContinuesPastInvalidTokenUsage` asserts the old behavior and
needs updating. Export continuing past a genuinely un-encodable session stays
covered by `TestExportContinuesPastInvalidManifestValue` and
`TestExportContinuesPastDeterministicRejection`.

I have this implemented with tests at both layers plus a server-level regression
test asserting the endpoint returns 200 rather than panicking, and can open a PR
if useful.

## Checklist

tlmaloney:
Fixes #1614.

A malformed `token_usage` blob takes down `GET /api/v1/sessions/{id}/messages`
entirely. `jsontext.Value` defers parsing to marshal time, so an invalid value
is stored without complaint and only fails when the API encodes the response:

```
json: cannot marshal from Go jsontext.Value: unexpected EOF within
"/messages/0/token_usage" after offset 396
```

`internal/server` installs no `recover()`, so `net/http` closes the connection
without a response. The session list still renders, the transcript comes back
blank, and exporting the same session to HTML still shows its content because
`humaExportSession` -> `GetAllMessages` -> `generateExportHTML` never marshals
the field. That asymmetry makes it look like a frontend rendering bug.

### Scope correction vs. the issue

Issue #1614 describes this as a SQLite-path bug. It is not — **all three
backends are affected**. The reporter was running `duckdb serve`, and
`internal/duckdb/messages.go` had no guard at all (it assigned
`[]byte(tokenUsage)` unconditionally, not even checking for empty).
`internal/postgres/messages.go` checked only for non-empty. Any mirror pushed or
built before this change still reproduces the panic through `pg serve` and
`duckdb serve`.

### Changes

**Write seam** — `ValidateAndSanitize` now blanks invalid `TokenUsage`. This is
the single path every session write passes through, so it covers every agent and
every write path (sync, upload, import) rather than only the Claude extraction
site at `internal/parser/claude.go:2671`, which stores gjson's `.Raw` verbatim.
A torn read of a JSONL line still being written yields a truncated `.Raw`;
`IsValid()` is already used for exactly this purpose in `hermes.go:899` and
`piebald.go:713`. Follows the file's stated sanitize-don't-drop policy: the
message keeps its content, only the unusable metadata is cleared.

**Read path** — `db.DecodeStoredTokenUsage` is exported and used by all three
scanners (`internal/db`, `internal/postgres`, `internal/duckdb`), so databases
and mirrors populated before the write guard render without a resync, and the
backends cannot drift again. This is the backend-parity rule in AGENTS.md.

**Observability** — `TokenUsageBlanked` is reported alongside the other
sanitize counters, so the drop shows up in the sync summary instead of being
silent. Silence is what made this hard to trace.

**Generated client** — regenerated after `token_usage_blanked` was added to
`SanitizeStats`; `npm run check:api` was failing the consistency gate.

### Behavior change worth a reviewer's attention

Invalid usage no longer rejects a session from artifact export, because the
value can no longer reach the segment encoder (`artifact_export_load.go:118`
loads through `scanMessages`). Previously one truncated blob cost the whole
session its export.

`TestExportContinuesPastInvalidTokenUsage` asserted that old behavior, so it now
describes an unreachable scenario. **I rewrote it rather than deleting it** —
it is now `TestExportSanitizesInvalidTokenUsage` and asserts the new outcome.
Export continuing past a genuinely un-encodable session remains covered by
`TestExportContinuesPastInvalidManifestValue` and
`TestExportContinuesPastDeterministicRejection`, both untouched and passing.
This is the one change here that flips an existing assertion, so it deserves
the closest look.

### Tests

Regression coverage at all three backends, each seeding the store directly
because the write seam now sanitizes the value before it could reach a mirror:

- `internal/db` — sanitize, idempotency, read-path, and `DecodeStoredTokenUsage`
  contract tests
- `internal/server` — endpoint returns 200 instead of panicking
- `internal/duckdb` — corrupts the mirror after push
- `internal/postgres` — writes the row directly (`pgtest` tag)

Every one of them was confirmed to **fail with its guard reverted**, reproducing
`cannot marshal from Go jsontext.Value: unexpected EOF`, so none are tautological.

### Verification

| Check | Result |
| --- | --- |
| `go vet` | clean |
| `make test-short` | 48 packages passing |
| `make test-postgres` | passing against a real PostgreSQL instance |
| frontend `vitest` | 159 files / 2457 tests passing |
| `npm run check:api` | exit 0 |

One pre-existing failure is unrelated and untouched by this change:
`cmd/agentsview TestRunDaemonArtifactExchangeConnectsToIPv6LocalhostListener`.
It is environmental — the test listens on `[::1]` but rewrites the endpoint host
to `localhost`, and on hosts whose `/etc/hosts` maps `localhost` to `127.0.0.1`
only (the Debian/Ubuntu default, where `::1` is named `ip6-localhost`) the dial
lands where nothing is listening. Its skip guard covers IPv6 *listening* being
unavailable, not resolution. It fails identically at tag `v0.41.1`.

### Workaround for affected users on an existing install

The read-side guard means no data surgery is needed — but for anyone pinned to a
released build, blank the unparseable rows in whichever store is being served
(`sessions.db` for `serve`, `sessions.duckdb` for `duckdb serve`):

```sql
UPDATE messages SET token_usage = ''
WHERE token_usage != '' AND json_valid(token_usage) = 0;
```

https://claude.ai/code/session_016BtS7hmmCDXSXvLRDxmkFx

### DIFF
--- a/internal/db/validate.go
+++ b/internal/db/validate.go
@@ -37,9 +37,19 @@ import (
 //     blanked. Empty-string handling is preserved as-is so downstream
 //     localTime treats a blanked timestamp as invalid.
 //
-// Message.TokenUsage (jsontext.Value) and transient
-// ToolResults.ContentRaw are intentionally not run through this pass. They are
-// raw provider payloads, not persisted display text. Persisted result content
+// Message.TokenUsage (jsontext.Value) is a raw provider payload, so it is
+// not run through the text contract below -- but it IS checked for JSON
+// validity and blanked when malformed. jsontext.Value defers parsing to
+// marshal time, so an invalid blob (e.g. a truncated gjson raw slice from a
+// torn read of a JSONL line still being written) is stored without
+// complaint and only fails much later, when the API marshals it. With no
+// recover() in internal/server that became an unrecovered handler panic and
+// a dropped connection on GET /api/v1/sessions/{id}/messages, while the
+// HTML export of the same session still rendered because it never marshals
+// the field. Blanking here follows the sanitize-don't-drop policy: the
+// message keeps its content and only the unusable metadata is cleared.
+// Transient ToolResults.ContentRaw remains outside this pass. It is a
+// raw provider payload, not persisted display text. Persisted result content
 // (tool_calls.result_content and tool_result_events.content) follows the same
 // text contract as Message.Content, with length fields reduced by the
 // stripped-byte delta so re-ingest rewrites historical poison rows into the
@@ -101,6 +111,7 @@ type ValidationStats struct {
 	TokensClamped        int
 	RoleCoerced          int
 	TimestampsBlanked    int
+	TokenUsageBlanked    int
 }
 
 // add accumulates another stats value into the receiver.
@@ -110,6 +121,7 @@ func (s *ValidationStats) add(o ValidationStats) {
 	s.TokensClamped += o.TokensClamped
 	s.RoleCoerced += o.RoleCoerced
 	s.TimestampsBlanked += o.TimestampsBlanked
+	s.TokenUsageBlanked += o.TokenUsageBlanked
 }
 
 // ValidateAndSanitize applies the central validation contract in place
@@ -161,6 +173,16 @@ func SanitizeMessage(m *Message) ValidationStats {
 	if !validMessageRole(m.Role) {
 		m.Role = ""
 		stats.RoleCoerced++
+	}
+
+	// Parsers store the provider's raw usage object (e.g. gjson's .Raw in
+	// extractClaudeTokenFields) without validating it. jsontext.Value only
+	// re-parses at marshal time, so a malformed blob survives the write and
+	// detonates in the API response instead. Blank it here, at the shared
+	// write seam, so every agent and every write path is covered.
+	if len(m.TokenUsage) > 0 && !m.TokenUsage.IsValid() {
+		m.TokenUsage = nil
+		stats.TokenUsageBlanked++
 	}
 
 	// Adjust ContentLength by the DELTA of bytes the sanitizer removed
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -1833,6 +1833,7 @@ func sanitizeBreakdownLines(s sync.SanitizeStats) []string {
 		{"tokens clamped", s.TokensClamped},
 		{"role coerced", s.RoleCoerced},
 		{"timestamps blanked", s.TimestampsBlanked},
+		{"token usage blanked", s.TokenUsageBlanked},
 	}
 	var out []string
 	for _, c := range cats {
--- a/frontend/src/lib/api/generated/models/SyncSanitizeStats.ts
+++ b/frontend/src/lib/api/generated/models/SyncSanitizeStats.ts
@@ -7,6 +7,7 @@ export type SyncSanitizeStats = {
   model_clamped?: number;
   role_coerced?: number;
   timestamps_blanked?: number;
+  token_usage_blanked?: number;
   tokens_clamped?: number;
 };
 
--- a/internal/artifact/export_test.go
+++ b/internal/artifact/export_test.go
@@ -441,7 +441,16 @@ func TestExportContinuesPastDeterministicRejection(t *testing.T) {
 	assert.Contains(t, rejection.Error, "message count exceeds 2")
 }
 
-func TestExportContinuesPastInvalidTokenUsage(t *testing.T) {
+// Invalid token usage no longer costs a session its artifact export. The
+// db layer now blanks a malformed usage blob at the write seam
+// (ValidateAndSanitize) and again on read (scanMessages, which this export
+// path loads through), so the value can never reach the segment encoder.
+// Previously one truncated usage object -- the same corruption that
+// panicked GET /api/v1/sessions/{id}/messages -- rejected the whole
+// session here. Export continuing past a genuinely un-encodable session
+// stays covered by TestExportContinuesPastInvalidManifestValue and
+// TestExportContinuesPastDeterministicRejection.
+func TestExportSanitizesInvalidTokenUsage(t *testing.T) {
 	t.Parallel()
 
 	database := testExportDB(t)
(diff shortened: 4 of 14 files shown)


mariusvniekerk:
Thanks, do you have enough data to be able to do a fix on the writer side too?

tlmaloney:
> Thanks, do you have enough data to be able to do a fix on the writer side too?

I will check - I observed the issue on a Windows machine at work, which is in a corporate environment. I'm making this fix on my home machine, where I have more tools at my disposal. I agree with the goal about making the writer more robust and preventing bad data. Let me see what I can do, I'll report back.

internal/db/validate.go
```diff
@@ -163,6 +175,16 @@ func SanitizeMessage(m *Message) ValidationStats {
 		stats.RoleCoerced++
 	}
 
+	// Parsers store the provider's raw usage object (e.g. gjson's .Raw in
+	// extractClaudeTokenFields) without validating it. jsontext.Value only
+	// re-parses at marshal time, so a malformed blob survives the write and
+	// detonates in the API response instead. Blank it here, at the shared
+	// write seam, so every agent and every write path is covered.
+	if len(m.TokenUsage) > 0 && !m.TokenUsage.IsValid() {
```
tlmaloney, line 183: @mariusvniekerk Is this what you mean by the write fix?

mariusvniekerk:
@tlmaloney mostly to prevent the need for the update kinds of things above.  if the go struct types we ultimately write into the database can't have invalid data in them by design then we're best off obviously.

tlmaloney:
> @tlmaloney mostly to prevent the need for the update kinds of things above. if the go struct types we ultimately write into the database can't have invalid data in them by design then we're best off obviously.

Agreed. Sorry, I think there's a misdiagnosis and fix isn't entirely appropriate. I'm going to close this PR and related issue and reopen!

PR #1618, opened 2026-09-04
fix(duckdb): restore session transcripts under duckdb serve (empty token_usage fails response marshaling)

### LINKED ISSUE #1617: duckdb serve renders no session transcripts in v0.42.0: empty token_usage fails response marshaling
Supersedes #1614, which misdiagnosed this as a rare data-corruption bug. The
real defect is far broader: **`duckdb serve` cannot render any session
transcript in v0.42.0.** Every session, every user, every request. No corrupt
data is involved.

## Environment

| Field | Value |
| --- | --- |
| agentsview version | v0.42.0 |
| OS / platform | Windows 10.0.22631.7517 (Windows 11 Enterprise, Build 22631) — not platform-specific, see below |
| Install method | not recorded |
| Agent | n/a — affects every agent |
| Model | n/a |

## What happens

Under `agentsview duckdb serve`, the session list renders normally but selecting
any session leaves the transcript pane blank. The request fails at the network
level (`ERR_EMPTY_RESPONSE` in Chrome), and the server log shows the handler
panicking while encoding the response:

```
http: panic serving 127.0.0.1:61982: error marshaling response for
GET /api/v1/sessions/{id}/messages 200: json: cannot marshal from Go
jsontext.Value: unexpected EOF within "/messages/0/token_usage" after offset 396
```

`internal/server` installs no `recover()`, so `net/http` closes the connection
without a response.

Exporting the same session to HTML renders its content correctly, which makes
this look like a frontend rendering bug. It is not — `humaExportSession` ->
`GetAllMessages` -> `generateExportHTML` renders server-side and never marshals
`token_usage`.

`agentsview serve` (SQLite) and `pg serve` are unaffected.

## Root cause

`token_usage` is declared `TEXT NOT NULL DEFAULT ''`, so nearly every message row
holds `''`. The DuckDB scanner assigns it **unconditionally**, while the SQLite
and PostgreSQL scanners guard for the empty case:

```go
// internal/duckdb/messages.go:296  -> non-nil, ZERO-LENGTH jsontext.Value
m.TokenUsage = []byte(tokenUsage)

// internal/db/messages.go, internal/postgres/messages.go -> stays nil
if tokenUsage != "" { m.TokenUsage = ... }
```

A zero-length `jsontext.Value` is not valid JSON, and `omitempty` does **not**
skip it, so marshaling fails for the very first message of every response.

The truncated-blob theory in #1614 was wrong. A zero-length value produces a
byte-identical error message to a truncated one, which is what misled the
original diagnosis. Both stores on the reporting machine were queried and
contained **zero** rows with invalid `token_usage`:

```sql
-- 0 rows in sessions.db AND in sessions.duckdb
SELECT COUNT(*) FROM messages
WHERE token_usage <> '' AND NOT json_valid(token_usage);
```

## This is a regression, and it is datable

The DuckDB scanner line is unchanged since `ac7cef31` (#609, the original DuckDB
backend). It was harmless for two releases. What broke it is `b85eae5a`
("feat: adopt Go 1.27 JSON v2 semantics", #1475, 2026-08-21), which changed the
field type:

```go
- TokenUsage json.RawMessage `json:"token_usage,omitempty"`
+ TokenUsage jsontext.Value  `json:"token_usage,omitempty"`
```

v1's `omitempty` dropped a zero-length slice; v2's does not. Verified directly:

```
v1 json.RawMessage("") + omitempty -> [{}]           err=<nil>
v2 jsontext.Value("")  + omitempty -> err=json: cannot marshal from Go
                                          jsontext.Value: unexpected EOF
```

`b85eae5a` first shipped in **v0.42.0**, which is the first affected release.

## Why CI did not catch it

`frontend/e2e/duckdb-backend.spec.ts` is the entire DuckDB end-to-end suite. It
checks `/api/v1/version` and counts sessions in the list — it never opens a
session, so the one broken endpoint is never requested. With the bug present
that test still passes.

The Go side has the same shape of gap: `runDuckDBServe`, `openDuckDBServeStore`,
and `loadDuckDBServeConfig` are all at 0% statement coverage.

## Reproduction

```
agentsview duckdb push
agentsview duckdb serve
# open any session in the UI -> blank transcript
curl http://127.0.0.1:8080/api/v1/sessions/<any-id>/messages   # connection dropped
```

No special data required. A mirror built from a healthy SQLite archive is enough.

## Fix

PR to follow, from `tlmaloney:fix/duckdb-serve-blank-transcripts`:

- Export `db.DecodeStoredTokenUsage` and use it in all three scanners, so empty
  yields `nil` and invalid is dropped, and the backends cannot drift again
  (the backend-parity rule in AGENTS.md).
- Blank invalid `TokenUsage` in `ValidateAndSanitize` — defensive hardening for
  genuinely malformed provider payloads, which is a real but separate concern
  from this bug.
- Regression coverage that asserts on a **clean** mirror, plus two e2e tests
  that open a session and hit the messages endpoint. With the guard reverted the
  pre-existing e2e test still passes while all three new tests fail.

Severity note: this makes `duckdb serve` unusable for its primary purpose in
v0.42.0, so it may warrant a patch release rather than waiting on the next
minor.

tlmaloney:
Fixes #1617. Supersedes #1614 (closed as misdiagnosed) and PR #1615, whose
description was written against that wrong diagnosis.

**`duckdb serve` cannot render any session transcript in v0.42.0.** Every
session, every user, every request. No corrupt data is involved.

## The bug

`token_usage` is declared `TEXT NOT NULL DEFAULT ''`, so nearly every message row
holds `''`. The DuckDB scanner assigned it **unconditionally**, while the SQLite
and PostgreSQL scanners guard the empty case:

```go
// internal/duckdb/messages.go:296  -> non-nil, ZERO-LENGTH jsontext.Value
m.TokenUsage = []byte(tokenUsage)

// internal/db/messages.go, internal/postgres/messages.go -> stays nil
if tokenUsage != "" { m.TokenUsage = ... }
```

A zero-length `jsontext.Value` is not valid JSON, and `omitempty` does **not**
skip it, so the very first message of every response fails to marshal:

```
json: cannot marshal from Go jsontext.Value: unexpected EOF within
"/messages/0/token_usage"
```

`internal/server` installs no `recover()`, so `net/http` drops the connection —
`ERR_EMPTY_RESPONSE`, blank transcript. HTML export of the same session still
works because `humaExportSession` -> `GetAllMessages` -> `generateExportHTML`
never marshals the field, which is what makes it look like a frontend bug.

## A datable regression

The scanner line is unchanged since `ac7cef31` (#609) and was harmless for two
releases. `b85eae5a` ("adopt Go 1.27 JSON v2 semantics", #1475) changed the field
type:

```go
- TokenUsage json.RawMessage `json:"token_usage,omitempty"`
+ TokenUsage jsontext.Value  `json:"token_usage,omitempty"`
```

v1's `omitempty` dropped a zero-length slice; v2's does not:

```
v1 json.RawMessage("") + omitempty -> [{}]           err=<nil>
v2 jsontext.Value("")  + omitempty -> err=cannot marshal ... unexpected EOF
```

First affected release is **v0.42.0**.

## The change

`db.DecodeStoredTokenUsage` is exported and used by all three scanners: `""`
yields `nil`, invalid JSON is dropped, valid values pass through. One
implementation means the backends cannot drift again, per the backend-parity
rule in AGENTS.md.

That is the whole fix. The net diff against `main` touches three scanners, one
existing test, and adds test files — no schema changes, no new API fields.

Dropping invalid values (as opposed to empty ones) is defense at the point of
consequence: reaching the response encoder with one panics the handler, and
there is no `recover()` to catch it.

### On the write side: deliberately not included

Earlier revisions of this branch also validated `TokenUsage` in
`ValidateAndSanitize`. That has been reverted in `6ca8f5be`, because it defends a
path that does not exist:

- `internal/parser/claude.go:168` validates every line with `gjson.ValidBytes`
  and **skips invalid ones**, so a torn read never reaches
  `extractClaudeTokenFields`. gjson would return an invalid `.Raw` for a
  truncated line, but that line is discarded first.
- Every other parser builds usage through `json.Marshal` (devin) or derives it
  from an already-valid value (`claude_websearch`).

`TokenUsage` is valid by construction at every write site, so the counter that
change added could never increment — and it cost a public API schema field
(`token_usage_blanked` on `SanitizeStats`) plus its generated client entry for
no reachable benefit. Happy to revisit it as its own PR if you consider
"never persist known-bad data" worth the schema surface.

## Tests

The read-path tests plant corrupt data, so **none of them could catch this bug.**
The coverage that matters asserts on a pristine mirror:

- `TestDuckCleanMirrorMessagesMarshal` — an absent usage value must be `nil`
  rather than zero-length, and a clean mirror must marshal.
- Two e2e tests in `duckdb-backend.spec.ts` — open a session and assert the
  transcript renders; hit the messages endpoint directly, taking the session id
  from `GET /api/v1/sessions` so the assertion is not coupled to UI rendering.

Plus corruption-based coverage at all three backends, an endpoint-level
regression test, and a contract test for `DecodeStoredTokenUsage`.

**Every new test was confirmed to fail with the guard reverted**, so none are
tautological. Most tellingly, with the bug present the pre-existing e2e test
still passes while both new ones fail:

```
✓ DuckDB backend › serves fixture sessions in read-only mode
✘ DuckDB backend › renders a session transcript
✘ DuckDB backend › serves session messages over the API
      apiRequestContext.get: socket hang up
```

`socket hang up` is the dropped connection — the Node-level form of the
`ERR_EMPTY_RESPONSE` users see.

## Why CI missed it

`duckdb-backend.spec.ts` was the entire DuckDB e2e suite: it checked
`/api/v1/version` and counted sessions in the list, and never opened a session,
so the one broken endpoint was never requested. `runDuckDBServe`,
`openDuckDBServeStore`, and `loadDuckDBServeConfig` are also all at 0% statement
coverage on the Go side — worth a follow-up beyond this PR.

## Verification

| Check | Result |
| --- | --- |
| `go vet` | clean |
| `make test-short` | 48 packages passing |
| `make test-postgres` | passing against a real PostgreSQL instance |
| `make e2e-duckdb` | 18 passed, 2 skipped |
| frontend `vitest` | 159 files / 2457 tests passing |
| `npm run check:api` | exit 0 |

One pre-existing failure is unrelated and untouched:
`cmd/agentsview TestRunDaemonArtifactExchangeConnectsToIPv6LocalhostListener`.
It listens on `[::1]` but rewrites the endpoint host to `localhost`, so it fails
on hosts whose `/etc/hosts` maps `localhost` to `127.0.0.1` only (the
Debian/Ubuntu default). Its skip guard covers IPv6 *listening* being
unavailable, not resolution. Fails identically at tag `v0.41.1`.

## One behavior change worth review

Invalid usage no longer rejects a session from artifact export, since the value
can no longer reach the segment encoder (`artifact_export_load.go:118` loads
through `scanMessages`). Previously one bad blob cost the whole session its
export. `TestExportContinuesPastInvalidTokenUsage` asserted that, so it now
describes an unreachable scenario; I rewrote it as
`TestExportSanitizesInvalidTokenUsage` rather than deleting it. Export
continuing past a genuinely un-encodable session stays covered by
`TestExportContinuesPastInvalidManifestValue` and
`TestExportContinuesPastDeterministicRejection`, both untouched and passing.

This follows from the read guard, not the reverted write-side change — verified
by disabling the write seam alone and confirming the rewritten test still
passes. It is the one change here that flips an existing assertion.

## Severity

This makes `duckdb serve` unusable for its primary purpose in v0.42.0, so it may
warrant a patch release rather than waiting on the next minor.

https://claude.ai/code/session_016BtS7hmmCDXSXvLRDxmkFx


tlmaloney:
@mariusvniekerk I'm all done with updates, so open for your review whenever you get a chance

tlmaloney:
The Benchmark Gate failure looks like runner noise rather than a regression from
this PR. Every other check passes (20 of 21, including e2e, the race detector,
all Windows lanes, integration and lint). Evidence below in case it saves a
maintainer the dig.

**The regressing benchmark cannot reach this PR's code.**

```
benchgate: 1 regression(s):
  internal/db.RecallEvidenceWindowSingleEventCalls-4: B/op regressed 1.47x
  (4.082Mi -> 5.983Mi, limit 1.35x)
```

`BuildRecallEvidenceWindow` issues a single query:

```sql
SELECT ordinal, role, content, source_uuid
FROM messages
```

It does not select `token_usage`, does not go through `scanMessages`, and never
calls `DecodeStoredTokenUsage`. This PR changes only those three scanners, so
there is no path from the diff to that benchmark's allocations.

**The metric that tripped is the noisy one.** Running it at `GOMAXPROCS=4` to
match CI, six times on identical code:

```
B/op:      2.69Mi  2.71Mi  2.75Mi  3.02Mi  2.92Mi  2.72Mi   <- 12% spread
allocs/op:   9322    9322    9322    9324    9323    9322   <- stable
```

The allocation *count* is deterministic while the *sizes* swing, which is the
signature of a growable buffer whose final capacity depends on GC and scheduling.
`B/op` is the metric the 1.35x limit caught.

**The run also reports its own captures breaking down:**

```
benchgate: baseline capture is corrupted (4 unparseable result line(s))
  /tmp/bench-old.txt:59,61,67,69: parsing iteration count: invalid syntax
benchgate: candidate capture is corrupted (2 unparseable result line(s))
  /tmp/bench-new.txt:59,68: parsing iteration count: invalid syntax
benchgate: invalid benchmark configuration:
  internal/db.GetMessagesFiveAgentEventCalls-4: sec/op needs at least 5 candidate samples, got 4
  internal/db.RecallEvidenceWindowFiveAgentEventCalls-4: sec/op needs at least 5 candidate samples, got 4
```

`RecallEvidenceWindowFiveAgentEventCalls` is a sibling of the regressing
benchmark, so the recall family specifically was emitting malformed or missing
results on this run.

A re-run of the Benchmark Gate job should clear it. I do not have `actions: write`
on this repo, so I cannot trigger one myself — happy to push a trivial commit to
retrigger the workflow instead if that is preferred.

mariusvniekerk:
@tlmaloney yeah, i've got a PR cleaning up some of the benchgate issues.

PR #1620, opened 2026-09-04
fix(cli): resolve Windows local timezone to IANA

### LINKED ISSUE #1619: usage daily fails on Windows: localTimezone returns non-IANA zone name
`agentsview usage daily` fails with HTTP 500 on Windows because `localTimezone()` returns the Windows timezone name (e.g. `Eastern Standard Time`) instead of an IANA name (`America/New_York`).

The server rejects it:

```
> agentsview usage daily
error: usage summary: HTTP 500: {"error":"internal error"}
```

Hitting the endpoint directly with the Windows name confirms it:

```
GET /api/v1/usage/summary?timezone=Eastern+Standard+Time
400: {"error":"invalid timezone: Eastern Standard Time"}
```

Passing an IANA name works fine:

```
GET /api/v1/usage/summary?timezone=America/New_York
200
```

## Source trace

`localTimezone()` at `cmd/agentsview/usage.go:655` calls `time.Now().Location().String()`. On Windows, Go's `time.Local` resolves to the Windows zone name, not IANA. Every caller of `localTimezone()` sends this to the server or to `time.LoadLocation()`, which rejects it:

- `cmd/agentsview/usage.go:135` — `runUsageDaily`
- `cmd/agentsview/usage.go:245` — `runUsageStatusline`
- `cmd/agentsview/usage_cursor.go:85`
- `cmd/agentsview/activity.go:445`
- `cmd/agentsview/archive_query_backend.go:181`

`internal/timeutil/timeutil.go:85` has `BestEffortLocalTimezone()` which validates via `time.LoadLocation` before returning, so it would return `""` instead of the bad name. But `localTimezone()` doesn't use it, and `BestEffortLocalTimezone()` itself doesn't map Windows names to IANA either — it just falls back to empty string, which callers treat as UTC.

## Possible fix

Map the Windows zone name to IANA. Go doesn't expose a mapping, but the registry at `HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Time Zones\<name>` has metadata, and the CLDR `windowsZones.xml` provides the canonical mapping. Alternatively, checking `$env:TZ` first (as `BestEffortLocalTimezone` does) and documenting the workaround would cover users who set it.

## Workaround

```
set TZ=America/New_York
agentsview usage daily
```

## Environment

- Windows 11 (build 26100)
- agentsview v0.42.0-9-ga5e1348cc-dirty
- Go embeds tzdata (`_ "time/tzdata"` at `cmd/agentsview/main.go:21`)

rodboev:
CLI usage and activity defaults now resolve the host timezone to a loadable IANA name, so Windows local-day reports reach the daemon with the correct calendar rules. Statusline date bounds use the same resolved timezone. Explicit timezone values and valid TZ overrides remain unchanged.

The shared time utility now uses an OS-aware resolver for Windows identities, including Eastern Standard Time, while retaining safe fallbacks for environments that cannot provide a named zone. This adds `github.com/thlib/go-timezone-local` for the Windows-to-IANA zone mapping. Daily usage, statusline usage, Cursor usage windows, direct activity reports, daemon activity requests, and stats metadata use one timezone decision.

This also allows `agentsview usage daily` to work on Windows; currently it returns `HTTP 500: {"error":"internal error"}`

Closes #1619

### DIFF
--- a/cmd/agentsview/activity_test.go
+++ b/cmd/agentsview/activity_test.go
@@ -47,6 +47,33 @@ func TestActivityReportCommand_Flags(t *testing.T) {
 	}
 }
 
+func TestResolveCLIActivitySelection(t *testing.T) {
+	t.Setenv("TZ", "America/New_York")
+	oldLocal := time.Local
+	time.Local = time.FixedZone("Eastern Standard Time", -5*60*60)
+	t.Cleanup(func() { time.Local = oldLocal })
+	oldNow := activityReportNow
+	activityReportNow = func() time.Time {
+		return time.Date(2026, 3, 9, 2, 30, 0, 0, time.UTC)
+	}
+	t.Cleanup(func() { activityReportNow = oldNow })
+
+	query, filter, err := resolveCLIActivitySelection(ActivityReportConfig{
+		Preset: "day",
+	})
+	require.NoError(t, err)
+	assert.Equal(t, "America/New_York", query.Timezone)
+	assert.Equal(t, "America/New_York", filter.Timezone)
+	assert.Equal(t, "2026-03-08T05:00:00Z", query.RangeStart.Format(time.RFC3339))
+
+	query, filter, err = resolveCLIActivitySelection(ActivityReportConfig{
+		Preset: "day", Date: "2026-03-09", Timezone: "Europe/Berlin",
+	})
+	require.NoError(t, err)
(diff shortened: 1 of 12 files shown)


mariusvniekerk:
Oh windows, why must you be so weird

PR #1626, opened 2026-09-04
fix(copilot): report usage without shutdown summaries

### LINKED ISSUE #1625: Copilot CLI usage omits sessions without shutdown summaries
Copilot CLI transcripts can persist assistant model and output-token data without a usable `session.shutdown` aggregate usage summary. AgentsView imported these sessions but did not create usage-reportable rows, so recent Copilot CLI usage could be absent or substantially undercounted.

The persisted format does not identify why a shutdown summary is absent. Its absence must not be treated as zero usage, failed ingestion, or proof that the session is still active.

The proposed fix reports known assistant output-token data when aggregate shutdown metrics are absent, and documents the limitations. It also reads observed per-request token fields from the documented local Copilot session store when available, while retaining `session.shutdown.totalNanoAiu` as the only authoritative reported-cost source.

mslynch:
Resolves #1625

Copilot CLI sessions now report known token usage even when no usable shutdown summary exists. The parser reads richer per-call usage from the local session store. Later transcript responses supply per-message output fallback; within the overlapping time range, only positive per-model transcript output remainders are added as aggregate estimates. Known output remains in session totals when the model is unknown, without inventing a model-priced usage row.

Session-store rows use request-level pricing bands without requiring a transcript message ordinal. A trace through GitHub's published CLI confirms that separate model calls remain separate database rows even within one turn. The primary-source pricing research and producer trace are recorded in `docs/internal/copilot-pricing-research-2026-09-07.md`.

Store tokens remain catalog estimates. Transcript shutdown billing totals retain the existing authoritative-cost treatment; estimates are not claims about plan allowances, discounts, or invoice charges. Existing Copilot sessions reparse, and the usage cache rebuilds facts with the corrected request classification.

Copilot shared-store updates now use the same general strategy as OpenCode: scan lightweight session metadata, then read payloads and refresh the archive only for changed sessions. Transcript hashes are cached and restored across engine restarts after validating current file-change metadata. Per-session usage fingerprints use the producer's indexed latest usage-row ID because usage appends do not advance `sessions.updated_at`. One append among either 8 or 800 sessions refreshes one session, rereads no unchanged transcript payloads, and reads only that session's usage rows. Candidate enumeration still grows with session count.

Startup and the first sync after each five-minute interval verify all usage rows. Copilot also joins the daemon's 15-minute scheduled reconciliation, so edits or deletions below an unchanged latest row ID do not need another filesystem event to be detected. The threshold and scheduled cadence are separate; with regular ticks, deferred edits can wait up to 20 minutes. Shared-store timestamps no longer become transcript timestamps, and missing transcripts remain in the persistent archive. The freshness logic is in `internal/parser/copilot_store_freshness.go`.

The official CLI writer can lose a usage insert under SQLite write contention and still persist later calls. Producer experiments now establish that gap and confirm native assistant events generate timestamps. The parser no longer treats the latest store timestamp as proof that all earlier output was captured. If missing rows later appear, they replace the aggregate remainder rather than adding to it. This remains a conservative lower bound: extra store-only calls can mask missing output for the same model, and the remainder cannot recover a call's input count or pricing band. The research note and usage documentation record the evidence and limits.


mariusvniekerk:
Superseded by the smaller replacement stack, in merge order:

1. #1683: transcript output without shutdown summaries.
2. #1684: observed request usage and pricing, including missing-row recovery without double counting.
3. #1685: shared-store refresh performance and transcript fingerprint reuse.

The stack reduces the combined additions from 1,905 to 1,285. It drops the standalone research essay, periodic historical-row verification, scheduler changes, and repeated edge-case tests. The producer evidence and accounting limits are summarized in the existing format-provenance document.

Please review the replacement PRs. This original branch remains intact for reference.

PR #1628, opened 2026-09-04
feat(parser): discover and link Cursor CLI subagent transcripts

prateek:
Cursor CLI subagent transcripts now appear under the session that launched
them. Agentsview imports `.jsonl` and `.txt` files from each session's
`subagents` directory, including files from S3 sources.

When several files share a session ID, full scans and file-change events now
choose the same copy within a project. A session's own transcript wins over a
copy under another parent. Otherwise, JSONL wins over text, with alphabetical
parent order breaking ties. Updating another copy no longer changes the
selected child's content or parent link. Review this selection logic in
`internal/parser/cursor_provider.go`.

Cursor's `Subagent` calls now count as Task calls. Existing archives re-parse
on the next sync; installations that only run `agentsview usage` need one
`agentsview sync` to import existing children. Discovery supports one
`subagents` level and links sessions rather than individual tool calls. A
child whose parent transcript is missing remains visible on its own.


prateek:
Addressed in cf037c24. The streaming walk now resolves a project's subagent files through the same per-project stem map as the batch walk: parents in sorted order, `.jsonl` over `.txt`, and any top-level transcript shadows a subagent copy. A child stem that appears under two parents therefore yields one source in both routes. Covered by `TestCursorDiscoveryDeduplicatesChildStemAcrossParentsInBothWalks`.

---
_via Prateek's agent (`check-pr`)_

wesm:
looking

PR #1637, opened 2026-09-05
fix(pricing): price Codex gpt-reserve as GPT-5.6 Luna

### LINKED ISSUE #1636: Codex gpt-reserve usage is unpriced despite being Luna Reserve
### agentsview version

Current main (pricing catalogs already include `gpt-5.6-luna`; this is a reported-name miss, not a missing Luna row).

### Install method

Built from source

### OS / platform

macOS 27.0 (arm64)

### Which agent and version

Codex (ChatGPT / Codex app sessions that enter Luna Reserve)

### Which model(s)

`gpt-reserve` (producer id) / GPT-5.6 Luna

Codex rollouts that continue under Luna Reserve persist `turn_context.payload.model` as `gpt-reserve`. Agentsview copies that string onto later messages and usage rows.

`gpt-5.6-luna` is already in the Pydantic and LiteLLM-backed catalogs. `gpt-reserve` does not match that row, so Reserve turns show as unpriced / `$0` even when token counts are present.

This is expected Codex identity, not a parser bug. OpenAI documents Luna Reserve as extra GPT-5.6 Luna usage after regular allowance is exhausted, with internal model id `gpt-reserve`. It is a separately metered fallback mode, not a distinct list-price SKU.

Expected: keep the stored / reported model as `gpt-reserve` (do not rename it in the UI or merge it into `gpt-5.6-luna` in breakdowns). Resolve it for pricing the same way `k2d6-agent` maps to K2.6 in #1337, using the existing `gpt-5.6-luna` catalog rates across SQLite, PostgreSQL, and DuckDB.

These estimates would be Luna list rates, not an OpenAI invoice. Luna Reserve is a Plus/Pro allowance, not API credits. That is the same class of catalog estimate Agentsview already applies to ordinary Codex Luna turns.

Out of scope: display aliases, collapsing Reserve and Luna into one usage series, or asking Codex to stop writing `gpt-reserve`.

Local workaround until this lands: a `[custom_model_pricing."gpt-reserve"]` row keyed to the producer string.

```json
{"type":"turn_context","payload":{"model":"gpt-reserve"}}
```

Agentsview then stores that model on subsequent messages. `internal/parser/codex.go` reads `payload.model` from each `turn_context` line.

First-party description: [Luna Reserve in Codex and ChatGPT Work](https://help.openai.com/en/articles/20001499-luna-reserve-in-codex-and-chatgpt-work).

### Steps to reproduce

1. Use Codex until regular usage is exhausted and Luna Reserve is active.
2. Continue a turn so the rollout writes `turn_context` with `model: "gpt-reserve"`.
3. Sync the session and open Usage (or `agentsview usage daily --agent codex --breakdown`).
4. Observe `gpt-reserve` listed as unpriced while `gpt-5.6-luna` rows in the same archive price normally.

### Checklist

BUKOWSKIREAL:
## Summary
- Codex Luna Reserve turns persist `gpt-reserve`. Usage reports kept that name unpriced even though `gpt-5.6-luna` is already in the catalog.
- Pricing now resolves `gpt-reserve` to the Luna catalog row at query time on SQLite, PostgreSQL, and DuckDB. The archived model string is unchanged, and an exact custom `gpt-reserve` rate still wins.
- SQLite usage caches rebuild (format version 8) because the pricing digest only hashes catalog rows and would otherwise keep the old zero costs.

Fixes #1636.

Review focus: `internal/pricing/supplemental.go` for the fixed alias, `duckPriceModelCaseSQL` for DuckDB CASE parity, and the usage-cache generation bump in `internal/db/usage_cache_schema.go`.

Made with [Cursor](https://cursor.com)


mariusvniekerk:
I'm not entirely if that is correct.  Us `gpt-reserve` guaranteed to be Luna?  Isn't it just general overflow?

BUKOWSKIREAL:
gpt-reserve is overflow in the quota sense. It is not a generic “keep running Sol/Terra” overflow.

Codex’s own TUI pins the authorized Reserve model id to `gpt-reserve` and labels it Luna Reserve:

```rust
pub(crate) const LUNA_RESERVE_MODEL: &str = "gpt-reserve";
pub(crate) const LUNA_MODEL: &str = "gpt-5.6-luna";
```

[`codex-rs/tui/src/model_catalog.rs`](https://github.com/openai/codex/blob/008bbd5884122dc95aaece19ecfe0fc6a59dcf36/codex-rs/tui/src/model_catalog.rs) (openai/codex `008bbd5884122dc95aaece19ecfe0fc6a59dcf36`).

The same file maps `gpt-reserve` to the display name "Luna Reserve".

The Reserve-only `/model` picker borrows Luna (or the saved ordinary slug) for presentation and effort, then forces routing back onto that authorized id:

```rust
preset.model = LUNA_RESERVE_MODEL.to_string();
```

Header copy: “Other models return when ordinary usage is available again.”
[`codex-rs/tui/src/chatwidget/luna_reserve_model.rs`](https://github.com/openai/codex/blob/008bbd5884122dc95aaece19ecfe0fc6a59dcf36/codex-rs/tui/src/chatwidget/luna_reserve_model.rs)

OpenAI’s help page describes the product the same way: Luna Reserve is extra usage with GPT-5.6 Luna after regular usage is exhausted, not a way to keep the stronger models.
https://help.openai.com/en/articles/20001499-luna-reserve-in-codex-and-chatgpt-work

BUKOWSKIREAL:
Same mapping in ChatGPT Work: the banner says the session is now using Luna, and the model picker shows **GPT-Reserve Medium**.

ChatGPT Work banner “You’re now using Luna” with the model picker set to GPT-Reserve Medium

mariusvniekerk:
Thanks. just wanted to be sure.

PR #1638, opened 2026-09-05
feat: add native Evener v2 session support

obra:
Adds Evener as a native file provider, with local discovery, configuration, session names, reasoning, tools, model and token usage, and fork/subagent relationships in the existing AgentsView UI. Reads semantic transcript v2 and optional metadata without requiring a running Evener hub.

Reuses existing source, capture, and remote-sync mechanisms. Composite freshness covers metadata and the immediate fork parent; incomplete transcript writes preserve archived content. Verified copied fork history is omitted using the existing Codex policy. When the parent cannot be verified, child history is retained and shared usage can appear in both sessions.

Structured model-switch facts from [Evener #889](https://github.com/prime-radiant-inc/evener/pull/889) update model context when available; older prose-only switches do not invent billing identities. Per-response identities take precedence. Older transcript formats are unsupported, and unsupported media is represented by a placeholder. Provider code, synthetic lifecycle fixtures, and pinned producer provenance are included; no storage migration or new transport is introduced.


obra:
@wesm what's your take on that 'high' finding from roborev?

wesm:
looking

wesm:
The Windows finding has a real code path behind it: importing an Evener session can make Agentsview inspect the session’s recorded directory on the receiving machine. Credential exposure depends on a malicious imported transcript and Windows network/authentication settings; we haven’t demonstrated that outcome, so I’d qualify the “High” wording.

The tool-output finding also identifies a real formatting problem. However, the suggested fix needs care: preserving structured JSON without handling Evener’s object results could turn visible output into blank output.

wesm:
Pushed 256ddc4df with a narrow fix for the remote-directory lookup.

When importing Evener sessions from another machine, Agentsview now derives the project name from the recorded path without looking up that directory on the receiving computer. Local sessions still use Git project discovery as before. There are no new settings, prompts, or extra steps.

The tradeoff is limited to remote project names: a session recorded under a repository’s subdirectory may use that subdirectory’s name instead of the Git repository name. Session importing continues to work.

The regression test checks that remote parsing skips the directory lookup, local parsing still performs it, and both preserve the session content. The existing Evener import tests also pass. The separate tool-output finding is unchanged.

wesm:
known limitation. i'll fix the CI on windows and merge this

wesm:
Accepted issues. Merging

Issue #1599, opened 2026-09-03
Windows Sync CI job hangs inside a SQLite cgo call until the package timeout


wesm:
## Symptom

The `Go Test (windows-latest, Sync)` CI job intermittently hits the 20-minute package timeout in `internal/sync`. Each time, one test is blocked inside a single cgo call into SQLite that never returns. Every other lane on the same commit passes, and the job passes on re-run.

## Occurrences

| Date | Branch | Run | Hung test | Blocked call |
| --- | --- | --- | --- | --- |
| 2026-08-27 | `main` @ 1940e41f1 | https://github.com/kenn-io/agentsview/actions/runs/33072104713/job/98516726190 | `TestOpenCodeFullPassSkipsAfterWatcherPassParse` | `sqlite3_step` inside a `Row.Scan` |
| 2026-09-02 | PR #1573 @ a707264d9 | https://github.com/kenn-io/agentsview/actions/runs/33698490860/job/100472437600 | `TestResyncAllMixedOpenCodeRootsKeepsSQLiteFallback` | `sqlite3_close_v2` while closing the writer pool of the replacement archive at the end of `resyncBuildLocked` |

Both runs used `windows-latest`, Go 1.27.0, and `github.com/mattn/go-sqlite3` v1.14.50.

## What the goroutine dump shows

In the 2026-09-02 run only eight goroutines were alive when the timeout fired: the test runner, the hung test, one parallel test waiting on `t.Parallel`, and the `database/sql` helper goroutines. The hung test's stack is:

```
github.com/mattn/go-sqlite3._Cfunc_sqlite3_close_v2
github.com/mattn/go-sqlite3.(*SQLiteConn).Close
database/sql.(*driverConn).finalClose
database/sql.(*DB).Close
go.kenn.io/agentsview/internal/db.(*DB).CloseContext   internal/db/db.go:4539
go.kenn.io/agentsview/internal/db.(*DB).Close
go.kenn.io/agentsview/internal/sync.(*Engine).resyncBuildLocked   internal/sync/engine.go:3663
go.kenn.io/agentsview/internal/sync.(*Engine).resyncAllWithOptionsLocked
go.kenn.io/agentsview/internal/sync.(*Engine).ResyncAll
```

No other goroutine was in a cgo call, so this is not a Go-side lock order problem between two connections. The SQLite call itself does not come back.

The 2026-08-27 run has the same shape with `sqlite3_step` instead of `sqlite3_close_v2`.

[...]

PR #1641, opened 2026-09-06
test(sync): reduce large archive fixture setup cost

rodboev:
This PR speeds up setup for two file-watching tests by inserting their archive rows in one transaction. They still check that changing or deleting a session works with both 1 and 10,000 unrelated sessions in the archive.

Repeated per-row writes made these tests expensive and contributed to the Sync package exhausting its 20-minute Windows CI budget in #1599. The change cuts their runtime by about 85% in alternating local comparisons. Their combined Windows CI time fell from 66.53 to 4.29 seconds across the base and replacement runs.

Refs #1599

### DIFF
--- a/internal/sync/watch_batch_sync_test.go
+++ b/internal/sync/watch_batch_sync_test.go
@@ -2,6 +2,7 @@ package sync
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


rodboev:
Rebuilding around these constraints:

  - Bound waiting and retain the original completion result and ownership.
  - Return promptly with failure when recovery encounters an unfinished native close.
  - Preserve ordered closing, existing WAL safety, and ordinary recovery after completion.
  - Inspect replacement-file cleanup explicitly rather than calling `removeTempDB` after replacement close fails.

This trades recovery availability during a pathological native stall for simpler ownership and saves about 600 lines.

rodboev:
The logs show both failures hit the package's 20-minute limit while the named tests had been running for less than a second. I'm replacing the SQLite close changes with faster setup for two costly watcher tests. The prototype cuts their runtime by about 85% locally while preserving the 10,000-row cases and existing assertions.

PR #1657, opened 2026-09-07
fix(cli): clarify listener and browser URL settings

### LINKED ISSUE #1653: Arguing about arguments
### agentsview version

v0.42.0

### Install method

GitHub release binary

### OS / platform

Fedora 43

### Which agent and version

N/A

### Which model(s)

N/A

# --public-port is really --tls-port

`--public-port` has little if anything to do with --public-url. It's role is to set the HTTPS port when `-proxy caddy` is used. I think it's poorly named. My 3c is that the argument name should tie it to --proxy-*, not to --public-*.

# consistency in argument naming

Why  name `-host` this way, but  `--proxy-bind-host` this way?

shouldn't it be either 
`--bind-host` and  `--proxy-bind-host`
OR 
`--host` and  `--proxy-host`
?

# [Rant Warning] What is --public-url *really* for?

The help says:
```
   --public-url string                   Public URL to trust and open for hostname or proxy access
```
"To trust and open" is a rather strange brew. 

It seems like  `--public-url`  conflates two somewhat unrelated things, and I'm not sure is either needed or makes sense:
1. It's a singleton duplicate of --public-origin - and thus not strictly needed
2. It Instructs agentsview what url to point the launched browser at. For this, it's poorly named.
3. It identifies the "primary url" you expect the app to be accessible from to non-local users, and logs it. But agentsview can't verify that address works since it doesn't control it. So what's the point of logging it as "public"?
what is the value of a "--print-this-value-on start" option?

For me,  printing out "public at https://0.0.0.0:9999", suggests that agentsview is listening at that address, which isn't the case at all. 

Printing out whitelisted origins can be helpful. But since agentsview has not visibility into routing at those hosts, it  shouldn't present them as its "public" facing url.

I also struggle to see when it would make sense to launch a browser to a public url which differs from the backend url. That either means
1.  you're not local and so can't launch a browser, or
2. you're local and can simply access the backend server directly.

So... what is this option for? Why does it mix Origin whitelisting, with browser launching, with reverse-proxies and network routing outside its control?

```text
⬆️
```

### Steps to reproduce

⬆️ 

### Checklist

wesm:
Server startup now labels the listening address and browser URL separately, so a configured URL no longer reads as a claim of public reachability. `--public-url` rejects wildcard bind addresses such as `0.0.0.0` and `::`, with guidance to use the bind flags instead.

Help and remote-access docs distinguish backend and Caddy listeners, the browser URL, and trust-only origins. They correct the Caddy bind default to loopback and explain how explicit URL ports, `--public-port`, and the 8443 default interact.

The existing flag names remain: `--public-url` still selects a browser destination and trusted origin, while `--public-port` applies to managed Caddy in both HTTP and HTTPS modes.

Closes #1653.


docs/commands.md
```diff
@@ -147,12 +147,12 @@ Running plain `agentsview` shows help instead of starting the web UI.
...
 | `--background`      | `false`     | Start `agentsview serve` as a managed background process |
 | `--replace`         | `false`     | Replace a running local daemon before starting           |
-| `--public-url`      |             | Public URL for hostname or proxy access                  |
+| `--public-url`      |             | Browser URL, also added to trusted origins                  |
 | `--public-origin`   |             | Trusted browser origin (repeatable/comma-separated)      |
 | `--proxy`           |             | Managed proxy mode (`caddy`)                             |
 | `--caddy-bin`       | `caddy`     | Caddy binary path                                        |
 | `--proxy-bind-host` | `127.0.0.1` | Interface for managed proxy                              |
-| `--public-port`     | `8443`      | External port for managed proxy                          |
+| `--public-port`     | URL port or `8443`      | Managed Caddy HTTP/HTTPS listener and URL port                          |
```
BroadlyWhitaker: What does "listener and URL port" mean? aren't those synonyms? how about just "listening port"?

docs/remote-access.md
```diff
@@ -330,14 +330,39 @@ Use header-based auth for normal API calls whenever possible.
...
 ## Public URL And Trusted Origins
 
-When you access AgentsView through a hostname or reverse proxy, tell the server
-about the public URL:
+The bind address, browser URL, and trusted origins serve different purposes:
+
+| Setting | Purpose |
+| --- | --- |
+| `--host` / `--port` | Choose the backend HTTP listener (default `127.0.0.1:8080`). |
+| `--proxy-bind-host` / `--public-port` | Choose the managed Caddy listener when `--proxy caddy` is enabled. |
```
BroadlyWhitaker: This shows naming is still a mess. this software is young, it's not too late to fix this: replace the official options from the help, without ripping out the code yet. 

you have `host` vs `bind-host` for the **same thing** on the one hand, and "--proxy-*" vs. "--public-" for options that configuring the same server.

docs/remote-access.md
```diff
@@ -330,14 +330,39 @@ Use header-based auth for normal API calls whenever possible.
...
 
 ```bash
 agentsview serve --public-url https://agents.example.com
 ```
 
-For additional trusted origins, use `--public-origin`:
+This setting does not configure DNS, port forwarding, certificates, or an
+external reverse proxy. The startup message labels it **browser URL**, separately
+from the backend's listening address; it does not confirm that the URL is
+reachable. Here, "public" means the client-facing address, which may be local
```
BroadlyWhitaker: But "client-facing address" is not a guarantee. It's just what address you told agentsview, to tell Caddy, it should expect to see. It's not agentsview telling the user where the server is, it's the user telling agentsview about network routing it has no way of knowing about.

```
"public" means the client-facing address, which maybe private
```

If this was model-generated, I don't think the model will ever do a good job of clearly explaining this.
And the difficulty with documenting this is partly because the option's role itself is muddled.

Feels a little like this is a "god" option, so that invocations are simple. but that makes understanding what's actually going on more difficult.

Notice how the explanation for this blew up in size - that's a smell. In code, if you need a 10 line block comment to explain your code - it's a sign you should probably rewrite the code.

docs/remote-access.md
```diff
@@ -461,14 +488,40 @@ agentsview serve \
...
 | `--tls-cert`        |             | TLS certificate file path                             |
 | `--tls-key`         |             | TLS key file path                                     |
 | `--allowed-subnet`  |             | Client CIDR allowlist (repeatable or comma-separated) |
 
 Caddy must already be installed and available on `PATH` unless you override it
 with `--caddy-bin`.
 
+### Listener And URL Ports
+
+`--public-port` applies only to managed Caddy. It controls Caddy's listening
```
BroadlyWhitaker:
```
`--public-port` applies only to managed Caddy. 
```

Then why doesn't passing it without also passing -proxy caddy raise an error? models like quick feedback for self-correction. Humans too.

docs/remote-access.md
```diff
@@ -461,14 +488,40 @@ agentsview serve \
...
+port and the port in the resolved `--public-url`; it does not change the backend
+`--port`. It works with both HTTP and HTTPS, as selected by the URL scheme.
+HTTPS requires `--tls-cert` and `--tls-key`; HTTP must omit them.
+
+- If the URL has an explicit port, Caddy uses that port.
+- If the URL omits a port, Caddy uses `--public-port`, or `8443` when the flag is
+  unset. Thus `--proxy caddy --public-url https://agents.example.com` resolves to
+  `https://agents.example.com:8443`, not port 443.
+- If both the URL and `--public-port` specify ports, they must match. For HTTPS
+  on port 443, use `--public-port 443` or put `:443` in the URL.
```
BroadlyWhitaker: This has all the dark hallmarks of AI automation. It Carefully documents complexity instead of cutting it out at the root.

You have two ways of specifying the same thing...
...and then a bunch of rules and logic on how to reconcile them...
...and then a bunch of documentation explaining the logic...
...and the a bunch of annoying assholes on github nitpicking you to death...

Better to nuke it from space and make the conflict impossible.

docs/remote-access.md
```diff
@@ -461,14 +488,40 @@ agentsview serve \
...
+
+Caddy binds to `127.0.0.1` by default, even when the URL contains a remote
+hostname. To accept LAN connections, set `--proxy-bind-host` and an
+`--allowed-subnet` as shown below. DNS and routing remain your responsibility.
```
BroadlyWhitaker: What is with this baroque logic? where's the simplicity?

so --public-url cmpetes duplicates --public-port, but then it still requires --proxy-bind-host for the ip/host part. This is a hodgepodge.

docs/remote-access.md
```diff
@@ -461,14 +488,40 @@ agentsview serve \
...
+
+With an external proxy, leave `--proxy` unset and put its browser-facing port
+directly in `--public-url`. `--public-port` has no effect in that mode. For
+example, if an existing proxy listens on 443 and forwards to backend port 8080:
```
BroadlyWhitaker: OMG, and there's **more** rules..

internal/config/config.go
```diff
@@ -54,7 +54,7 @@ type ProxyConfig struct {
 	Bin string `json:"bin,omitempty" toml:"bin"`
 	// BindHost is the local interface/IP the proxy binds to.
 	BindHost string `json:"bind_host,omitempty" toml:"bind_host"`
-	// PublicPort is the external port exposed by the proxy.
+	// PublicPort selects the managed proxy listener port and public URL port.
```
BroadlyWhitaker: Compare with above:
```
- If the URL has an explicit port, Caddy uses that port.
- If the URL omits a port, Caddy uses `--public-port`, or `8443` when the flag is
  unset. Thus `--proxy caddy --public-url https://agents.example.com` resolves to
  `https://agents.example.com:8443`, not port 443.
```
vs.
```
PublicPort selects the managed proxy listener port and public URL port.
```
so one of these is at least somewhat wrong.

Imagine Shiva the multi-armed Indian goddess was assigned to be traffic cop. How many conflicting directions are there?

This is where LLM code still sucks balls. it creates complexity that slowly corrodes projects from the inside. Incrementally, it's no big deal. But you end up with heavy, clumsy, subtly not-quite-right tools. And beating it back requires inordinate effort from pedantic humans. It requires so much caring, for little reward, to resist the vibe currents dragging your code base down.

internal/config/config.go
```diff
@@ -1988,16 +1988,16 @@ func (f *stringListFlag) Type() string {
 // RegisterServeFlags registers serve-command flags on fs.
 // The caller must call fs.Parse before passing fs to Load.
 func RegisterServeFlags(fs *flag.FlagSet) {
-	fs.String("host", "127.0.0.1", "Host to bind to")
-	fs.Int("port", 8080, "Port to listen on")
+	fs.String("host", "127.0.0.1", "Interface/IP for the backend HTTP server to bind")
```
BroadlyWhitaker: host can never be an interface. a network interface would be eth0, or enp1s0, or ppp0. you don't listen on those. You probably wanted "Host/IP", if hostnames are resolved along the way.

internal/config/config.go
```diff
@@ -1988,16 +1988,16 @@ func (f *stringListFlag) Type() string {
...
+	fs.Int("port", 8080, "Port for the backend HTTP server to listen on")
```
BroadlyWhitaker: It's only the "backend" server when you use a proxy. Many local users won't. I prefer "web server" and "reverse-proxy" as more stable unambiguous terms.

internal/config/config.go
```diff
@@ -1988,16 +1988,16 @@ func (f *stringListFlag) Type() string {
...
 	fs.String(
 		"public-url", "",
-		"Public URL to trust and open for hostname or proxy access",
+		"Browser URL, also added to trusted origins; does not bind a listener",
 	)
 	fs.Var(
 		&stringListFlag{},
 		"public-origin",
-		"Trusted browser origin to allow for remote or proxied access (repeatable or comma-separated)",
+		"Trusted origin for Host/Origin checks; does not change the browser URL (repeatable or comma-separated)",
```
BroadlyWhitaker: LLM smell.  Adding obscure qualifications as a way to "resolve" a more fundamental design decisions.

internal/config/config.go
```diff
@@ -2077,11 +2077,11 @@ func RegisterServePFlags(fs *pflag.FlagSet) {
 	)
 	fs.String(
 		"proxy-bind-host", "",
-		"Local interface/IP for managed Caddy to bind (default: 0.0.0.0)",
+		"Local interface/IP for managed Caddy to bind (default: 127.0.0.1)",
 	)
 	fs.Int(
 		"public-port", 0,
-		"External port for the public URL in managed Caddy mode (default: 8443)",
+		"Managed Caddy HTTP/HTTPS listener port; also sets the public URL port (default: URL port or 8443)",
```
BroadlyWhitaker: You have `-proxy VALUE`, with `caddy` being the only value currently supported. Is it a good choice to reify Caddy into all the help strings and documentation prose? if yes, why have `-proxy` take a value at all for now?

README.md
```diff
@@ -113,9 +113,13 @@ agentsview serve --public-url http://127.0.0.1:18080
 agentsview serve --public-url https://your-workspace.exe.dev
 ```
 
-Use `--public-origin` (repeatable or comma-separated) to trust additional
-browser origins. If you expose the UI beyond loopback, also enable
-`--require-auth`.
+`--public-url` also selects the URL opened on startup; it does not configure a
+listener or external proxy. Use `--no-browser` on a headless server, or use
+`--public-origin` (repeatable or comma-separated) on its own for trust-only
+configuration. If you expose the UI beyond loopback, also enable
```
BroadlyWhitaker: LLM smell: "it does not configure a listener or external proxy"

i.e
"One thing worth flagging is that this pencil sharpener does not make coffee or file your taxes. It is however is load-bearing for pencil sharpening. It isn't just that the pencils need sharpening, it's that the entire stationary-industrial complex requires rethinking".

README.md
```diff
@@ -113,9 +113,13 @@ agentsview serve --public-url http://127.0.0.1:18080
...
+`--require-auth`. See
+[Remote Access](https://agentsview.io/remote-access/#public-url-and-trusted-origins)
+for the listener, browser URL, and managed Caddy port rules.
```
BroadlyWhitaker: I was having fun. Let me know if it's ultimately just aggrevation, rather then somewhat helpful...  No trolling intended.

BroadlyWhitaker:
@wesm, I added some post-merge curmudgeon comments. Hope you don't mind.

PR #1663, opened 2026-09-08
feat(db): add tool-result image retention and db strip --images

### LINKED ISSUE #1622: Feature request: remove archived tool images and keep them excluded on future syncs
### Problem / motivation

I use AgentsView to review conversations with coding agents. Image-heavy tool results can take far more disk space than the text I want to keep. I would like to remove images from the AgentsView archive while preserving conversation text and textual tool output, without changing the original agent transcripts or standalone image files.

A read-only inspection of my macOS desktop installation, AgentsView v0.41.1 (a902515a), found:

- `sessions.db`: 19.96 GB (decimal).
- Inline `data:image/...;base64,...` payloads in `tool_calls.result_content`: 9.57 GB.
- The same aggregate image payload size in `tool_result_events.content`: another 9.57 GB. Combined, these account for about 96% of the database file size.
- 2,239 image-bearing tool-result rows in each table, across 151 sessions. Most image bytes came from `exec` calls categorized as `Bash`.
- SQLite reported zero freelist pages, so reclaiming free pages alone would not address the main storage cost.

These measurements describe v0.41.1, not a claim that the latest development branch still duplicates tool results. Related: #1375 (storage maintenance/compression) and #1586 (duplicate tool-result storage). This request specifically concerns selective image retention, including after synchronization.

### Proposed solution

Please add an opt-in way to remove inline image payloads from existing tool results and persist that retention policy for future ingestion.

Desired behavior:

1. Let users select sessions, or the whole archive, and preview how many tool images and bytes would be removed before applying cleanup.
2. Preserve user/assistant text, textual tool-result blocks, tool-call metadata, and the order of the conversation. Replace image blocks with a small readable placeholder if needed.
3. Change only AgentsView's stored content. Leave original Codex/other agent transcripts and standalone image files untouched.
4. Apply the policy during incremental ingestion and full resync/rebuild, so images are not silently imported again from the original transcripts. A restart or upgrade should preserve the preference.
5. Support cleanup of existing archived content even when its original transcript file is no longer available; do not require deleting or reconstructing those sessions.
6. Provide a supported space-reclamation step and report the actual disk space recovered. A desktop setting/action and a CLI or API equivalent would make this accessible without manual SQLite edits.

The initial scope could be tool-result images only, leaving user-uploaded images separately controllable. Keeping image data remains useful for other users, so this should be opt-in.

### Alternatives considered

- `result_content_blocked_categories`: blocking `Bash` would also discard textual command output that is valuable for reviewing conversations. An image-only policy would be more precise.
- Session deletion / `agentsview prune`: removes more than the image payloads; in v0.41.1, `prune` also removes original session files.
- `db compact`: useful for reclaiming free pages, but the storage-maintenance documentation correctly distinguishes it from filtering live payloads.
- Manually editing the database once: does not establish a supported retention policy, so a later full reparse could restore images from source transcripts.
- Compression or deduplication: complementary improvements, but users should also be able to choose not to retain bulky tool images at all.

No transcript contents, screenshots, session identifiers, or private paths are attached.

rodboev:
`agentsview db strip --images` removes supported inline image payloads from archived tool results while preserving surrounding text, event identity, and original source files. Project and date filters also reach trashed and source-missing sessions. Each changed session is rescanned for secrets in the same transaction so detections and their offsets match the remaining content. Run `agentsview db compact` afterward to reclaim SQLite free pages; content savings and disk reclamation are reported separately.

Set `tool_result_images = "drop"` in `config.toml` to apply the same policy to future writes. This is a TOML-only setting; desktop Settings and the settings API do not expose it. The default `keep` preserves existing provider decoding. Supported Codex image arrays become structured placeholders with media type and decoded byte size, displayed alongside their text in the UI.

Large Codex imports project events before scratch storage, and late results preserve raw digests before projection so replay deduplication remains stable. Combined summaries strip both labeled and anonymous sections, including summaries rebuilt from older events. Full resync projects only copied archived sessions; freshly parsed sessions are already projected. Review the staged sink and incremental result writer for these storage boundaries. Windows sync tests are split across two exhaustive jobs to fit the existing 20-minute timeout.

Closes #1622


rodboev:
This is part of a plan for making Docbank integration possible, per @mariusvniekerk's idea in https://github.com/kenn-io/agentsview/issues/1622#issuecomment-5544840925, roughly:

- slice 1: this image retention policy + `db strip --images` (and `tool_result_images` predicate)
- slice 2: `db migrate --images`, offload to the embedded Docbank vault
- slice 3: offload on ingest (images never land inline, via `tool_result_images = "offload"`)
- slice 4: local render cache with age eviction (keep full-size locally for N days)
- slice 5: OCR / searchable image text

GUI is getting worked in throughout, but the idea is to make this CLI-first while exposing HTTP routes.

wesm:
reviewing

Issue #1627, opened 2026-09-04
Cursor CLI transcript fidelity: subagent linking, legacy tool results, turn timestamps, and richer local stores


prateek:
## TL;DR
agentsview currently omits some Cursor CLI data that exists locally, including subagent transcripts and legacy tool results, and presents inferred timestamps as exact.

The roadmap addresses these gaps in independent stages, starting with discovery and parent linking for `subagents/` transcripts. Existing archives gain child sessions on their next sync.

## Problem / motivation

agentsview imports Cursor CLI sessions from `~/.cursor/projects/<project>/agent-transcripts/`, but several gaps have different causes.

- Cursor writes delegated transcripts to `agent-transcripts/<parent>/subagents/<child>.jsonl`. Batch and streaming walks, watcher classification, find-by-ID, and the S3 layout filter currently reject this path, so parents show a `Subagent` call without the child session.
- One macOS sample held 814 parent transcripts, 109 of them delegating, and 967 children. 933 children still had a parent file on disk, and every child sat exactly one level deep.
- The legacy text parser discards parsed `[Tool result]` blocks.
- Parent and child user messages begin with a `<timestamp>` tag, but the parser discards it and uses file mtime as the exact session start and end. The sample contained tags in 810 of 814 parents and 964 of 967 children.
- Cursor names the delegation tool `Subagent`, while the taxonomy matches lowercase `subagent`, so these calls appear as `Other`.

A Cursor forum thread reports tool inputs present but outputs absent in CLI JSONL: https://forum.cursor.com/t/accessing-the-full-agent-transcript-in-cursor/157311. In the inspected transcripts, no `tool_result` blocks were present and every `tool_use` block had `"id": null`; JSONL alone therefore cannot join results to calls, and parent-child relationships must come from the file layout.

Cursor also stores data in the CLI `store.db`, IDE `state.vscdb`, protobuf `toolCallBinary` payloads, SDK databases, and `agent-tools/*.txt` sidecars. The IDE parser already reads `toolFormerData.result` and real timestamps, but it does not decode `toolCallBinary` or model subagent relationships.

Since #1589, blank `tool_calls.result_content` can indicate deduplication. The text remains in `tool_result_events`, and the API reconstructs it.

## Proposed solution

Implement the work as independently reviewable changes:

[...]

PR #1664, opened 2026-09-08
feat(parser): read Cursor CLI store.db transcripts

rodboev:
Cursor CLI sessions now gain assistant reasoning and producer timestamps from matching `~/.cursor/chats/<workspace>/<agent-id>/store.db` files while keeping the transcript's session identity. Previously, AgentsView read only the text or JSONL transcript under `~/.cursor/projects`.

The provider reads the database and its WAL in one read-only transaction, follows only the tree selected by `latestRootBlobId`, and aligns decoded turns with transcript messages. The composite fingerprint detects store changes even when the transcript is unchanged.

Chats directory scan failures and readable stores with missing required tables or columns, or unsupported metadata or blob structure leave transcripts available for initial import and later updates, with a warning. A failed scan retains usable cached store locations and retries on the next discovery pass. Database open/read errors remain retryable source errors that preserve existing archive content. Temporary path-access failures retain cached stores and newly observed store candidates for retry, and store fingerprint failures prevent freshness skips.

Format evidence is limited to one installed Cursor CLI capture. Native tool-call/result joining, store-only discovery, encrypted payloads, and cross-version field stability remain outside this change. Store enrichment is local and automatically associates chats only with a resolved `.cursor/projects` root. Custom roots such as `.cursor/archive` remain transcript-only. The chats directory stays outside remote, SSH, and S3 transfer targets. Store and WAL events map back to the transcript, and shared-memory events are ignored by the provider.

Refs #1627

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -1040,6 +1040,43 @@ preservation of archived messages for OpenCode, Kilo, MiMoCode, and Icodemate.
   Cursor IDE is retired through stored-source-hint tombstones on `state.vscdb`
   change events and through complete-container ownership reconciliation.
 
+### Cursor CLI store store.db
+
+Cursor CLI store format. A per-session SQLite store under
+  `~/.cursor/chats/<workspace-hash>/<agent-id>/store.db`, with `blobs` and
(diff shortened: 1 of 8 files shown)


wesm:
reviewing. i would love to get my hands on a large synthetic cursor database for benchmarking!

PR #1660, opened 2026-09-08
feat(parser): support Open Code Review sessions

rodboev:
Open Code Review sessions now appear in AgentsView from `.opencodereview/sessions`, with review metadata, thinking, tool results, structured review comments, recorded token usage, and resume links. The new provider supports custom session directories and replaces parsed sessions when appended tool results update earlier calls.

The parser counts new user turns from main and plan requests, excludes auxiliary prompts, and avoids replaying compressed history. It treats `task_done` as a control signal and separates uncached input from cached tokens for pricing. Upstream fallback or proxy usage records can omit whether prompt totals include cached tokens; those records may undercount. The pinned format evidence and limitations are recorded in `docs/internal/session-format-sources.md`.

Closes #1612

PR #1678, opened 2026-09-08
fix(sidebar): let agent names use available row width

### LINKED ISSUE #1673: Keep long agent names distinguishable in the session sidebar
Long agent names lose their distinguishing text in the session sidebar. @thibaudcolas reported that OpenCode and Open Code Review are hard to tell apart while testing PR #1660. Screenshot from the original report:

Session sidebar truncating OpenCode and Open Code Review agent names

Source: https://github.com/kenn-io/agentsview/pull/1660#issuecomment-5583286507

`frontend/src/lib/components/sidebar/SessionItem.svelte` applies `max-width: 52px`, `white-space: nowrap`, `overflow: hidden` and `text-overflow: ellipsis` to every `.agent-tag`. Custom agent labels use the same element, so this also affects other long names. The rule is identical at PR #1660's head and main commit `c34055661f2deb35facf9f53485fa46e878b376d`.

To reproduce, view OpenCode and Open Code Review sessions together in the sidebar using PR #1660's build. Compare their displayed labels with the full provider names. Other sufficiently long provider or custom labels reach the same width limit.

Let agent names use available row space so users can distinguish them while scanning sessions. Check narrow sidebar widths and rows with other metadata. Expose the full name when truncation remains necessary, though a tooltip alone would still require hovering to distinguish sessions.

This is a shared sidebar layout issue and can be handled separately from the Open Code Review parser.

rodboev:
Long agent names use more of the session sidebar, so OpenCode and Open Code Review remain distinguishable while you scan rows. Names that still need truncation show their full label in a tooltip on hover.

`SessionItem.svelte` capped every agent label at 52px even when the sidebar had more room. This change lets the label use a bounded share of the row while preserving the flexible session title, machine metadata, entrypoint resolution and suppression, row actions, and responsive sidebar limits. Long entrypoint badges stay on one line and ellipsize within that bound.

The reproduction came from @thibaudcolas while testing https://github.com/kenn-io/agentsview/pull/1660#issuecomment-5583286507.

Closes #1673

### DIFF
--- a/frontend/e2e/sidebar-agent-names.spec.ts
+++ b/frontend/e2e/sidebar-agent-names.spec.ts
@@ -1,0 +1,332 @@
+import { expect, test, type Page, type TestInfo } from "@playwright/test";
+
+const SIDEBAR_WIDTH_KEY = "agentsview-sidebar-width";
+const REVIEW_LABEL = "Open Code Review";
+const REGISTRY_LABEL = "Opencode";
+
+interface SidebarFixtureRow {
+  id: string;
+  parent_session_id: null;
+  relationship_type: null;
+  project: string;
+  machine: string;
+  agent: string;
+  agent_label?: string | null;
+  entrypoint?: string | null;
+  display_name: string;
+  first_message: string;
+  started_at: string;
+  ended_at: string;
+  created_at: string;
+  message_count: number;
+  user_message_count: number;
+  is_automated: boolean;
+  is_teammate: boolean;
+}
+
+interface SidebarFixtureResponse {
+  sessions: SidebarFixtureRow[];
+  next_cursor: null;
+  total: number;
+}
+
+const now = "2026-09-08T12:00:00Z";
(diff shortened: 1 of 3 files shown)


rodboev:
Narrowing this to native browser tooltip only to stay consistent with AV conventions in other areas of the UI, rather than custom drawn tooltips. Also, the custom tooltips are too big relative to the area, and that compounds if we show them only if the `agent-tag` is truncated because then we'd have to do more math, which hurts performance when doing it across many session rows.

Tooltips will be native and unconditional, which keeps them small enough to activate on `agent-tag` hover without being jarring or introducing complexity.

Also extending this to `session-name`, `session-project`, and `entrypoint-tag` since those get cut off too, and more frequently than `agent-tag`. Skipping `machine-tag` even though it's `text-overflow: ellipsis`, since in practice it seems not to get truncated and we don't need stray tooltip popups.

PR #1679, opened 2026-09-09
fix(frontend): add preformatted rendering for unknown XML blocks (#1672)

### LINKED ISSUE #1672: Add an option to render unknown XML blocks as preformatted text
Diffs inside unknown XML-style prompt tags render as Markdown lists and headings in the session viewer. @thibaudcolas reported this while testing Open Code Review in https://github.com/kenn-io/agentsview/pull/1660#issuecomment-5583286507

Session viewer interpreting diff deletions as list items and Python comments as headings

I reproduced it through the shared `renderMarkdown` function with this input:

```text
<current_file_diff>
diff --git a/example.py b/example.py
--- a/example.py
+++ b/example.py
@@ -1,2 +1 @@

-    # FIXME: replace this
-    old_value
+    new_value

</current_file_diff>
```

The output contains lists and an `<h1>` for the deleted comment. Replacing the wrapper with `<review_context>` or `<policy>` produces the same result. Any provider that supplies this text can hit the issue.

`frontend/src/lib/utils/markdown.ts` preserves unknown tags as escaped text, but it tokenizes the enclosed content as Markdown first. The relevant renderer code is identical at PR #1660's head and main commit `c34055661f2deb35facf9f53485fa46e878b376d`.

Following @mariusvniekerk's suggestion, add a display preference to render unknown XML blocks as preformatted text, including their tags. Keep current rendering as the default, since some prompts intentionally put Markdown inside those tags. Recognize complete blocks before Markdown tokenization, preserve whitespace, and include the preference in the render cache key. Existing HTML sanitization, shell wrappers, code fences and inline code need to retain their behavior.

This follows the tag-preservation work in https://github.com/kenn-io/agentsview/pull/892 and belongs in the shared renderer. Parser output can remain unchanged.

rodboev:
Selecting "Render unknown XML blocks as preformatted text" in Appearance settings makes complete unknown XML blocks display as literal preformatted text with their opening and closing tags, so diff text keeps its literal lines. The default renderer keeps its current behavior for prompts that intentionally use Markdown inside unknown tags.

This adds a browser-local preference, passes it through all six shared Markdown consumers, recognizes complete blocks before Markdown tokenization, preserves whitespace inside the block, and includes the preference in the renderer cache key. The block scanner keeps Markdown prose, known HTML, inline code, fenced code, and mismatched tags on their existing paths, including malformed nested blocks whose inner tags close first. Existing HTML sanitization, shell wrappers, asset links, and parser output remain unchanged. The current workaround requires the provider text to include a fenced code block, which stored provider output cannot add after capture.

The shared behavior lives in `markdown.ts`, with the preference in `ui.svelte.ts` and the control in Appearance settings. The reproduction came from a user report on Discord and @thibaudcolas's report on https://github.com/kenn-io/agentsview/pull/1660; https://github.com/kenn-io/agentsview/pull/892 supplies the existing custom XML tag preservation path.

The enabled session view renders the issue fixture as an XML block using the active app background, text color, and border. Regular code blocks keep their existing palette.

Unknown XML in the light theme

Unknown XML in the dark theme

Closes #1672


mariusvniekerk (approved):

mariusvniekerk:
https://github.com/kenn-io/agentsview/pull/1679#issuecomment-5635664519 is VERY VERY likely to force the agent to try and build a psycho xml parser, be warned.

PR #1680, opened 2026-09-09
fix(identity): preserve local history when adopting installation IDs

wesm:
Keep local session history together across updates, restarts, and hostname changes. AgentsView owns the existing plain ID file, `telemetry-install-id`, and creates it even with telemetry disabled; PostHog consumes that ID. The display label defaults to the current hostname or an explicit `local_machine_name`. Creating an identity does not force a config rewrite.

Upgrade moves the archive's recorded local sessions, worktree rules, and project/source metadata to the installation ID in one transaction. Session IDs, content, and curation stay intact. Adopted hostnames remain aliases for saved filters, URLs, and generated insights. Read handlers use the shared filter resolver without rewriting request URLs.

Unknown hostname keys remain untouched and are logged once; startup continues. Operators can inspect them with `db adopt-machine --list` and explicitly adopt history they own. Matching labels never prove ownership. Conflicting worktree rules still require reconciliation, and retired hostnames no longer grant artifact export ownership.

Local `session_sources` entries must omit `machine`; remove any old hostname setting. PostgreSQL republishes adopted history incrementally and accepts alias updates; in a shared mirror, the latest push wins a hostname-alias collision. DuckDB rebuilds once when its default machine key changes. Label edits take effect after a daemon restart and the next mirror push. Copying the data directory copies its identity; deleting the ID file creates a new installation without automatically claiming the previous one's history.


mariusvniekerk:
Should we use the identity record we generate for posthog for this?

wesm:
looking

wesm:
Agreed to handle the broader identity change in a follow-up agent session. Promote the existing persisted telemetry installation ID to an application-owned installation identity, with PostHog consuming it, and keep the human-readable machine name as a separate display label. Identity creation and use must work when telemetry is disabled.

The ID should survive AgentsView updates, binary replacements, restarts, and network changes because it lives in the data directory. A fresh data directory or deleted identity file creates a new identity; copying the data directory copies its identity. It identifies an installation, not physical hardware.

This PR remains the hostname-stabilization fix. The follow-up will adopt the installation ID for machine identity. Historical hostname aliases still require a separate evidence-based repair; the new ID alone cannot establish their ownership.

PR #1684, opened 2026-09-09
fix(copilot): import observed request usage without double counting

mariusvniekerk:
Copilot sessions now import observed per-call tokens from session-store.db, so usage can appear before shutdown. Each row receives request-level catalog pricing without requiring a transcript message ordinal. The latest shutdown reported cost remains authoritative.

A failed earlier store insert can coexist with a later successful row. Overlapping transcript output contributes only its positive per-model excess over store output; a recovered row replaces that excess without double counting. This remains a lower bound when extra store-only calls mask missing output, and it cannot recover missing input or request bands.

This is the second replacement for #1626. The dependent performance PR bounds refresh work for the shared store.

Merge order: #1683, then #1684, then #1685.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -597,7 +597,9 @@ add an archived or maintained mirror without replacing the original identity.
 
 - **Format:** Flat session JSONL or a session directory containing
   `events.jsonl`.
+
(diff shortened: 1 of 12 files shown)


wesm:
looking

PR #1685, opened 2026-09-09
perf(copilot): refresh shared-store usage without rereading transcripts

mariusvniekerk:
A Copilot store update now rereads usage payloads only for sessions whose latest row ID changed. Unchanged transcript hashes are reused across provider instances and engine restarts. Shared database timestamps affect freshness, not session timestamps.

The producer leaves sessions.updated_at unchanged when inserting usage, so the refresh uses the producer's session-and-row-ID index. Startup builds the store hashes once. Runtime updates follow appends and latest-row removal; historical edits below an unchanged maximum ID wait for a new engine or CLI sync. Missing or incomplete usage schemas are cached as empty usage for the current SQLite state, so metadata-only writes leave unchanged sessions skipped. Unreadable store markers and operational read failures remain retryable.

### DIFF
--- a/docs/internal/session-format-sources.md
+++ b/docs/internal/session-format-sources.md
@@ -679,6 +679,15 @@ add an archived or maintained mirror without replacing the original identity.
   calls can mask missing output, and missing input or request bands cannot be
   recovered. A recovered row replaces that excess without adding it twice.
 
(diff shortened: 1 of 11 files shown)


wesm:
looking

PR #1689, opened 2026-09-09
fix(db): scope usage rollup install reads to the snapshot

### LINKED ISSUE #1688: Idle serve daemon burns most of a machine's CPU: the usage-cache backfill re-reads every installed rollup row for each 256-session batch
### agentsview version

agentsview v0.42.0 (commit ff8fb4e8, built 2026-09-01T19:37:16Z), and still present on `main` at 988c7406.

### Install method

Built from source

### OS / platform

Ubuntu 24.04.4 LTS (x86_64), 24 cores

### Which agent and version

n/a (usage cache / serve daemon)

### Which model(s)

n/a

On a large archive the serve daemon sits at eight to nine cores with no clients connected and startup sync long finished. Expected an idle daemon to be idle.

The cost is the usage-cache backfill. `readUsageRollupInstalls` in `internal/db/usage_rollup.go` filters only on `timezone_key`:

```sql
FROM usage_rollup_timezones tz JOIN usage_rollup_installs i
  ON i.timezone_id = tz.id WHERE tz.timezone_key = ?
```

A backfill pass correctly narrows each batch to 256 sessions and calls `Ensure` per batch, but that read is unscoped, so every batch materializes *all* installed rows into four maps and discards them — twice per batch whenever anything is stale. The pass is therefore quadratic in archive size.

Measured on a 128,343-session archive with 141,454 installed rollup rows: one daemon lifetime spent 3,161 CPU-seconds and 27,798 GC cycles, nearly all of it in a five-and-a-half-minute window that ended when the backfill gave up with `could not stabilize after 3 attempts`. The allocation profile attributed 57% of allocation to `readUsageRollupInstalls` alone. Each of the 502 batches read 141,454 rows to verify 256. Against the 256 MiB soft memory limit the serve daemon sets for itself, that allocation rate is what turns a slow pass into sustained garbage collection.

The archive's cache had not finished a backfill in three days, so it was being rebuilt from scratch and abandoned on every daemon start.

Query plan for the read today:

```
SCAN i
```

### Steps to reproduce

1. Build an archive large enough that installed rollup rows greatly outnumber a 256-session batch (this was 128,343 sessions / 141,454 installed rows; the effect scales with the ratio, so it shows up well below that).
2. Start `agentsview serve` and leave it with no clients connected.
3. Watch CPU during the usage-cache backfill, or take a heap profile and look at the share of allocation under `readUsageRollupInstalls`.

Scoping the read to the sessions a snapshot can actually consult turns the scan into one seek per batched session against the unique `(timezone_id, session_id)` index — 257 rows instead of 141,454 per batch, plan `SEARCH i USING COVERING INDEX (timezone_id=? AND session_id=?)`. Rebuilt against the same archive, the pass completes in 4 minutes 6 seconds for 1,349 CPU-seconds and 14,397 GC cycles, and `readUsageRollupInstalls` falls to 0.6% of allocation. That was the difference between never finishing and finishing on this archive.

I have that change ready with a regression test and will open a PR against this issue.

### Checklist

srosro:
On a large archive the serve daemon burned most of a machine's CPU with no clients connected and startup sync long finished. Measured on a 24-core machine against a 128,343-session archive with 141,454 installed rollup rows, one daemon lifetime spent 3,161 CPU-seconds and 27,798 GC cycles, nearly all of it inside a five-and-a-half-minute window that ended when the backfill gave up with `could not stabilize after 3 attempts` — eight to nine cores sustained, of which the allocation profile attributed 57% to `readUsageRollupInstalls` alone. That read filtered only on `timezone_key`, so although a backfill pass correctly narrows each batch to 256 sessions and calls `Ensure` per batch, each of the 502 batches materialized all 141,454 installed rows into four maps and discarded them, twice per batch whenever anything was stale. The pass is quadratic in archive size, and against the 256 MiB soft memory limit the serve daemon sets for itself, that allocation rate is what turns a slow pass into sustained garbage collection.

This scopes the read to the sessions a snapshot can actually consult, binding them and the cursor row through the package's existing `maxSQLVars` and `inPlaceholders` seam. The unique `(timezone_id, session_id)` index turns the table scan into one seek per batched session, so a batched verification materializes 257 rows instead of 141,454 and the plan changes from `SCAN i` to `SEARCH i USING COVERING INDEX (timezone_id=? AND session_id=?)`.

Iterating `Versions` alone is sufficient because `Sessions` and `Versions` are appended one-for-one at capture, filtered by the same predicate in `dropDeleted`, reordered together, and sliced together per batch. Snapshots above the bind limit — the whole-archive verification at the end of a pass, and the foreground usage read — keep the unscoped query, which returns a superset of the same rows; that is safe because `installs` is only ever indexed by key and is never ranged over. Rollup contents, install revisions, staleness semantics, batch size, and the retry policy are unchanged, and no schema or column changes are involved, so the PostgreSQL and DuckDB paths need no matching update.

Reviewers should look at the bind gate in `readUsageRollupInstalls`, which reserves one parameter for the timezone predicate, and at `internal/db/usage_rollup_lifecycle_test.go`, where a real-database regression proves a batched verification does not materialize an installed session outside its batch.

Rebuilt and run against that same archive, the pass now completes in 4 minutes 6 seconds for 1,349 CPU-seconds and 14,397 GC cycles, and `readUsageRollupInstalls` falls to 0.6% of allocation; the remaining 63.5% is `loadUsageRollupFacts` building the rollups, which is linear and one-time. That archive's cache had not finished a backfill in three days, so it was being rebuilt from scratch and abandoned on every daemon start. This change does not alter how often a pass restarts — on an archive whose transcripts are appended continuously, a pass can still exhaust its three attempts against `errUsageCacheSourceChanged` — it just loses far less work each time, which on this archive was the difference between never finishing and finishing.

Closes #1688

https://claude.ai/code/session_01H6sY6VBHwhnwjuZq8kczcK

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -71,6 +71,13 @@ description: Release history for AgentsView
   print coarse phase updates. Redirected output omits terminal control
   sequences and per-session refresh lines. Interactive terminals retain live
   progress. (#1645)
+- Stop the server from consuming most of a machine's CPU while it sits idle on
+  a large archive. The usage-cache backfill verified each 256-session batch by
+  reading every installed rollup row, so the work grew with the square of the
+  archive size and the discarded rows kept the garbage collector saturated. The
+  backfill now reads only the rows for the sessions it is verifying. On a
+  128,000-session archive a pass that previously burned eight to nine cores for
+  several minutes and then gave up now finishes in about four minutes.
 - Price Codex Luna Reserve turns that persist as `gpt-reserve` using the
   existing GPT-5.6 Luna catalog rates. Usage reports still list `gpt-reserve`
   as the reported model. Existing SQLite usage caches rebuild so previously
--- a/internal/db/usage_rollup.go
+++ b/internal/db/usage_rollup.go
@@ -442,12 +442,30 @@ func readUsageRollupInstalls(
 	snapshot usageQuerySnapshot, fills map[string]usageFillResult,
 	pricingHash string,
 ) (map[string]usageRollupInstall, map[string]bool, error) {
-	rows, err := conn.QueryContext(ctx, `SELECT i.id, i.session_id,
+	query := `SELECT i.id, i.session_id,
 		i.fact_install_revision, i.install_revision, i.cached_at,
 		i.source_sync_marker, i.source_transcript_rev, i.usage_event_fingerprint,
 		i.baked_agent, i.baked_started_at, i.pricing_hash
 		FROM usage_rollup_timezones tz JOIN usage_rollup_installs i
-		  ON i.timezone_id = tz.id WHERE tz.timezone_key = ?`, identity.Key)
+		  ON i.timezone_id = tz.id WHERE tz.timezone_key = ?`
+	args := []any{identity.Key}
+	// Scope the read to the sessions this snapshot can consult, so a batched
+	// backfill pass stays linear in archive size instead of re-reading every
+	// installed row per batch. Sessions and Versions carry the same IDs in the
+	// same order from capture through ordering and batch slicing. Archive-scale
+	// snapshots exceed the bind limit and read unscoped, which returns a
+	// superset of the same rows.
+	scope := make([]string, 1, len(snapshot.Versions)+1)
+	scope[0] = usageRollupCursorSessionID
+	for _, version := range snapshot.Versions {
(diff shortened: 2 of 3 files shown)


wesm:
Thank you, will review and merge

PR #1690, opened 2026-09-10
fix(pricing): correct Codex costs on Amazon Bedrock

primshnick:
Fix missing or incorrect cost estimates for Codex sessions using GPT-5.4, GPT-5.6 Luna, GPT-5.6 Terra, or GPT-6 Astra through Amazon Bedrock.

These sessions report names such as `openai.gpt-6-astra`. Agentsview now uses Bedrock prices for those exact names, wherever they are reported. Direct OpenAI names such as `gpt-6-astra` continue to use their own pricing lookup.

Use historical Bedrock prices when available, so older Luna and Terra sessions use the prices from before the July 30 cut. Include Astra's Bedrock prices in the app so costs are available before an online price refresh. Recalculate cached usage totals with the corrected pricing.

Keep custom price overrides, regional prices such as GovCloud, and the model names shown in usage breakdowns.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -58,9 +58,10 @@ description: Release history for AgentsView
 
 **Bug fixes**
 
-- Price Codex Astra turns reported as `openai.gpt-6-astra` using the
-  `gpt-6-astra` catalog rates while keeping the reported model name in usage
-  breakdowns, including before a live pricing refresh succeeds.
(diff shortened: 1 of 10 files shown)


wesm:
looking

mariusvniekerk:
Does this pricing apply only for Bedrock?

mariusvniekerk (approved):

PR #1694, opened 2026-09-10
feat(data): add tool-result image cleanup endpoints and panel

rodboev:
Open Settings, choose Tool-result images, and click Preview to see how many stored tool results still hold inline image payloads, broken down by project with stored and decoded byte totals. Removing them replaces each payload with a readable placeholder. Until now `agentsview db strip --images` was the only way to do this, so it was out of reach for anyone not running the CLI.

Tool-result images settings panel showing a preview broken down by project

Two routes carry it under `/api/v1/data`: `strip-images/preview` reports the selection without writing, and `strip-images` performs it. Both accept an optional project substring and a `YYYY-MM-DD` bound on session end. The apply route requires `confirmed: true` in the body and returns 400 otherwise, so a bare POST cannot destroy stored content, and the panel pairs that with a dialog stating the rewrite cannot be undone. Both routes are localhost-only and reject a remote backend the same way archive compaction already does; preview also works against a read-only archive, and only the apply requires a writable one.

The apply route uses a new `tryArchiveWrite` helper rather than the existing `serializeArchiveWrite`. `serializeArchiveWrite` blocks until the engine's exclusive lock frees, which suits background scheduled work but leaves a user waiting behind a running sync pass. `tryArchiveWrite` calls `engine.TryRunExclusive` and returns 409 immediately when the lock is held, matching the foreground compact runner the daemon already injects. The five existing `serializeArchiveWrite` call sites are untouched. Removal rewrites stored rows in place, so the archive file keeps its size until `agentsview db compact` runs, which the panel states.

Refs #1718


mariusvniekerk (approved):

mariusvniekerk:
I wonder if we should have selective pruning so that you can basically age out old images. Basically keep month retention.

rodboev:
Yeah definitely, that's probably the primary use case.

Then the DateRangePicker from the homepage gets reused unchanged.

wesm:
rebasing

PR #1703, opened 2026-09-10
refactor(poller): share background job scheduling

jzila:
Claude usage collection will also require polling. This moves pricing refresh
onto a shared scheduler so both jobs can use the same timing and retry behavior.

Jobs are configured at startup, with jitter, cooldown, capped retry backoff,
Retry-After support, and manual triggering. Pricing remains the only job; it
refreshes at startup and serializes writes with resync. State stays in memory,
with no new configuration, API routes, or tables.

Failed pricing refreshes now retry after 1.5, 3, 6, and 12 hours before reaching
the usual 24-hour interval, with up to five minutes of jitter. Previously,
failures waited for the next daily tick. Shutdown waits for pricing work to
finish, including any pending sync/resync lock wait.

Two CI test expectations also match current session URLs and Windows temporary
paths.

Review `internal/poller` and `cmd/agentsview/pricing_schedule.go`.

### DIFF
--- a/cmd/agentsview/main.go
+++ b/cmd/agentsview/main.go
@@ -459,7 +459,11 @@ func runServe(cfg config.Config, opts serveOptions) {
 		pricingRefreshRunner = engine
 	}
 	seedPricing(database, pricingRefreshRunner)
-	poller.Start(ctx, pricingRefreshJob(database, pricingRefreshRunner))
+	scheduler := poller.Start(ctx, pricingRefreshJob(database, pricingRefreshRunner))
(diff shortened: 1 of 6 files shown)


jzila:
@mariusvniekerk I feel much better about this. I took a machete to what the agents produced and positioned it properly for subsequent integration into use cases that might need to poll, like Claude usage window tracking.

mariusvniekerk:
Yeah that's much less. Thanks

wesm:
looking

PR #1712, opened 2026-09-10
feat(usage): estimated energy mode alongside cost and tokens

jzila:
This adds a third mode to the Usage page's Cost/Tokens picker: Energy. It switches the summary cards, the cost/time chart, the attribution panel and treemap, top sessions, the pairwise comparison, and the session-detail usage breakdown to an estimated watt-hour figure for the same rows, computed at read time across SQLite, DuckDB, and PostgreSQL. Top sessions can now be ranked by energy the same way it already ranks by cost or tokens. A figure that includes usage with no published rate for its model is marked with a trailing asterisk instead of quietly reading as a smaller, complete number. A Settings > Preferences panel lets a user pick which point on the estimate's low/mid/high band the app reports. A small marker next to every energy figure explains what it means on hover.

No vendor publishes a real "Wh per token" number, so the estimator leans on something that does exist: list price. Anthropic and OpenAI both charge less for cache reads and more for output than for a plain input token, and those ratios already track how expensive each token type is to serve. Two outside attempts at estimating Claude's energy use lean on the same assumption. That is reassuring, not proof; list price also carries margin and demand that energy draw does not.

The estimator has three steps, all in `internal/energy`. First, a small committed dataset of published energy measurements (an Epoch estimate for ChatGPT, Google's own Gemini disclosure, two independent third-party Claude estimators, and a couple of GPU benchmark papers) gets normalized onto one scale, watt-hours per million output tokens, using each row's own input/output price ratio to fold input tokens into an output-token-equivalent count. Second, those normalized points get a weighted least-squares fit in log-log space, output-token price against output-token energy, so a vendor's list price converts into a watt-hour figure. A vendor with three or more calibration rows of its own gets its own fit; everyone else falls back to the pooled fit across every vendor. Third, a session or usage row's token counts by type get converted into an output-token-equivalent using the model's own price ratios, then multiplied by the fitted watt-hour rate for that model.

The result is not presented as precise. The low and high scenarios come from the fit's own residual spread rather than a hand-picked multiplier, and both the doc and the in-app estimate marker say plainly that this is anchored on a handful of measurements and carries something like 2x uncertainty either way. A model with no catalog price gets no energy figure, the same way it gets no cost figure today. An operator can override the fitted rate for a specific model in config.toml if they have a better number for their own deployment.

Reasoning and thinking tokens from Claude and Codex are already counted inside output tokens by those providers, so the estimator does not add a separate term for them; this is checked against the evidence already tracked in `docs/internal/session-format-sources.md`. One thing worth flagging for review: a late pass re-checked every calibration source against what it actually said, rather than against an earlier summary of it, and that changed the pooled fit's price sensitivity from roughly 0.49 to roughly 0.22. The methodology doc and its generated tables reflect the corrected numbers, so they are worth reading next to the code rather than skimmed.

For the model itself, start with `internal/energy/fit.go` and `calibration.go`, then `internal/energy/estimate.go` for how a token mix becomes a watt-hour number, and `docs/internal/energy-model.md` for the worked examples and full source list. On the read path, `internal/db/energy.go`, `internal/duckdb/energy.go`, and `internal/postgres/energy.go` each wire the same estimator into their existing cost computation rather than growing a parallel cache.

There is one exception to the read-time-only rule, and it is deliberate. The usage-facts rollup gains a single column, `energy_billable_output_tokens`, and `usageCacheFormatVersion` moves from 11 to 12 in `internal/db/usage_cache_schema.go`. A reasoning-only row can have zero output tokens and nonzero reasoning tokens, and the cost pipeline already falls back to the reasoning count for that case. The energy estimate needs the same fallback applied per fact, before facts merge into a rollup row. Once two facts are summed, the merged counters no longer say which fact was reasoning-only, so recomputing the fallback at read time from the sums gets the wrong answer for exactly the sessions where it matters. Storing the per-fact value is the only way to keep it correct. The cost of the bump is a one-time cold rebuild of the usage cache on the first read after upgrading. The cache is a disposable sibling database, the SQLite archive is untouched, and nothing is lost, but on an archive with tens of thousands of sessions that first Usage page load can take several seconds to a couple of minutes before it is warm again. A handful of correctness details live in that same layer, all worth a close look: the scenario a user picks in Settings can change while a usage read is already in flight on another goroutine, so every backend now builds one estimator up front per request and guards the underlying config with a lock; `duckdb serve` previously ignored a configured scenario or override entirely and always estimated at mid, which is now fixed the same way the SQLite and PostgreSQL servers already worked; the per-model override and vendor-family lookup now key on the pricing catalog's matched pattern rather than whatever model name the cost resolver happened to return, since the two can diverge after alias handling and a mismatch would make a correctly written override silently do nothing; and a row whose output tokens are zero but reasoning tokens are not (an edge case the cost pipeline already has a fallback for) now estimates at the reasoning-token count instead of reading as exactly zero energy for work that genuinely happened. A session that delegates to Claude Task-tool subagents now folds their energy into the parent session's total the same way it already folds in their cost, in `internal/service/session_usage_subagents.go`; that combining logic predates energy and previously dropped it entirely for any session with delegated work. On the frontend, `frontend/src/lib/energy.ts` holds the formatting and the TV-equivalence wording, and `EnergyEstimateMark.svelte` is the one component every energy figure on the page shares for its tooltip.

One known gap: the scenario picker in Settings does not currently propagate to a Usage page tab that was already open before the change, so a second tab can keep showing figures from the old scenario until it is reloaded. Fixing that properly means either the backend pushing a change notification or the frontend polling for one, and neither felt worth the added complexity for a preference a user sets once and rarely revisits. Reviewers should treat this as accepted for now rather than something the PR tries to hide.


mariusvniekerk:
I'm not really sure we want to incorporate this feature given that by its very estimation, the numbers are iffy. Prices are a pretty solid approximation for energy usage already, and so we do not need this.

This kind of tool probably fits more as something adjacent to agentsview than being incorporated into it.

Also, please tone it down with your PR descriptions. They are incredibly long.

jzila:
Completely fair, this one was mostly for fun :)

jzila:
@mariusvniekerk I closed a bunch of PRs. The agents went off the rails and put the PRs up for review on their own accord. That's on me, I am still annealing to the right level of trust in the newer models. Trying to fix and reimplement the remaining ones.

PR #1714, opened 2026-09-10
fix(parser): apply Claude queue links incrementally

naveenspark:
## Summary

- Apply Claude queue and progress subagent links incrementally.
- Preserve deterministic ordering and existing first-link precedence.
- Keep link-only appends on the incremental sync path.

## Why

Active Claude sessions append queue and progress events. Replacing the stored session for each event repeatedly parses large JSONL files and drives sustained CPU use.

## Tests

- go test ./internal/parser -count=1
- focused incremental sync tests for Claude queue and progress links
- focused database incremental-link tests
- go vet ./internal/parser


wesm:
looking

PR #1720, opened 2026-09-11
fix(appearance): unify zoom controls with a configurable default

### LINKED ISSUE #1697: Persist appearance preferences
### Problem / motivation

First, thanks for the wonderful tool, it's been a mainstay for a while now!

Second, I find myself constantly re-adjusting the text size option to be 120% rather than the default 100% and it doesn't appear that I can persist this in the `config.toml` file. It's worth noting that if I make the size change in the settings menu, then it persists, but if I do it from the little status bar at the bottom, then it doesn't and that the two compound each other if adjusted simultaneously.

### Proposed solution

The ability to be able to persist appearance options like text size among the others available.

### Alternatives considered

Maybe there's an alternative here that I'm not privvy to, but even for the sake of being portable across multiple machines, a configuration option would seem to make sense.

rodboev:
Appearance, the desktop status bar, and desktop shortcuts use one Zoom preference in place of the separate Text size and Desktop zoom controls. Existing local preferences migrate to it, and all controls use the same supported percentages.

The optional `zoom_level` in `config.toml` supplies a default for browsers and desktop webviews without a local preference. Interface changes stay in that client’s local storage, so clients with different displays keep independent choices. Manual config edits require a daemon restart and client reload; explicit settings API updates retain validated TOML persistence.

Fixed typeahead menus, context menus, and chart tooltips keep their viewport positioning while their contents scale with the interface. The date picker retains its existing scale because the pinned component fixes its panel width. The pinned kit-ui integration and browser coverage are documented in `DESIGN.md`.

Closes #1697


wesm:
looking

PR #1722, opened 2026-09-11
feat(parser): support Cline CLI sessions

brynjmsdlnn:
Cline CLI task sessions were not previously surfaced in AgentsView. This adds discovery, synchronization, and parsing support for sessions stored in `~/.cline/data/sessions/`.

### What it does
- Discovers task sessions from `<sessionId>/<sessionId>.json` (metadata, exit code) and `<sessionId>/<sessionId>.messages.json` (transcript), skipping dot- and underscore-prefixed directories.
- Parses user messages, assistant text, extended thinking blocks, tool calls, and tool results (handling both string and array content blocks).
- Extracts per-message token metrics (`inputTokens`, `outputTokens`, `cacheReadTokens`, `cacheWriteTokens`) from assistant turns for multi-step breakdown and peak context tracking.
- Preserves authoritative `metadata.totalCost` (including explicit zero costs) via non-duplicating aggregate usage events.
- Normalizes Cline tools (e.g. `read_files`, `replace_in_file`) into canonical taxonomy categories.
- Coalesces continuation snapshots for the same Cline teammate/subagent into one stable session, so the UI does not show duplicate rows for every transcript snapshot; genuinely distinct runs remain separate.
- Adds Cline-specific HTTP and SSH sync resolvers that emit curated session files, preserving empty file lists when all sessions are deleted so client mirrors can evict stale files without recursive root archiving.
- Adds Cline provider configuration (`CLINE_DIR` env var, `cline_dirs` config key), UI metadata, and format provenance documentation in `docs/internal/session-format-sources.md`.

### Safety and correctness hardening
- **Safe SSH & HTTP remote sync**: Preserves empty file-scoped targets when all sessions are deleted so client mirrors can authorize eviction without triggering recursive archiving of the root.
- **Physical containment & symlink guards**: Enforces physical root containment on configured roots, direct sessions roots, `data`, and `data/sessions` ancestors across both HTTP remote sync and SSH resolvers (`curatedRoot` and `[ -L "$target" ]`), and rejects symlinked session directories or files escaping the configured Cline root.
- **Root scanning guard**: Eliminates fallback to scanning the configured root when `data/sessions` is absent in `clineResolveSessionsDir`, preventing non-session directories like `~/.cline/settings` from being traversed.
- **Tar injection protection**: Rejects backslashes (`*'\'*`) in SSH Cline session IDs to prevent GNU tar escape processing from introducing path traversal during remote archive creation.
- **Terminal tool completion**: Handles `attempt_completion` as an explicit terminal tool call in session classification, ensuring successful Cline tasks ending on completion are marked `TerminationClean` rather than `TerminationToolCallPending`.
- **Stable teammate identity and reconciliation**: Derives continuation-chain IDs from stable chain roots rather than winner order, reuses persisted identities from the continuation chain when an earlier snapshot disappears, and tombstones missing teammate sessions while allowing them to revive when files return.
- **Remote identity stability**: Normalizes stored Cline IDs before parsing so remote host prefixes are applied exactly once and teammate links remain unchanged across resyncs.
- **Multi-run link safety**: Keeps `team_run_task` links only when a teammate run can be identified unambiguously, avoiding links to the wrong session when multiple runs share an agent ID.
- **Path containment & traversal guards**: Added strict validation rejecting dot-prefixed directories (`.secret`), directory separators, and `..` traversal in session IDs, plus physical containment checks that reject symlinked session directories or files escaping the configured root.
- **Symlink and freshness handling**: Rejects unsafe Cline primary files, preserves provider-specific JSONL companion symlink behavior, and uses stable tool-call locations so later tool results are attached after parsed-message slice growth.
- **Content-block decoding**: Preserves raw JSON payloads in `ContentRaw` while normalizing array-shaped `tool_result` content blocks.

### Tradeoffs and Limits
- Session discovery targets Cline CLI task directories (`~/.cline/data/sessions/<id>/`). Cline VS Code extension sessions stored in extension `globalStorage` are not tracked by this provider.
- Teammate continuation snapshots are intentionally represented as one logical UI session. Independent runs for the same subagent are retained separately, but ambiguous `team_run_task` calls are left unlinked rather than guessed.
- Token metrics are recorded at the turn level when present; aggregate usage events carry authoritative total cost without duplicating tokens.

### Where to look
- `internal/parser/cline.go`: Session and message parsing, teammate continuation coalescing, stable identity recovery, token metrics extraction, authoritative cost preservation, and tool call pairing.
- `internal/parser/cline_provider.go`: Directory discovery, physical path classification, symlink rejection, and traversal guards.
- `internal/sync/engine.go`: Cline stored-identity hints and complete-source teammate reconciliation.
- `internal/remotesync/resolve.go`: Remote target resolution, ancestor symlink validation, and empty-root retention for stale eviction.
- `internal/ssh/resolve.go`: SSH resolution script emitting curated session files, ancestor symlink validation, and backslash rejection.
- `docs/internal/session-format-sources.md`: Format provenance and upstream Cline schema evidence.

### DIFF
--- a/cmd/agentsview/cli.go
+++ b/cmd/agentsview/cli.go
@@ -926,6 +926,7 @@ func writeRootHelp(w io.Writer, root *cobra.Command) {
 	fmt.Fprintln(w, "  COPILOT_DIR             Copilot sessions or exported JetBrains Copilot directory")
 	fmt.Fprintln(w, "  GEMINI_DIR              Gemini CLI directory")
 	fmt.Fprintln(w, "  OPENCODE_DIR            OpenCode data directory")
+	fmt.Fprintln(w, "  CLINE_DIR               Cline sessions directory")
 	fmt.Fprintln(w, "  CURSOR_PROJECTS_DIR     Cursor projects directory")
 	fmt.Fprintln(w, "  IFLOW_DIR               iFlow projects directory")
 	fmt.Fprintln(w, "  AMP_DIR                 Amp threads directory")
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1529,6 +1529,7 @@ agentsview help
 | `COWORK_DIR`                      | (platform-specific)                                  | Claude Desktop cowork sessions directory                                                            |
 | `CODEX_SESSIONS_DIR`              | `~/.codex/sessions`                                  | Codex sessions directory                                                                            |
 | `CODEX_HOME`                      | unset                                                | Codex home that re-roots the default `sessions/` and `archived_sessions/` discovery paths           |
+| `CLINE_DIR`                       | `~/.cline`                                           | Cline sessions directory (discovers under `<root>/data/sessions/` or direct sessions root)          |
 | `COMMANDCODE_PROJECTS_DIR`        | `~/.commandcode/projects`                            | Command Code projects directory                                                                     |
 | `COPILOT_DIR`                     | `~/.copilot`                                         | Copilot CLI sessions directory                                                                      |
 | `CORTEX_DIR`                      | `~/.snowflake/cortex/conversations`                  | Cortex Code conversations directory                                                                 |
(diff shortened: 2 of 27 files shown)


mariusvniekerk:
Taking over here, this one has ballooned out of hand

brynjmsdlnn:
Thanks for taking over this PR! I really appreciate it. I reached the point where the latter part of the review was getting beyond what I could confidently follow, and I wasn't able to continue working through it, so I'm glad you're taking it from here. 😅

Hopefully the remaining changes won't give you too many headaches. Thanks again for taking the time to work through all of this!

PR #1727, opened 2026-09-11
test(timing): reject new sub-second polling budgets

### LINKED ISSUE #1702: Roadmap: deterministic timing tests with testing/synctest
Go tests across the tree wait on concurrent work by giving it a fixed budget and polling until the budget runs out. Roughly 260 such waits exist today. A budget is a guess about how fast the machine is, so on a loaded Windows runner the slow ones fail and the failure has nothing to do with the change under test. `testing/synctest` replaces the guess with a synchronization point. This issue tracks the direction so each focused pull request can stay reviewable without losing the larger context.  ## Outcomes  - Tests that wait on in-process work finish when the work finishes, with no budget to tune and no dependence on how many cores the runner has free. - Tests that genuinely depend on the operating system keep an explicit budget, and the budget is visible at the call site rather than hidden in a per-file constant. - The Windows job stops contributing timing failures that are unrelated to the code being tested. - New budgets do not quietly reappear as the tree grows.  ## Delivery sequence  1. Convert `cmd/agentsview`. The package already uses synctest in several files, so the pattern needs no introduction, and its database-backed tests prove the approach against the storage layer.   PR: #1708 2. Convert `internal/sync`, excluding the filesystem watcher tests. Largest concentration of waits in the tree, and the watchers sit on the other side of the boundary below.   PR: #1709 3. Convert the remaining packages. Smaller and more uniform than the first two.   PR: #1710 4. Clear the budgets that survive conversion and add a check that fails on a new bare one. Ship together since the check holds only once nothing is left to climb back from.
  PR: https://github.com/kenn-io/agentsview/pull/1727  Conversions land per package. One pull request per call site is the wrong unit at this count.  ## Design constraints  - The boundary is where a goroutine parks, not which package it lives in. A wait converts when every goroutine it depends on either runs to completion or parks on a Go channel, mutex, or timer. - A goroutine that stays parked on an operating system event source for the life of a test cannot be bubbled, and a bubble containing one never finishes. Filesystem watchers and the FSEvents tests are that shape and keep budgets. - SQLite through cgo works inside a bubble, so database-backed tests are convertible. - Conversions change tests only. Production code and behavior stay as they are. - A converted test must fail for the same reasons it failed before. Removing a budget removes a class of false failures, never a real assertion.  ## Adjacent  - #1700 and #1701: one instance of the class, addressed by raising that file's budget. Step 1 stacks on #1701 and converts that file with the rest of the package. - Playwright flakes on the e2e job are a separate source with a separate fix, tracked on their own.

rodboev:
Lint and pre-commit checks now reject new sub-second literal budgets in `Eventually` and `Never`, completing the prevention step after the conversions in #1708, #1709, and #1710.

The Go checker evaluates budget expressions and preserves the remaining integration waits through explicit allowances tied to test names, durations, and occurrence counts. Named constants and polling intervals stay outside the rule.

Closes #1702


mariusvniekerk:
Yeah, this is great. I love it.

Issue #1718, opened 2026-09-11
Roadmap: move tool-result images out of SQLite into the embedded Docbank vault


rodboev:
AgentsView stores tool-result images inline as base64 in `tool_calls.result_content` and `tool_result_events.content`, so an archive keeps growing with image bytes even when the user wants only the conversation text. The measurement in #1622 found 9.57 GB of inline image payloads in each of those two tables on a 19.96 GB `sessions.db`, about 96% of the file, with zero freelist pages to reclaim.

Those bytes belong in the Docbank vault AgentsView already embeds under `dataDir/artifacts`. mariusvniekerk proposed that incorporation point on #1622, which turned an opt-in cleanup request into a storage move: rows keep a placeholder carrying media type, decoded size, and content hash, the bytes live in the vault under their own namespace and retention rules, and the viewer reads them back through a daemon route. The cleanup commands and the settings surface for them have shipped. This issue tracks the rest, which makes offload the ingest path rather than a maintenance step and then builds on the stored bytes.

## Outcomes

- A new session never writes an inline image into SQLite, so the archive stops growing with base64 that a reference covers.
- A crash during ingest leaves an unreferenced vault object rather than a row pointing at bytes that were never written.
- The PostgreSQL push, the DuckDB mirror, artifact export, and session export carry the reference or the placeholder without special casing.
- Opening an image stays fast locally without a second store of inline bytes growing behind the vault.
- Text inside a stored image is searchable, keyed to the image rather than copied into the message body.

## Delivery sequence

1. Retention policy and `db strip --images`. Shipped in #1663. `tool_result_images = "keep" | "drop"` carries through full imports, incremental writes, late results, remote import, and full resync, and the command rewrites existing rows, reaching trashed, source-missing, and orphan-event rows that a resync cannot.
  PR #1693
  PR #1694
2. `db migrate --images`, which moves retained inline payloads into the embedded vault and publishes the reference.
  PR #1696
3. Offload on ingest. Add a third policy mode so images never land inline: the parse path decodes and stores at insert time and rows carry references from the start. The vault write becomes durable before the row referencing it commits, which mirrors Docbank's own bytes-before-reference rule. Each downstream backend needs checking, and a PostgreSQL or CockroachDB deployment has no local vault, so remote viewers get placeholder text, a documented limitation of the same class as other backend capability gaps.
  PR #1711
[...]

PR #1729, opened 2026-09-11
fix(sync): offload tool-result images at ingest

rodboev:
When you set `tool_result_images = "offload"`, AgentsView moves supported tool-result images into its asset store during ingestion and keeps readable references in the archive. New sessions retain image access without storing image bytes inside tool-result text.

The setting applies to new and resynced sessions, incremental updates, and late tool results. AgentsView publishes each asset before saving its reference, and keeps the inline image if publication fails. The Settings control accepts `offload`, and existing inline images can be moved with `agentsview db migrate --images`.

Refs #1718 (slice 3b)


wesm:
Kind of makes me think that this should be the default for new archives

PR #1730, opened 2026-09-11
feat: derive hosted raw captures into PostgreSQL sessions

salmonumbrella:
Adds opt-in server-side derivation for hosted raw sync. Accepted captures now feed a bounded worker, an isolated Linux parser, and a transactional PostgreSQL projection. Hosted browsing no longer needs a local SQLite archive or `pg push` to turn raw captures into sessions. Local sync and hosted derivation share the preparation path; the existing artifact wire format stays unchanged.

The hosted store enforces tenant isolation and maps storage IDs to public session IDs. Copies with the same provider identity and content share a session. Curation survives matching copies, conflicting copies remain explicit, and deleting a source removes only its contribution. Publication is transactional; retries and pending-signal maintenance are bounded.

Deployment remains one tenant per server instance, with explicit owner provisioning and a restricted runtime role. Disabling derivation preserves hosted reads. Embedding work is queued durably; its consumer, device-enrollment UI, and broader lifecycle controls remain separate work.

Continues item 4 of #1352 after #1607. Operator setup and limits are in `docs/hosted-raw-sync.md` and `docs/pg-sync.md`.


wesm:
rebasing

wesm:
Pushed targeted fixes while keeping the foundations used by the later PRs.

- Piebald chats now have identities tied to their capture source. Chat 1 on two devices stays two separate sessions, and forks still point to the right parent.
- A shorter exact-prefix copy now shares the longer compatible transcript instead of creating a conflict. Each device keeps its actual captured revision, so removing the longer source restores the shorter copy. Conflicting messages or metadata still produce variants.
- Usage-only storage can clear an old automated-session label when a fresh prompt shows an ordinary conversation.
- Selecting an upload for parsing no longer expires browsing cursors. Expired cursors return a client error, and the sidebar and activity report reload instead of repeatedly using the stale cursor.
- Hosted derivation rejects image offload with a clear instruction to use keep or drop, instead of silently retaining inline images.

The source records, tenant protections, curation, and embedding queue remain in place for the dependent PRs. This does not resolve the full-revision storage growth or complete the deployment and retention work tracked in #1352.

The PostgreSQL integration suite, database and ingest suites, hosted CLI checks, and 254 frontend tests passed. The broader CLI run hit an IPv6 localhost artifact-exchange failure that also reproduces with the original code.

wesm:
NOT AN ISSUE

PR #1736, opened 2026-09-12
feat: build hosted embeddings and activate complete search generations

salmonumbrella:
Hosted PostgreSQL can now build embeddings from projected sessions and serve semantic and hybrid search without a local SQLite archive. The server owns encoding, durable retries, and current-source coverage. Unchanged chunks can reuse vector values after a raw append replaces their physical session IDs.

Operators provision a versioned generation with an owner connection, then enable its worker on the restricted runtime target. A rebuild keeps the previous generation serving until the replacement has complete coverage. Each search uses one generation and its matching encoder throughout the request. Usage-only mode clears derived embedding content; keyword search remains available when the active embedding profile cannot be used.

This implements item 5 of #1352 and depends on #1730. While that parent is open, this PR against main includes its PostgreSQL projection changes as well. The embeddings feature itself is commit 5be5df10; the parent must land first, then this branch can be refreshed to leave only the embeddings diff. Generation storage, worker lifecycle, and configured-runtime acceptance tests are the main review areas. Retention and garbage collection remain part of roadmap item 6.


wesm:
this will probably have to wait until after 0.43.0 is released (should be this weekend)

PR #1757, opened 2026-09-13
feat(usage): show Codex rate limits and sparklines

jzila:
Codex rollout files already include rate-limit observations. This shows them
in Usage alongside token and cost totals.

Current windows, reset times, and credits are grouped by machine and limit.
Sparklines follow the selected dates, with at most 512 samples per limit;
current readings remain independent of that range. Observations use the
existing SQLite sync lifecycle. Missing values remain unknown. This is
Codex-only and unavailable in remote mode, with no additional polling or
account inference.

Review `internal/parser/codex.go`, `internal/db/rate_limits.go`, and
`frontend/src/lib/components/usage/RateLimitsSection.svelte`.


jzila:
@mariusvniekerk want to take an initial look if this is directionally aligned? I would have two major follow-ups:

1. Claude support (with a poller)
2. Desktop notification support, in both the browser and app

PR #1758, opened 2026-09-13
Include complete project identity in reporting exports

wesm:
Reporting exports now include project identity evidence alongside combined usage cells. An archive-scoped label key does not identify a repository: unrelated repositories can share a display name, and a separate catalog read can describe a different snapshot.

This completes the unreleased reporting schema 4 contract: joint cells and their identity evidence ship together. It does not extend a released schema-4 contract or introduce schema 5. Default schema 3 is unchanged. Development consumers must use the complete shape and refresh digests from earlier development builds.

The `joint.projects` catalog is keyed by the cell's project key and covered by the hour digest. A project resolves only when every session contributing activity or canonical usage has matching, consistent source identity. Missing evidence remains unknown; conflicting identities remain ambiguous. Neither borrows an identity from a known sibling. Identity-only changes therefore trigger ordinary digest-based correction discovery.

The exporter reuses stored session snapshots and the existing identity resolver, reading them in bounded batches once per export. It does not inspect live checkouts, expose raw paths or session identifiers, or change cell grouping and accounting. Consumers still own authorization; opaque machine-root keys alone cannot prove a configured path matches. The synthetic SQLite benchmark now includes resolved repository snapshots at its one- and five-minute precisions.


wesm:
rejected, schema v4 was never released

PR #1765, opened 2026-09-14
feat(timing): separate measured tool time from unattributed turn time (#1753)

### LINKED ISSUE #1753: Separate turn time spent on thinking, messages, and tool calls
# Summary

Session and turn durations are currently derived from the gap between consecutive message timestamps, so a turn containing a bash call displays its entire elapsed time as bash run time.
This makes slow sessions look like slow tool executions even when most of the time was model thinking or response generation.
The viewer should attribute turn time to activity types: thinking, assistant message generation, and tool calls.
That way, someone diagnosing a slow session can see what actually consumed the time.

# Background

A turn's elapsed time is really a mix of activities: model thinking before visible output, assistant text generation, waiting for tool calls to execute, and unattributed gaps between entries.
Today the inter-message gap for a turn with tool calls gets reported as tool execution time, which misattributes the slowest and most common source of delay: thinking.

The messages table already stores per-message timestamps plus has_thinking, thinking_text, and has_tool_use, but durations are not stored per activity, and parsers collapse thinking, text, and tool_use blocks of one assistant message into a single row, so within-turn boundaries are lost at sync time.

Source formats differ in what is recoverable: Claude Code JSONL has per-entry timestamps for thinking, tool_use, and tool_result entries, while other agents may only report one timestamp per assistant message.
The breakdown should degrade gracefully: attribute what the source supports, and label the remainder as unattributed instead of guessing.

# Version

532fd65

# Acceptance Criteria

## Must

- [ ] Turn time is attributed across thinking, assistant message, and tool call activities instead of one gap assigned to a single activity
- [ ] The session detail view shows the per-turn activity time breakdown
- [ ] Time that cannot be attributed from source data is shown as unattributed rather than counted as tool time
- [ ] Unit tests cover attribution for a turn containing thinking followed by a tool call
- [ ] The per-turn activity breakdown is available with matching aggregation across all supported backends (SQLite, PostgreSQL, DuckDB)

## Should

- [ ] Claude Code sessions (per-entry JSONL timestamps) produce a fully attributed breakdown; other formats degrade to unattributed
- [ ] Session-level analytics (for example active duration) can report the same activity breakdown
- [ ] The breakdown names the specific tool that consumed the time within a turn

---

Created with [create-issue](https://github.com/tomzx/agents/blob/db0e46d6c0cf5e10b1e8a8c89a9c7f6040dbb0a3/skills/create-issue/SKILL.md) via glm-5.3-flash (`db0e46d`)

rodboev:
Session timing now shows where a turn's time went. Session Vitals separates measured tool execution from the time the archive cannot assign to a phase.

The payload comes from one reducer shared by SQLite, PostgreSQL and DuckDB. It clips and merges paired execution intervals, removes the old whole-turn call fallback, and carries the remainder as unattributed time. Existing transcript and navigation behavior stays intact.

The timing route and live session updates carry the same fields. The UI links each activity row to its transcript ordinal. Claude rows without stored phase endpoints remain unattributed, so the view does not infer model time from content flags.

Paired `tool_execution` events currently come from the Copilot, Grok and OpenCode Review parser paths. Closed child session boundaries also measure linked Task calls. Calls without either boundary remain unknown, and their unresolved turn time stays unattributed.

| Before | After |
| --- | --- |
| Session Vitals before timing activity | Session Vitals with timing activity |

Screenshots show the previous completed-turn and timeline view beside the new Turn activity section with measured tool time and unattributed time.

Refs #1753


wesm:
rebasing

rodboev:
Fixing the roborev findings, they should have shown up on the latest review again but didn't

PR #1766, opened 2026-09-14
feat(server): add bounded image byte cache with age eviction

rodboev:
AgentsView keeps recently served image bytes in memory for up to seven days. Reopening a stored image can reuse bytes already read by the asset route, while the durable copy stays under dataDir/assets.

Entries expire seven days after creation. The cache holds 64 entries and 64 MiB, clears with the daemon, and falls back to the durable file after expiry or a source change. Asset validation and response headers stay the same. The cache is built on jellydator/ttlcache/v3, the same library Forge uses, which owns the expiry, the entry cap, and the byte bound; a one-minute sweep releases expired bytes.

The browser owns its HTTP cache and Blob URLs, so hit rates and visible latency remain unmeasured. Builds on [PR 1696](https://github.com/kenn-io/agentsview/pull/1696) and [PR 1729](https://github.com/kenn-io/agentsview/pull/1729).

Refs #1718

Also carried on this branch: the roborev review guidelines gain a rule against recommending gates in front of reads of user-owned files.


cpcloud:
If the image is read again after 7 days, does it get recached?

rodboev:
Yes. After 7 days the clock starts over.

cpcloud (approved):
LGTM.

PR #1773, opened 2026-09-14
feat: add Augure CLI/Desktop agents, agent remap rules, and duplicate-session suppression

scross01:
## Background: the Augure agents

[Augure](https://augureai.ca) ("the sovereign Canadian coding agent") ships
three clients, and each stores its sessions in a different place:

- **Augure CLI** — a rebranded fork of codex-rs (the same situation as
  TraeX, which agentsview already supports). It writes byte-compatible
  Codex rollout JSONL under `~/.augure/sessions/YYYY/MM/DD/`. Adding it is a
  registry entry plus a relabel of the shared Codex-format provider onto an
  `augure:` ID namespace, mirroring `internal/parser/traex.go`.
- **Augure Desktop v2** (goose-based, pre-beta) — an Electron shell whose
  conversations land in the **standard Goose store**
  (`~/.local/share/goose/sessions/sessions.db`). Nothing new to parse: the
  existing `goose` agent already ingests them. The only decision is labeling.
- **Augure Desktop v3** (current beta) — replaced the goose core with a fork
  of Hermes Agent (Nous Research). The runtime renames Hermes's `~/.hermes`
  to `~/.augure-desktop` and writes a Hermes-schema `state.db`
  (`schema_version` 26 on an independent version line). The existing Hermes
  state-DB parser already handles the shape, so this reuses the Hermes
  provider through a spec seam, relabeling onto `augure-desktop:`.

All three findings were verified against live installs (CLI 1.0.6, Desktop
2.x, Desktop 3.0.0-beta.7) on 2026-09-11; details are recorded in
`docs/internal/session-format-sources.md`.

## Why the enhancements are needed

**Agent support alone is not enough — the stores overlap.** Desktop v3
migrates conversations out of the shared Goose store without deleting them,
and early Augure CLI releases wrote Codex-format rollouts into the stock
Codex store. Measured against a live archive on 2026-09-14:

- 113 of 114 `goose` sessions have an `augure-desktop:` twin with the same
  `started_at` second — the same conversation ingested twice from two
  stores.
- 24 of those `goose` sessions carry Augure-hosted model slugs
  (`ossington-*`, `rosedale-*`, `tofino-*`); 90 have none and are genuine
  Goose.
- 5 archived `codex` sessions carry Augure model slugs.
- The twins are **partial duplicates**: the old-store copy of
  `goose:20251023_10` has 5 messages / 3,067 output tokens; its
  `augure-desktop:` twin has 4 messages / 0 tokens — migration copied the
  transcript but not the usage. The inverse also exists: a session resumed
  in the new client after migration makes the new copy a prefix plus
  continuation.

Investigation proved **no schema-level discriminator exists** in
the Goose store, so the parser cannot separate Augure-hosted sessions from
genuine Goose ones. The user can: message model slugs are a reliable
post-parse signal. Hence the three enhancements on this branch:

1. **Agent remap rules** (`feat(data): add agent remap rules…`) —
   user-defined rules (source agent + optional model GLOB + optional ID
   prefix) → target agent, applied to archived sessions with preview/apply
   and a consistency token, mirroring the existing worktree project-mapping
   feature. The parser phase stays untouched; users tune classification to
   their own setups (e.g. `goose` + `ossington-*|rosedale-*|tofino-*` →
   `augure-desktop`). Rules are stored in a new `agent_remap_rules` table,
   served through settings routes, applied incrementally on new sessions,
   and mirrored to PostgreSQL/DuckDB.
2. **Duplicate-session groups** (`feat(data): … duplicate-session
   suppression`) — client migrations duplicate sessions across stores, so
   the same interaction can be counted twice. This adds deterministic
   duplicate-group detection (same machine + started second + first-message
   prefix), a **presence-aware suppression** rule in the usage cache so a
   duplicate is only suppressed when its canonical copy actually carries
   usage (never losing usage that exists on only one side), UI indicators
   (sidebar badge `⧉2`, detail banner, Data-page review card), and PG/DuckDB
   parity. Nothing is ever deleted — the archive's hard rule — roles are
   re-derived on every rebuild.
3. **Typeahead selectors on the remap form** (`feat(data): add typeahead
   selectors…`, `fix(data): show raw slug case…`) — the remap form
   originally asked for raw slugs (`goose`, `augure-desktop:`, ID prefixes)
   as free text; a typo silently creates a rule that matches nothing. The
   source/target/prefix fields are now searchable dropdowns fed by the
   archive's agents (`/api/v1/agents`) merged with the built-in catalog,
   sorted by session count, with free-text preserved for agents outside the
   catalog and prefix rows displayed in the actual ID-prefix case
   (`goose:`), matching what the matcher compares against.

## Reviewer notes

- The remap and duplicate-group work is agent-agnostic — nothing in
  `internal/db`, `internal/sync`, or `internal/server` reads Augure code;
  the Augure agents merely motivated it.
- Parser behavior for existing agents is unchanged; the Codex/Hermes
  providers are reused via relabel seams, following the established TraeX
  pattern.
- The usage-cache format version was bumped (v12 → v13) for the suppression
  change, per the divergence contract in `docs/agents/storage.md`.
- Commit `502dc3f2` is large because the remap rules and duplicate-group
  features landed together.

## Verification

- `go build ./... && go vet ./...` clean; full db suite green with `fts5`
  tags; server/sync suites green.
- Frontend: `npm run check` 0 errors / 0 warnings; `npx vitest run` 164
  files / 2,733 tests pass, 0 unhandled errors; `npm run check:api` clean.


scross01:
This PR will likely need some discussion, it's not a typical new agent parser addition. 

Here's an example of the Agent remapping rules and duplicate session identification.
 
agent-remap-rules

mariusvniekerk (changes requested):
This PR conflates a whole bunch of different things. Can you split this into multiple pull requests? As it is, we're not going to merge it. You should make a separate one for the agent and we'll see about the general remap stuff. But that should not be part of this work.

mariusvniekerk:
Also we just released a rework of the remapping stuff.

scross01:
@mariusvniekerk  I figured this may be a bit contentious. There some dependency between the features implemented here to prevent over-reporting usage, but I'll take a look at breaking it up, or reducing the scope. It's probably such and edge case that it may not be worth the enhancements, but maybe there are other cases that could use this:

For background - Augure released there initial desktop and cli agent tools based Goose and Codex respectively, but didn't change where the tools stored the sessions, so for anyone already using goose and codex the sessions got mixed in, so in AgentView today Augure usage looks like Goose and Codex usage.  

Augure have improved (fixed) their clients to now use a separate sessions store, but the upgrade also copies across the sessions that where in the goose/codex stores (but also leaving the original content in the original store), so by adding the Augure parser AgentViews starts double reporting the sessions as both Augure and Codex or Goose usage.  

The two proposed features - Agent Remap, and Duplicate/Canonical Sessions, help address this by a) attributing usage for an agent that is using a specific model pattern to a different agent (i.e. the forked versions of goose and codex with the augure models) and b) for sessions that have been copied from one agent session store to another (due to the upgrade of the augure clients) the duplicate/canonical logic picks one copy (currently by message count) so usage is only counted once.

I'll explore if the new remapping framework could help with the Agent Remap case, althought the actual need for this may be an audience of one - how may people actually used the prior versions of the Augure clients, and also use both Goose and Codex, and use AgentsView - probably not a big number.   Easiest path forward is probably to just support the latest versions of the Augure clients and ignore their messy history. I'll submit that as a separate PR.

scross01:
Closing this PR in favour of separated feature specific PRs

PR #1792, opened 2026-09-16
feat(parser): support DeepSeek Harness v3 sessions

dqtz5vpvj9-create:
## What changed

agentsview now recognizes the generation-addressed DeepSeek Harness session
logs written by current releases (`session.vN.jsonl` / `session.vN.jsonl.zstd`)
and selects the newest generation when one session directory retains several
immutable files. Header validation accepts generations 0 through 3.

The parser also understands the event vocabulary those generations added:
`system/message`, `assistant/attempt`, `model/selection`, embedded
`assistant/message` streams, `tool/ptc-dispatch*`, `deliverables/presented`, and
`subagent/catalog`. It reads embedded stream usage and finish reasons, supports
the `{op, startSeq, endSeq}` replacement spelling, accepts `file` content
blocks, and applies the last `session/end-seed {inherited: true}` marker as the
inherited cut so seeded child sessions do not import their parent transcript.

Version-0 behavior is preserved, including packed assistant rows, compressed
`sourceEventSeqs` ranges, and surface replacements spelled with `start`/`end`.

## Limits

- agentsview reads each on-disk generation directly and does not run DeepSeek
  Harness's own migrations.
- Generation-2 request-header system prompts are not imported as transcript
  rows, and a failed assistant attempt with no final `assistant/message` is not
  reconstructed.
- A generation newer than 3 is passed over during discovery, so a session
  that also retains a readable generation still indexes from that log.
  Opening such a log directly reports an unsupported-version error.

Reviewers should start with `internal/parser/deepseek_harness_provider.go` for
discovery and generation selection, then `deepseek_harness_format.go` for
header, envelope, and filename handling, and `deepseek_harness.go` for event
folding, embedded streams, and the inherited seed cut.


dss09:
Verified this against a real archive — it closes the gap that 0.43.0 still leaves open. Also, the one red CI check has a one-line cause.

## The gap this fixes on top of 0.43.0

Current DeepSeek Harness releases write the session log as `session.v3.jsonl.zstd` (version zero keeps the suffix-only name, every later generation carries a `vN` component — see `dsh-session-persistence-jsonl`). 0.43.0 matches only `session.jsonl` / `session.jsonl.zstd`, so:

- a session whose only file is the v3 generation is invisible entirely;
- a session holding both a v0 and a v3 file is indexed from the v0 file only, silently truncating everything written after the upgrade.

The second case is the nastier one: the session shows up, but with stale content and a stale `ended_at`.

## A/B on identical input

Same machine, same session root, a fresh isolated database per binary (`AGENTSVIEW_DATA_DIR=… AGENTSVIEW_NO_DAEMON=1`), only the binary differs:

| session | 0.43.0 | this PR |
|---|---|---|
| `3254e96a…` | absent | 84 |
| `5750d47f…` | absent | 76 |
| `71ee14b8…` | absent | 107 |
| `9af31aa3…` | 74 | 74 |
| `session-39fbfb22…` | 675 | 675 |
| `session-8a4cd4ea…` | **absent** | **249** |
| `session-aac2df0d…` | 155 | 155 |
| `session-bf20a618…` | **17** | **419** |
| `session-c8ddfc12…` | **47** | **388** |
| `session-eb0b4dc1…` | absent | 399 |
| `session-ef0640cb…` | **0** | **275** |
| `session-f2a13bb9…` | **absent** | **0** |
| `smoke` | 4 | 4 |
| `smoke-1789548358` | 5 | 5 |

Totals: **8 sessions / 903 messages → 14 sessions**. Sessions whose data lives entirely in the v0 file (`9af31aa3…`, `session-39fbfb22…`, `session-aac2df0d…`, `smoke`, `smoke-1789548358`) are byte-identical before and after — no v0 regression observed. `session-f2a13bb9…` is legitimately near-empty (327-byte log), it just becomes visible now.

Also ran the parser suite locally: all 27 `TestDeepSeekHarness*` pass, including `TestDeepSeekHarnessV3SessionParses` and `TestDeepSeekHarnessProviderPrefersNewestGeneration`.

## The failing check

`run / lint` is the only red job (13/14 green, including the full Go matrix, race detector, integration and e2e). Cause:

```
internal/parser/deepseek_harness_provider.go:143:1: func isDeepSeekHarnessSourcePath is unused (unused)
```

It has no callers left — `isPreferredDeepSeekHarnessSourcePath` now does the generation-aware work — so it looks like a leftover from the pre-refactor path helper. Deleting it should take `lint` green.

Issue #1793, opened 2026-09-16
serve: an explicit --port should refuse a taken port (or take --strict-port) instead of moving to the next free one


Joi:
## What happens

`agentsview serve --host 127.0.0.1 --port 8080` on a Mac where another
process already listens on 8080 does not fail. It prints

```
Port 8080 in use, using 8081
agentsview v0.43.0 listening at http://127.0.0.1:8081 (started in 1.2s)
```

and runs on 8081. In v0.43.0 this is `prepareServeRuntimeConfig` in
`cmd/agentsview/serve_runtime.go`: `server.FindAvailablePort(cfg.Host,
cfg.Port)` picks the next free port whether `cfg.Port` came from the config
default or from an explicit `--port`.

## Why it hurts under a supervisor

I run the daemon under launchd with `KeepAlive`, pinned to `--host 127.0.0.1
--port 8080`, and expose it to my tailnet with `tailscale serve --tcp=8080
tcp://127.0.0.1:8080`. A collector on another machine pulls the archive from
`http://<tailnet-ip>:8080`, and the config lists that origin in
`public_origins`.

On one machine a different service held 8080. The daemon came up healthy on
8081 and the whole chain around it kept pointing at 8080: the `tailscale
serve` forwarder, the collector's `[[remote_hosts]]` url, the install
script's health probe (which waited 15 minutes on 8080 and then rolled the
install back), and `public_origins`, which `rewriteConfiguredPublicURLPort`
only adjusts when `public_url` is set. Nothing was down, but nothing
reachable either, and the only evidence was one stdout line.

For an interactive `agentsview` that opens a browser, picking a free port is
the right default. For a supervised `serve` with an explicit `--port`, a
moved port is a silent misconfiguration.

## Ask

One of:

- an explicit `--port` refuses to start when the port is taken (exit
  non-zero with the holder in the message), and only the default or `port =
  0` falls back to `FindAvailablePort`; or
- a `--strict-port` flag (or `strict_port = true` in config) with that
  behaviour, leaving the current default alone.

Either lets a supervisor see the failure instead of a working daemon on the
wrong port. I have worked around it on my side by checking the daemon log
for the `listening at` line and failing the install when the port differs,
but the daemon knows first.

v0.43.0, macOS, brew cask.

rodboev:
Implemented in https://github.com/kenn-io/agentsview/pull/1801.

PR #1801, opened 2026-09-16
fix(serve): reject collisions on an explicit port

rodboev:
An explicit nonzero `--port` now fails when occupied, so supervisors see a failure instead of a daemon running on the wrong port. Default and config-file ports retain automatic fallback; `--port 0` still chooses an available port. The shared runtime check covers `serve`, `pg serve`, and `duckdb serve`, including background launches.

Replacement checks an explicit port before stopping the running daemon. Its bind address can be reused, including aliases and narrowing a wildcard bind. Widening on the same explicit port requires stopping the daemon first because other interfaces may have unrelated listeners.

The daemon records the original `--port` value so `agentsview update` preserves that choice, including `0`. Explicit ports remain strict after restart, and a configured `public_url` for a forwarder stays unchanged when it differs from the CLI port. Without a recorded explicit port, update uses the hidden `--restart-port` flag to retry the previous listening port while retaining fallback and the configured public URL's port rewrite. Older runtime records cannot recover the original CLI choice.

Review the runtime metadata in `daemon_runtime.go`, restart arguments in `update.go`, port selection and URL rewriting in `serve_runtime.go`, and replacement preflight in `serve_lifecycle.go` together.

CI and local hooks use the same available Huma checker revision. CI prepares the SQLite headers before loading the API packages.

Refs #1793


wesm:
reviewing

Issue #1794, opened 2026-09-16
CLI JSON exposes machine keys, not machine labels (session list, usage daily --breakdown, export sessions)


Joi:
agentsview v0.43.0 (Homebrew cask), macOS, one local machine plus HTTP remote hosts.

The REST API resolves machine keys to names: `GET /api/v1/machines` returns `machine_labels` (`{"4a70f2a8…": "macazbd"}`). The CLI JSON surfaces do not:

- `agentsview session list --json` puts the key in each session's `machine` field and carries no label map.
- `agentsview usage daily --json --breakdown` puts the key in `machineBreakdowns[].machineName`.
- `agentsview export sessions --json` has no machine field at all.

So a script that wants "sessions per machine" or "spend per machine" from the CLI has to make a separate authenticated REST call to `/api/v1/machines` just for the names, or print hashes. For a collector with several remote hosts the names are the point.

Two small options, either would do:

1. Add a top-level `machine_labels` object to `session list --json` and `usage daily --json` (same shape as `/api/v1/machines`), or
2. Add `machine_label` next to `machine` on each session row and `machineLabel` next to `machineName` in the breakdown rows.

Repro:

    agentsview session list --json --since 3d --limit 1 | jq '.sessions[0].machine'
    # "4a70f2a82b531088b36678f1b96a47f8"
    agentsview usage daily --json --breakdown --no-sync --since 2026-09-14 | jq '.daily[0].machineBreakdowns[0].machineName'
    # "4a70f2a82b531088b36678f1b96a47f8"
    curl -s -H "Authorization: Bearer $TOKEN" http://127.0.0.1:8080/api/v1/machines | jq .machine_labels
    # {"4a70f2a82b531088b36678f1b96a47f8": "macazbd"}

PR #1802, opened 2026-09-16
fix(cli): include machine labels in JSON reports

rodboev:
`session list --json` and `usage daily --json --breakdown` now include a top-level `machine_labels` map, so scripts can resolve machine keys without a second authenticated request. Existing row keys stay unchanged.

Direct reads use the selected store's label catalog. Daemon-backed reads use the existing `/api/v1/machines` route. Each document includes labels only for machine keys in its rows. Label failures leave a complete document with an empty map and write the reason to stderr. The CLI gap was reported in #1794.

Refs #1794


mariusvniekerk:
This feels like it needs better types than map[string]string .

PR #1811, opened 2026-09-17
docs(skills): add a human-invoked sync-harness-icons skill

mariusvniekerk:
Adds a human-invoked repository skill, `sync-harness-icons`, for the maintenance loop that follows a batch of new agent parsers: find which supported harnesses have no brand mark, source a monochrome vector for each, add it to the landing-page chip grid, mirror it into kit-ui's `HarnessIcon` set, and move the Forge pin. That loop ran twice this month (#1577 and #1810, with kenn-io/kit-ui#54 and kenn-io/kit-ui#63) from session history and memory; this writes it down so the next run is one invocation.

The skill has `disable-model-invocation: true`, so an agent only follows it when a person asks. It ships with `audit.py`, which compares the provider headings in the format provenance doc with the grid chips, the sprite symbols, and the kit-ui registry, and prints a review list rather than a verdict: name matching cannot tell a new harness from a CLI or IDE variant of a brand already on the grid, so intentional omissions live in an explicit list in the script with a reason each.

Against current main the audit reports exactly the five harnesses #1810 adds; against that branch it reports none. The instructions were checked by a fresh agent walking a hypothetical new harness through all five sections, and the gaps it found (normalization method, chip placement, the kit-ui docs row format, how to obtain the merge commit, the Forge test command and lockfile location) are fixed in this version.

Reviewers should look at the sourcing rules in section 2, which encode the licence and fallback order used so far, and at `INTENTIONALLY_UNLISTED` in `audit.py`.


cpcloud:
Skill seems super long, but it's probably fine.

cpcloud (approved):

PR #1826, opened 2026-09-17
fix(ci): pin huma-check to its merged revision

rodboev:
CI and pre-commit hooks can resolve the Huma API checker again after the [failed CI run](https://github.com/kenn-io/agentsview/actions/runs/35255484176). Both now pin the merged revision from [kit#84](https://github.com/kenn-io/kit/pull/84).

The old pin named a commit from before the squash merge, which Go rejects as an unknown revision. The replacement has the identical source tree, so the checker behavior stays the same.


mariusvniekerk:
merged a similar one already

Issue #1821, opened 2026-09-17
Session export spends 94% of time in peer lookups due to query shape preventing index use


rodboev:
Session export spends 94% of its wall time in Claude snapshot peer lookups because the query shape prevents SQLite from using `idx_messages_claude_snapshot`. On a 4.5 GB archive with 2,196 sessions, `export sessions --all` takes 91.5 seconds. Three changes to `internal/db/session_export.go` reduce that to 2.2 seconds.

## Problem

`sessionExportClaudeSnapshotPeers` builds batches of up to 250 key pairs and searches for matching usage rows using OR-chained predicates:

```sql
(m.claude_message_id = ? AND m.claude_request_id = ?)
OR (m.claude_message_id = ? AND m.claude_request_id = ?)
OR ...
```

SQLite cannot use `idx_messages_claude_snapshot` for this shape. `EXPLAIN QUERY PLAN` shows `SCAN m USING COVERING INDEX idx_messages_usage_session_covering`, which reads every eligible usage row per batch. With 86 batches on a 2,196-session archive, the query scans the full usage index 86 times.

A secondary cost comes from the `last_activity_sort` expression, which computes `MAX(julianday(m.timestamp))` as a correlated subquery in the WHERE, ORDER BY, and watermark queries. Each page re-evaluates this for all candidate sessions. On subsequent pages, the main query contains 6 correlated scalar subqueries, each scanning that session's messages through `idx_messages_velocity`.

## Reproduction

```bash
# Baseline (current main)
time agentsview export sessions --all --format ndjson > /dev/null

# Per-function breakdown using Go overlay test (see investigation notes):
# querySessionExportUsageRows: 17.5s (94%)
# sessionExportClaudeSnapshotPeers: 16.7s (90%)
# querySessionExportRows: 0.11s
# loadPricingMapFrom: 0.06s
# all other phases: < 0.1s combined
```

Measured on a 4.5 GB archive, 2,196 eligible sessions, 5 pages at limit 500. The numbers scale super-linearly with session count because each batch scans the full usage index.

## Fixes

Three changes, all in `internal/db/session_export.go`:

[...]

PR #1828, opened 2026-09-17
perf(db): reuse activity sorting across session export pages

rodboev:
All-session export now computes the activity sort once per read transaction and reuses it across pages, which removes repeated correlated message scans from large exports.

This adds a transaction-local activity table and descending `(last_activity_sort, id)` index for `ExportAllSessionSummaries`. The watermark and page queries use the materialized values, while single-page callers keep their existing path. Filters, cursor order, output fields, and transaction cleanup stay unchanged.

On the reported 4.5 GB archive with 2,196 sessions, the complete export improvement went from 91.5 seconds to 2.2 seconds, a 42x reduction. This slice removes repeated activity sorting across the five export pages. The companion Claude peer lookup fix is [PR #1825](https://github.com/kenn-io/agentsview/pull/1825).

The reported repro came from #1821. This PR covers the all-session activity-sort slice and complements the separate Claude snapshot peer lookup fix tracked by the same issue.

Refs #1821


wesm:
looking

PR #1830, opened 2026-09-17
perf(reporting): load digest source data once for the whole range

### LINKED ISSUE #1823: Digest export reloads all source data 30 times instead of once
The 30-day digest export reloads all source data from the database 30 times, once per day. Each call to `ExportReportingDay` opens its own read transaction, queries all active sessions, loads their full message histories (with no timestamp bound), loads usage rows with 14-hour padding on each side, and loads standalone usage. Across 30 days this means:

- 30 independent read transactions
- 150 source-loading function calls (5 per day)
- Sessions spanning multiple days have their full message history loaded repeatedly
- Adjacent days' 14-hour usage padding overlaps by 28 hours, reading 2.09x the combined window

On a 4.5 GB archive the digest takes 22.8 seconds. The per-day cost averages 0.76 seconds, with the heaviest day at 1.55 seconds. At benchmark scale (15.7 GB corpus, 6 runs of two digest workloads) this is roughly 16 minutes.

The digest command (`cmd/agentsview/export_reporting.go:185-206`) calls `ExportReportingDay` in a loop. `ExportReportingDay` (`internal/db/reporting_export.go:33-99`) opens a transaction, calls `reportingHoursFromSnapshot`, and commits. `reportingHoursFromSnapshot` (lines 102-270) loads five source tables, then loops over 24 hours calling `PairActivityEvents` and `AggregateCandidatesWithJointActivity` per hour.

A shared read transaction across all 30 days would load source data once and slice it per day. The source-loading functions (`activityReportSessionsFrom`, `activityReportActivityFrom`, `reportingUsageSessionsFrom`, `activityReportUsageCandidatesFrom`, `reportingStandaloneUsageCandidatesFrom`) are at `reporting_export.go:128-167` and `activityreport.go:507-655,1196-1340`.

Constraints: daily boundaries must survive the change. Session eligibility, first-seen counts, usage survivor selection, and cost allocation currently use daily boundaries. A multi-day source load must preserve those daily decisions. Model inheritance across adjacent messages within a day must not cross day boundaries. Usage deduplication across the whole month could suppress rows that daily exports retain.

Estimated saving: ~85% of digest time (loading dominates processing). At benchmark scale, ~14 minutes per run.

Related: #1821 fixed the session export path. The digest path has the same class of problem (redundant database work on each iteration) at a different scale.

rodboev:
agentsview export digest --from A --to B now reads the archive source rows once for the requested date range. Long session histories and overlapping usage windows no longer repeat the same reads for every date.

The database opens one SQLite read transaction, loads the union of source rows, and builds each date with its existing eligibility, deduplication, cost allocation and first-seen rules. Hour and day exports keep their current output.

The source stays in memory until the digest finishes, and long sessions can make it larger than a single-day export. The implementation retains one full rendered day and compact digest results while removing repeated source acquisition.

Closes #1823


wesm:
Added a note explaining why the test keeps its saved output size and checksum, plus the command to run when those values need updating after an intentional output change.

Day and digest exports now share the same data-loading code. A bug in that shared code could affect both, so comparing their results alone is not enough. Keeping the saved result from before the change gives us an independent check. This commit only adds comments.

PR #1833, opened 2026-09-17
fix(sync): cache source failures to stop retrying unchanged broken files

### LINKED ISSUE #1820: Idle daemon burns ~1 core on large archive: 4k-file watcher loop, 46 TiB rchar in 6.5 days (v0.42.0; still ~1 core on v0.43.0)
### agentsview version

- **v0.42.0** for the 6.5-day idle observation (binary that had been running since boot)
- **v0.43.0** (`9be7745ad1906ee24e04eb05bb86c872ef0939a1`, built 2026-09-14T02:30:01Z) after a restart/upgrade — high CPU is reduced in the 4k-file stampede sense, but the daemon still sits at about **one full core** on this archive

### Install method

GitHub release binary, run as a systemd user service:

```
agentsview serve --host <ip> --port 4092 --public-url http://seamus:4092 --no-browser
```

### OS / platform

Ubuntu 24.04.4 LTS (x86_64), 8 CPUs, 64 GiB RAM. Host load was ~15–16 while otherwise idle.

### Which agent and version

n/a as a single-agent parse bug. Archive mixes Claude Code, Codex, OpenCode, Grok, and (until deleted) Piebald. ~**14.5k sessions**, `~/.agentsview/sessions.db` **~5 GB**. Extra config is only custom OpenCode dirs:

```toml
[agents.opencode]
  dirs = ["~/.local/share/opencode", "~/.config/seamus/opencode"]
```

Related parser failure that fed the same retry loop: #1819 (Piebald `chats.current_directory`).

### Which model(s)

n/a

Idle `agentsview serve` was the machine's baseline CPU hog. Nobody was using the UI.

#### v0.42.0, 6.5 days uptime (PID lived the whole boot)

| Metric | Value |
|---|---|
| Live CPU (`top`) | **~90% of one core** continuously |
| CPU time | **94 hours** in **158 hours** uptime (~60% of one core for the entire uptime) |
| RSS | 1.1 GiB |
| `rchar` | **~46 TiB** |
| `syscr` | **~11.4 billion** |
| `read_bytes` / `write_bytes` | ~171 GiB / ~757 GiB |
| `debug.log` | **96 MB** of repeating sync errors |

FDs: `sessions.db` + WAL, `usage-cache-v7-*.db`, inotify.

Watcher log, **every ~5 seconds, forever**:

```
watcher: 4229 file(s) changed, triggering sync
sync: 3 file(s) updated
watcher callback: changed-path sync incomplete: 130 source or archive failures
```

4,229 paths treated as changed, **3** actually updated, **130** failures, then the next 5s batch is 4,229 again.

Repeating failure classes in that loop:

1. Missing Grok sidecars, same paths every cycle:
   ```
   sync error: stat ~/.grok/sessions/<encoded-cwd>/<id>/summary.json: no such file or directory
   ```
2. Piebald schema mismatch (now #1819):
   ```
   sync error: loading piebald chat N: no such column: c.current_directory
   ```

This looks like a **failure-driven resync loop** plus **over-broad change detection**: failed paths never get a backoff / “don't retry until identity changes” bit, and the watcher reports thousands of changes when only a handful of files need work. v0.43.0 notes “SQLite-backed sessions stop resyncing in a loop because of AgentsView's own database reads” — the 4,229 stampede matches that class of bug, and it was still happening on **v0.42.0**.

#### After restart onto v0.43.0 (same archive)

- Startup did a **full resync** (`Data version changed, running full resync...`) of **14503 sessions**. HTTP stayed closed for ~10 minutes.
- The 4,229-file stampede stopped. Watcher batches dropped to **~40–55 files / 5s** while the Piebald failures were still present, then quieter after those files were removed.
- It did **not** become idle. Same process, ~38 minutes after start (mostly post-resync, serving HTTP 200):

| Metric | Value |
|---|---|
| `ps` `%CPU` (lifetime) | **90.9%** |
| 5s `/proc/<pid>/stat` sample | **~101% of one core** (incl. children; none at sample time) |
| `rchar` | **621 GiB** in 38 minutes |
| `read_bytes` / `write_bytes` | ~26 GiB / ~39 GiB |
| RSS | ~360 MiB |

It also logs `polling 3 unwatched root(s)` about every two minutes. I do not know which roots those are.

**Expected:** an idle daemon on a large but quiet archive should be near 0% CPU. Stat/schema failures should be sticky until the source path's identity (mtime/size/inode) changes. A watcher tick should not claim thousands of changed files when only 3 need updates. `debug.log` should not grow at tens of MB/day from the same errors.

No session transcript needed. Representative watcher cycle from v0.42.0 (redacted paths):

```
watcher: 4229 file(s) changed, triggering sync
sync: 3 file(s) updated
watcher callback: changed-path sync incomplete: 130 source or archive failures
sync error: stat /home/…/.grok/sessions/…/summary.json: no such file or directory
sync error: loading piebald chat 10: no such column: c.current_directory
```

### Steps to reproduce

1. Build a large local archive (~10k+ sessions, multi-GB `sessions.db`) with several agent roots, including some that 404/stat-fail (Grok sessions without `summary.json` is enough).
2. Run `agentsview serve` as a long-lived daemon. Do not open the UI.
3. Watch `top` and `~/.agentsview/debug.log`.
4. On v0.42.0: watcher reports thousands of changed files every 5s, CPU stays ~1 core, `rchar` grows without bound.
5. On v0.43.0: the thousand-file batches are gone, but CPU and `rchar` on this 5 GB / 14.5k-session archive still look like a tight loop rather than an idle watcher.

### Suggested direction

- Treat parser/stat failures as sticky until the source file's identity changes.
- Keep AgentsView's own `sessions.db` / WAL / usage-cache out of watch roots (v0.43.0 may already do this; v0.42.0 did not behave that way).
- Cap or coalesce changed-path batches so a noisy tree cannot schedule a 4k-file sync every 5s.
- Log *which* unwatched roots are being polled, and how expensive that poll is.

Happy to attach more `debug.log` excerpts or `agentsview doctor sync` output.

### Checklist

- [x] I searched existing issues (`high CPU`, `idle CPU`, `busy loop`, `watcher`) — no match besides the related Piebald schema issue #1819

rodboev:
An idle daemon no longer repeats the same failed work on every watcher cycle. Before this change, sync parsed a broken source file again on each pass, and a failed watcher batch retried at once whenever an unrelated file event arrived. The change bounds that repeated work; it does not claim a fixed CPU result for every archive.

## Failure cache

Sync now remembers a source whose content is malformed or whose file is missing, and skips it until the file's mtime changes. The records live in a small `internal/sync/failurecache` package backed by a new `source_failures` table. The cached identity is `{mtime, missing}` and the cache owns the comparison, so the ordinary skip cache and its readers are unchanged. The table is created with `CREATE TABLE IF NOT EXISTS` and needs no data-version bump.

- Timeouts, permission errors, I/O errors and canceled passes are never cached. `sourceParseFailureCacheable` decides this from the error text. It is the most fragile part of the change and the first place to look.
- Aider, composite, multi-file and hash-aware sources never use the cache, because one mtime does not describe their freshness. S3 and virtual paths are excluded too.
- Failures reach the archive even when the pass that found them did not complete, without promoting that pass's ordinary skip entries.
- A forced parse, a session row with a stale data version, a full resync, a watcher overflow and a per-path skip clear all forget the failure, so the source is parsed again.
- A resync build that fails in the worker process does not pass its failures back to the daemon. Each broken file is parsed once more on the next pass and cached again. An in-process resync that aborts flushes them to the live archive.
- Entries have no maximum age, and entries for missing sources are not pruned.

## Watcher backoff

A retained retry keeps its backoff delay when unrelated file events arrive. Explicit retries through `QueueRetryBatch`, authoritative reconciliation and `OpenDispatch` still dispatch at once. An `immediate` flag travels with the pending batch to carry that urgency; the wake handling in the `watcher.go` loop is where the two cases split.

## Poll logs

Unwatched-root polling now logs the roots it checked for each agent and how long the poll took.

This can be split if desired, see https://github.com/kenn-io/agentsview/pull/1833#issuecomment-5722539204. Closes #1820


rodboev:
Reworking this fix to target a series of root causes that will combine two here and two in separate slices

rodboev:
I planned this as 3 slices but all of them ended up in the same PR so that's why all the title changes. This is the synthesized result which splits across three lanes:

1. **Skip cache records failures** (core fix):
  Files that fail to parse or stat get cached at their current mtime and the watcher won't retry them until the file actually changes.
2. **Backoff survives concurrent events** (defense-in-depth):
  When a retained batch is backing off, new filesystem events no longer reset the timer to 5 seconds.
3. **Poll logs name what they're polling** (observability):
  Unwatched-poll log lines now include which roots and how long each pass took.

They fix the issue in complementary ways, and I can split them up if you prefer to review individually.

PR #1834, opened 2026-09-17
fix(desktop): restore the main window on macOS reopen

prateek:
Clicking the Dock icon or activating via Cmd-Tab with no visible windows is delivered as RunEvent::Reopen. Without a handler the close-to-tray window could not come back without quitting the app.


prateek:

- #1834 ◀

PR #1838, opened 2026-09-18
fix(parser): support older Piebald app.db schemas

### LINKED ISSUE #1819: Piebald parser requires chats.current_directory; older app.db has worktree_path only and retries forever
### agentsview version

agentsview v0.43.0 (commit 9be7745ad1906ee24e04eb05bb86c872ef0939a1, built 2026-09-14T02:30:01Z)

### Install method

GitHub release binary

### OS / platform

Ubuntu 24.04.4 LTS (x86_64)

### Which agent and version

Piebald v0.5.1 (Web). Data dir: `~/.local/share/piebald/app.db` (default Linux path). Last Diesel migration in that DB: `202606250000000000` applied 2026-07-14.

### Which model(s)

n/a — parse/schema bug, not model-specific

`agentsview serve` repeatedly failed to load every non-empty Piebald chat:

```
sync error: loading piebald chat 10: no such column: c.current_directory
sync error: loading piebald chat 11: no such column: c.current_directory
… (same for chats 14, 15, 16, 19, 20, 21, 22, 48, 56, 57)
watcher callback: watch root reconciliation failed processing page: 12 failures
```

Those 12 chats are the full set of non-deleted Piebald chats with `message_count > 0`. The failure did not settle: the watcher retried the same 12 chats about every 20 seconds (`watcher: N file(s) changed, triggering sync` → same 12 `current_directory` errors → `watch root reconciliation failed processing page: 12 failures`).

**Cause:** `internal/parser/piebald.go` `piebaldChatSelect` unconditionally selects `c.current_directory`:

```go
SELECT c.id,
       COALESCE(c.title, ''),
       c.created_at,
       COALESCE(c.updated_at, c.created_at),
       c.message_count,
       COALESCE(c.current_directory, ''),
       COALESCE(c.worktree_path, ''),
       COALESCE(c.branch_name, ''),
       COALESCE(p.directory, ''),
       COALESCE(p.name, '')
FROM chats c
LEFT JOIN projects p ON p.id = c.project_id
```

This Piebald `chats` table (v0.5.1 / last written 2026-07-14) has `worktree_path`, `branch_name`, and `project_id`, but **no `current_directory` column**. Relevant columns:

```
id, title, …, message_count, permission_mode, …, project_id, branch_name, worktree_path, is_read_only, …
```

The in-repo Piebald test fixtures create `current_directory TEXT`, so they do not cover this older/real schema.

**Expected:**
- Load chats from a Piebald `app.db` that has `worktree_path` but not `current_directory` (treat the missing column as empty).
- Schema/parse failures should not be retried every watcher tick. Sticky-fail until the source DB identity (path + mtime/size) changes, otherwise a leftover unused Piebald install keeps the daemon busy forever.

I no longer use Piebald, so I deleted `~/.local/share/piebald/` locally. Filing this so the parser does not assume a column that older `app.db` files do not have.

Cannot attach `app.db` (private chats). Repro schema for `chats` is: include `worktree_path`, `branch_name`, `project_id`; omit `current_directory`. The existing test helper in `internal/parser/piebald_test.go` / `internal/sync/piebald_integration_test.go` currently creates `current_directory TEXT` — dropping that column from a fixture should reproduce `no such column: c.current_directory` on `loadOnePiebaldChat`.

### Steps to reproduce

1. Install Piebald v0.5.1 (or any build whose `chats` table has `worktree_path` but not `current_directory`).
2. Have at least one non-deleted chat with `message_count > 0` in `~/.local/share/piebald/app.db`.
3. Run `agentsview serve` (or `agentsview sync`).
4. Watch `~/.agentsview/debug.log`: the same N chats fail on `c.current_directory` every watcher cycle.

### Checklist

- [x] I searched existing issues (`piebald`, `current_directory`, `no such column`) — no match

rodboev:
Piebald chats from older app.db files now load when current_directory is absent, so existing worktree paths still reach the archive. The parser probes the chats schema before building its query and keeps current behavior when the column exists.

The sync engine remembers a failed Piebald source by database path, size, and modification time. Repeated ordinary passes keep the failure visible without repeating member parsing or duplicate parse-error logging. A database or WAL change or explicit resync retries it.

The report came from [@colinmollenhour's issue](https://github.com/kenn-io/agentsview/issues/1819). Its private database wasn't attached, so the format inventory records that schema claim as unverified.

Closes #1819


wesm:
looking

wesm:
false positive

Issue #1824, opened 2026-09-17
Support Euro currency


cdeil:
### Problem / motivation

Hi, could you please add support to show cost information in EUR?

I think currently only USD is supported.

rodboev:
I think we can do it with a manually-entered exchange rate to start off, then add automatic rates in separate slice. We'd have to choose a rate source and decide what happens when it can't refresh.

cdeil:
Or just bundle a small master data table of historic monthly rates in the app and assume last value for later dates. To spare users having to look up and edit settings. Docs could explain that it's not super precise matching your bank statement. At least that's how I do it in my personal app where I use agentsview DB as one of many input data sources.

rodboev:
Yeah, the tricky part is when to apply the conversion. The first PR will use one rate for all displayed costs. A follow-up can use the historical rate for each period, converting costs before adding them up. That needs separate calculation logic.

rodboev:
Closing per https://github.com/kenn-io/agentsview/pull/1839#issuecomment-5724408474:

> Supporting conversion would mean taking on exchange-rate sourcing, historical rates, and ongoing maintenance across currencies. That is outside the scope of this project.

PR #1839, opened 2026-09-18
feat(usage): add EUR cost display

rodboev:
Choose EUR in Settings and enter an exchange rate to see costs in euros throughout the UI. This replaces repeated manual conversion of totals, charts and session costs.

The browser saves the currency and EUR-per-USD rate separately from language. All displayed dates use that rate, which you update yourself. Shared cost formatting converts the display; billing, API values and exports retain USD units.

{10DE733A-105F-4E28-9E33-21F228B2CE0A}{9BC9B005-1E02-44B3-BC4A-A400C4407514}

Refs #1824


mariusvniekerk:
Given that most price rates are not quoted in Euros, how does this even work?

mariusvniekerk:
Closing this PR because the conversion approach is not acceptable. Applying one manually supplied USD-to-EUR exchange rate to every displayed date makes historical euro totals change whenever that setting changes. It does not account for the exchange rate applicable to the underlying usage or billing period.

Please do not proceed with this static-rate implementation. EUR support needs an agreed approach to exchange-rate sourcing and historical conversion before implementation. We would rather leave this feature out than ship it in its current form.

mariusvniekerk:
To clarify the broader scope decision: AgentsView should not own currency conversion. This is not a request to replace the static EUR rate with a more elaborate implementation.

Supporting conversion would mean taking on exchange-rate sourcing, historical rates, and ongoing maintenance across currencies. That is outside the scope of this project.

AgentsView already supports exporting data. Users who need costs in another currency can convert that exported data with their own tools, using the rates and accounting rules appropriate to their needs. We will keep costs in USD and leave currency conversion downstream of the export.

rodboev:
> EUR support needs an agreed approach to exchange-rate sourcing and historical conversion before implementation. We would rather leave this feature out than ship it in its current form.

Yeah, that was the gap I mentioned in https://github.com/kenn-io/agentsview/issues/1824#issuecomment-5723212913. I thought a manual-rate first pass could be useful, but I can see the that it was kicking the can down a road the product is unlikely to follow. Thanks for clarifying.

PR #1846, opened 2026-09-18
feat(parser): add Augure Code (Codex-based CLI) agent

scross01:
## What

Registers Augure Code, the Codex-based CLI from Augure (augureai.ca), as its
own agent in agentsview, independent from Codex.

- Sessions are Codex-format rollout JSONL under `~/.augure/sessions`, minted
  under the new `augure-code:` ID prefix and labeled "Augure Code".
- Reuses the shared Codex parser through the existing TraeX-style "fork
  relabel" pattern: an agent spec + relabel provider, no new parsing logic.
- Resume uses the vendor's own `augure resume <id>` CLI with the Codex-shaped
  `-m` model flag, on both the server and the frontend fallback.
- Sync's format-shaped branches (duplicate resolution, layout preference,
  reconciliation identity, parse-diff mtime) treat it as a Codex-format fork
  via `isCodexFormatAgent`; the Codex-only `session_index.jsonl` and S3
  branches stay Codex-only because Augure writes no index file and has no S3
  layout.
- Archive curation includes `augure-code` in the unmarked-user-row cleanup,
  since the fork shares the Codex unpaired-notification message shape.
- Configurable via `AUGURE_CODE_SESSIONS_DIR` (docs/configuration.md), with a
  format-provenance entry in docs/internal/session-format-sources.md.
- Models observed are proprietary Augure slugs (ossington-5, rosedale-1),
  which price as unpriced until catalog coverage appears.

## Why

Augure ships a Codex-based CLI (verified on CLI 1.0.2-1.0.6: rollout JSONL
with `originator: "codex-tui"`, `rate_limits.limit_id: "codex"`, Codex field
shapes throughout, plus Augure's own `legacy_migration.json` documenting a
one-time import from a stock `~/.codex` layout). Today agentsview attributes
that usage to Codex, so a user of both tools cannot see Augure usage separately
— filtering, colors, per-agent usage, and resume all route through the codex
agent. Registering the fork as its own agent gives accurate per-agent usage
while reusing the proven Codex parser, the same way TraeX already integrates.

## Scope

Parser registration and seams only. No schema changes, no migrations, no
session-model changes (PostgreSQL/DuckDB mirrors unaffected), no remap or
duplicate/canonical machinery. Historic-session handling (agent remap rules,
duplicate/canonical detection) is deliberately out of scope and planned as
separate proposals, since Augure's earlier clients mixed sessions into the
goose/codex stores before moving to their own stores.


scross01:
This PR partially re-implements #1773

PR #1847, opened 2026-09-18
feat(parser): add Augure Desktop (Hermes-based) agent

scross01:
## What

Registers Augure Desktop 3 (beta), the Hermes-based desktop client from Augure
(augureai.ca), as its own agent in agentsview.

- Sessions live in a Hermes-schema `state.db` at `~/.augure-desktop` (macOS/
  Linux) / `%LOCALAPPDATA%\augure-desktop` (Windows), minted under the new
  `augure-desktop:` ID prefix and labeled "Augure Desktop".
- Reuses the Hermes state-DB provider through a `hermesProviderSpec` seam; the
  provider claims only roots carrying the `.augure-desktop` fork marker (the
  store's own root name, never the schema shape), so a stock Hermes-shaped
  store stays with Hermes.
- Discovery is agent-keyed, so the fork's and Hermes's independently minted
  session IDs can't collide.
- Sync treats it like Hermes for fingerprint-hash freshness and provider
  fingerprint file info.
- `session export` dispatches the Hermes JSONL writer for both agents,
  deriving the raw session ID from the agent-aware prefix table and falling
  back across the same transcript layout.
- Remote sync is explicitly excluded for this agent until safe Hermes-schema
  snapshotting exists (live WAL-backed state.db + non-transcript application
  state), with a resolver test pinning the forbidden boundary.
- Configurable via `AUGURE_DESKTOP_SESSIONS_DIR` (docs/configuration.md), with
  a format-provenance entry and README agents row.

## Why

Augure Desktop 3.0.0-beta is a Hermes Agent fork (the app bundles a 340-file
`hermes_*` Python runtime and `hermes_constants.py` declaring
`DEFAULT_HERMES_HOME_DIRNAME = ".augure-desktop"`; its `schema_version` was 26
the same day stock Hermes measured 30 — same table family, independent version
lines). Without registration that usage is invisible to agentsview. Reusing
the Hermes provider through a narrow spec seam gives full support — parsing,
usage/cost, search, export — with a small, low-risk diff, and the fork-marker
root gate plus remote-sync exclusion keep the integration conservative until
snapshotting is proven.

## Scope

Parser registration and seams only. No schema changes, no migrations, no
session-model changes (PostgreSQL/DuckDB mirrors unaffected), no remap or
duplicate/canonical machinery. Historic-session handling (agent remap rules,
duplicate/canonical detection) is deliberately out of scope and planned as
separate proposals, since Augure's earlier clients mixed sessions into the
goose/codex stores before moving to their own stores.


scross01:
This PR partially re-implements #1773

PR #1849, opened 2026-09-18
fix(e2e): wait for data responses without reading their bodies

rodboev:
Data-mode end-to-end waits now check request completion without retrieving unused response bodies, removing the Chromium protocol call that failed in [CI run #5218](https://github.com/kenn-io/agentsview/actions/runs/35362841734).

The helper asserts successful completion before the existing checks verify project reclassification, the mutation count, and the saved mapping rule. This keeps the same coverage while removing the dependency on Chromium retaining response payloads.


rodboev:
Superseded by https://github.com/kenn-io/agentsview/pull/1850, which merged the same response-completion fix and the broader fixture cleanup.

PR #1863, opened 2026-09-20
feat(analytics): split activity timeline by role with absolute and percent views

tomzx:
# What

- The "Activity by Day and Hour" timeline now splits each day's message bar into three stacked segments when the Messages metric is selected: user messages at the bottom in green, assistant messages in blue, and a gray Other segment on top for message rows that belong to neither role (tool-result carrier rows and system-injected rows), with a matching color legend in the header.
- Adds an Absolute/Percent toggle for the messages view. Percent normalizes every bar to the full plot height over the full message total, so the role mix per day stays comparable across days with very different volumes.
- Sessions view keeps a single bar per day; the toggle and legend only appear for Messages.
- Tooltips report the user/assistant/other split as counts in Absolute view and as percentages of the total in Percent view.

# Why

Messages were previously drawn as one indistinguishable bar, so there was no way to see the balance between user and assistant activity over time, and low-volume days flattened out entirely. The percent view makes the role-mix trend visible independently of daily volume. The Other segment keeps the painted stack consistent with the message total, since the backend counts tool-result carrier and system rows in the total but outside the user/assistant split.

# Design decisions

- Three `Bar` segments with array-valued `y` ranges (`[bottom, top]`) instead of layerchart's series/stack layout, because the chart renders one custom marks layer per date and this keeps the existing brush, selection, and keyboard machinery untouched.
- The Other segment reuses the existing `--chart-series-other` token and `shared_other` label, so no new color or locale keys were introduced beyond the tooltip's `other` parameter.
- Percent mode normalizes over the full message total, not the user/assistant split, so the three segments always sum to 100% and bars fill the plot.
- Only the user segment is focusable (`role="button"` with the existing arrow-key range selection); the assistant and other segments are `aria-hidden`, keeping one tab stop and one screen-reader label per day. Hovering any segment shows the shared tooltip.
- Kept the uniform 1px bar radius so segments render as `rect` elements: per-segment rounded caps switch layerchart to `path` output, which would change the DOM the component tests assert on.

# How to test

1. Open Analytics and find "Activity by Day and Hour" with Messages selected: bars are stacked green-blue-gray, the legend swatches match, and hovering a bar shows the total plus the user/assistant/other split.
2. Switch to Percent: every bar becomes full height and only the segment boundaries move, showing the day-by-day role trend.
3. Switch to Sessions: a single bar per day, and the toggle and legend hide.

# Visual proof

**Absolute (counts):**

absolute view

**Percent (share per day):**

percent view

*Captured with Playwright against the built app with seeded fixture data.*

---

Created with [create-pr](https://github.com/tomzx/agents/blob/8bd9ed89b8b39ee2d4b7ba96ff4f40563b6cbcb0/skills/create-pr/SKILL.md) via glm-5.3-flash (`8bd9ed8`)


frontend/src/lib/components/analytics/ActivityTimeline.svelte
```diff
@@ -555,6 +709,18 @@
...
 
+  .timeline-container :global(.bar-user) {
+    fill: var(--accent-green);
+  }
+
+  .timeline-container :global(.bar-assistant) {
+    fill: var(--accent-blue);
+  }
+
+  .timeline-container :global(.bar-user:focus-visible + .bar-assistant) {
```
tomzx: The focus ring only wraps the bottom (user) segment, so this sibling rule raises the assistant segment's opacity too and the full stacked bar highlights during keyboard range selection.

Posted with [create-pr](https://github.com/tomzx/agents/blob/8bd9ed89b8b39ee2d4b7ba96ff4f40563b6cbcb0/skills/create-pr/SKILL.md) via glm-5.3-flash (`8bd9ed8`)

tomzx:
Fixed in f6b9feed: the timeline now renders a third gray Other segment on top of the assistant segment covering tool-result carrier and system-injected rows, so the stack always reaches the total the tooltip reports and carrier-only buckets stay visible. Percent view normalizes against the full message total (segments sum to 100%), the tooltip and legend report the other count via the existing `shared_other` label, and a regression test covers carrier-only buckets along with stack-geometry cases for split gaps.

Posted with [handle-pr-reviewer-feedback](https://github.com/tomzx/agents/blob/8bd9ed89b8b39ee2d4b7ba96ff4f40563b6cbcb0/skills/handle-pr-reviewer-feedback/SKILL.md) via glm-5.3-flash (`8bd9ed8`)

PR #1870, opened 2026-09-20
perf(clickhouse): persist usage prices and aggregate daily reads

mariusvniekerk:
Daily usage reads from ClickHouse now aggregate stored request prices in SQL instead of transferring and pricing every event in Go. Push computes prices with the existing Go rules, preserving request rounding, historical rates, token tiers, provider adjustments, and pricing provenance.

Prices are keyed by their inputs and a digest of the pricing catalog, billing policy, and record format. Catalog changes take effect immediately through request-time pricing while the next push fills the new generation. Reader-specific overrides also use request-time pricing. Backfills use bounded batches, reads stream their results, and monetary sums preserve exact overflow checks and error identity.

This adds two derived tables in ClickHouse schema version 2. Run `agentsview clickhouse push` before using the updated reader against an existing mirror. The first push fills the price tables; older pricing generations remain available to concurrent readers. Serving stays read-only, and SQLite remains the archive. Review the pricing keys and contexts in `usage_prices.go` and the grouping and exception handling in `usage.go`.

The activity machine filter also keeps friendly labels visible and adds a short ID only when labels collide. The ClickHouse setup example documents native LZ4 compression.


mariusvniekerk:
The finding about pricing before stale-row deletion is a false positive at `3dc2387f0`.

The pricing query reads [`usage_normalized` and groups by `price_key`](https://github.com/kenn-io/agentsview/blob/3dc2387f0731abbb7addd6b1881c12d33ed49f29/internal/clickhouse/usage_prices_push.go#L85-L99). It does not read the deduplicated `usage_localized` rows. Survivor selection happens in the later [`usage_ranked` / `usage_localized` stages](https://github.com/kenn-io/agentsview/blob/3dc2387f0731abbb7addd6b1881c12d33ed49f29/internal/clickhouse/usage.go#L946-L958), so an older event sharing a deduplication key cannot suppress the new pricing input.

I checked this with a temporary integration probe against an isolated ClickHouse 25.8 instance. After an initial push, the probe inserted a replacement message at a different ordinal with the same source UUID and different token usage, leaving the old row present. It confirmed two normalized rows in one deduplication group, selected the new unpriced input, and persisted its price before stale-row deletion. The probe passed against the unchanged PR head and was removed afterward.

No production change is needed for this finding.

Issue #1872, opened 2026-09-21
Make conversation history usable as agent memory


salmonumbrella:
## The situation

AgentsView already stores conversation transcripts, supports semantic and hybrid search, and exposes search plus transcript reads over MCP. Coding agents can use those pieces, but the current recall skill does not give them a complete proactive workflow for recovering prior decisions, corrections, and implementation context.

## What works today, and why it is not enough

An agent can manually search the archive, inspect matching sessions, and read message windows. That works when the operator already knows to ask for history. It does not consistently trigger before the agent guesses, and it leaves search scope, evidence reads, source quality, fallbacks, installation, and diagnostics up to each client.

## Ask

Make conversation recall a supported agent workflow: ship the proactive skill and bounded search agent, align MCP search and evidence reads with that workflow, package lifecycle integration for Claude Code and Codex, add actionable readiness diagnostics, and verify the behavior end to end against local and remote archives.

PR #1875, opened 2026-09-21
feat(mcp): add bounded recall search contract

salmonumbrella:
Agents using MCP can now search for several literal terms within one user/assistant exchange and narrow recall by exact session, branch, project, agent, or date. The response reports the requested and effective mode, applied filters, exclusions, and whether another candidate page exists; exact current-session exclusion happens before the limit so unrelated recent sessions remain eligible.

SQLite and PostgreSQL share the same exchange grouping, scope, ordering, and pagination behavior. The daemon HTTP path and generated clients carry the full contract, while unsupported stores reject `terms` explicitly.

Refs #1872


mariusvniekerk:
Doing a bit of cleanup work here.

PR #1876, opened 2026-09-21
feat(mcp): add multi-concept semantic recall

salmonumbrella:
`search_content` can now accept two to five semantic `concepts` and return only sessions with evidence for all of them, even when the matches live in different exchanges. Results rank by the mean of each concept's strongest score and include a supporting conversation range per concept; filters and exclusions apply to every search leg, and candidate fan-out is bounded.

The shared service path keeps SQLite and PostgreSQL behavior aligned, while MCP, HTTP, and generated clients expose the same contract.

Depends on #1875.
Refs #1872.


mariusvniekerk:
So I'm taking a bunch of these concepts for better semantic search and incorporating them into the shared `kit` library. And I'll rework this PR to make use of the functionality that's going to land into kit.

PR #1879, opened 2026-09-21
Add Spanish as an interface language

cpcloud:
## TL;DR

AgentsView had six interface languages and no Spanish. Español is now a choice in Settings, browsers that report `es` or any `es-*` variant get Spanish on first load, and the catalog covers every message key.

## How

Spanish plugs into the same places as the other locales: the inlang project settings, the supported-locale list, the language picker, and two new catalogs (`messages/es.json` and `messages/find/es.json`). The browser-language matcher maps `es` and every `es-*` tag to `es`.

The translation is neutral software Spanish. Instructions address the reader informally, buttons use the infinitive, and product names, model names, paths, commands, and raw API values stay in English, as in the French catalog. Search keywords for the Language setting list Spanish terms first and keep the English ones so both match.

A handful of user-facing strings still bypassed the catalogs: the empty state in the Trends line chart and the fallback error text in the usage, analytics, trends, activity, insights, and settings stores. Those now come from messages, so every language benefits.

## Reviewer notes

- **Zero is plural in Spanish.** CLDR puts 0 in `other` for Spanish but in `one` for French. The plural test asserts "0 llamadas a herramienta" so this cannot regress silently.
- **Key alignment is enforced.** The i18n tests fail if any catalog, including the new Spanish one, drifts from the English key set.
- **Playwright covers the real switch.** `e2e/spanish-locale.spec.ts` loads the app with an `es-MX` browser locale and also switches to Español from Settings, then checks the settings chrome and the Sessions view render Spanish.
- **Focus area:** the catalog is long. Spot checks on Recall (rendered as "Recordar"), Trash ("Papelera"), and Quality ("Calidad") wording are welcome.


cpcloud:
As a second-language speaker of Spanish, I spot checked about 10 phrases in the 2k+-line catalog, and things look good to me.

If we want to add regional Spanish (e.g., Latin America) then that can be done as a follow-up.

PR #1883, opened 2026-09-21
Show query duration and a request timeline in the refresh control

cpcloud:
## TL;DR

The refresh control on the Activity, Analytics, Quality, Usage, and Recall pages said "Updated just now" with no hint of how long the refresh took, and its width tracked the text so the toolbar shifted on every tick. It now reads "Updated just now · 232 ms" at one fixed width, and hovering it opens a timeline of every query the page ran with server, transfer, and render segments per request.

## How

The wall clock runs from the first fetch a page starts to the moment it applies the last response. Each store records that as its query duration, so pages that fan out parallel panels report real elapsed time, not a sum of requests.

Per-request phases come from the shared API client. `orvalFetch` stamps when it sent the request, got headers, and finished the body, then attaches the stamps to the parsed response through a WeakMap. A store combines a stamp with its own apply time to build a step with Server, Transfer, and Render segments. The tooltip lays those steps on one axis with its origin at the first request, picks 1-2-5 tick intervals, and snaps any start within two pixels of zero onto the zero line so a burst of parallel requests shows one flush edge.

kit-ui #67 gave `RefreshControl` an `ageWidthSamples` prop that lays every candidate label into a hidden grid cell and sizes the box to the widest, plus an `ageTooltip` snippet slot. This branch pins that merge commit.

## Reviewer notes

- **Whole seconds only.** Under a second shows milliseconds, one second and up shows whole seconds, past a minute shows minutes and seconds. No decimals.
- **No placeholder without a duration.** The label drops back to the plain age string. The width samples still include the longest duration form so the box holds its size.
- **Activity maps SSE phases to steps.** The activity report streams phase markers. `ReportPhaseTimer` turns them into sessions, usage, scan, and finalize rows instead of one opaque bar.
- **Usage render starts at the later body.** Summary and the window summary arrive in parallel, so the render segment begins at the later of their body-complete stamps.
- **Focus area:** `querySegmentsFrom` and `queryStepFrom` in `frontend/src/lib/utils/refresh.ts`, and the tick and snap math in the shared `RefreshControl.svelte`. The jsdom tests read computed `left` and `width` styles because jsdom normalizes the raw style string.
- **Eight catalogs carry the 28 new keys,** including the Spanish and Azerbaijani ones that landed on main while this was in flight.

## Follow-up

The segments are client-observed. Splitting server time into database time versus the rest needs a `Server-Timing` header from the Go handlers and is a separate change.


mariusvniekerk:
It seems like a lot of lines added for measuring duration of the refresh?

cpcloud:
I guess? Most of it is not measurement, but UI:

image

mariusvniekerk (approved):

PR #1886, opened 2026-09-21
Serve semantic search from ClickHouse behind a replica vector contract

cpcloud:
## TL;DR

A ClickHouse mirror could hold every session but answered semantic and hybrid search with "not supported", so a team serving the dashboard from ClickHouse lost the search mode the SQLite and PostgreSQL paths have. This adds a replica-neutral vector search contract, makes ClickHouse its second implementation behind PostgreSQL, and moves the local export of embeddings onto kit's new `sqlitevec.Snapshot` API.

> [!IMPORTANT]
> Draft until kenn-io/kit#94 merges. `go.mod` pins kit to that PR's head commit as a pseudo-version; the final step is bumping to the released kit version.

## How

`storage.VectorSearchProvider` is what a replica backend implements to serve vectors: list pushed generations, and open a `db.VectorSearcher` for a config fingerprint or say why it cannot. `storage.VectorSearchStore` is the two setters a replica store exposes for the searcher and the unavailable reason. One gate in `cmd/agentsview/replica_vector_search.go` runs at serve startup for every replica and owns the config checks, encoder construction, and the operator-facing reason text. PostgreSQL moved onto that gate with no behavior change.

ClickHouse gets a vector phase inside `clickhouse push` and four `ReplacingMergeTree` tables keyed by the generation's config fingerprint and by the pushing archive. ClickHouse has no serial ids and no row-level upsert, so ownership follows the session tables' rule (a session's vectors live while its session row from this archive is resident) instead of PostgreSQL's owner markers. A change-scoped watch push widens to the whole generation until the archive records one clean generation-wide pass, and eviction never deletes rows another archive still records. Ranking is an exact `cosineDistance` scan in a shape the experimental `vector_similarity` index could later accelerate; the index itself is not created because it needs per-dimension tables and a result cap below the candidate pool.

The last commit replaces the raw SQL the export used to run against kit's sqlite-vec bookkeeping tables with kit's `Snapshot`, so the freshness rule lives in one place. It also swaps two private chunk rollups for `vector.RollupByDocument` and two hand-rolled single-text encoders for `vector.EncodeOne`.

## Reviewer notes

- **Behavior change for existing ClickHouse users.** A machine with `[vector]` enabled starts pushing embeddings on its next `clickhouse push`. `push_vectors = false` or `--no-vectors` opts out, matching PostgreSQL.
- **Schema version 1 to 2 only adds tables.** `EnsureSchema` upgrades an existing mirror in place. An older binary against an upgraded mirror ignores the new tables.
- **Focus area:** `internal/clickhouse/vector_push.go`. The completion marker, promotion rule, and ownership-aware eviction are where a wrong assumption would lose or leak vectors. `vector_push_chtest_test.go` covers interrupted passes, promoted exports that are not ready, and a two-archive handoff.
- **Semantic and hybrid `SearchContent` are still per backend.** PostgreSQL and ClickHouse each carry about 500 lines of the same orchestration around different SQL. A third backend is the point to lift it into `internal/db`; two did not justify it.
- **Setup runbook.** `docs/clickhouse-sync.md` gains a numbered sequence naming the output that confirms each step, written so an agent can run it unattended.

## Links

- kenn-io/kit#94 (the kit API this depends on)
- #1858 (the storage contract this builds on)


cpcloud:
Seems pretty out of scope unless the existing backends have this behavior, which I don't think they do. Looking.

